PR #2, opened 2017-06-26
Lined up description of arguments for clSetKernelArg with the order the arguments are passed in

kathawala:
For function

` cl_int clSetKernelArg ( 	cl_kernel kernel,
  	cl_uint arg_index,
  	size_t arg_size,
  	const void *arg_value)`

the arguments are described in the following order

- kernel
- arg_index
- arg_value
- arg_size

I've just changed the order of the descriptions to match the actual order of the arguments

- kernel
- arg_index
- arg_size
- arg_value

PR #17, opened 2018-11-05
describe clSetKernelArg arguments in the order they are passed

bashbaug:
This change is an updated version of the very old pull request #2.

It's been updated to work with the newer asciidoctor toolchain.  If/when this change is accepted we can close PR #2 as well.


alycm:
LGTM.

PR #27, opened 2018-11-13
Fix dead link to the INRIA paper defining ulp(x)

### LINKED ISSUE #26: Dead link for the attribution of ulp definition
https://github.com/KhronosGroup/OpenCL-Docs/blob/cb56131b1a912c59d7c229bcd2b80a68a6a0dddd/OpenCL_C.txt#L9093-L9096

The link is dead. It should probably be updated to point to https://hal.inria.fr/inria-00070503/document, or more generally using the recommended way to cite this paper:

> Jean-Michel Muller. On the definition of ulp(x). RR-5504, INRIA. 2005, pp.16. <inria-00070503>

kpet:
Fixes #26.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9091,9 +9091,13 @@ Moreover, ulp(NaN) is NaN.
 ====
 
 _Attribution: This definition was taken with consent from Jean-Michel Muller
-with slight clarification for behavior at zero.
-Refer to
-ftp://ftp.inria.fr/INRIA/publication/publi-pdf/RR/RR-5504.pdf[_ftp://ftp.inria.fr/INRIA/publication/publi-pdf/RR/RR-5504.pdf_]._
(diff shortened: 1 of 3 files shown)


bashbaug:
Would you mind fixing this in the OpenCL C++ kernel language spec as well?

https://github.com/KhronosGroup/OpenCL-Docs/blob/master/cxx/numerical_compliance/relative_error_as_ulps.txt#L17

The same reference is also in the OpenCL SPIR-V Environment Spec.  It doesn't include a dead link, so it's nominally OK, or we could add a link there as well for consistency.

https://github.com/KhronosGroup/OpenCL-Docs/blob/master/env/references.txt#L41

kpet:
Hmm, I should have grep'ed around. Thanks for spotting these. Will update the PR.

bashbaug:
Looks good - thanks!

PR #39, opened 2019-01-08
Fix typos.

PiotrFusik:


bashbaug:
These all look like spelling fixes - thanks for catching these!

I'll leave this PR open for one more day just to get a few more eyeballs on it, but in general LGTM.

PR #42, opened 2019-01-16
Convert extension template and Makefile to asciidoctor

oddhack:
Extension HTML docs now use the same toolchain and CSS as the OpenCL API
HTML doc.

Something people might care about is that the individual extension
HTML/PDF now don't have a TOC. This could probably be fixed with some
fairly straightforward conditionals in the documents to check whether
they are being included in the unified extensions doc or built
standalone.

This works as is, but it would probably make more sense to use the
parent Makefile to generate these documents as well as the API, C++,
etc. specs.

Closes internal issue 162.


oddhack:
@outofcontrol this PR is confused about the CLA too. Clicking on the link says "You have signed the CLA" but it still shows as "Some checks haven't completed yet". Feels like some process is hung up on the CLA server where it communicates with github. Note this is the same repository as the other problem PR, which cleared up much more quickly.

outofcontrol:
Just asked cla-assistant to recheck this repository. Seems that cleared this one up. Perhaps cla-assistant is having a bad hump-day?

oddhack:
Include all extensions targets in top-level Makefile and update extensions/README. Resolve conflicts.

bashbaug:
This was approved by the working group in the February 5th call.

Issue #33, opened 2018-11-30
Common float built-in functions have no precision requirements


EwanC:
Section 6.12.4 of the 1.2 spec defines common built-in functions that operate on floating point types. However unlike the maths functions from 6.12.2 which have their errors defined as ULPs in section 7.4, there doesn't seem to be a precision requirement for common functions.

The CTS however does check for a certain level of precision. [Degrees](https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_degrees.c#L81) and [radians](https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_radians.c#L81) have a ULP error, while [mix](https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_mix.c#L33
) and [smoothstep](https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_smoothstep.c#L73) look like they use an absolute error.

Should these CTS precision requirements be defined in the spec somewhere? And if so how should the error thresholds be translated for half in the cl_khr_fp16 extension?

alycm:
A collection of notes on this w.r.t. the most recent set of specifications:

As the original report states: the CTS uses a ULP of 2.0 for `degrees` and `radians` for both `float` and `double`.  This could be handled by adding that value to tables 7.1 and 7.3 of the OpenCL C 2.0 language specification, and tables 5 and 6 of the OpenCL 2.2 SPIR-V Environment specification.  Also as the original report states, the CTS isn't a reference for `cl_khr_fp16`, but I will make the arbitrary observation that other builtins with FP32 accuracy of 2 ULP also have FP16 accuracy of 2 ULP.

The CTS using absolute errors for `mix` and `smoothstep` is interesting, because the specifications only use absolute errors for a few operations when unsafe optimizations are enabled.  Default-mode operations always use ULP for errors (so far).

This sentence from the OpenCL C specification:

>The mix and smoothstep functions can be implemented using contractions such as mad or fma.

has implications for embedded profile as there  `mad` has "implementation defined" accuracy.   This probably requires that `mix` and `smoothstep` also have "implementation defined" accuracy in the embedded profile to match.

[...]

PR #44, opened 2019-01-23
Define error tolerances for common functions

EwanC:
Define accepted error thresholds for the common class of functions from table 6.12

Taken from how the CTS is doing verification, however coming up with values for `mix()` and `smoothstep()` is problematic for half and double, since they're only tested against 32-bit float in the CTS using an absolute error.

I've put 'Implementation defined' down for these as a placeholder due the below spec note, but not sure if that's the correct solution.

> The mix and smoothstep functions can be implemented using contractions
> such as mad or fma

Related to [issue 33](https://github.com/KhronosGroup/OpenCL-Docs/issues/33)

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9148,10 +9148,12 @@ currently undefined.
 | *atanh*               | {leq} 5 ulp
 | *cbrt*                | {leq} 2 ulp
 | *ceil*                | Correctly rounded
+| *clamp*               | 0 ulp
 | *copysign*            | 0 ulp
 | *cos*                 | {leq} 4 ulp
 | *cosh*                | {leq} 4 ulp
 | *cospi*               | {leq} 4 ulp
+| *degrees*             | {leq} 2 ulp
 | *erfc*                | {leq} 16 ulp
 | *erf*                 | {leq} 16 ulp
 | *exp*                 | {leq} 3 ulp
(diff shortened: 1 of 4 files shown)


OpenCL_C.txt
```diff
@@ -9178,25 +9180,32 @@ currently undefined.
 | *mad*                 | Implemented either as a correctly rounded fma or
                           as a multiply followed by an add both of which are
                           correctly rounded
+| *max*                 | 0 ulp
 | *maxmag*              | 0 ulp
+| *min*                 | 0 ulp
 | *minmag*              | 0 ulp
+| *mix*                 | absolute error tolerance of 1e-3
```
EwanC: https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_mix.c#L33

OpenCL_C.txt
```diff
@@ -9178,25 +9180,32 @@ currently undefined.
...
 | *rint*                | Correctly rounded
 | *rootn*               | {leq} 16 ulp
 | *round*               | Correctly rounded
 | *rsqrt*               | {leq} 2 ulp
+| *sign*                | 0 ulp
 | *sin*                 | {leq} 4 ulp
 | *sincos*              | {leq} 4 ulp for sine and cosine values
 | *sinh*                | {leq} 4 ulp
 | *sinpi*               | {leq} 4 ulp
+| *smoothstep*          | absolute error tolerance of 1e-5
```
EwanC: https://github.com/KhronosGroup/OpenCL-CTS/blob/cl22_trunk/test_conformance/commonfns/test_smoothstep.c#L73

bashbaug:
Merged by resolving conflicts and pushing as 68a5a10, as per working group discussion.

PR #52, opened 2019-02-15
clean up the descriptions for cl_khr_gl_sharing

bashbaug:
This change rewrites the intro sections to Chapters 10 and 11 in the OpenCL Extensions Spec so it is more clear that they are describing functions in the **cl_khr_gl_sharing** extension.

These changes should address internal issue 157.


bashbaug:
Thanks Jon!  @alycm can you give this one final sanity check review and merge?

Issue #54, opened 2019-02-21
Disallowing blocks as function arguments


AnastasiaStulova:
According to OpenCL C spec v2.0 we should be able to compile the following:

```
int foo(int (^ bl)(void)) {
  return bl();
}

int get21() {
  return foo(^{return 21;});
}

int get42() {
  return foo(^{return 42;});
}
```

However, the only way to implement this in general case is to use function pointers because even if the blocks passed to `foo` are statically known there are multiple of them and therefore the call into block can't be resolved in traditional way.

Some options to resolve this issue are:
1. Inlining
2. Add switch and case with a call to corresponding block function and replace the argument to block by the argument indicating what case from the switch to pick. 

As for (1) we don't want to always enforce inlining of all functions on the way to block parameter because it can be inefficient. This might need to enable optimizations (to resolve function pointers in general case)?

As for (2) it has an extra overhead of switch and case logic. Additionally it will rely on compiler analysis and optimizations to determine what blocks are being passes at each call.

Currently upstream Clang generates function pointers that can't be handled by the SPIRV-LLVM Translator: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/102

Considering that no good alternatives were found to support this in the compiler would it be possible to disallow this?

See some discussions on Clang review: https://reviews.llvm.org/D43783

bashbaug:
I dug up the old Bugzilla that proposed the block restrictions restrictions that are currently in the OpenCL C spec (12024).  Interestingly, blocks were initially disallowed as function arguments, but this was ultimately relaxed (comment 6).

I don't have a strong opinion and would be fine disallowing blocks as arguments to user-defined functions (we'd obviously need to allow passing a block to `enqueue_kernel()`, etc), I'm just observing that this isn't the first time that this issue has been raised.

bashbaug:
I just accepted PR #59 so I am closing this issue.

PR #59, opened 2019-03-12
#54 Disallow blocks as function arguments

AnastasiaStulova:


bashbaug:
Accepting as discussed in the March 12th teleconference.

PR #62, opened 2019-03-12
move acknowledgements to the end the API, C, and C++ specs

bashbaug:
I think we should consider moving the "Acknowledgements" section of the specs to the end vs. the beginning.  This avoids needing to scroll through a few pages before getting to the content, especially for the HTML versions of the specs.

### DIFF
--- a/api/acknowledgements.txt
+++ b/api/acknowledgements.txt
@@ -2,7 +2,8 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
-*Acknowledgements*
+[float]
+= Acknowledgements
 
 The OpenCL specification is the result of the contributions of many people,
 representing a cross section of the desktop, hand-held, and embedded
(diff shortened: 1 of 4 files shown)


api/acknowledgements.txt
```diff
@@ -2,7 +2,8 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
-*Acknowledgements*
+[float]
```
alycm: I have tried building both with and without this `[float]` attribute.  The only visible effect that I saw is inclusion in the table of contents or not, why remove it?  The effect is most visible (to me) for the HTML output as I'm more likely to use the table of contents to browse there.

bashbaug: I was concerned that without the `[float]` attribute the chapter numbering might be messed up, but everything looks good without it, so I've removed it.  The "Acknowledgements" now shows up in the table of contents, without a chapter number, which seems reasonable.  Thanks!

alycm: Okay, thanks.

Issue #66, opened 2019-03-26
Old-style atomics with generic address space


davidrohr:
Porting some of our OpenCL 1.2 code to OpenCL 2.0, I encountered some problems with atomic operations. After getting rid of the __global and __shared specifiers to pointers, the code fails to compile.

The reason is that the "old-style" atomics we are using, e.g. atomic_add, yield in clang:
address space mismatch in 1st argument ('__generic int *'), parameter type must be 'volatile __global int *'

Unfortunately, switching to the new atomics in C11/C++11 style is not straight forward.
In other parts of the code, where we exclude concurrent access to the integers, we use binary operators (&,  |, ^), which are not available for the C11 atomics like atomic_uint.

Basically, we rely on the fact that the integers behave as normal integers when the pointers are dereferenced, but we need the atomic operations on them in addition (we make sure ourselves, that this will not collide). This is a feature available with the old OpenCL 1.2. atomics, but not with the new C11 atomics.

In my opinion, all operations of "normal" integers should be available also for atomic_uint, etc, or the old-style atomic operations like atomic_add should work with pointers to the generic address space in OpenCL 2.0.

presentfactory:
I am having a similar problem to this due to an algorithm I had written before using the 1.2-style atomics where in most cases non-atomic access would be fine and more performant, but only requiring atomics on the same variables in specific contexts. Additionally due to the fact that there's no specification from what I can see on the size of the new atomic types, I am not even sure how to properly allocate memory for them, for example a buffer of `uint`s one would usually allocate `sizeof(cl_uint) * size` bytes, but if they are `atomic_uint`s, then there's not much that can be done other than hope it matches the size of the non-atomic type (which it should really with how GPUs work usually and given that they're guaranteed to be lockfree).

I'd say there should be some way to use the old behavior or allow for "casting" of the new atomics to their respective types, perhaps as a pointer or something to modify it through non-atomically. Another option could be that since the OpenCL standard for access of non-active union members, perhaps a `uint` unioned with an `atomic_uint` could be defined to work properly depending on which is accessed, but I have no idea really.

AnastasiaStulova:
Also to address the problem with host allocation that @presentfactory mentions would it make sense to allow `_Atomic` specifier? This would mean that all atomic types are represented as regular types with extra atomic specifier. Would this resolve the confusion about allocation of atomic type buffers on the host because it would be just be using the basic type.

`_Atomic int = atomic_int = cl_int`

[...]

Issue #65, opened 2019-03-25
Procedure for evolution of language spec


AnastasiaStulova:
Even though the specs are published some time ago it is desirable to have the ability to improve them and fix the bugs.

The question is how to change existing specs and propagate changes to tools and testing (in particular CTS) minimizing negative impact.

Summary of previous discussions:
- If changes affect the diagnostics i.e. some code would be rejected or accepted differently from the older specs. What should happen to CTS and how to dial with the drivers that are already shipped that will behave differently to the new drivers?
- Adding an extension for each modification in a language doesn't seem like a great language design choice.
- Changing language version for a group of features could be an option  but (a) how to define the amount of changes that is enough (b) how long would it take to wait for a new version release.
- Clang extension is not portable and has some limitations i.e. ideally it shouldn't be big changes.

Some early observations:
- Clang has one language mode for each standard i.e. -std=c++11. It documents revision that it implements and deltas between releases however there is no formal procedure for this.
- TODO: Check the procedure with ISO.

AnastasiaStulova:
One idea could be to use minor version and let compiler add implicit macro indicating which minor version it implements for a certain standard.

Example: If we are to publish changes to CL2.0

We could add a macro `__CL_VERSION_MINOR__` that could have values
`CL_MINOR_VERSION_CL_2_0_0` for original spec
`CL_MINOR_VERSION_CL_2_0_1` for the spec with the changes to the original revision
`CL_MINOR_VERSION_CL_2_0_2` ...

This will allow to query what compiler supports and condition the code to make sure it's portable between drivers if needed.

We would need to define the process for publishing the minor spec versions. The CTS can use the defined implicit macros to condition the code too. It is relatively easy compiler change because it would still need to support one language mode. However, we would need to think of the process for the macro to be available to the users (is it at the beginning of the new version development phase or at the end).

keryell:
Perhaps we could use something similar to C++20 feature testing macros https://en.cppreference.com/w/cpp/feature_test ?

davidrohr:
I agree that it should be transparent to the programmer as long as it is related to minor revisions / bug fixes. For real bug-fixes, I am not even sure how prominently I would present them, in particular when they are backward compatible. In particular general minor improvements should not become OpenCL extensions. This would only complicate things in my opinion. Extension make most sense for something that might not be generally available also in the future.

[...]

PR #71, opened 2019-04-09
#66 Allow operations on C11 atomic types.

AnastasiaStulova:
This is the frontend only feature. Removing the restriction
won't result in incorrect use of atomic types. There is a
convenience feature since using regular operations
is more concise than builtin functions.


kpet:
Is it worth stating explicitly the `memory_order` and `memory_scope` that apply when operators are used? Should we add new conformance tests?

AnastasiaStulova:
OK, alternatively we can reference C11 spec?

My understanding is we are not testing C standard features in OpenCL CTS. But I don't know whether this is our "official" position wrt testing language changes in CTS.

bashbaug:
I think this is very much related to the discussion in #65.

This seems like a very useful feature, but I am concerned that if we accept this PR and make behavior that was previously explicitly forbidden now valid, we may cause shipping implementations to become non-conformant.

PR #73, opened 2019-04-25
Disconnected minor fixes

alycm:
Various small fixes to things I noticed while working on other things.

### DIFF
--- a/api/opencl_platform_layer.txt
+++ b/api/opencl_platform_layer.txt
@@ -755,7 +755,8 @@ device except for the following queries:
         _OpenCL<space><major_version.minor_version><space><vendor-specific
         information>_
 
-        The major_version.minor_version value returned will be 2.2.
+        The major_version.minor_version value returned will be one of 1.0,
+        1.1, 1.2, 2.0, 2.1 or 2.2.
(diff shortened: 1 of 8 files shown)


api/opencl_platform_layer.txt
```diff
@@ -1385,7 +1386,7 @@ on one or more devices specified in the context.
   | cl_bool
       | Specifies whether the user is responsible for synchronization
         between OpenCL and other APIs.
-        Please refer to the specific sections in the OpenCL 2.0 extension
+        Please refer to the specific sections in the OpenCL 2.2 extension
```
kpet: Should we start referring to the unified specification instead?

alycm: Yes, I'll reword this.  Those the unified specification is still v2.2-10.

bashbaug:
These changes look good to me.  Did you want to change the text to refer to the unified extension spec?

I don't see a strong reason to wait so I'll accept this PR tomorrow unless you'd rather wait - thanks!

api/opencl_platform_layer.txt
alycm: Done, I know just say "OpenCL Extension Specification".

PR #78, opened 2019-05-12
WIP: Very initial prototype of a small number of unified API spec changes

alycm:


alycm:
This is still WIP as I may break this into many PRs to ease review.

alycm:
After discussion I've removed WIP, this PR will be reviewed as is.

api/opencl_runtime_layer.asciidoc
```diff
@@ -304,19 +392,49 @@ Otherwise, it returns one of the following errors:
     by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
     required by the OpenCL implementation on the host.
+--
+
+[open,refpage='clSetCommandQueueProperty',desc='Enable or disable properties of a command-queue.',type='protos']
+--
+To enable or disable the properties of a command-queue, call the function
+
+include::{generated}/api/protos/clSetCommandQueueProperty.txt[]
```
bashbaug: The description of this API was transferred faithfully from the older version of the OpenCL spec, but I think we should add a stronger warning that it may not behave as expected in newer versions of OpenCL.  For example, the unified headers currently has the comment:

```c
    /*
     *  WARNING:
     *     This API introduces mutable state into the OpenCL implementation. It has been REMOVED
     *  to better facilitate thread safety.  The 1.0 API is not thread safe. It is not tested by the
     *  OpenCL 1.1 conformance test, and consequently may not work or may not work dependably.
     *  It is likely to be non-performant. Use of this API is not advised. Use at your own risk.
     *
     *  Software developers previously relying on this API are instructed to set the command queue
     *  properties when creating the queue, instead.
     */
```

Perhaps this text should also be included in the spec, after the "NOTE" below?

api/appendix_f.asciidoc
```diff
@@ -16,195 +16,321 @@ This section lists OpenCL error codes and their meanings.
...
 
 | {CL_INVALID_EVENT_WAIT_LIST_anchor}
+
+include::{generated}/api/version-notes/CL_INVALID_EVENT_WAIT_LIST.asciidoc[]
 | Returned when the specified event wait list or number of events in the wait list is not valid.
 
 // This is currently defined in cl.h, but it's not a core API error code.
 //| {CL_INVALID_GL_OBJECT_anchor}
+
+include::{generated}/api/version-notes/CL_INVALID_GL_OBJECT.asciidoc[]
```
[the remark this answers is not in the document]
alycm: Oops, yes, I'll fix this.

api/opencl_architecture.asciidoc
```diff
@@ -1027,12 +1041,33 @@ The various SVM mechanisms to access host memory from the work-items
...
+OpenCL 1.x uses a relaxed consistency memory model; i.e. the state of memory
+visible to a work-item is not guaranteed to be consistent across the collection
+of work-items at all times.
+
+Within a work-item memory has load / store consistency.
+Local memory is consistent across work-items in a single work-group at a
+work-group barrier.
+Global memory is consistent across work-items in a single work-group at a
+work-group barrier, but there are no guarantees of memory consistency between
+different work-groups executing a kernel.
```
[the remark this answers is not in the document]
alycm: Hmmm, good point.  Perhaps something like the below text.  But as you say, I'll have that in it's own PR as it changes words, rather than bury it in here.

> Local memory is consistent across work-items in a single work-group at a work-group barrier if `CLK_LOCAL_MEM_FENCE` is set in the barrier memory fence flags.
> Global memory is consistent across work-items in a single work-group at a work-group barrier if `CLK_GLOBAL_MEM_FENCE` is set in the barrier memory fence flags.
> There are no guarantees of memory consistency between different work-groups executing a kernel.

api/opencl_runtime_layer.asciidoc
alycm: Thanks, this is helpful.   Handling thread safety language is still on the TODO list, in particular due to the differences between 1.0 and 1.1+.

api/introduction.asciidoc
```diff
@@ -105,3 +105,20 @@ When the _revision_ number changes there may be an impact on the behavior of
...
+
+This document specifies all versions of the OpenCL API.
+
+There are three ways that an OpenCL feature may be described in terms of what
+versions of OpenCL support that feature.
+
+  * Missing before _major.minor_: Features that were introduced in
+    version _major.minor_.  Implementations of an earlier version of OpenCL
+    will not provide these features.
+  * Deprecated by _major.minor_: Features that were deprecated
```
AlexeySachkov: Does it make sense to add something like `Missing after` or `Removed by`? Just to be able to clearly state when particular feature was removed

api/opencl_architecture.asciidoc
```diff
@@ -789,13 +792,19 @@ memory:
 
 The constant address space is disjoint from the generic address space.
 
+NOTE: The generic address space is <<unified-spec, missing before>> version
```
AlexeySachkov: Strictly speaking, generic address space **is** documented by [OpenCL 1.2 spec](https://www.khronos.org/registry/OpenCL/specs/opencl-1.2.pdf), from section 6.5 Address Space Qualifiers:

> If the type of an object is qualified by an address space name, the object is allocated in the
specified address name; otherwise, the object is allocated in the generic address space.

> The generic address space name for arguments to a function in a program, or local variables of a
function is `__private`. 

But apparently it has different meaning, so, not sure if we need to change anything here

scripts/gen_version_notes.py
```diff
@@ -0,0 +1,140 @@
...
+# hierarchy, and then find all the ${entry_type} within each hierarchy:
+def process_xml(spec, entry_type, note_printer):
+    numberOfEntries = 0
+    numberOfNewEntries = 0
+    numberOfDeprecatedEntries = 0
+
+    for feature in spec.findall('.//feature/require/%s/../..' % entry_type):
+        for entry in feature.findall('.//%s' % entry_type):
+            name = entry.get('name')
+            #print('found ' + entry_type + ': ' + name)
```
AlexeySachkov:
```suggestion

```

scripts/gen_version_notes.py
```diff
@@ -0,0 +1,140 @@
...
+        '-o',
+        action='store',
+        dest='directory',
+        default='.',
+        help='Create target and related files in specified directory')
+
+    args = parser.parse_args()
+
+    specpath = args.registry
+    #specpath = "https://raw.githubusercontent.com/KhronosGroup/OpenCL-Registry/master/xml/cl.xml"
```
AlexeySachkov:
```suggestion

```

api/introduction.asciidoc
alycm: Yes, I think that would make sense.  The problem is that OpenCL doesn't seem to have explicitly removed anything.  I have created an issue to consider this point: https://github.com/KhronosGroup/OpenCL-Docs/issues/153

api/opencl_architecture.asciidoc
alycm: I hadn't ever made that connection before.  I think this is just a coincidence of wording and that the use of `generic address space` in that paragraph could have also just said `private address space` directly.

In the context of a unified specification I think the current text is clear enough, it is talking about an actual address space called "generic", and that is what was missing before 2.0.

I will resolve this comment, but if you think the wording should be changed then please file an issue and I'll happily look at it a bit more.

api/opencl_architecture.asciidoc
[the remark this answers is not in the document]
alycm: I've put this into an issue, but it will be seconds of work to put the above wording into a PR (just not extending this one).

https://github.com/KhronosGroup/OpenCL-Docs/issues/154

api/opencl_runtime_layer.asciidoc
alycm: I've filed this as an issue so that I could look at the thread-safety descriptions all at once: https://github.com/KhronosGroup/OpenCL-Docs/issues/155

PR #79, opened 2019-05-13
Tweak the generator script to allow anonymous <member> declarations

oddhack:
(e.g. no <name> field), and modify the cl_image_desc XML definition to format more cleanly (internal issue 177). This has no impact on spec language, just generated code formatting.

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -626,11 +626,13 @@ class OutputGenerator:
             paramdecl = indent + ' '.join(paramdecl.split())
         return paramdecl
 
-    # getCParamTypeLength - return the length of the type field is an indented, formatted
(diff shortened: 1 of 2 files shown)


oddhack:
@bashbaug this could go in whenever you're comfortable with it - very minor tweakage.

bashbaug:
This is much better - thanks!

The one small issue I see is that the `#ifdef` and `#endif` are indented differently now:

```c
    #ifdef __GNUC__
    __extension__   /* Prevents warnings about anonymous union in -pedantic builds */
#endif
```

I think we can fix this by indenting the `#endif` in the XML schema so it lines up with the corresponding `#ifdef` and the `__extension__`, but that's no reason to hold up this PR.

PR #80, opened 2019-05-20
Add Subgroup to the list of valid Memory Scopes for 2.1/2.2

kpet:

### DIFF
--- a/env/opencl2_1.txt
+++ b/env/opencl2_1.txt
@@ -87,6 +87,7 @@ _Scope_ for _Memory_ is generally limited to:
   * *CrossDevice*
   * *Device*
   * *Workgroup*
+  * *Subgroup*
   * *Invocation*
 
 _Scope_ for _Execution_ for the *OpGroupAsyncCopy* and *OpGroupWaitEvents*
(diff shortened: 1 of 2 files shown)


bashbaug:
I think there is an issue here, but I wonder if this is the wrong fix.

For OpenCL 2.1 or newer, or any implementation supporting subgroups, couldn't **Subgroup** be a valid _Scope_ for _Memory_?  It would correspond to an atomic with `memory_scope_sub_group`.

kpet:
Agree this is missing. I was planning to deal with all the subgroup-related issues later but this was probably a bad idea :). I'll repurpose this PR to add the missing _Subgroup_ _Scope_ and look into reorganising the rules and solving the other issues later instead.

bashbaug:
> look into reorganising the rules and solving the other issues later instead.

Heads up, I'm already about halfway through refactoring these parts of the environment spec already, for other reasons (see internal issue 156).  I appreciate the help, but let's be sure we aren't duplicating work - thanks!

kpet:
Right, hadn't caught up with it yet. I will pause work on the environment spec until you're done with that. Do you want to get this in or would you rather I close it?

kpet:
This will be addressed by @bashbaug 's #84. Closing.

PR #83, opened 2019-05-28
Add static refpages

oddhack:
All of these are for extensions, but not easily extractable from the 
OpenCL Extensions document at present. There are a
few minor tweaks to the spec sources and changes to the Makefile to
incorporate the static files into the refpage build, and create the
refpage navigation and sidebar indices.

The static pages are imported largely unchanged from the Docbook
versions, aside from reformatting to match the other pages. This means
that there are a lot of out-of-date references to explicit section and
table numbers in other documents, and that they have incomplete coverage
of extensions (in particular, the EXTENSION page is not accurate about
what extensions may be enabled).

However, once we've merged this, I'll be able to replace the old
refpages on khronos.org, and we can move forward from there as the WG needs to.

N.b. this also adds a couple of extracted index pages in the spec sources. These are to replace Docbook index pages that just didn't exist in the spec. At present they are not included in the spec builds, although they could be tweaked and included - they would make sense as section indices, too.

Big thanks to the Stream Computing folks and https://github.com/OpenCL/man for their tools to convert Docbook -> asciidoc!


api/e_appendix.txt
```diff
@@ -3,7 +3,11 @@
 // http://creativecommons.org/licenses/by/4.0/
 
 [appendix]
-= Changes
+[[changes_to_opencl]]
+= Changes to OpenCL
+
+The following sections describe changes to the OpenCL API and OpenCL C which
+were made in each successive version.
```
bashbaug: This is phrased awkwardly.  Consider something like: "changes to the OpenCL API and OpenCL C Language specifications, which were made..."

bashbaug:
There are a lot of changes in this PR, so I didn't review everything in detail, but I did a quick look through and verified that I was able to build all of the man pages with these changes.  Nice work!

Two questions:

1. Luckily, we don't make changes to the extensions spec very often, but if we do fix bugs or add a new extension I presume we should make "manual" updates to the man pages similar to the changes in this PR?

1. Longer term, do you have any suggestions how to structure the extensions spec differently so at least some of the extensions man pages could be generated automatically?

I'll leave this PR open for one more day, but I hope we are able to approve and merge it in tomorrow's teleconference.  Thanks!

api/e_appendix.txt
oddhack: Should be fixed - I added a bit of verbosity in the subsection headers to make it more clear where the changes applied.

oddhack:
> Two questions:
> 
> 1. Luckily, we don't make changes to the extensions spec very often, but if we do fix bugs or add a new extension I presume we should make "manual" updates to the man pages similar to the changes in this PR?

Unfortunately, that would be required to keep them in sync (to the extent they're in sync now - I did some minor cleanup on the pages as part of this PR, but many of them are quite old - this probably mostly affects cross-references).

> 2. Longer term, do you have any suggestions how to structure the extensions spec differently so at least some of the extensions man pages could be generated automatically?

I think the extension-string refpages (the 'cl_khr_*' pages) could be generated automatically from cl.xml, for the most part. We do something like that with the Vulkan extension appendices today. That's about 40% of the static pages.

The refpages are extracted as a single contiguous block of asciidoctor markup. So or the rest of the pages - the descriptions of individual commands and types added by extensions - it would take significant rewriting of the extensions docs. They are structured as summaries of (all the command) (all the structures) (all the enums) followed by a list of deltas to the base spec they extension modifies, describing the new APIs in more detail. So in addition to refactoring the pages to be individual descriptions of APIs, a bunch of language from the core spec would have to be manually imported to the ext/* documents so that the refpages didn't read "changes to / modify this section" followed by sentence fragments. Alternatively, a Vulkan-like model could be done where the extension language is folded into the core spec, protected by asciidoctor ifdefs - then there's a single source base that's more immune to future changes to the core. Either one of these is a lot of work and a significant change to how the extensions are published.

> I'll leave this PR open for one more day, but I hope we are able to approve and merge it in tomorrow's teleconference. Thanks!

Great. Once the PR is merged, I will update the published reference pages with a new OpenCL 2.2 area built off this repo, and put instructions for doing it into a README or BUILD document at the repo root for future use, if someone else needs to update them.

bashbaug:
Merging as discussed in the June 11th teleconference.

PR #89, opened 2019-06-19
Clarify 1.2 validation rules for Memory Semantics

kpet:
Memory Semantics is used to specify both memory-order
constraints and the storage classes they apply to.

Make it explicit that only the memory-order constraint
bits are restricted. The storage class can vary
(WorkgroupMemory or CrossWorkgroupMemory).


kpet:
This begs the question: do we want 1.2 validation rules for the storage classes? Can we define them more generally?

bashbaug:
I've created issue #91 so we don't lose track of the question regarding validation rules for storage classes after merging this PR.

kpet:
Thanks!

PR #95, opened 2019-07-01
OpenCL XML Fixes

bashbaug:
This PR fixes a few issues I've found generating code from the OpenCL XML file:

* For several APIs, the "origin" and "region" arguments are of pointer type and include an array size in square brackets, which is redundant/incorrect.  I've removed the square brackets, which is consistent with the OpenCL headers.

* For clUnloadCompiler, I've added an explicit unnamed parameter of type "void", vs. no parameters.  This eliminates a special-case of no paramters, and is also consistent with the OpenCL headers and the OpenCL 1.0 specification.

* Several deprecated APIs were included in section for the API they were deprecated in rather than the section for the API that added them.  I've fixed a few of these, but there may be others.

* I fixed a small typo in my recent PR where the name of the unused enum block was not updated correctly.

I've built the specs and online man pages with these changes with no issues.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -2778,8 +2778,8 @@ server's OpenCL/api-docs repository.
             <param><type>cl_command_queue</type>                        <name>command_queue</name></param>
             <param><type>cl_mem</type>                                  <name>image</name></param>
             <param><type>cl_bool</type>                                 <name>blocking_read</name></param>
-            <param>const <type>size_t</type>*                           <name>origin[3]</name></param>
-            <param>const <type>size_t</type>*                           <name>region[3]</name></param>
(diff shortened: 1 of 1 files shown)


oddhack:
Incidentally, this reminds me that I made some progress towards being able to autogenerate C headers from cl.xml, but kind of stalled out because getting all the deprecation stuff right looked like a great deal of work. Do you think there's interest in moving forward with that?

alycm:
@oddhack Ideally some of the unified specification marking of deprecation will come from cl.xml, but I haven't looked into how to do that yet.  That would certainly require cl.xml to be both accurate and complete in recording what is deprecated and when.

PR #96, opened 2019-07-02
Correct the rendering of the phrase `C++`

alycm:
In some instances the phrase `C++` was rendered as `C`, due to the `++`
sometimes being considered as markup.  This change modifies all
instances to use a macro, whether they happen to currently render
incorrectly or not.


bashbaug:
LGTM.  Since this is a straightforward change I don't see any reason to wait to merge.

Issue #92, opened 2019-06-25
Create anchors for internal links that are missing them


alycm:
PR https://github.com/KhronosGroup/OpenCL-Docs/pull/75 introduced a significant amount of internal linking to the API specification.  A few of the links, however, have no anchor currently.  Generally this is because they are simple values such as `CL_TRUE` that aren't explicitly defined anywhere, or error codes that are defined for each API entry point that may use them.

A probably incomplete list of links without anchors:
* `CL_TRUE`
* `CL_FALSE`
* `CL_NONE`
* `CL_SOME_ERROR_CODE`

It should be possible to do some simple processing on the HTML output to find the list of internal links without a corresponding anchor.

PR #98, opened 2019-07-09
add error code table to appendix

bashbaug:
This change adds a table of error codes and their meanings as an appendix.  This should fix most of the "missing links" described in #92 - only a handful of enums such as `CL_NONE`, `CL_TRUE`, and `CL_FALSE` will be missing (that I know of!).

I've also renamed the files in the "api" directory from a .txt to a .asciidoc extension.  Because of the dictionaries and other generated code they don't render perfectly outside of the spec toolchain, but I still think this is better than nothing.


bashbaug:
Hmm, I wrote a quick python script to look for missing anchors and I found a few more missing anchors than I was expecting.

This PR _does_ fix a bunch of missing links, but I'll need to create another PR to fix the remaining issues.

api/appendix_f.asciidoc
```diff
@@ -0,0 +1,210 @@
...
+| {CL_INVALID_VALUE_anchor}
+| This is a generic error that is returned when a specified value is not a valid value.
+
+| {CL_INVALID_WORK_DIMENSION_anchor}
+| Returned by {clEnqueueNDRangeKernel} when the specified work dimension is not valid.
+
+| {CL_INVALID_WORK_GROUP_SIZE_anchor}
+| Returned by {clEnqueueNDRangeKernel} when the specified total work group size is not valid for the specified kernel or device.
+
+// This is a strange error code - replace by CL_INVALID_WORK_GROUP_SIZE?
```
alycm: It is strange.  Probably so named to match `CL_DEVICE_MAX_WORK_ITEM_SIZES` which is also a strange name.

The error codes probably could be combined with no disruption as the entry point that can return it already can return both.  It is currently well described when to use which code though.

No action required in this PR, I'm just commenting.

alycm:
Oh dear, I should have just merged this as soon as I reviewed rather than wait for further comments.   The file renames are automatic conflicts.

I can merge as soon as the conflicts are resolved.

bashbaug:
I've rebased and resolved the merge conflict.  Thanks!

PR #99, opened 2019-07-10
Fix refpage links not opening

### LINKED ISSUE #93: Links from 2.2 refpages don't open properly
Reported by Neil Trevett: When opening HTML links in the new OpenCL 2.2 refpages, they just result in a blank pane instead of the target document. Probably the right thing to do here is add target= attributes on the hyperlinks, so they open in a new tab. However, some people find this annoying, so raising the issue before deciding how to proceed.

oddhack:
This adds '^' to generated asciidoctor link: text, causing the resulting
HTML to open external hyperlinks in a new page (see https://github.com/asciidoctor/asciidoctor/issues/2071). While that works, there
are some related issues I'm exploring with landing on the wrong anchor
in the opened document when it's a CL specification, so marking this WIP
for the moment.

Closes #93 (once it's deWIPed).

### DIFF
--- a/man/copyright-ccby.txt
+++ b/man/copyright-ccby.txt
@@ -2,5 +2,5 @@
 
 Copyright (c) 2014-2019 Khronos Group. This work is licensed under a
 http://creativecommons.org/licenses/by/4.0/[Creative Commons
-Attribution 4.0 International License].
+Attribution 4.0 International License^].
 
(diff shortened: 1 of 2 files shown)


alycm:
The link changes here are fine, are the "related issues I'm exploring with landing on the wrong anchor" the issue described in https://github.com/KhronosGroup/OpenCL-Docs/issues/102 ?

oddhack:
> The link changes here are fine, are the "related issues I'm exploring with landing on the wrong anchor" the issue described in #102 ?

Yes. We could accept this at any time, although it won't have any visible effect until the spec HTML is updated and the refpages are regenerated in the registry.

PR #101, opened 2019-07-11
cl_khr_icd: reflect changes in definition of the dispatch table

kpet:
The dispatch table declaration no longer is an implementation
detail of the Khronos ICD loader and is now provided with the
OpenCL headers.


bashbaug:
LGTM - merging.

Issue #102, opened 2019-07-11
Update refpages to match latest API spec update


oddhack:
Working through #99 I realized that there seem to be a lot of missing HTML anchors in the 2.2 API spec in the registry, which AFAIK was put there in https://github.com/KhronosGroup/OpenCL-Registry/pull/60 . As a result, links from the refpages like https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#clGetDeviceIDs , which should open on the clGetDeviceIDs function, just open at the start of the spec.

I'm not sure what's caused this. The anchors come from the asciidoctor fragments built from cl.xml automatically, and I've no idea why they wouldn't propagate through to the output - maybe there was something odd about the toolchain, or some unusual options passed in the build? A spec I just built from OpenCL-Docs master branch doesn't suffer this problem. @mkinsner , I think this was your PR - do you recall anything unusual about generating the spec?

@bashbaug I think the best course of action would be to regenerate a spec from the last approved commit to OpenCL-Docs, ensure the anchors are present, and update the registry. Could you give that a try? If not, could you point me at the commit to do this from? Presumably current master, but I'm not keeping up with OpenCL WG activity so am unsure. Probably the other specs should be rebuilt at the same time, to be on the safe side.

mkinsner:
We don't think the anchors were merged as of the most recent maintenance release.  I have the action to push a new maintenance version, including the anchors.

oddhack:
Spec has been updated; need to update refpages now.

oddhack:
This is fixed. I had to push a couple of minor changes to the Makefile to properly regenerate the refpages - the renaming of files in api/ to .asciidoc wasn't reflected in the refpage extraction process.

PR #106, opened 2019-07-12
more API and enum link fixes

bashbaug:
This is a follow-on change to #98 that adds anchors for all API and enum links except for `CL_TRUE`, `CL_FALSE`, and `CL_NONE`.  I'll probably add another table as an appendix for these enums for completeness, but I'd like to get these changes merged first.

Most of the missing anchors were easy to add, but a few could use a more detailed review:

* `CL_MEM_KERNEL_READ_AND_WRITE`: The description for this enum was missing, so I added it.  As best I can tell, it is only used when querying enum formats via `clGetSupportedImageFormats`, and it is never used when creating a new memory object.
* Image Channel Orders: These were listed in the spec, but also didn't have a description, so I updated the "List of supported Image Channel Order Values" table to describe them.
* Sampler Properties: These were missing descriptions too, so I updated the "List of supported sampler creation properties" table to describe them.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -6032,7 +6032,7 @@ include::{generated}/api/protos/clGetProgramBuildInfo.txt[]
 
         {CL_BUILD_NONE_anchor} - The build status returned if no {clBuildProgram},
         {clCompileProgram} or {clLinkProgram} has been performed on the
-        specified _program_ object for _device).
+        specified _program_ object for _device_).
 
         {CL_BUILD_ERROR_anchor} - The build status returned if {clBuildProgram},
         {clCompileProgram} or {clLinkProgram} - whichever was performed last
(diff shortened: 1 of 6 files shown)


bashbaug:
Note that this PR includes a fix for a typo in the table of error codes (`CL_QUEUE_PROFILING_ENABLED` -> `CL_QUEUE_PROFILING_ENABLE`) that is currently causing this enum to appear unformatted in the spec and without a link.  If this PR is not accepted in the next day or so, I'd like to cherry pick this fix before the specs are rebuilt and posted (see #102, heads up @mkinsner ).  Thanks!

api/opencl_runtime_layer.asciidoc
```diff
@@ -5987,68 +6026,68 @@ include::{generated}/api/protos/clGetProgramBuildInfo.txt[]
...
       | Returns the build, compile or link status, whichever was performed
-        last on program for device.
+        last on the specified _program_ object for _device_.
 
         This can be one of the following:
 
         {CL_BUILD_NONE_anchor} - The build status returned if no {clBuildProgram},
         {clCompileProgram} or {clLinkProgram} has been performed on the
-        specified program object for device.
+        specified _program_ object for _device).
```
alycm: `_device_`

alycm: `_device_`

bashbaug: Fixed - thanks!

PR #110, opened 2019-07-18
Adjust an include to fix build

alycm:
A file got renamed so an include from a seperate folder needs updated.


mkinsner:
Confirmed that the fix works

PR #116, opened 2019-08-10
Add the missing static 'enums' refpage & fix most dangling links

oddhack:


oddhack:
@bashbaug could you review and sign off on this? Minimal changes to spec markup, but fixes a number of broken links in the refpages.

oddhack:
It's straightforward to add aliases - look for the [open] block for clGetDeviceInfo and add cl_device_info to the 'alias' attribute (or create it, if not already present). It might be necessary to edit the static part of the TOC (in man/toctail) depending on the page.

Generating (the majority of) the enums page would also be straightforward, though it would take a little time to write a new generator for the purpose. I don't think we could get the informative footnotes added automatically. It would probably still be useful to have an entry for cl_device_info in the enums page that links out.

Mixing and matching these two approaches would be more work, since whatever manual aliases were added, would have to be filtered out of the autogenerated aliases.

oddhack:
If you want me to pursue any of that, please open a tracking issue, as this discussion will go away once I accept the PR.

PR #118, opened 2019-08-15
allow parent devices to be passed to clRetainDevice and clReleaseDevice

### LINKED ISSUE #68: Contradiction in clRetainDevice and clReleaseDevice
The description of `clRetainDevice` says:

> clRetainDevice returns CL_SUCCESS if the function is executed successfully or the device is a root-level device.

But, in the list of error codes, it also says that it may return:

> CL_INVALID_DEVICE if device is not a valid sub-device created by a call to clCreateSubDevices.

`clReleaseDevice` is has a similar contradiction.  To resolve the contradiction I think the error condition should be removed, so it is valid to retain and release a root-level device.

bashbaug:
This removes the contradiction between the description of these APIs and the error condition for `CL_INVALID_DEVICE`.

I've also reworded a few other error conditions for `CL_INVALID_DEVICE` for consistency.

Fixes #68.

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1002,7 +1002,7 @@ device except for the following queries:
 successfully.
 Otherwise, it returns one of the following errors:
 
-  * {CL_INVALID_DEVICE} if _device_ is not valid.
+  * {CL_INVALID_DEVICE} if _device_ is not a valid device.
   * {CL_INVALID_VALUE} if _param_name_ is not one of the supported values or
     if size in bytes specified by _param_value_size_ is < size of return
     type as specified in the <<device-queries-table, Device Queries>> table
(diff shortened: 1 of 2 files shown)


alycm:
> Makes sense. This begs the question though: should we be thinking about centralising error codes and their description? Does asciidoctor allow macros to do that kind of things?

@bashbaug did add the [error code appendix](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#error_codes) very recently, so yes the error code descriptions can be centralized.  But I'm not sure fully centralizing them would be better, whether by macros or otherwise.  An error code may always mean the same thing, but there may be multiple reasons why that error code is returned.  In this PR, for example, there are two types of situations that result in `CL_INVALID_DEVICE`.

Would you like to create an issue about it and then we can merge this PR?

kpet:
That was a good change but not quite what I meant and as you say, things are a bit more complex than that in some cases. I didn't intend this to block this PR. I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/120 which should hopefully clarify what I meant.

PR #119, opened 2019-08-16
documentation for cl_khr_srgb_image_writes

### LINKED ISSUE #20: cl_khr_srgb_image_writes is documented inconsistently
I've found two possible issues relating to the `cl_khr_srgb_image_writes` extension:

1. This extension is currently documented in the OpenCL C spec, not in the OpenCL extensions spec, which seems strange.  I think it should be documented in the OpenCL extensions spec, instead.

1. This extension isn't currently documented in the SPIR-V environment spec and I think it should be.

bashbaug:
Fixes #20.

Removes description of cl_khr_srgb_image_writes from the language specs (OpenCL C and OpenCL C++), for consistency.  Instead, the behavior of this extension is described in the extensions spec, similar to all other extensions.

Also includes a few minor edits to the API spec that also refer to this extension.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -2043,9 +2043,8 @@ Kernel Read Or Write>> table.
     objects.
 
 10::
-    sRGB channel order support is not required for 1D image buffers.
-    Writes to images with sRGB channel orders requires device support of the
-    *cl_khr_srgb_image_writes* extension.
+    Read sRGB channel order support is optional for 1D image buffers.
+    Write sRGB channel order support is optional for all image types.
 
 For 1D, 1D image from buffer, 2D, 3D image objects, 1D and 2D image array
 objects, the mandated minimum list of image formats that can be read from
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -7022,15 +7022,8 @@ of the following values:
 
 The built-in image read functions will perform sRGB to linear RGB
 conversions if the image is an sRGB image.
-Writing to sRGB images from a kernel is an optional extension.
-The *cl_khr_srgb_image_writes* extension will be reported in the
-`CL_DEVICE_EXTENSIONS` string if a device supports writing to sRGB images
-using *write_imagef*.
-*clGetSupportedImageFormats* will return the supported sRGB images if
-`CL_MEM_READ_WRITE` or `CL_MEM_WRITE_ONLY` is specified in _flags_ argument
-and the device supports writing to an sRGB image.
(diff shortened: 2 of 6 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -2043,9 +2043,8 @@ Kernel Read Or Write>> table.
     objects.
 
 10::
-    sRGB channel order support is not required for 1D image buffers.
-    Writes to images with sRGB channel orders requires device support of the
-    *cl_khr_srgb_image_writes* extension.
+    Read sRGB channel order support is optional for 1D image buffers.
+    Write sRGB channel order support is optional for all image types.
```
alycm: I'm wondering about this footnote change.  It effectively promotes an extension to optional-core.  So previously the absense `cl_khr_srgb_image_writes` would tell the programmer something, but now an implementation could report support without reporting the extension.

If any code out their checks for the extension before using sRGB writes, and  a future implementation supports sRGB writes without reporting the extension then the "wrong" path will be taken.   The cross-product of those two "if" may be extremely small and I may be overthinking this.  I have no real examples of why this would matter.

I also can't think why sRGB writes would have been completely excluded before.   If this footnote had used your proposed wording all along it would make more sense.

alycm: I've made this comment because I followed your notes in https://github.com/KhronosGroup/OpenCL-Docs/issues/20, but I'm just want to confirm that we're okay promoting `cl_khr_srgb_image_writes` during maintenance.

bashbaug: Yeah, I remain confused by this extension. :smiley:

I believe the scenario we are discussing is:

* An implementation supports writes to sRGB images by including them in the set of writeable image formats returned by `clGetSupportedImageFormats`, but...
* The implementation does not report support for `cl_khr_srgb_image_writes`.

Is this valid or invalid?

Said another way, does an implementation **have to** support the `cl_khr_srgb_image_writes` extension if it supports writing to sRGB images?

I will check the CTS tests to see if it checks for this condition.

bashbaug:
I've created PR https://github.com/KhronosGroup/OpenCL-CTS/pull/449 to update the CTS tests to remove the check for `cl_khr_srgb_image_writes`, as we discussed in the last teleconference.

I think this change is ready to go.

api/opencl_runtime_layer.asciidoc
alycm: We discussed this on the teleconference, essentially if the image type is in the list of supported images (which it would be with this extension, and may be without) then it's fine.  Resolving.

Issue #74, opened 2019-04-25
Additional Requirements Beyond C99 TC2: Wording around NaN is confusing


mantognini:
https://github.com/KhronosGroup/OpenCL-Docs/blob/6163e510aba70daced330b7d7f54bc9b5b416252/OpenCL_C.txt#L9943-L9953

I think there are a few typos in this paragraph, specifically in the 3rd and 4th sentences. Could they simply be rewritten into:

> A *signaling* NaN shall be converted to a *non-signaling* NaN.

bashbaug:
I did some spec archaeology to try to figure out what sentences 3 and 4 _should_ say and was surprised to find that this text has been around since OpenCL 1.0, so it doesn't appear that it's just a typo that managed to sneak in somewhere along the line.

It took me a while to figure out what this was trying to say, but I think it's something like the following (using [RFC 2110](https://tools.ietf.org/pdf/rfc2119.pdf) terms):

1. Functions that return a NaN with more than one NaN operand **must** return one of the NaN operands.
1. A signaling NaN **may** be converted to a non-signaling NaN.
1. A non-signaling NaN **must** only be converted to a non-signaling NaN.
1. A signaling NaN **should** be converted to a non-signaling NaN, but **must** be converted to a NaN.
1. When a NaN is converted to another NaN, the sign of the NaN and the NaN payload bits **may** become undefined.

I don't think the 3rd and 4th sentences can be simplified into...

> A _signaling_ NaN shall be converted to a _non-signaling_ NaN.

... since the only requirement is that a signaling NaN remains a NaN, not that it must be converted to a non-signaling NaN.

All that said, is condition (1) really necessary?  Is the returned NaN subject to the conversion rules (2) - (5)?  I guess it would be needed if an implementation wanted to do: `f( sNaN, qNaN ) -> sNaN`?

Note that this text was inherited by the OpenCL SPIR-V environment spec and needs to be fixed there, too.

Tagging @b-sumner since he usually has good insight into issues like these.

b-sumner:
We have explicit (convert_*, (cast), etc.) and implicit conversions, and at least some of them are functions and it isn't clear which rules are supposed to apply to the explicit conversion functions.  I think we should make that clear.

Personally I am not a fan of the first rule (the only one currently addressing the behavior of built-in functions).   I see little value in imposing this burden on implementations.  Can we replace must with should?

bashbaug:
[...]

PR #128, opened 2019-09-06
reworded NaN behavior in Additional Requirements Beyond C99 TC2

bashbaug:
Possible fix for #74.

Changes the NaN description in "Additional Requirements Beyond C99 TC2" from:

> Functions that return a NaN with more than one NaN operand shall return one of the NaN operands. Functions that return a NaN operand may silence the NaN if it is a signaling NaN. A non-signaling NaN shall be converted to a non-signaling NaN. A signaling NaN shall be converted to a NaN, and should be converted to a non-signaling NaN. How the rest of the NaN payload bits or the sign of NaN is converted is undefined.

to:

> For functions returning NaN, a signaling NaN should be converted to a non-signaling NaN, but must remain a NaN. A non-signaling NaN may be converted to a different NaN, but must remain a non-signaling NaN. When one NaN is converted to a different NaN, the conversion may assign any value to the NaN sign bit and NaN payload bits.

Changes were made to the OpenCL C, OpenCL C++, and OpenCL SPIR-V Environment specs.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -10010,15 +10010,7 @@ except <<additional-requirements-beyond-c99-tc2,where noted below>>.
 [[additional-requirements-beyond-c99-tc2]]
 === Additional Requirements Beyond C99 TC2
 
-Functions that return a NaN with more than one NaN operand shall return one
-of the NaN operands.
-Functions that return a NaN operand may silence the NaN if it is a signaling
-NaN.
-A non-signaling NaN shall be converted to a non-signaling NaN.
-A signaling NaN shall be converted to a NaN, and should be converted to a
-non-signaling NaN.
-How the rest of the NaN payload bits or the sign of NaN is converted is
-undefined.
+All functions that return a NaN should return a quiet NaN.
 
 *half_<funcname>* functions behave identically to the function of the same
 name without the *half_* prefix.
--- a/cxx/numerical_compliance/edge_case_behavior.txt
+++ b/cxx/numerical_compliance/edge_case_behavior.txt
@@ -10,11 +10,7 @@ The edge case behavior of the math functions (see the <<math-functions, _Math Fu
 [[additional-requirements-beyond-isoiec-9899tc2]]
 ==== Additional Requirements Beyond ISO/IEC 9899:TC2
 
-Functions that return a NaN with more than one NaN operand shall return one of the NaN operands.
-Functions that return a NaN operand may silence the NaN if it is a signaling NaN.
-A non-signaling NaN shall be converted to a non-signaling NaN.
-A signaling NaN shall be converted to a NaN, and should be converted to a non-signaling NaN.
-How the rest of the NaN payload bits or the sign of NaN is converted is undefined.
+All functions that return a NaN should return a quiet NaN.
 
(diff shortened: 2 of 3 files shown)


bashbaug:
@b-sumner would you mind reviewing this proposed change?  Thanks!

b-sumner:
I think it would be clearer to talk about the return value required for specific argument values and avoid the use of "converted" completely.  E.g. "When the return value of a function given one or more NaN arguments is a NaN, then the return value should be a quiet NaN."  I believe this allows the return payload and sign to be arbitrary.

bashbaug:
Hi @b-sumner , it wasn't clear to me that the conversion rules only apply to functions returning a NaN operand or whether it applies to any function returning a NaN.  Is that the case - does this section only apply to functions returning a NaN operand?  If so, I agree it can be simplified.

b-sumner:
Right.  We need a general statement about the type of NaN a function shuld return.  And then a statement about what should happen if any of the arguments are NaN if not otherwise specified.

b-sumner:
Actually, I'm wondering if we want any more than this:

A function returning a NaN should return a quiet NaN.  

If we really want to say something about what happens with NaN arguments, I wouldn't want to go beyond:

Unless otherwise specified, a function returning a NaN result when given one or more NaN arguments should(? or may?) return a quieted copy of one of those NaN arguments.

bashbaug:
> A function returning a NaN should return a quiet NaN.

This is certainly simpler, which is attractive!  I also like that it doesn't discuss "converting" NaNs, which never felt quite right.

Thinking out loud, the cases that this debatably doesn't cover are:

1. The quiet NaN may be any quiet NaN, since any NaN sign bit or NaN payload bits are valid.
1. The function may return a quiet NaN if an operand is a signaling NaN, but the function should not return a signaling NaN when no operands are signaling NaNs.

I think these cases are already covered in the C99 specification, though.  Is that true?  If so, is this the only "Additional Requirements Beyond C99 TC2"?

Note that there is a "should" recommendation about NaN arguments in the C99 spec that is stronger than the suggestion above, so I don't think any additional text is needed in the OpenCL specs for it:

From C99 F.9, clause 13:

> If a function with one or more NaN arguments returns a NaN result, the result should be
the same as one of the NaN arguments (after possible type conversion), except perhaps
for the sign.

bashbaug:
I've switched to the much simpler text:

> Functions returning a NaN should return a quiet NaN.

@b-sumner, do you still think this is sufficient?  Thanks!

b-sumner:
There's a numbering mismatch between functions and "a" quiet NaN.  That's why I suggested "A function returning a NaN should return a quiet NaN."

bashbaug:
How about:

> All functions that return a NaN should return a quiet NaN.

I lean away from "a function" since it doesn't seem strong enough - the recommendation to return a quiet NaN applies to "all functions" and not to just "a function".  We're really splitting hairs here though, so if there's a preference for "a function" (or any more discussion, really) I'm fine with the "a function" wording, and I think it's an improvement over what is in the spec today.  Thanks!

b-sumner:
You're right, sorry for splitting hairs.  I'm fine with "Any function" instead of "A function" if you would prefer that.

bashbaug:
Great, sounds like we're all happy with:

> All functions that return a NaN should return a quiet NaN.

I think this is good to go.

PR #129, opened 2019-09-06
fix minor asciidoc formatting error in conversion built-ins

bashbaug:
Removes some asciidoctor formatting that slipped into a source block for the conversion built-in functions.


oddhack:
LGTM.

Issue #50, opened 2019-02-12
Variadic Macros Support for OpenCL


davidrohr:
I am wondering why the OpenCL standard forbids variadic macros. They are actually allowed by a couple of OpenCL implementations I have been using, but they are disabled in clang (yielding "error: variadic macros not supported in OpenCL")
I do not understand the reasoning behind disabling variadic macros, in particular since OpenCL compilers usually are based on A C / C++ frontent, which usually has a preprocessor that would actually support variadic macros.

I agree variadic macros do not belong to a modern coding style in the direction of C++20, but I have some code where they are widely used, in particular when you try to have common code segments for different GPU APIs, where you do not have access to the full C++ feature set.

Prime examples are
- Wrappers around printf with some additional checks / functionality (I agree this could also be done with variadic templates nowadays)
- Conditional disabling of parts of the code via defining #define DEBUG(...) to nothing, which is much cleaner than #ifdes everywhere in the code.
- At several places we create some code fragments automatically using macro string concattenation (##) and variadic macros.

My suggestion is to simply allow variadic macros in the code, since they do not increase compiler complexity anyway. At least, there could be an OpenCL extension or in case of clang a compile-time switch to allow variadic macros. My current workaround is to run the .cl file through the -std=c99 frontend of clang for the preprocessor phase and then I am giving the preprocessed file to clang again with -std=c++

bashbaug:
+1, I've never understood why this restriction exists, either.

At the very least it would be nice to support variadic macros via an offline compiler, say when compiling to SPIR-V.

I suspect supporting variadic macros via an online compiler would require an extension, just to formalize that some implementations support this functionality, while others don't.

AnastasiaStulova:
I don't think compiler actually knows if it's offline or online compilation. Also I wouldn't like the code rejected differently depending on that. I would just allow it unconditionally.

bashbaug:
> compiler actually knows if it's offline or online compilation.

I think this depends what "the compiler" is.  If it's a newer version of open source CLANG, I can see how it might not know if it is being used for offline or online compilation.  However, we have been shipping snapshots of CLANG with our drivers for years, which only get used for online compilation (meaning: for programs created using `clCreateProgramWithSource`) and report errors for kernels that use variadic macros:

````
1:1:15: error: variadic macros not supported in OpenCL
#define MAGIC(...)  5
              ^
````

[...]

PR #134, opened 2019-09-17
Added language extension doc and variadic macros ext

AnastasiaStulova:
This PR adds a new document to describe the language extensions and a folder for individual extensions (https://github.com/KhronosGroup/OpenCL-Docs/issues/65). 

There is also the first extension added to allow the variadic macros (https://github.com/KhronosGroup/OpenCL-Docs/issues/50).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,6 +260,30 @@ else
 	$(QUIET)mv $(PDFDIR)/$(EXTENSIONSSPEC)-optimized.pdf $@
 endif
 
+# Language Extensions spec
+EXTSPEC = OpenCL_LangExt
+EXTSPECSRC = $(EXTSPEC).txt $(GENDEPENDS) \
+    $(shell grep ^include:: $(EXTSPEC).txt | sed -e 's/^include:://' -e 's/\[\]/ /' | xargs echo)
+
+exthtml: $(HTMLDIR)/$(EXTSPEC).html $(EXTSPECSRC)
+
+$(HTMLDIR)/$(EXTSPEC).html: $(EXTSPECSRC) katexinst
+	$(QUIET)$(ASCIIDOCTOR) -b html5 $(ADOCOPTS) $(ADOCHTMLOPTS) -o $@ $(EXTSPEC).txt
+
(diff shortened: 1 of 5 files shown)


AnastasiaStulova:
I think I should add Acknowledgements section.

Also would it make sense to rename the doc:
Specification -> Documentation
to make it stronger that it's not a ratified formal document?

Makefile
```diff
@@ -260,6 +260,30 @@ else
 	$(QUIET)mv $(PDFDIR)/$(EXTENSIONSSPEC)-optimized.pdf $@
 endif
 
+# Language Extensions spec
+EXTSPEC = OpenCL_LangExt
```
bashbaug: As with the "C++ for OpenCL" document, this PR also needs different variables and makefile targets than "EXTSPEC", "exthtml", and "extpdf".

OpenCL_LangExt.txt
```diff
@@ -0,0 +1,37 @@
...
+:cpp: pass:[C++]
+:cpp14: pass:[C++14]
+
+// Various special / math symbols. This is easier to edit with than Unicode.
+include::config/attribs.txt[]
+
+// type of the source code in the document
+:language: {basebackend@docbook:c++:cpp}
+
+include::copyrights.txt[]
```
bashbaug: Confirming: This document shouldn't have the Khronos spec copyright since it is not a ratified Khronos specification?

langext/intro.txt
```diff
@@ -0,0 +1,33 @@
...
+
+[cols="5,4,2,4",options="header",]
+|====
+| *Extension Name*
+| *Feature macro*
+| *FM Value*
+| *OpenCL versions*
+
+| <<opencl_c_variadic_macro,Variadic Macros>>
+| __opencl_c_variadic_macro
```
bashbaug: Should the feature macro be named `__opencl_variadic_macro`, which leaves the door open for them being used in C++ for OpenCL as well?  Or is there a reason to prefer `__opencl_c_variadic_macro` explicitly?

bashbaug: OK, that works.  Have you given any thought what would happen if a language extension did apply to C++ for OpenCL as well?

PR #137, opened 2019-09-17
Fix typo + copyright year in XML header

Perksey:
Does what it says on the tin.


bashbaug:
Thanks for fixing these!

PR #136, opened 2019-09-17
update extension template

bashbaug:
* Adds a link to the OpenCL-Docs repo.
* Adds a link to the OpenCL specification registry.
* Adds a link to the OpenCL enumerant registry.
* Minor updates for the asciidoctor toolchain.

PR #139, opened 2019-10-02
A few small changes to the extension template

kpet:
Best rebased.


kpet:
I was on the fence since these were rather minor. I'll do that. I hadn't spotted #136, will review.

PR #140, opened 2019-10-03
Reserve 5 bits in cl_mem_flags for Arm

kpet:
Also mark the remaining bits in the upper 32-bit region as unused
(cl_mem_flags is defined as a cl_bitfield which is a cl_ulong,
guaranteed to be 64-bit).


bashbaug:
This is the first time that I know of where we will be assigning bits in the upper 32-bit region of a bitfield, but I think this should work fine.

@oddhack any concerns before merging?

oddhack:
Seems fine. If we eventually autogenerate headers we might need to tag the bitmask type as being 64 bits, but no need for that now, and no defined schema for it.

PR #152, opened 2019-10-25
Add a conformance tests section to the extension template

kpet:

### DIFF
--- a/extensions/cl_extension_template.asciidoc
+++ b/extensions/cl_extension_template.asciidoc
@@ -404,11 +404,9 @@ successive revisions of the specification.
 ****
 
 . The new *clShinyNewFunction* entrypoint must be called and succeed.
-. The value returned for +CL_FANCY_QUERY+ must be in the valid range.
-. Calling *clShinyNewFunction* twice with incompatible flags must fail
- - Call *clShinyNewFunction* with +CL_NEW_FLAG+ set, it must succeed
- - Call *clShinyNewFunction* again with +CL_NEW_INCOMPATIBLE_FLAG+ set, it must
-   fail
+. The value returned for `CL_FANCY_QUERY` must be in the valid range.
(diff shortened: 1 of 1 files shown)


extensions/cl_extension_template.asciidoc
```diff
@@ -387,6 +387,29 @@ extensions, or the optional interactions are documented as Modifications
...
+If it is not possible to provide any tests for the specification, a
+justification must be provided here.
+
+Test descriptions must be numbered and their order maintained across
+successive revisions of the specification.
+
+****
+
+. The new *clShinyNewFunction* entrypoint must be called and succeed.
+. The value returned for +CL_FANCY_QUERY+ must be in the valid range.
```
bashbaug: Minor: We should be recommending back-ticks instead of plus signs for mono-space formatting since plus signs only work for asciidoc but not for asciidoctor, but back-ticks work for both.

extensions/cl_extension_template.asciidoc
```diff
@@ -387,6 +387,29 @@ extensions, or the optional interactions are documented as Modifications
...
+. Calling *clShinyNewFunction* twice with incompatible flags must fail
+ - Call *clShinyNewFunction* with +CL_NEW_FLAG+ set, it must succeed
+ - Call *clShinyNewFunction* again with +CL_NEW_INCOMPATIBLE_FLAG+ set, it must
+   fail
```
bashbaug: I don't think we've traditionally done "negative testing" where we check for an error.  Do we want to recommend this?

extensions/cl_extension_template.asciidoc
kpet: Will do.

extensions/cl_extension_template.asciidoc
kpet: We do have some in the conformance tests (and probably should have more IMHO) but that's another discussion. I'll pick a less controversial example here.

PR #162, opened 2019-11-13
Fix minor typos related to sub_group_barrier

mantognini:
`subgroup_barrier` (with one underscore) was used instead of `sub_group_barrier` (with two underscores).


alycm:
Hmmm, this is now merged, this is the commit: https://github.com/KhronosGroup/OpenCL-Docs/commit/d9bc0f680390f63961302a1580c96374cd2dda4e

I had to do it manually, due to file rename from another PR, though I didn't have to resolve any conflicts at all, so I don't know why GitHub couldn't handle that.

I expected this PR to still close automatically, but clearly it hasn't.  So I'll close it myself, but again, this is merged!

PR #163, opened 2019-11-13
Clarify default value of program-scope/static variables

mantognini:
This patch clarifies that OpenCL program-scope/static variables in the
global address space have an appropriate null value by default,
similar to objects with static storage duration in C99 6.7.8p10.

The previous wording was not clear regarding the default value (is the
variable uninitialized?) when no constant expression is given. This
clarifies the intent.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -2063,8 +2063,8 @@ program.
 These variables are not shared across devices.
 They have distinct storage.
 
-Program scope and `static` variables in the `global` address space may be
-initialized, but only with constant expressions.
+Program scope and `static` variables in the `global` address space are zero
+initialized by default. A constant expression may be given as an initializer.
 
 Examples:
 


bashbaug:
I'm not sure that all implementations will zero-initialize their program scope variables by default.  The wording and clarification looks good, but I think we'd need to check what implementations are doing before accepting this change.

FWIW, I checked and we do have a CTS test for "uninitialized program scope global variables", but it is only checking that program scope global variables may be declared without an initializer, and it still executes a "writer" kernel that assigns to the program scope global variable before executing a "reader" kernel to check the contents of the program scope global variable.

https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/test_progvar.cpp#L1463

kpet:
I agree with @bashbaug. Strictly speaking, this imposes a new constraint on implementations that needs to be tested.

@mantognini Could you create a CTS issue and write a test please?

mantognini:
While some implementations might not zero-initialise program scope variable by default, I think OpenCL users expect those variables to be zero-initialised. The main argument for this clarification is to close the gap between C and OpenCL C and therefore make the latter more convenient to use.

I would argue that this potential difference between C and OpenCL is unnecessary. Clarifying it would also make C++ for OpenCL more intuitive.

As for the CTS, I think it would be better to first reach a consensus on the specification before taking any additional step.

kpet:
@mantognini I think we're all in agreement that it's a good change to make. The problem is that the behaviour wasn't really specified so there is a possibility that some implementations didn't zero-initialise program-scope variables by default. We just need to make sure we give these implementers a way of spotting the change (a.k.a. conformance tests) and generally manage the transition (and the approach to that might depend on whether there is a practical problem or not).

mantognini:
Alright then. I've created OpenCL-CTS#498 for now. I'll have a look at the code and try to create a patch for it.

kpet:
@bashbaug and others, we have a CTS test now (https://github.com/KhronosGroup/OpenCL-CTS/pull/544). How many implementations do we want to test that on before merging and changing the spec?

Jeremy-Kemp:
@mantognini thanks clarifying the initialisation wording and creating a CTS test.

I've run the new test on our implementation and am happy with it. @kpet @bashbaug would having 1 more implementation run the test be sufficient?

bashbaug:
Merging to keep the spec in sync with the CTS tests (https://github.com/KhronosGroup/OpenCL-CTS/pull/544).  Thanks!

PR #182, opened 2020-01-10
Increase the number of levels in the API spec table of contents

kpet:
3 levels is what we had prior to the move to asciidoc. This makes
the spec quite a bit more navigable.

### DIFF
--- a/OpenCL_API.txt
+++ b/OpenCL_API.txt
@@ -8,7 +8,7 @@ Khronos{R} OpenCL Working Group
 :data-uri:
 :icons: font
 :toc2:
-:toclevels: 2
+:toclevels: 3
 :max-width: 100
 :numbered:
 :imagewidth: 800


bashbaug:
This looks good to me and I agree it helps with spec navigation, but I'd like one more +1 before merging.

Observations:
* We may want to revisit a few section titles to remove redundancy after this change, such as the sub-sections of Chapter 3.
* The OpenCL C spec already uses three levels in its table of contents.
* The OpenCL SPIR-V Environment spec currently uses two levels in its table of contents.  This could be increased also, but a few section titles (such as the names of the supported extensions) are long and should be revisited.
* The OpenCL Extensions spec currently uses one level in its table of contents but will need a lot of updates to section titles, so I would not recommend making a change to it now.

kpet:
Yes, agree that some titles will need a bit of tweaking.

bashbaug:
Thanks for filing #184 to track the section titles.  Merging!

PR #185, opened 2020-01-14
Update copyright notices to 2020

alycm:
Note that I only updated Khronos copyright dates (primarily this means that I didn't update the dates in the Intel extensions as they are copyright Intel).


bashbaug:
Thanks for taking care of this!  Merging.

PR #188, opened 2020-01-16
fix a few extension issues  in the XML file

bashbaug:
I've found a few issues with the definitions of several extension functions in the cl.xml file.  The affected extensions are:

* Several of the DX interop extensions, e.g. `cl_khr_d3d10_sharing`.
* `cl_khr_create_command_queue`
* `cl_qcom_ext_host_ptr`
* `cl_arm_import_memory`


bashbaug:
Thanks for the prompt review - merging.

PR #192, opened 2020-01-23
Updates cl.xml for the cl_intel_unified_shared_memory extension

bashbaug:
These changes are consistent with the PR for the OpenCL headers, see: https://github.com/KhronosGroup/OpenCL-Headers/pull/65

The extension spec is a touch out-of-date on github, but can be found [here](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/USM/cl_intel_unified_shared_memory.asciidoc).  I will be updating it to the revision described by these changes (or newer) shortly.

I have scripts that use the API definitions so I am reasonably confident that they are correct.  I'm less confident regarding the typedefs and enum changes, but I have verified that the OpenCL specs still build correctly with these changes.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -188,6 +188,12 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_uint</type>          <name>cl_profiling_info</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_sampler_properties</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_kernel_exec_info</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_unified_shared_memory_capabilities_intel</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_properties_intel</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_alloc_flags_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_mem_info_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_unified_shared_memory_type_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_mem_advice_intel"/></name>;</type>
 
             <comment>Structure types</comment>
         <type category="struct" name="cl_dx9_surface_info_khr">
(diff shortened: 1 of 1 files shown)


Kerilk:
Hello,

In this PR you wrote the void keyword as <type>void</type>. Until now, in the xml file, the type markup was not used around void.

xml/cl.xml
```diff
@@ -188,6 +188,13 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_uint</type>          <name>cl_profiling_info</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_sampler_properties</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_kernel_exec_info</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_unified_shared_memory_capabilities_intel</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_properties_intel</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_alloc_flags_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_mem_info_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_unified_shared_memory_type_intel</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_migration_flags_intel</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_mem_advice_intel"/></name>;</type>
```
Kerilk: You have a trailing "/ at the end of cl_mem_advice_intel

bashbaug: Fixed - thanks!

bashbaug:
@oddhack, can you provide advice when the `<type>` markup should be used, and specifically whether it should be used around `void`?  I've removed the type markup around `void` for my changes, since that's consistent with most usage of `void` in the XML file, but there are still a few other places where the type markup is still used.  Thanks!

Kerilk:
Not exactly in your pull request, but maybe not worth doing another, there is an incoherence here:
https://github.com/bashbaug/OpenCL-Docs/blob/8dc3f950951c0642c54334808da2eabfc0db500e/xml/cl.xml#L1631-L1635
```xml
           <unused start="0x4170" end="0x417F"/>
```
should be, unless I am mistaken,
```xml
           <unused start="0x4170" end="0x417D"/>
```

oddhack:
> @oddhack, can you provide advice when the `<type>` markup should be used, and specifically whether it should be used around `void`?

We use it consistently in the Vulkan and XR XML now. It is not necessary for C header generation and our other internal uses, though, because there's no actual dependency to satisfy to define the type before use. That is the purpose of the \<type> tag. So I'd say it's your call to use it or not, but best to be consistent so the question doesn't keep coming up.

bashbaug:
@Kerilk thanks, I fixed the inconsistency in the 0x4170 enum range also.

I'd like to address the `<type>` markup inconsistency as a separate PR though.  I've filed #196 to track this.

Kerilk:
@bashbaug You're welcome. I deleted the comment because I have a whole list of them and it will really be out of the scope of the pull request.

bashbaug:
Just FYI, I have the type markup changes done to address #196, but they build upon this change, so I'd like to merge this first.

xml/cl.xml
Kerilk: You forgot the to remove the ">". Sorry I didn't catch that earlier.

bashbaug: I've added this to the "more fixes" PR #196.

Issue #150, opened 2019-10-24
Simplify ULP Definitions for Geometric Functions


bashbaug:
Pull request #43 defined ULP definitions for geometric functions, and pull request #142 fixed a few asciidoc quirks in the way the ULP definitions needed to be formatted.

The ULP definitions are a little complicated though, for example:

Function | ULP
---------| ---
`distance` | 3 + (1.5 * n) + (0.5 * (n - 1)) ulp, for gentype with vector width n

Is there any reason this cannot be simplified?

```
3 + (1.5 * n) + (0.5 * (n - 1))
= 3 + 1.5n + (0.5n - 0.5)
= 2.5 + 2n
```

This example uses `distance`, but I think this could apply to many (if not all) of the geometric functions.

EwanC:
Think presenting the final simplified ULP is defintely more readible for users.

As a side-effect I'm wondering if it's worth also record somewhere in this repo how we arrived at these expressions for our own reference. Currently the working is in CTS comments, which adds a level
of indirection which could go stale. As in the case of my [!43](https://github.com/KhronosGroup/OpenCL-Docs/pull/43) MR links which point to
the cl22_trunk CTS branch which no longer exists.

Could copy the CTS comments into asciidoc comments, e.g for `distance()`.
```c
 float maxUlps = 3.0f +                              // error in sqrt
        ( 1.5f * (float) sizes[size] +      // cumulative error for multiplications  (a-b+0.5ulp)**2 = (a-b)**2 + a*0.5ulp + b*0.5 ulp + 0.5 ulp for multiplication
         0.5f * (float) (sizes[size]-1));    // cumulative error for additions
```

```
////
3           ULP error in sqrt
1.5 * n     cumulative error for multiplications (a-b+0.5ulp)**2 = (a-b)**2 + a*0.5ulp + b*0.5 ulp + 0.5 ulp for multiplication
0.5 * (n-1) cumulative error for additions

= 3 + (1.5 * n) + (0.5 * (n - 1))
= 3 + 1.5n + (0.5n - 0.5)
= 2.5 + 2n
////
| *distance*            | {leq} 2.5 + 2n ulp, for gentype with vector width n

```

bashbaug:
Fixed by #194 - closing.

PR #194, opened 2020-01-24
Simplify Geometric builtin ULP definitions

EwanC:
Condensed the ULP formulas used to define the ULP bounds of geometric builtin functions, and included comments on the derivation taken from the CTS.

Also fixed a bug in the `dot()` error where I had misread the [CTS code](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/geometrics/test_geometrics.cpp#L432) as `(2 * (n - 1))` when it should've been `2n - 1`.

Addresses [Github Issue 150](https://github.com/KhronosGroup/OpenCL-Docs/issues/150)

Used lines comments in table rather than block comments due to [Asciidoctor bug](https://github.com/asciidoctor/asciidoctor/issues/1681)

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9509,10 +9509,20 @@ currently undefined.
 | *cos*                 | {leq} 4 ulp
 | *cosh*                | {leq} 4 ulp
 | *cospi*               | {leq} 4 ulp
-| *cross*               | absolute error tolerance of 'max * max * (3 * FLT_EPSILON)', where max is the maximum input operand value
+// 3 operations from the 2 multiplications and 1 subtraction per component
+| *cross*               | absolute error tolerance of 'max * max * (3 * FLT_EPSILON)' per vector component, where _max_ is the maximum input operand magnitude
 | *degrees*             | {leq} 2 ulp
-| *distance*            | {leq} 3 + (1.5 * n) + (0.5 * (n - 1)) ulp, for gentype with vector width n
-| *dot*                 | absolute error tolerance of 'max * max * (2 * (n - 1)) * FLT_EPSILON', For vector width n and maximum input operand value 'max'
+// 3           ULP error in sqrt
+// 1.5 * n     cumulative error for multiplications
+// 0.5 * (n-1) cumulative error for additions
+//
+// = 3 + (1.5 * n) + (0.5 * (n - 1))
+// = 3 + 1.5n + (0.5n - 0.5)
+// = 2.5 + 2n
+| *distance*            | {leq} 2.5 + 2n ulp, for gentype with vector width _n_
+// n + n-1  Number of operations from n multiples and (n-1) additions
+// 2n - 1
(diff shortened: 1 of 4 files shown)


OpenCL_C.txt
```diff
@@ -9588,9 +9613,33 @@ currently undefined.
...
+// = 8192 + 1.5n + (0.5n - 0.5)
+// = 8191.5 + 2n
+| *fast_distance*       | {leq} 8191.5 + 2n ulp, for gentype with vector width _n_
+
+// 8192        ULP error in half_sqrt
+// 0.5 * n     cumulative error for multiplications
+// 0.5 * (n-1) cumulative error for additions
+//
+// = 8192 + (0.5 * n) + (0.5 * (n - 1))
+// = 8192.5 + 0.5n + (0.5n - 0.5)
```
alycm:
```suggestion
// = 8192 + 0.5n + (0.5n - 0.5)
```

ext/cl_khr_fp16.asciidoc
```diff
@@ -1491,20 +1491,30 @@ is the infinitely precise result.
...
 | *cross*
-| absolute error tolerance of 'max * max * (3 * HLF_EPSILON)', where max is the maximum input operand value
+| absolute error tolerance of 'max * max * (3 * HLF_EPSILON)' per vector component, where _max_ is the maximum input operand magnitude
 | Implementation-defined
 
 | *degrees*
 | \<= 2 ulp
 | \<= 2 ulp
 
+// 0.5         ULP error in sqrt
```
alycm: Presumably the equation here was derived from the CTS, but fp16 `sqrt` is "correctly rounded" for FULL profile.

OpenCL_C.txt
EwanC: nice catch, fixed

ext/cl_khr_fp16.asciidoc
EwanC: So I've been assuming that "correctly rounded" incurs a 0.5 ULP cost based on multiply/add/subtract operations being "correctly rounded" in the spec and allotted this value in other geometric error calculations. This assumption may be wrong though...

For example [single precision normalize](https://github.com/KhronosGroup/OpenCL-CTS/blob/ef832c330c95196e4c8a8de0d95a87f2eb8da967/test_conformance/geometrics/test_geometrics.cpp#L1055) has 2.5 ULP assigned for the 2 ULP error in rsqrt plus the error in a multiply.

alycm: Yes, that makes sense.

bashbaug:
I followed the simplification logic in the OpenCL C spec and this all looks good.  Thank you for addressing this!

The error bounds for `cross` and `dot` are a little strange, but the descriptions in this PR are consistent with the previous descriptions and the CTS tests.

I'll wait another day or so just in case anyone else wants to review and approve this, then merge.

PR #197, opened 2020-01-28
use type markup consistently in the XML file

### LINKED ISSUE #193: Missing Spec Description for conversion of half to image_channel_type: CL_UNORM_INT8 / CL_UNORM_INT16 / CL_SNORM_INT8 / CL_SNORM_INT16.
Section 8.3 Image Addressing and Filtering: Conversion Rules.
The section on conversion of image write(or read) data to (or from) image_channel_types gives details of conversion to int/float/uint datatypes but not for half datatypes. It should be a part of the extension documentation.
It can be extrapolated from the conversion rules for float, but it should be explicitly given.
Reference: https://www.khronos.org/registry/OpenCL/specs/2.2/pdf/OpenCL_Ext.pdf
and https://www.khronos.org/registry/OpenCL/specs/2.2/pdf/OpenCL_C.pdf

bashbaug:
This PR contains a few more fixes to the XML file, and fixes #193.  There should be no functional changes.

* Use the `<type>` markup consistently in the XML file.
* One minor fix for the name of an argument for clCreateFromGLsyncKHR.
* Fixed, added, or edited a few enum block comments for consistency.

The one place where the `<type>` markup is not currently used is for function pointers, e.g.:

```
<param>void (CL_CALLBACK*                      <name>pfn_free_func</name>)(cl_command_queue, cl_uint, void *[], void *)</param>
```

I think if we wanted to fix this we'd do so by creating a typedef for each function pointer type, then referencing the new type using the `<type>` markup.  If we decide to go this route I'd prefer to make this change as part of a separate PR.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -4475,8 +4475,10 @@ server's OpenCL/api-docs repository.
             <require>
                 <type name="CL/cl.h"/>
             </require>
-            <require comment="Interop tokens">
+            <require comment="cl_device_info">
                 <enum name="CL_DEVICE_TERMINATE_CAPABILITY_KHR"/>
+            </require>
+            <require comment="cl_context_properties">
                 <enum name="CL_CONTEXT_TERMINATE_KHR"/>
             </require>
             <require>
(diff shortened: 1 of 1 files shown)


Kerilk:
As you bring up callbacks, in addition to using the type markup, maybe being consistent about having the parameter names would be beneficial. Some callback in the xml are defined as (line 3108):
```xml
<param>void (CL_CALLBACK* <name>pfn_free_func</name>)(cl_command_queue queue, cl_uint num_svm_pointers, void* svm_pointers[], void* user_data)</param>
```
In other they are defined as (line 2435):
```
<param>void (CL_CALLBACK* <name>pfn_notify</name>)(const char*, const void*, size_t, void*)</param>
```
In the doc the names are always defined. I would advocate for having the parameters name.

bashbaug:
Thanks, I added argument names to the function pointer callbacks, consistent with the headers.

I did find one discrepancy between the headers and the spec and filed #198 to track it.

Issue #201, opened 2020-01-29
Required image formats in embedded profile


kpet:
The unified specification doesn't document any required image format for implementations that support the embedded profile. The OpenCL 1.2 specification did have a list of required formats.

kpet:
I've opened #202 to cover read_only and write_only images, taking the list of formats from the OpenCL 1.2 specification.

What about read_write images? This doesn't seem to be optional in the embedded profile. Should we at least require one format? We could require the intersection of the list of required formats for read_write images in the full profile and the list of required formats for the embedded profile.

Jeremy-Kemp:
I think starting with the intersection of the required embedded profile formats with the required read_write formats for full profile sounds like a good direction to go in. 

We probably wouldn't want to require additional formats in read_write for embedded over what is required for read _or_ write.

bashbaug:
I just merged #202.  I'm keeping this issue open though, until we've resolved the issue regarding read_write formats.

bashbaug:
@kpet @Jeremy-Kemp , I think we fixed this as part of the OpenCL 3.0 spec changes.  The latest API spec says:

> For embedded profiles devices that support reading from and writing to the same image object from the same kernel instance (see CL_​DEVICE_​MAX_​READ_​WRITE_​IMAGE_​ARGS) there is no required minimum list of supported image formats.

Do you think that is sufficient to close this issue, or are additional changes needed?  Thanks!

Jeremy-Kemp:
Yes, I think we can close this as we already addressed it. Thanks!

bashbaug:
Great, thanks for confirming.  Closing!

PR #202, opened 2020-01-29
Document required formats for {read,write}_only images in embedded profile

kpet:
Adapt the wording from the unified spec to allow for optionality.

Required formats for read_write images aren't covered by this change.

Contributes to #201.

### DIFF
--- a/api/embedded_profile.asciidoc
+++ b/api/embedded_profile.asciidoc
@@ -305,3 +305,30 @@ Device Queries>> table is {CL_TRUE}, the values assigned to
 must be greater than or equal to the minimum values specified in the
 <<embedded-device-queries-table, OpenCL Embedded Device Queries>> table.
 
+For 1D, 1D image from buffer, 2D, 3D image objects, 1D and 2D image array
+objects, when supported, the mandated minimum list of image formats that can
+be read from and written to by different kernel instances when correctly ordered
+by event dependencies and that must be supported by all devices that support
+images is described in the <<embedded-required-image-read-or-write-formats-table>> table.
+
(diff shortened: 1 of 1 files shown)


api/embedded_profile.asciidoc
```diff
@@ -305,3 +305,30 @@ Device Queries>> table is {CL_TRUE}, the values assigned to
...
+  | {CL_RGBA}
+      | {CL_UNORM_INT8} +
+        {CL_UNORM_INT16} +
+        {CL_SIGNED_INT8} +
+        {CL_SIGNED_INT16} +
+        {CL_SIGNED_INT32} +
+        {CL_UNSIGNED_INT8} +
+        {CL_UNSIGNED_INT16} +
+        {CL_UNSIGNED_INT32} +
+        {CL_HALF_FLOAT}^1^ +
```
bashbaug: Is this footnote required?

I don't see a similar footnote in the OpenCL 1.2 spec or for the full profile, and it feels strange to me to tie the `CL_HALF_FLOAT` image storage format to the `cl_khr_fp16` extension that is otherwise all about computation on the fp16 ("`half`") type.

I'd prefer to either remove the footnote or remove the `CL_HALF_FLOAT` image format from the minimum list of required formats unconditionally.

kpet: It's actually not required. I'll remove it.

kpet: BTW, removing `CL_HALF_FLOAT` from the list of formats wouldn't have been an option since the format was required by OpenCL 1.2.

bashbaug: Thanks!  Latest changes LGTM.

PR #203, opened 2020-01-29
Add Vendor ID enum for devices without a PCIe ID

EwanC:
For [CL_DEVICE_VENDOR_ID](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_DEVICE_VENDOR_ID) the OpenCL spec says
> A unique device vendor identifier. An example of a unique device identifier could be the PCIe ID.

However as an ISV Codeplay doesn't have a PCIe ID to base this on. To solve this problem we've used [Vulkan vendor ID](https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#extensions-vendor-id) as a reference, to introduce an enum which ISVs can reserve above the threshold of largest PCI vendor ID (0x10000).

To stay in-sync with vk.xml Codeplay would like to reserve `0x10004`, see https://github.com/KhronosGroup/Vulkan-Docs/pull/1168

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -318,7 +318,7 @@ include::{generated}/api/version-notes/CL_DEVICE_TYPE.asciidoc[]
         {CL_DEVICE_TYPE_CPU}, {CL_DEVICE_TYPE_GPU}, {CL_DEVICE_TYPE_ACCELERATOR},
         {CL_DEVICE_TYPE_DEFAULT}, a combination of the above types, or
         {CL_DEVICE_TYPE_CUSTOM}.
-| {CL_DEVICE_VENDOR_ID_anchor}
+| {CL_DEVICE_VENDOR_ID_anchor}^4^
 
 include::{generated}/api/version-notes/CL_DEVICE_VENDOR_ID.asciidoc[]
   | cl_uint
@@ -444,7 +444,7 @@ include::{generated}/api/version-notes/CL_DEVICE_IMAGE_SUPPORT.asciidoc[]
   | cl_bool
       | Is {CL_TRUE} if images are supported by the OpenCL device and {CL_FALSE}
         otherwise.
-| {CL_DEVICE_MAX_READ_IMAGE_ARGS_anchor}^4^
+| {CL_DEVICE_MAX_READ_IMAGE_ARGS_anchor}^5^
 
 include::{generated}/api/version-notes/CL_DEVICE_MAX_READ_IMAGE_ARGS.asciidoc[]
   | cl_uint
@@ -460,7 +460,7 @@ include::{generated}/api/version-notes/CL_DEVICE_MAX_WRITE_IMAGE_ARGS.asciidoc[]
         write_only qualifier.
         The minimum value is 64 if {CL_DEVICE_IMAGE_SUPPORT} is {CL_TRUE}, the
         value is 0 otherwise.
-| {CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS_anchor}^5^
+| {CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS_anchor}^6^
 
 include::{generated}/api/version-notes/CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS.asciidoc[]
   | cl_uint
@@ -612,7 +612,7 @@ include::{generated}/api/version-notes/CL_DEVICE_MIN_DATA_TYPE_ALIGN_SIZE.asciid
       | The minimum value is the size (in bytes) of the largest OpenCL data
         type supported by the device (`long16` in FULL profile, `long16` or
         `int16` in EMBEDDED profile).
-| {CL_DEVICE_SINGLE_FP_CONFIG_anchor}^6^
+| {CL_DEVICE_SINGLE_FP_CONFIG_anchor}^7^
 
 include::{generated}/api/version-notes/CL_DEVICE_SINGLE_FP_CONFIG.asciidoc[]
   | cl_device_fp_config
@@ -647,7 +647,7 @@ include::{generated}/api/version-notes/CL_DEVICE_SINGLE_FP_CONFIG.asciidoc[]
         For the embedded profile, see the
         <<embedded-profile-single-fp-config-requirements, dedicated table>>.
 
-| {CL_DEVICE_DOUBLE_FP_CONFIG_anchor}^7^
+| {CL_DEVICE_DOUBLE_FP_CONFIG_anchor}^8^
 
 include::{generated}/api/version-notes/CL_DEVICE_DOUBLE_FP_CONFIG.asciidoc[]
 Also see extension *cl_khr_fp64*.
@@ -926,7 +926,7 @@ include::{generated}/api/version-notes/CL_DRIVER_VERSION.asciidoc[]
   | char[]
       | OpenCL software driver version string.
         Follows a vendor-specific format.
-| {CL_DEVICE_PROFILE_anchor}^8^
+| {CL_DEVICE_PROFILE_anchor}^9^
 
 include::{generated}/api/version-notes/CL_DEVICE_PROFILE.asciidoc[]
   | char[]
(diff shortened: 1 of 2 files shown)


EwanC:
Do I need to create a matching PR for https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl.h ?

xml/cl.xml
```diff
@@ -992,6 +992,12 @@ server's OpenCL/api-docs repository.
         <enum bitpos="0"            name="CL_MEM_ALLOC_WRITE_COMBINED_INTEL"/>
     </enums>
 
+    <enums start="0x10000" end="0x1FFFF" name="cl_khronos_vendor_id" vendor="Khronos">
+        <unused start="0x10000" end="0x10004" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
```
alycm:
```suggestion
        <unused start="0x10000" end="0x10003" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
```

api/opencl_platform_layer.asciidoc
```diff
@@ -1154,6 +1154,21 @@ include::{generated}/api/version-notes/CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_P
 |====
 
 4::
+    OpenCL adopters must report a valid vendor ID for their implementation.
+    If there is no valid PCI vendor ID defined for the physical device,
+    implementations must obtain a Khronos vendor ID. To reserve a Khronos vendor
+    ID, propose a merge request against cl.xml in the master branch of the
+    OpenCL-Docs project. The merge must define a new enumerant by adding an
```
oddhack: Do you want to say something about the awkward "reserve in vk.xml and mirror to cl.xml" strategy here?

oddhack:
> I don't think this needs to go in the headers, none of the PCIe IDs are in there.

We thought that for Vulkan too, but we were wrong, and did get enough requests to put them in.

Since the vk.xml reservation has already been merged, albeit won't be published for a week or so, this seems fine to put in now.

api/opencl_platform_layer.asciidoc
EwanC: Good point, part of this awkwardness is that it won't be appropriate to reserve the new cl.xml vendor ID in vk.xml before mirroring if a vendor doesn't implement Vulkan.

Added these sentences in a new commit:

> To keep Khronos vendor IDs synchronized across standards, the proposed ID value should not collide with the ID of another vendor already reserved in vk.xml, from the master branch of the Vulkan-Docs project. If an adopter has already reserved a Khronos vendor ID in vk.xml, then that value should be duplicated in cl.xml.

EwanC:
> 
> 
> > I don't think this needs to go in the headers, none of the PCIe IDs are in there.
> 
> We thought that for Vulkan too, but we were wrong, and did get enough requests to put them in.
> 
> Since the vk.xml reservation has already been merged, albeit won't be published for a week or so, this seems fine to put in now.

https://github.com/KhronosGroup/OpenCL-Headers/pull/71

xml/cl.xml
```diff
@@ -1004,6 +1004,12 @@ server's OpenCL/api-docs repository.
         <enum bitpos="0"            name="CL_MEM_ALLOC_WRITE_COMBINED_INTEL"/>
     </enums>
 
+    <enums start="0x10000" end="0x1FFFF" name="cl_khronos_vendor_id" vendor="Khronos">
```
alycm: I've just realized that `cl_khronos_vendor_id` isn't a defined in the spec (or this XML file), because a `cl_uint` is returned for this query.

Could put `cl_uint` here and that doesn't affect the current spec, but all the other enums have a type.  Adding a type in a maintenance spec release seems a bit sneaky though.

EwanC: To avoid any risk I made `cl_khronos_vendor_id` the comment attribute and `cl_uint` the name attribute. Based on the [schema](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/xml/registry.rnc#L176) for `<enums>` that seems valid, didn't update the spec language but think it still makes sense.

xml/cl.xml
```diff
@@ -1016,6 +1016,12 @@ server's OpenCL/api-docs repository.
         <enum bitpos="0"            name="CL_MEM_ALLOC_WRITE_COMBINED_INTEL"/>
     </enums>
 
+    <enums start="0x10000" end="0x1FFFF" name="cl_uint" vendor="Khronos" comment="cl_khronos_vendor_id">
+        <unused start="0x10000" end="0x10003" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
+        <enum value="0x10004" name="CL_KHRONOS_VENDOR_ID_CODEPLAY" comment="Codeplay Software Ltd. Matches Khronos Vendor ID from vk.xml"/>
+        <unused start="0x10005" end="0x1FFFF" comment="This is the next unused available Khronos vendor ID"/>
```
oddhack: Per previous comment, I'd like to see one of two things happen: alter the process to require reservation in vk.xml first in all cases, or split this into a range for shared Vulkan-OpenCL assignment (say, 0x10000 - 0x100FF) and a range for OpenCL-only assignment (say, 0x10100-0x101FF), with a comment noting that reservations in the former range must be coordinated with vk.xml and reservations in the latter range may not be usable in other Khronos APIs. Something like this:

```
    <enums start="0x10000" end="0x100FF" name="cl_uint" vendor="Khronos" comment="cl_khronos_vendor_id - shared Vulkan IDs which must be reserved first in Vulkan's vk.xml">
        <enum value="0x10004" name="CL_KHRONOS_VENDOR_ID_CODEPLAY" comment="Codeplay Software Ltd. Matches Khronos Vendor ID from vk.xml"/>
    </enums>
    <enums start="0x10100" end="0x101FF" name="cl_uint" vendor="Khronos" comment="cl_khronos_vendor_id - OpenCL-only IDs">
        <unused start="0x10100" end="0x101FF" comment="Vendor IDs in this range can be reserved here and are usable only with OpenCL implementations"/>
    </enums>
```
If neither of these things is done, eventually someone will reserve a CL vendor ID which conflicts with a Vulkan vendor ID for another vendor.

api/opencl_platform_layer.asciidoc
EwanC: That is a situation which seems avoidable, thanks for the possible solutions. 

In favour of splitting the ranges since it mitigates the cross repository interaction, and requiring reservation in vk.xml first feels like it violates [this Vulkan spec wording](https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#extensions-vendor-id)

> Please do not try to reserve vendor IDs unless you are making a good faith effort to develop a Vulkan implementation and require one for that purpose.

Will update this PR to reflect that.

EwanC: Actually, after some discussion internally we're going to revisit mandating the id be reserved in vk.xml and come back to this PR

oddhack: I think we could alter the Vulkan spec wording you quote, to allow for a global Khronos vendor ID. The intent wasn't to prevent useful cross-API functionality.

xml/cl.xml
EwanC: Related PR https://github.com/KhronosGroup/OpenCL-Docs/pull/235

api/opencl_platform_layer.asciidoc
oddhack: We've accepted an MR internally that allows reserving vendor IDs for other Khronos APIs in vk.xml, and that will go out in the next spec update, next week. Since that's locked now, this can move forward - we plan to publish a spec update next Tuesday (03/17).

xml/cl.xml
EwanC: updated to use expeimental enum from [235](https://github.com/KhronosGroup/OpenCL-Docs/pull/235) and headers [MR 75](https://github.com/KhronosGroup/OpenCL-Headers/pull/75)

api/opencl_platform_layer.asciidoc
EwanC: great thanks, pushed new language to reflect that

bashbaug:
@alycm, is there more to discuss here, or are these changes ready to be merged?  I don't want to merge them too early and I see you just added the "agenda" label.  Thanks!

alycm:
@bashbaug Nothing more to discuss, we're happy with the wording.  Happy for this to merge any time.

oddhack:
@bashbaug @alycm the changes in the Vulkan spec allowing this reservation system have been pushed out in the 1.2.135 spec update. Thanks for working through the process!

PR #209, opened 2020-02-06
Further annotations for unified API spec

alycm:


api/embedded_profile.asciidoc
```diff
@@ -145,7 +149,7 @@ For {CL_UNORM_INT_101010_2}
   * 3 must convert to 1.0f (for A)
 --
 
-The following optional extensions defined in the OpenCL 2.2 Extension
+The following optional extensions defined in the OpenCL Extension
```
bashbaug: Do we need to call this list out explicitly?  Earlier, this section said that:

> The optional extensions defined in the OpenCL Extension Specification apply to both profiles.

alycm: You're right, the `cl_khr_xxx` extensions do not need to be listed (or limited to this list).  The `cles_khr_xxx` notes should remain somewhere though.  I'll change this.

alycm: I've cut this list completely, and moved the "doubles implies support for longs" comment elsewhere.

PR #212, opened 2020-02-07
Minor cleanups.

alycm:
Just some things I spotted while working on other text.

### DIFF
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -66,7 +66,7 @@ implemented with OpenCL.
 image::images/platform_model.png[align="center", title="Platform Model ... one host plus one or more compute devices each with one or more compute units composed of one or more processing elements."]
 
 Programmers provide programs in the form of SPIR-V source binaries, OpenCL C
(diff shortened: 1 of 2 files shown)


api/opencl_architecture.asciidoc
```diff
@@ -312,7 +312,7 @@ Each work-items global ID is an N-dimensional tuple.
 The global ID components are values in the range from F, to F plus the
 number of elements in that dimension minus one.
 
-If a kernel is created from OpenCL 2.0 or SPIR-V, the size of work-groups in
+If a kernel is created from OpenCL C 2.0 or SPIR-V, the size of work-groups in
```
bashbaug: This sentence is more correct now than it was before, but it still feels incomplete.  Presumably kernels created from implementation-defined binary objects may also support non-uniform work groups, at least in some cases?

alycm: Yes, that makes sense.  Likewise I see no reason why a builtin kernel couldn't support non-uniform work groups.

I've flipped the wording to

> Unless a kernel comes from a source that disallows it, e.g. OpenCL C 1.x or using `-cl-uniform-work-group-size`, ...

PR #215, opened 2020-02-11
Remove monkey patch to asciidoctor-pdf

kpet:
The fix has made it to 1.5.0 and the patch doesn't work with
newer versions of ruby.


bashbaug:
Probably best for @oddhack to review this one.

oddhack:
The difficulty with the asciidoctor toolchain has always been these alpha-status version dependencies. If they've fixed this particular problem that's great, but then the README should also be updated to require the corresponding version of asciidoctor-pdf. Once you've done that and verified the PDF build looks fine, I think @bashbaug should just go ahead and accept it - I don't have any additional value to add.

I'm putting together a Docker image which has a known good asciidoctor toolchain for building the Vulkan spec (Ryan has already done this for OpenXR), and plan to work around all the installation stuff by changing our README from 5 pages of OS-dependent instructions to "either use the Docker image to build, or install the equivalents in your environment based on the Dockerfile - we're out of the business of trying to tell you the fine details". If that works out, OpenCL could make use of it too.

PR #216, opened 2020-02-11
Update sectnumoffset-treeprocessor.rb

kpet:
...so that it works with newer versions of ruby.

### DIFF
--- a/config/sectnumoffset-treeprocessor.rb
+++ b/config/sectnumoffset-treeprocessor.rb
@@ -1,8 +1,8 @@
 # This extension script is taken from the Asciidoctor extensions lab [1], under
 # the MIT license [2].
 #
-# [1] https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/18bdf628f46d9c98920c20413d8d99bf5ea622a8/lib/sectnumoffset-treeprocessor.rb
-# [2] https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/18bdf628f46d9c98920c20413d8d99bf5ea622a8/LICENSE.adoc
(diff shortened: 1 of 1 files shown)


bashbaug:
Probably best for @oddhack to review this one.

oddhack:
Sorry, missed the review comment but I wouldn't have had much to add anyway. I can't assess whether this breaks compatibility with older versions or not; if it doesn't obviously no problem, if it does you could just update the required minimum versions in the build docs.

bashbaug:
For the travis builds (which we should really merge, even if we don't merge the automatic deployments) I'm currently using:

```yaml
rvm:
  - 2.3.3

before_install:
  - sudo apt-get install -y libpango1.0-dev ghostscript
  - gem install asciidoctor -v 1.5.8
  - gem install coderay json-schema
  - gem install ttfunk -v 1.5.1
  - gem install asciidoctor-pdf -v 1.5.0.beta.8 --pre
  - MATHEMATICAL_SKIP_STRDUP=1 gem install asciidoctor-mathematical -v 0.2.2
```

A few of these packages are a touch newer than the versions described in the README.

@kpet, how does this compare to the versions you are using?  Should we just update the README with these versions?

kpet:
I'm using:
- Ruby 2.5.7p206
- asciidoctor 2.0.10
- asciidoctor-mathematical 0.3.1
- asciidoctor-pdf 1.5.0
- ttfunk 1.5.1

I suggest we merge the Travis config and update the README to the versions used by the Travis config (we should also make sure we fix all the versions there) before merging this. I don't think there is a single distribution that provides all the packages needed and everyone is presumably using gem so I don't think we should worry too much about bumping the required versions.

bashbaug:
FYI, I've created PR https://github.com/KhronosGroup/OpenCL-Docs/pull/233 with the Travis config and README updates.

PR #218, opened 2020-02-12
Specification for cl_khr_extended_versioning

kpet:


kpet:
Here's a first attempt at getting the specification for `cl_khr_extended_versioning` into the main extension specification. A few things I think we should discuss:
- Individual extension specifications are currently grouped by the OpenCL version required. This means chapter numbers change as we insert extensions.
- I've added a section with metadata about the extension. I'd like to do the same for all other extensions (at the very least add version information) but before I do that, I'd rather get a bit of feedback on the approach to reduce back and forth on the whole document.

bashbaug:
I have a few style opinions:

* I think we should eliminate any notion of tying an extension to an OpenCL version and just add the extensions to the document in numerical order.  The specification text has already eliminated this, I think, and the only reference to specific OpenCL versions is in the asciidoc source.  This will keep chapter numbers consistent as extensions are added.

* The only exception we may want to consider are the extensions to the SPIR-V environment, which we may want to either keep as the last chapter, or could consider moving to an Appendix.

* I like some of the extension metadata, such as the version of the OpenCL specification it was written against, but some of the other metadata is debatably extraneous.  A good example is the "issues", I think: this is useful information, but it is purely informational, so does it belong in the spec?  As a compromise, could it be conditionally compiled into the spec (disabled by default)?

* Could we move the "contributors" to the main spec acknowledgements vs. including it for each individual extension?

* At least in the common case for ratified and final extensions, could we omit the "extension status"?

A few other comments:

* Please add this extension to the "quick reference table" in quick_reference.asciidoc.

* We purposefully haven't included definitions for the enum values for prior KHR extensions.  I do see the value in including them, and we include them in vendor extensions, but if we're going to add them we should do so consistently.

In general though, this all looks very good.  Thanks!

kpet:
Picking this up again.

>I think we should eliminate any notion of tying an extension to an OpenCL version and just add the extensions to the document in numerical order.

Agree, and this change has now been done in another PR.

> [...] some of the other metadata is debatably extraneous. A good example is the "issues", I think: this is useful information, but it is purely informational, so does it belong in the spec? As a compromise, could it be conditionally compiled into the spec (disabled by default)?

I think it's useful info to answer the "what-were-they-thinking-when-they-designed-this type questions". I take your point though and I agree this may end up taking quite a bit of space in the spec. Given that we maintain the extension spec as a separate document, I'm not sure it's too big a deal. It seems to be working fine for Vulkan (these are in the appendix where extensions are described).

>Could we move the "contributors" to the main spec acknowledgements vs. including it for each individual extension?

I disagree with this. I find it very useful to be able to know what companies/people were behind specific Vulkan extensions for example.

>At least in the common case for ratified and final extensions, could we omit the "extension status"?

Yup, makes sense. I'll remove that.

>Please add this extension to the "quick reference table" in quick_reference.asciidoc.

Will do.

>We purposefully haven't included definitions for the enum values for prior KHR extensions. I do see the value in including them, and we include them in vendor extensions, but if we're going to add them we should do so consistently.

I think we should always do it. It's our best bet to guard against enum conflicts.

kpet:
@bashbaug I think this is ready for another round of review when you have time.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,286 @@
...
+from the individual fields.
+
+[source,c]
+----
+
+typedef cl_uint cl_version_khr;
+
+#define CL_VERSION_MAJOR_BITS_KHR (10)
+#define CL_VERSION_MINOR_BITS_KHR (10)
+#define CL_VERSION_PATCH_BITS_KHR (12)
```
alycm, line 67: For the main specification these values are in the XML as constants, this could be too.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,286 @@
...
+----
+
+==== Name and version
+
+This extension adds a structure that can be used to describe a combination of a
+name alongside a version number:
+
+[source,c]
+----
+#define CL_NAME_VERSION_MAX_NAME_SIZE_KHR 64
```
alycm, line 92: For the main specification this value is in the XML as a constant, this could be too.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,286 @@
...
+
+// NOTA: the structure below must be packed so that the OpenCL implementation
+//       and application use the same layout in memory.
+typedef struct _cl_name_version_khr {
+    cl_version_khr version;
+    char name[CL_NAME_VERSION_MAX_NAME_SIZE_KHR];
+} cl_name_version_khr;
```
alycm, line 99: And again, this struct type could be in the XML.

alycm (commented):
Sorry I didn't come back to this sooner.  One real open question here for me.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,286 @@
...
+==== Name and version
+
+This extension adds a structure that can be used to describe a combination of a
+name alongside a version number:
+
+[source,c]
+----
+#define CL_NAME_VERSION_MAX_NAME_SIZE_KHR 64
+
+// NOTA: the structure below must be packed so that the OpenCL implementation
```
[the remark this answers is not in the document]
kpet: That was short for Nota Bene in Latin :). Since we decided to remove packing, I'll remove that line altogether.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,286 @@
...
+| Return Type
+| Description
+
+| `CL_DEVICE_NUMERIC_VERSION_KHR`
+| `cl_version_khr`
+| Returns detailed (major, minor, patch) numeric version information. The major
+  and minor version numbers returned must match those returned via
+  `CL_DEVICE_VERSION`.
+
+| `CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR`
```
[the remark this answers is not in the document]
kpet: I think it's too late for this. We've ratified it as is and it has shipped in multiple software releases I know of.

ext/cl_khr_extended_versioning.asciidoc
kpet: Yup, done.

ext/cl_khr_extended_versioning.asciidoc
kpet: Done

ext/cl_khr_extended_versioning.asciidoc
kpet: Done

kpet:
@alycm Thanks for the review. I've rebased it and addressed your comments.

ext/cl_khr_extended_versioning.asciidoc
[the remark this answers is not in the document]
alycm: That's unfortunate, but yes, it sounds too late.

kpet:
@alycm @bashbaug I've rebased this again. I would appreciate if we could get it out of the way :).

OpenCL_Ext.txt
```diff
@@ -48,6 +48,7 @@ include::ext/cl_khr_int64_atomics.asciidoc[]
 include::ext/cl_khr_select_fprounding_mode.asciidoc[]
 include::ext/cl_khr_gl_sharing__context.asciidoc[]
 include::ext/cl_khr_gl_sharing__memobjs.asciidoc[]
+include::ext/cl_khr_extended_versioning.asciidoc[]
```
bashbaug, line 51: I'd recommend adding this extension to the end of the doc, after `cl_khr_device_uuid`, to keep reasonably consistent chapter numbers.

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,284 @@
...
+for applications to process strings to test for the presence of an extension or
+intermediate language or built-in kernel.
+
+=== General information
+
+==== Name Strings
+
+`cl_khr_extended_versioning`
+
+==== Contributors
```
bashbaug, line 22: See previous comment regarding extension metadata like contributors, issues, etc.  We haven't included this information in other KHR extensions.  Is this a practice we want to continue into the future?

ext/cl_khr_extended_versioning.asciidoc
```diff
@@ -0,0 +1,284 @@
...
+`CL_NAME_VERSION_MAX_NAME_SIZE_KHR - 1`.
+
+=== New API Enums
+
+Accepted value for the _param_name_ parameter to *clGetPlatformInfo*:
+
+[source,c]
+----
+CL_PLATFORM_NUMERIC_VERSION_KHR         0x0906
+CL_PLATFORM_EXTENSIONS_WITH_VERSION_KHR 0x0907
```
bashbaug, line 111: See previous comment regarding including enum values in the extension spec.

ext/quick_reference.asciidoc
```diff
@@ -169,4 +169,8 @@
 | Create Command Queues with Different Throttle Policies
 | Extension
 
+| <<cl_khr_extended_versioning,cl_khr_extended_versioning>>
+| Extend versioning of platform, devices, extensions, etc.
+| Extension
```
bashbaug, line 174: Please keep this table in alphabetical order.

xml/cl.xml
```diff
@@ -5718,6 +5736,28 @@ server's OpenCL/api-docs repository.
             <require comment="cl_channel_type">
                 <enum name="CL_UNORM_INT16"/>
                 <enum name="CL_FLOAT"/>
+        </extension>
```
bashbaug, line 5739: There is a missing XML tag before this line (`</require>`).

xml/cl.xml
```diff
@@ -5718,6 +5736,28 @@ server's OpenCL/api-docs repository.
...
+        <extension name="cl_khr_extended_versioning" supported="opencl">
+            <require comment="Types">
+                <type name="cl_version_khr"/>
+                <type name="cl_name_version_khr"/>
+            </require>
+            <require comment="Constants">
+                <enum name="CL_NAME_VERSION_MAX_NAME_SIZE_KHR"/>
```
bashbaug, line 5746: Since the `cl_name_version_khr` structure depends on the `CL_NAME_VERSION_MAX_NAME_SIZE_KHR` define I think it will make code generation simpler if this "Constants" section is swapped with the preceding "Types" section.

bashbaug (changes requested):
Could you please have one last look at some of the style comments I left previously -  https://github.com/KhronosGroup/OpenCL-Docs/pull/218#issuecomment-586492421?

Some of these are admittedly subjective, but this PR is doing things a bit inconsistently with the ways extensions have been documented in the past and I would like to be sure we are purposefully deviating.

I've also left a few specific comments inline.  Thanks!

OpenCL_Ext.txt
kpet: Good point. I can't think of a reason why I added it there (probably came out of a conflict resolution).

ext/quick_reference.asciidoc
kpet: Hadn't spotted the ordering. Done.

xml/cl.xml
kpet: Done.

xml/cl.xml
kpet: Not sure we should be relying on that in code generation but I don't mind.

ext/cl_khr_extended_versioning.asciidoc
kpet: I think my position hasn't changed here and I think we should do that for all extensions.

ext/cl_khr_extended_versioning.asciidoc
kpet: I'm all for consistency. Are you proposing we don't add them here and do it all later or keep them here and create issues to follow up?

bashbaug: Either could work but I have a slight preference for this so the spec sources stay consistent:

> we don't add them here and do it all later

I can't find the commit ID off-hand but I think I removed extension enum assignments from a handful of extensions previously.

kpet: Ok, I've removed all of them.

bashbaug (approved):
Thanks, all my issues are resolved.

Let's see if we can merge this one in today's teleconference.

bashbaug:
Merging as discussed in the November 10th teleconference.

PR #227, opened 2020-02-25
Clarified clSetKernelArg _arg_size value for memory objects

raunraun:


api/opencl_runtime_layer.asciidoc
```diff
@@ -6985,8 +6985,8 @@ include::{generated}/api/version-notes/clSetKernelArg.asciidoc[]
     leftmost argument to _n_ - 1, where _n_ is the total number of arguments
     declared by a kernel (see below).
   * _arg_size_ specifies the size of the argument value.
-    If the argument is a memory object, the size is the size of the memory
-    object.
+    If the argument is a memory object, the _arg_size_ value must be equal to
```
kpet: s/the _arg_size_ value/_arg_size_/?

raunraun: I think I copied and pasted from the 2 lines below (cl_sampler, cl_command_queue) so we're consistent throughout the section now.  Is changing all instances of that wording any more clear?

kpet: I hadn't spotted the other instances (out of the default diff view). I don't think it hurts clarity really but I find what I proposed a bit easier to read. I'm equally happy for this change to go in unmodified (what you improved is way more important).

PR #231, opened 2020-02-28
Define ULP overflow leniency

EwanC:
Add wording clarifying the current CTS behaviour permitting arithmetic operations
to overflow to infinity within ULP threshold error of a finite reference.

Motivated by CTS PR https://github.com/KhronosGroup/OpenCL-CTS/pull/600

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -11022,6 +11022,11 @@ The reference value used to compute the ULP value of an arithmetic operation
 is the infinitely precise result.
 0 ulp is used for math functions that do not require rounding.
 
+Result overflow within the specified ULP error is permitted. Math functions are
+allowed to return infinity for a finite reference value when the next
+floating-point number that would be representable after the finite maximum, if
+there was sufficient range, meets ULP error tolerance.
+
 [[table-ulp-float-math]]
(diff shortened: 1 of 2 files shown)


bashbaug (commented):
This makes sense to me.  Would you mind making a similar change to the OpenCL SPIR-V environment spec?  It has a similar "relative error as ULPs" section.  Thanks!

jlewis-austin:
I might be reading it wrong, but the second sentence looks incomplete. Suggest replacing "Allowing math
functions" with "Math functions are allowed"

EwanC:
Thanks for the review @bashbaug & @jlewis-austin, I've updated the PR with "Math functions are allowed ..." wording and duplicated it in the SPIR-V environment spec.

Only thing I'm unsure about is whether it should be "Math **instructions**" instead in the SPIR-V environment spec?

EwanC:
> Only thing I'm unsure about is whether it should be "Math **instructions**" instead in the SPIR-V environment spec?

Pushed a new commit updating the wording to **instructions**

jlewis-austin:
LGTM

bashbaug (approved):
LGTM!

PR #233, opened 2020-02-28
add travis continuous integration for spec builds

bashbaug:
This change adds Travis continuous integration to verify that the specs build correctly and updates the README to reflect the package versions used by the Travis build.  This change does NOT add automatic deployment of the built specs, but this could be added at a later date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,29 +18,4 @@ before_install:
 
 script:
   - git describe --tags --dirty
-  # This produces an error because travis-ci works with a "detached head".
-  #- git symbolic-ref --short HEAD
   - make api c env ext
-
-#after_success:
-#  - cd out
-#  - git init
-#  - git config user.name "${GH_USER_NAME}"
-#  - git config user.email "${GH_USER_EMAIL}"
-#  - git add html pdf ; git commit -m "Deploy to Github Pages"
-#  - git push -f "https://${GH_TOKEN}@${GH_REF}" master:gh-pages
-
-after_success:
-  - mkdir deploy
-  - cp -R out/html out/pdf deploy
-
-deploy:
-  verbose: true
-  provider: pages
-  local_dir: deploy
-  skip_cleanup: true
-  github_token: $GH_TOKEN
-  #keep_history: true
-  on:
-    #branch: master
-    branch: travis-ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ rvm:
 before_install:
   - sudo apt-get install -y libpango1.0-dev ghostscript
   - gem install asciidoctor -v 1.5.8
-  - gem install coderay json-schema
+  - gem install coderay -v 1.1.1
   - gem install ttfunk -v 1.5.1
   - gem install asciidoctor-pdf -v 1.5.0.beta.8 --pre
   - MATHEMATICAL_SKIP_STRDUP=1 gem install asciidoctor-mathematical -v 0.2.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,46 @@
+dist: xenial
+
+language: ruby
+
+git:
+  depth: 200
+
+rvm:
+  - 2.3.3
+
+before_install:
+  - sudo apt-get install -y libpango1.0-dev ghostscript
+  - gem install asciidoctor -v 1.5.8
+  - gem install coderay json-schema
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+git:
+  depth: 200
+
+rvm:
+  - 2.3.3
+
+before_install:
+  - sudo apt-get install -y libpango1.0-dev ghostscript
+  - gem install asciidoctor -v 1.5.8
+  - gem install coderay json-schema
```
kpet: Do we want fixed versions there are well?

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+  - gem install ttfunk -v 1.5.1
+  - gem install asciidoctor-pdf -v 1.5.0.beta.8 --pre
```
kpet: asciidoctor-pdf 1.5.0 does exit now. Should we move to that instead of using a beta version?

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+  - MATHEMATICAL_SKIP_STRDUP=1 gem install asciidoctor-mathematical -v 0.2.2
+
+script:
+  - git describe --tags --dirty
+  - make api c env ext
```
kpet: Should we add C++ for OpenCL there as well?

.travis.yml
bashbaug: I'm pretty flexible about this, but I'd lean towards using the latest packages as long as they are working fine.  Let me know what you prefer.

.travis.yml
bashbaug: We will need to use version 1.5.0.beta.8 at least until #215 is merged, due to an issue with the asciidoctor-pdf-monkeypatch script.  If we merge #215 first, I can look at switching to 1.5.0 as part of this PR.

.travis.yml
bashbaug: Yes, I'll add this.  Actually, I think we can just run `make`, and pick up everything by default?

.travis.yml
bashbaug: I added the explicit versioning for coderay.  For json-schema, it actually seems that it isn't required, so I've removed it completely.

.travis.yml
bashbaug: I looked at some options here:

* Just running `make` doesn't do anything, so that's not an option.
* Running `make all` builds more, but also doesn't include C++ for OpenCL.

I'm inclined to leave this as-is for now, and we can always update the build targets for the CI builds later.

bashbaug:
I think we're in a bit of a deadlock right now with these changes and the changes in #215 and #216.  @kpet, any suggestions to break the deadlock?  Should we just pick one to go first?  If we merge #215 (and maybe #216) first, the main change I'll make to this PR is to try the non-beta Asciidoctor v1.5.0 vs. v1.5.0.beta.8.

.travis.yml
kpet: 1.5.0 comes after 1.5.0.beta.8 ;)

.travis.yml
kpet: Yes, let's keep what you have, it's better than nothing and easy to build on top of.

kpet:
I'm not sure I'm following. Doesn't 1.5.0 come after 1.5.0.beta.8? My understanding is that we can just move to 1.5.0 which I think still works with the other PRs and then merge those PRs. I haven't tested this though. TBH, if we're fine with the lot we can merge all three of them in quick succession with no practical adverse consequences.

bashbaug:
Sorry, I probably didn't explain that very well.

[Here](https://travis-ci.org/github/bashbaug/OpenCL-Docs/builds/662051227?utm_medium=notification&utm_source=github_status) is a build that uses asciidoctor v1.5.0 and fails.  I _think_ merging #215 and/or #216 will fix it, but I'm not 100% sure.

Let's do this: merge #215 and #216.  That should "just work", with the existing toolchain, including the one in this PR.  Assuming it does, I'll try asciidoctor v1.5.0, and if it works I'll update it and the README so it doesn't use the pre-release v1.5.0.beta.8.  Sound good?

bashbaug:
OK, I've rebased after merging #215 and #216 and asciidoctor-pdf v1.5.0 works now.  Cool!  I've updated the travis script and README to reflect this.  I think these changes are ready to go.  Thanks!

PR #236, opened 2020-03-11
Document difference of C++ for OpenCL wrt OpenCL C

AnastasiaStulova:
Added a new section into C++ for OpenCL documentation with initial description of the behavior that is different to OpenCL C.

### DIFF
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -1,0 +1,129 @@
+// Copyright 2019-2020 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+=== Difference to OpenCL C
+
+{cpp} for OpenCL aims to achieve backwards compatibility with OpenCL C
+for the majority of features. However, it is a different
+language that is derived from {cpp} and therefore it inherits
+its fundamental design principles. Hence {cpp} for OpenCL
+deviates from OpenCL C in the same areas where {cpp} deviates
+from C.
+
+This section describes the main differences between {cpp} for
+OpenCL and OpenCL C.
+
+==== Implicit Conversions
+
+{cpp} is much stricter about conversions between types,
+especially those that are performed implicitly by the compiler.
+For example it is not possible to convert a `const` object
+to non-`const` implicitly. For details please refer to {cpp}17
+(`ISO/IEC 14882:2017`) `[conv]`.
+
+[source,c]
+----------
+void foo(){
+  const int *ptrconst;
+  int *ptr = ptrconst; // invalid initialization discards const qualifier
+}
+----------
+
+The same applies to narrowing conversions in
+initialization lists ({cpp}17 `[dcl.init.list]`). 
+
+[source,c]
+----------
+struct mytype {
+ int i;
+};
+void foo(uint par){
+  mytype var = {
+    .i = par // narrowing from uint to int is disallowed
+  };
+}
+----------
+
+Some compilers allow silencing this error using a flag
+(e.g. in Clang `-Wno-error=c++11-narrowing` can be used).
+
+Among other common conversions that will not be compiled in C++
+mode there are pointer to integer or integer to pointer
+type conversions.
+
+[source,c]
+----------
+void foo(){
+  int *ptr;
+  int i = ptr; // incompatible pointer to integer conversion
+}
+----------
+
+==== Null Literal
+
+In C and OpenCL C the null literal is defined using other
+language features as it is not represented explicitly i.e.
+commonly it is defined as
+
+[source,c]
+----------
+#define NULL (void*)0
+----------
+
+In {cpp} there is an explicit builtin literal `nullptr`
+that should be used instead ({cpp}17 `[lex.nullptr]`).
+
+{cpp} for OpenCL does not define `NULL` and therefore any
+source code using it must be modified to use `nullptr`
+instead. However as a workaround to avoid large modifications
+`NULL` can also be defined/aliased to `nullptr` in custom
+headers or using command line flag `-D`. It is not recommended
+to reuse the C definition of `NULL` in {cpp} for OpenCL as it may
+cause compilation failures in cases that work for C.
+
+[source,c]
+----------
+void foo(){
+ int *ptr = NULL; // invalid initialization of int* with void*
+}
+----------
+
+==== Use of restrict
+
+{cpp}17 does not support `restrict` and therefore {cpp} for OpenCL
+can not support it either. Some compilers might provide extensions
+with some functionality of `restrict` in {cpp}, e.g. `__restrict`
+in Clang.
+
+This feature only affects optimizations and the source
+code can be modified by removing it. As a workaround to avoid manual
+modifications macro substitutions can be used to either remove the
+keyword during the preprocessing by defining `restrict` as an empty
+macro or mapping it to another similar compiler features, e.g.
+`__restrict` in Clang. This can be done in headers or using `-D`
+compilation flag.
+
+==== Limitations of goto statements
+
+{cpp} is more restrictive with respect to entering the scope of
+variables than C. It is not possible to jump forward over a variable
+declaration statement apart from some exceptions detailed in {cpp}17
+`[stmt.dcl]`.
+
+[source,c]
+----------
+if (cond)
+  goto label;
+int n = foo();
+label:  // invalid: jumping forward over declaration of n
+   // ...
+----------
+
+==== Use of Clang Blocks
+
+Clang Blocks that are defined by the Objective-C language are not supported
+and their use can be replaced by lambdas ({cpp}17 `[expr.prim.lambda]`).
+
+This means that builtin functions using blocks, such as `enqueue_kernel`,
+are not supported in {cpp} for OpenCL.
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -98,7 +98,7 @@ in Clang.
 
 This feature only affects optimizations and the source
 code can be modified by removing it. As a workaround to avoid manual
-modifications macro substitutions can be used to either remove the
+modifications, macro substitutions can be used to either remove the
 keyword during the preprocessing by defining `restrict` as an empty
 macro or mapping it to another similar compiler features, e.g.
 `__restrict` in Clang. This can be done in headers or using `-D`
--- a/CXX_for_OpenCL.txt
+++ b/CXX_for_OpenCL.txt
@@ -32,7 +32,15 @@ include::copyrights-ccby.txt[]
 
 include::cxx4opencl/intro.txt[]
 
+<<<<
+
 include::cxx4opencl/restrictions.txt[]
+
+<<<<
+
+include::cxx4opencl/diff2openclc.txt[]
+
+<<<<
 
 include::cxx4opencl/address_spaces.txt[]
 


cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
+=== Difference to OpenCL C
+
+{cpp} for OpenCL targets backwards compatibility to OpenCL C
```
svenvh:
```suggestion
{cpp} for OpenCL aims to achieve backwards compatibility with OpenCL C
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+in the majority of features. However, it is a different
```
svenvh:
```suggestion
for the majority of features. However, it is a different
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+language that is derived from {cpp} and therefore it inherits
+its fundamental design principles. Hence {cpp} for OpenCL will
```
svenvh:
```suggestion
its fundamental design principles. Hence {cpp} for OpenCL
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+deviate from OpenCL C in the same areas where {cpp} deviates
```
svenvh:
```suggestion
deviates from OpenCL C in the same areas where {cpp} deviates
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+from C.
+
+In this section some of the most common areas where it is
+expected to behave differently are described.
```
svenvh:
```suggestion
This section describes the main differences between {cpp} for
OpenCL and OpenCL C.
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+
+==== Implicit Conversions
+
+{cpp} is much more strict about conversions between types
```
svenvh:
```suggestion
{cpp} is much stricter about conversions between types,
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+especially those that are performed implicitly by the compiler.
+For example it is not possible to convert non-`const` object
```
svenvh:
```suggestion
For example it is not possible to convert a non-`const` object
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+to `const` explicitly. For details please refer to {cpp}17
+(`ISO/IEC 14882:2017`) `[conv]`.
+
+[source,c]
+----------
+void foo(){
+  const int* ptrconst;
+  int* ptr = ptrconst; // rejected in C++ for OpenCL
```
svenvh:
```suggestion
  const int *ptrconst;
  int *ptr = ptrconst; // rejected in C++ for OpenCL
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+}
+----------
+
+The same situation applies to narrowing conversions in
```
svenvh:
```suggestion
The same applies to narrowing conversions in
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+initialization lists ({cpp}17 `[dcl.init.list]`). 
+
+[source,c]
+----------
+struct mytype {
+ int i;
+};
+void foo(uint par){
+  mytype var = {
+    .i = par
```
svenvh: It would help if you add a comment to highlight the issue.

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+  };
+}
+----------
+
+Note that some compilers will allow to silence this error
+using some flags (e. g. in Clang `-Wno-error=c++11-narrowing`
+can be used).
```
svenvh:
```suggestion
Some compilers allow silencing this error using a flag
(e.g. in Clang `-Wno-error=c++11-narrowing` can be used).
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+
+Among other common conversions that won't be compiled in C++
```
svenvh:
```suggestion
Among other common conversions that will not be compiled in C++
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+mode there are pointer to integers or integer to pointer
+conversions.
+
+[source,c]
+----------
+void foo(){
+  int* ptr;
```
svenvh:
```suggestion
  int *ptr;
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+  int i = ptr;
```
svenvh: Again, would be useful to highlight the issue with a comment.

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+}
+----------
+
+==== Null Literal
+
+In C and OpenCL C null literal is being defined using other
```
svenvh:
```suggestion
In C and OpenCL C the null literal is defined using other
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+
+In {cpp} there is an explicit builtin literal `nullptr`
+that should be used instead ({cpp}17 `[lex.nullptr]`).
+
+{cpp} for OpenCL does not define `NULL` and therefore any
+source code using it must be modified to use `nullptr`
+instead. However as a workaround to avoid large modifications
+`NULL` can also be defined/aliased to `nullptr` in custom
+headers or using command line flag `-D`. It is not recommended
+to reuse C definition of `NULL` in {cpp} for OpenCL as it may
```
svenvh:
```suggestion
to reuse the C definition of `NULL` in {cpp} for OpenCL as it may
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+cause compilation failures in cases that work for C.
+
+[source,c]
+----------
+void foo(){
+ int* ptr = NULL; // inivalid initialization of int* with void*
```
svenvh:
```suggestion
 int *ptr = NULL; // invalid initialization of int* with void*
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+}
+----------
+
+==== Use of restrict
+
+C++17 does not support `restrict` and therefore {cpp} for OpenCL
+can not support it either. Some compilers might provide extensions
+with some functionality of `restrict` in {cpp} e. g. `__restrict`
```
svenvh:
```suggestion
with some functionality of `restrict` in {cpp}, e.g. `__restrict`
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+
+In this section some of the most common areas where it is
+expected to behave differently are described.
+
+==== Implicit Conversions
+
+{cpp} is much more strict about conversions between types
+especially those that are performed implicitly by the compiler.
+For example it is not possible to convert non-`const` object
+to `const` explicitly. For details please refer to {cpp}17
```
mantognini: I believe the terms in this sentence are swapped.

```suggestion
For example it is not possible to convert a `const` object
to non-`const` implicitly. For details please refer to {cpp}17
```

I'm also suggesting using _implicitly_ as it might be possible to use `const_cast` for this in some specific cases (which is outside the scope of this doc, I think).

cxx4opencl/diff2openclc.txt
mantognini: I would suggest adding an example such as `(int2)(4.2f, 3.14f)` as well.

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+}
+----------
+
+Note that some compilers will allow to silence this error
+using some flags (e. g. in Clang `-Wno-error=c++11-narrowing`
+can be used).
+
+Among other common conversions that won't be compiled in C++
+mode there are pointer to integers or integer to pointer
+conversions.
```
mantognini:
```suggestion
mode there are pointer to integer or integer to pointer
type conversions.
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+[source,c]
+----------
+void foo(){
+ int* ptr = NULL; // inivalid initialization of int* with void*
+}
+----------
+
+==== Use of restrict
+
+C++17 does not support `restrict` and therefore {cpp} for OpenCL
```
mantognini: I'm not familiar with the markup language used in this file, so maybe it's all fine. I've noticed you use `{cpp}` everywhere but here you use `C++17`. Should this be replaced by `{cpp}17` or something?

```suggestion
{cpp}17 does not support `restrict` and therefore {cpp} for OpenCL
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+   // ...
+----------
+
+==== Use of Clang Blocks
+
+Clang Blocks that are defined by ObjC language are not supported
+and their use can be replaced by lambdas ({cpp}17 `[expr.prim.lambda]`).
+
+This means that builtin function `enqueue_kernel` with OpenCL C
+prototype `OpenCL v2.0 s6.13.17` is not supported in {cpp} for OpenCL.
```
mantognini: There are other BIF that rely on blocks, possibly in extensions. I would suggest generalising this statement a bit, maybe as follows: 

```suggestion
This means that builtin functions using blocks, such as `enqueue_kernel`,
are not supported in {cpp} for OpenCL.
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
...
+if (cond)
+  goto label;
+int n = foo();
+label:  // invalid: jumping forward over declaration of n
+   // ...
+----------
+
+==== Use of Clang Blocks
+
+Clang Blocks that are defined by ObjC language are not supported
```
mantognini:
```suggestion
Clang Blocks that are defined by the Objective-C language are not supported
```

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,120 @@
+=== Difference to OpenCL C
```
mantognini: I've no idea if this is a requirement but other files seem to have a copyright header.

cxx4opencl/diff2openclc.txt
AnastasiaStulova: Just FYI I would still like to enumerate those from the core spec here.

cxx4opencl/diff2openclc.txt
AnastasiaStulova: Btw I realized that this rule is just for the initialization list but I am not sure vector initialization falls under this category. For now I will avoid adding a vector example until I investigate this further.

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,121 @@
...
+----------
+struct mytype {
+ int i;
+};
+void foo(uint par){
+  mytype var = {
+    .i = par // narrowing from uint to int is disallowed
+  };
+}
+----------
```
kpet: This example should be simplified. The structure isn't really needed.

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,121 @@
...
+ int *ptr = NULL; // invalid initialization of int* with void*
+}
+----------
+
+==== Use of restrict
+
+{cpp}17 does not support `restrict` and therefore {cpp} for OpenCL
+can not support it either. Some compilers might provide extensions
+with some functionality of `restrict` in {cpp}, e.g. `__restrict`
+in Clang.
```
kpet: Maybe worth a comment explaining that defining `restrict` to an empty macro or `__restrict` is possible.

cxx4opencl/diff2openclc.txt
AnastasiaStulova: Hmm... I don't know how to show it without... Can you paste an example you have in mind?

cxx4opencl/diff2openclc.txt
AnastasiaStulova: I don't like the official language documentation is the best place to describe workarounds like this... Besides, it is pretty straight forward to get.

AnastasiaStulova: Also developers might prefer to alias it to __restrict or modify source code. There can be many approaches depending on the situation.

cxx4opencl/diff2openclc.txt
kpet: Oh, I'd read a bit too quickly. The restriction only applies to initialisation lists. The example is fine as is.

cxx4opencl/diff2openclc.txt
kpet: Well, we've done it for `NULL`...

>developers might prefer to alias it to __restrict or modify source code

Yes, as said in my comment.

AnastasiaStulova: Fair point! I have added a paragraph explaining that... although generally I think we should not go too much in this direction...

cxx4opencl/diff2openclc.txt
```diff
@@ -0,0 +1,129 @@
...
+
+This feature only affects optimizations and the source
+code can be modified by removing it. As a workaround to avoid manual
+modifications macro substitutions can be used to either remove the
```
svenvh, line 101:
```suggestion
modifications, macro substitutions can be used to either remove the
```

PR #244, opened 2020-03-22
add missing error code CL_CONTEXT_TERMINATED_KHR

bashbaug:
see: https://github.com/KhronosGroup/OpenCL-Headers/issues/69

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -516,8 +516,12 @@ server's OpenCL/api-docs repository.
             <unused start="-1108" end="-1120"/>
     </enums>
 
-    <enums start="-1121" end="-9999" name="ErrorCodes.1121" vendor="Khronos" comment="RESERVED FOR FUTURE ALLOCATIONS BY KHRONOS">
-            <unused start="-1121" end="-9999"/>
+    <enums start="-1121" end="-1121" name="ErrorCodes.1121" vendor="Khronos" comment="For cl_khr_terminate_context">
+        <enum value="-1121"         name="CL_CONTEXT_TERMINATED_KHR"/>
+    </enums>
+
+    <enums start="-1122" end="-9999" name="ErrorCodes.future" vendor="Khronos" comment="RESERVED FOR FUTURE ALLOCATIONS BY KHRONOS">
+            <unused start="-1122" end="-9999"/>
     </enums>
 
     <enums name="cl_bool" vendor="Khronos" comment="Boolean values">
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -516,8 +516,12 @@ server's OpenCL/api-docs repository.
...
     </enums>
 
-    <enums start="-1121" end="-9999" name="ErrorCodes.1121" vendor="Khronos" comment="RESERVED FOR FUTURE ALLOCATIONS BY KHRONOS">
-            <unused start="-1121" end="-9999"/>
+    <enums start="-1121" end="-1121" name="ErrorCodes.1121" vendor="Khronos" comment="For cl_khr_terminate_context">
+        <enum value="-1121"         name="CL_CONTEXT_TERMINATED_KHR"/>
+    </enums>
+
+    <enums start="-1122" end="-9999" name="ErrorCodes.future" vendor="Khronos" comment="RESERVED FOR FUTURE ALLOCATIONS BY KHRONOS">
+            <unused start="-1122" end="-9999"/>
```
kpet: Do we want to allocate a bigger block to KHR extensions as part of this to avoid creating too many small blocks?

bashbaug: I think this is a good idea, although it's not strictly required.

For whatever reason we currently allocate enums in blocks of 16, but error codes are allocated one at a time.  If we decide to allocate error codes in blocks, I think the only thing we need to decide is the block granularity.  The obvious block sizes are 16 (consistent with non-error enums) or 10 (since error codes are currently expressed as base 10 integers).  Any preference?  I have a slight leaning towards blocks of 10, but -1120 is divisible by both 10 and 16 so we have options.

Note, if we decide to allocate blocks then it shouldn't change the value of the `CL_CONTEXT_TERMINATED_KHR` enum, so we could expand the allocation to a block in a subsequent PR.

kpet: Yes, definitely not a strict requirement. I don't think the size of the block matters too much given how chaotic previous allocations were :).

I'm fine to merge this as is if you prefer allocating a block in a separate PR. Your call.

oddhack: The block size actually dates back to SGI ca. 1990 in the original OpenGL enum reservation system :-) It is convenient to have it come out as a multiple of 16 since the enums are reserved in hex, and often a single block will last a vendor for many years, but the block size doesn't really matter.

alycm: In the interest of merging this XML change I've filed an issue (https://github.com/KhronosGroup/OpenCL-Docs/issues/248) about whether to reserve error codes in blocks.  Obviously if we do change that it should be soon given it is currently numerically convenient to do so.

PR #252, opened 2020-03-31
Remove Duplicated cl_sampler_properties

FranklandJack:
Remove the `cl_sampler_properties` `CL_SAMPLER_MIP_FILTER_MODE`,
`CL_SAMPLER_LOD_MIN` and `CL_SAMPLER_LOD_MAX` which are duplicated
under `CL_VERSION_2_0` when they should only be required under
the `ck_khr_mipmap_image` extension.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1290,8 +1290,11 @@ server's OpenCL/api-docs repository.
         <enum value="0x1153"        name="CL_SAMPLER_ADDRESSING_MODE"/>
         <enum value="0x1154"        name="CL_SAMPLER_FILTER_MODE"/>
         <enum value="0x1155"        name="CL_SAMPLER_MIP_FILTER_MODE"/>
+        <enum value="0x1155"        name="CL_SAMPLER_MIP_FILTER_MODE_KHR"/>
         <enum value="0x1156"        name="CL_SAMPLER_LOD_MIN"/>
+        <enum value="0x1156"        name="CL_SAMPLER_LOD_MIN_KHR"/>
         <enum value="0x1157"        name="CL_SAMPLER_LOD_MAX"/>
+        <enum value="0x1157"        name="CL_SAMPLER_LOD_MAX_KHR"/>
         <enum value="0x1158"        name="CL_SAMPLER_PROPERTIES"/>
             <unused start="0x1159" end="0x115F" comment="Reserved for cl_sampler_info"/>
         <enum value="0x1160"        name="CL_PROGRAM_REFERENCE_COUNT"/>
@@ -4839,9 +4842,9 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl.h"/>
             </require>
             <require comment="cl_sampler_properties">
-                <enum name="CL_SAMPLER_MIP_FILTER_MODE"/>
-                <enum name="CL_SAMPLER_LOD_MIN"/>
-                <enum name="CL_SAMPLER_LOD_MAX"/>
+                <enum name="CL_SAMPLER_MIP_FILTER_MODE_KHR"/>
+                <enum name="CL_SAMPLER_LOD_MIN_KHR"/>
+                <enum name="CL_SAMPLER_LOD_MAX_KHR"/>
             </require>
         </extension>
         <extension name="cl_img_use_gralloc_ptr" supported="opencl">


bashbaug:
Hello, sorry for the very slow reply.

This all sounded vaguely familiar so I looked through the OpenCL headers and found:

```c
#ifdef CL_VERSION_2_0
/* These enumerants are for the cl_khr_mipmap_image extension.
   They have since been added to cl_ext.h with an appropriate
   KHR suffix, but are left here for backwards compatibility. */
#define CL_SAMPLER_MIP_FILTER_MODE                  0x1155
#define CL_SAMPLER_LOD_MIN                          0x1156
#define CL_SAMPLER_LOD_MAX                          0x1157
#endif
```

How would you feel about eliminating the duplication by adding KHR-suffixed aliases for these enums, then changing the enums associated with `cl_khr_mipmap_image` to use the suffixed aliases?  I think that would be most consistent with the way the enums are defined in the headers.

FranklandJack:
> Hello, sorry for the very slow reply.
> 
> This all sounded vaguely familiar so I looked through the OpenCL headers and found:
> 
> ```c
> #ifdef CL_VERSION_2_0
> /* These enumerants are for the cl_khr_mipmap_image extension.
>    They have since been added to cl_ext.h with an appropriate
>    KHR suffix, but are left here for backwards compatibility. */
> #define CL_SAMPLER_MIP_FILTER_MODE                  0x1155
> #define CL_SAMPLER_LOD_MIN                          0x1156
> #define CL_SAMPLER_LOD_MAX                          0x1157
> #endif
> ```
> 
> How would you feel about eliminating the duplication by adding KHR-suffixed aliases for these enums, then changing the enums associated with `cl_khr_mipmap_image` to use the suffixed aliases? I think that would be most consistent with the way the enums are defined in the headers.

Thanks for your comments, can I just clarify, are you asking whether instead of removing the duplication we should simply append the `_khr` suffix to `cl.xml` under `cl_khr_mipmap_image` and keep the non `_khr` enum values as well so that they match the definitions in `cl.h` and `cl_ext.h`, that is the `_khr` in `cl.xml` would be for the definitions in `cl_ext.h` and the non `_khr` in `cl.xml` would be for the definitions in `cl.h`?

bashbaug:
> are you asking whether instead of removing the duplication we should simply append the `_khr` suffix [...]

Yes, that's the idea.  We did something similar with the `cl_kernel_sub_group_info` enums.  You can look at `CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE_KHR` for an example.  Would this work for you?

FranklandJack:
> > are you asking whether instead of removing the duplication we should simply append the `_khr` suffix [...]
> 
> Yes, that's the idea. We did something similar with the `cl_kernel_sub_group_info` enums. You can look at `CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE_KHR` for an example. Would this work for you?

Thank you for the clarification, make sense now, will update the patch.

FranklandJack:
@bashbaug Have updated the patch to reflect the changes you suggested, let me know if this is what you meant.

xml/cl.xml
```diff
@@ -4836,9 +4836,9 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl.h"/>
             </require>
             <require comment="cl_sampler_properties">
-                <enum name="CL_SAMPLER_MIP_FILTER_MODE"/>
-                <enum name="CL_SAMPLER_LOD_MIN"/>
-                <enum name="CL_SAMPLER_LOD_MAX"/>
+                <enum name="CL_SAMPLER_MIP_FILTER_MODE_KHR"/>
+                <enum name="CL_SAMPLER_LOD_MIN_KHR"/>
+                <enum name="CL_SAMPLER_LOD_MAX_KHR"/>
```
bashbaug, line 4841: Thanks - yes, this is exactly what I had in mind.

The last thing to change is to add these suffixed enums as aliases for the un-suffixed enums in the `<enums>` section.  Something like:

```xml
        <enum value="0x1156"        name="CL_SAMPLER_LOD_MIN"/>
        <enum value="0x1156"        name="CL_SAMPLER_LOD_MIN_KHR"/>
```

FranklandJack: Thanks for your feedback, have updated the `<enums>` section as well. https://github.com/KhronosGroup/OpenCL-Docs/pull/252/commits/f1c9cba11a7bc1dcea327bd4a16071179402b7b7#diff-62465512a1e1a33f8cdaee21f4a5486cR1293-R1297

PR #257, opened 2020-04-20
Fix vector literal example rendering

svenvh:
Disable AsciiDoc's macro expansion.  It is not needed for this
example, and actually causes a problem due to `((` and `))` being
interpreted.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -562,12 +562,11 @@ replicated across all lanes of the vector.
 Examples:
 
 [source,c]
-[subs="verbatim,quotes,macros"]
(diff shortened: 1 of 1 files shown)


bashbaug:
Strange, the output looks fine on one of my systems even without this change, but not on another, and not via the Travis builds.

Are there any other places where we should apply a similar fix?  Just asking, I don't see any reason to hold up merging this change ASAP.  Thanks!

svenvh:
> Are there any other places where we should apply a similar fix?

Had a quick look already, but couldn't find any other places that needed fixing.

PR #259, opened 2020-04-21
Add initial version to all extension specifications

kpet:
Addresses https://gitlab.khronos.org/opencl/OpenCL-Docs/issues/194.


kpet:
Thanks for the reviews!

1. Agree it's ambiguous. I'll change the wording.

2. I think it's probably a good thing to have the provisional to final transition be application-visible. Maybe we could use 1.0.0 for the final version and start with say 0.9.0 for now? I'll say that the first version is provisional.

3. Hmm, yes, but we are not describing this one in the same way as other extensions. Should we add a dedicated chapter?

bashbaug:
Thanks, latest changes LGTM!  Merging.

PR #260, opened 2020-04-23
[C++ for OpenCL] Increased ToC level

AnastasiaStulova:


CXX_for_OpenCL.txt
```diff
@@ -8,7 +8,7 @@ Khronos{R} OpenCL Working Group
 :data-uri:
 :icons: font
 :toc2:
-:toclevels: 2
+:toclevels: 4
```
svenvh, line 11: Any reason to have 4?  The OpenCL C spec uses 3.

AnastasiaStulova: No big reason. Now updated to 3. :)

PR #272, opened 2020-04-28
Update extension template to use versions instead of revisions

kpet:
Assign arbitrary versions to existing revisions.

### DIFF
--- a/extensions/cl_extension_template.asciidoc
+++ b/extensions/cl_extension_template.asciidoc
@@ -125,7 +125,7 @@ Promoters, or any other relevant status information.
 == Version
 
 Built On: {docdate} +
-Revision: 3
+Version: Major.Minor.Patch
 
 ****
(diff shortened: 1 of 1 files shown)


kpet:
>I do wonder if we should assign the extension versions with a bit more thought. [...]

I did think about all this when drafting this change. I agree that some of the changes should probably have been patch changes but I didn't want to go through the git history to figure which was which (and run the risk of getting into patch vs. minor arguments on specific changes :)). I don't think it matters too much for historical changes but I agree we should use a policy that corresponds to the suggested use of these version numbers for real extensions:
- major versions can break compability
- minor versions for backwards-compatible feature additions
- patch version for small tweaks

bashbaug:
Works for me.  No reason to hold off merging this, I think.  We can always retroactively assign different version numbers in the extension template if someone believes very strongly they should be different :-).

PR #277, opened 2020-05-07
API: Update clEnqueueNDRange description for OpenCL 1.0

pierremoreau:
In OpenCL 1.0, `clEnqueueNDRange`’s `global_work_offset` had to be `NULL`. This information was missing from the unified API specification.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -8061,6 +8061,7 @@ include::{generated}/api/version-notes/clEnqueueNDRangeKernel.asciidoc[]
     values that describe the offset used to calculate the global ID of a
     work-item.
     If _global_work_offset_ is `NULL`, the global IDs start at offset (0, 0, 0).
+    _global_work_offset_ must be `NULL` <<unified-spec, before>> version 1.1.
   * _global_work_size_ points to an array of _work_dim_ unsigned values that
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -8061,6 +8061,7 @@ include::{generated}/api/version-notes/clEnqueueNDRangeKernel.asciidoc[]
     values that describe the offset used to calculate the global ID of a
     work-item.
     If _global_work_offset_ is `NULL`, the global IDs start at offset (0, 0, 0).
+    _global_work_offset_ has to be `NULL` <<unified-spec, before>> version 1.1.
```
bashbaug, line 8064: Nice catch.  This looks good to me, but could I suggest rephrasing slightly to say "_global_work_offset_ **must** be `NULL`" rather than "**has to** be `NULL`", to be more consistent with terminology in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)?  I know we don't follow RFC 2119 consistently in the spec today, but I think this something we _should_ be doing, and if we use this phrasing here now it's one fewer place we'll need to fix in the future.

Thanks!

pierremoreau: Good point! I updated the specification text and the commit message.

pierremoreau:
Thank you for the review!

alycm:
Thanks for catching this!

PR #278, opened 2020-05-08
all built-in variables must be in the input storage class

### LINKED ISSUE #269: SPIR-V OpenCL environment spec does not describe Builtin decoration
The SPIR-V core specs reference the client API to describe Builtin decoration operand (see https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#BuiltIn) but the environment specs does not mention any of them.

Among the ambiguities:

- Kernel Execution Model related builtin decoration are not described (I'm not sure if more is needed to be said as SPIR-V core seems rather explicit)
- The environment spec does not specify which storage class an OpVariables should use when decorated by a Builtin decoration (leads to some ambiguities like here https://github.com/intel/llvm/pull/1581#discussion_r414731009)

bashbaug:
Fixes #269.

This change documents in the OpenCL SPIR-V Environment Spec that all variables decorated with **BuiltIn** must be in the **Input** storage class.

There's probably a bit more we could say here - do specific built-in variables need to be specific types, for example? - but documenting that they must be in the input storage class is a good first step.

### DIFF
--- a/env/common_properties.asciidoc
+++ b/env/common_properties.asciidoc
@@ -384,3 +384,9 @@ API.
 
 Environments that support extensions or optional features may allow
 additional types in an entry point's parameter list.
+
+=== Built-in Variables
+
+An *OpVariable* in a SPIR-V module with the *BuiltIn* decoration represents
+a built-in variable.
+All built-in variables must be in the *Input* storage class.


bashbaug:
@Naghasan or @alycm, would one of you mind reviewing this please?  Thanks!

bashbaug:
> shouldn't we also say something about the type ?

Yeah, I agree.

I've filed https://github.com/KhronosGroup/OpenCL-Docs/issues/279 to track this.  Let's get this change in, then we can tackle the question of types next.

Issue #274, opened 2020-04-29
Clarify that build status should be set before calling build completion callback


jrprice:
The [`async_build` test in CTS](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_async_build.cpp#L46) is currently written in a way which assumes that the build completion callback provided to `clBuildProgram` is called *before* the build status is changed from `CL_BUILD_IN_PROGRESS` to `CL_BUILD_SUCCESS`/`CL_BUILD_ERROR` (i.e. it starts a build, loops until build status != "in progress", and then checks that the callback has been fired).

This seems strange, as it implies that the application cannot check the program build status from inside the callback itself, which is likely something a developer would want to do. The spec doesn't explicitly state in which order these things should happen.

I'd like to propose that we clarify the spec to explicitly state that the build status will be set *before* the callback is called, so that applications can safely check the build status from inside the callback. I've modified the CTS test in question to test this behavior (see [here](https://github.com/jrprice/OpenCL-CTS/blob/async-build/test_conformance/compiler/test_async_build.cpp#L94)), and verified that it still passes on NVIDIA and Qualcomm drivers.

CC @kpet.

bashbaug:
This makes sense to me.  I've checked our CPU and GPU implementations and they also set the build status before invoking the callback, so they also should pass your modified CTS test.

alycm:
Our implementation also calls the callback last (i.e. after build status will be set).

And I agree that checking for success/error in the callback makes sense and getting "in progress" there would not be helpful.

You don't mention `clCompileProgram`/`clLinkProgram` but I assume that the same spec clarification would apply there as well.

jrprice:
> You don't mention `clCompileProgram`/`clLinkProgram` but I assume that the same spec clarification would apply there as well.

The description for `clLinkProgram` already contains this text, which I think is sufficient:

> Once the linker has completed, the pfn_notify callback function is called which returns the program object returned by clLinkProgram. The application can query the link status and log for this program object. 

I'll propose similar wording for `clBuildProgram` and `clCompileProgram`.

PR #285, opened 2020-05-11
Clarify that program build callbacks can query build status

jrprice:
Fixes https://github.com/KhronosGroup/OpenCL-Docs/issues/274


jrprice:
> This doesn't really guarantee that the build is complete. It is valid to query the build status or log before the build is complete.

Fair point. I've added "... to determine if the build was successful". I think this, in combination with the "will be called when the program executable has been built" text from a previous bullet, provides the guarantees an application needs for this to be useful, but let me know if you want even more explicitness.

jrprice:
Works for me, I've switched to that suggestion almost verbatim.

PR #316, opened 2020-06-08
Clarify description of CL_DEVICE_OPENCL_C_FEATURES

alycm:


alycm:
> Is there a chance an OpenCL 3.0 feature would return a non-zero "patch" number, e.g. 3.0.1?

Given 3.0 is currently provisional there is definitely the possibility of that, yes.

Perhaps that means that there should be a table of the expected values so that implementations are consistent.

Alternatively, I'm not sure that the patch number will ever provide useful information, once a spec is non-provisional it doesn't seem like an entire new feature macro would be added in a patch release.  So MAJOR.MINOR.0 seems just as helpful, and saves arbitrary different patch numbers for different macros.

bashbaug:
The scenario I'm considering is a bugfix/clarification regarding one of the feature macros.  Let's say it was ambiguous whether some OpenCL C built-in function applied to `__opencl_c_3d_image_writes` that we clarified in a patch version of the spec.  In that case, the compiler could use the feature macro value to indicate that the built-in function can be unambiguously used.  Admittedly this should not be common.

Let's go with the current text for now.  We can always relax it later.  Thanks!

PR #319, opened 2020-06-08
fix minor bugs regarding image conversion rules

### LINKED ISSUE #97: Incorrect image conversion rules in OpenCL 1.2 and C language specs
Section 8.3 of the OpenCL Specification (1.2) and OpenCL C 2.0 Specification incorrectly specifies some of the conversion rules. Most of the issues are corrected in the OpenCL™ SPIR-V Environment Specification and The OpenCL™ C++ 1.0 Specification. 

1. Conversion algorithms are overly prescriptive, e.g. (in 8.3.1.1 of OCL 1.2 ) _normalized float value = (float)c / 255.0f_ dictates the division be done in float math, while not specifying the rounding mode. Section 9.4 of the SPIR-V spec relaxes this to _normalized_float_value(x)=round_to_float(x/255)_
2. Rounding rules for conversions from float to normalized integers are wrong, e.g. (in 8.3.1.2 of OCL 1.2) in the rule _fabs(fpreferred – fapprox) must be <= 0.6_, _fpreferred_ and _fapprox_ are defined as integer values, so the 0.6 limit makes no sense.  The SPIR-V spec corrects this to compare with the exact value rather than the rounded one.

Minor issues remaining in the SPIR-V and OpenCL++ specs:

1.  In SPIR-V 9.4.1, Converting Floating-point Values to Normalized Integer Channel Data Types, the sentence beginning _The conversions from half precision floating-point values to normalized integer values_ should not refer to half precision.
2.  In SPIR-V 9.4.1, the sentence _The precision of the above conversions is <= 1.5 ulp except for the following cases._ should use the word accuracy, not precision.

bashbaug:
Fixes #97.

- relaxed conversion algorithms from int to float
- fixed precision descriptions to refer to exact values
- fixed one section which erroneously referred to half precision
- fixed several cases of precision vs. accuracy

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -11687,25 +11687,25 @@ These conversions are performed as follows:
 
 `CL_UNORM_INT8` (8-bit unsigned integer) {rightarrow} `float`
 
-  :: normalized `float` value = `(float)c / 255.0f`
+  :: normalized `float` value = `round_to_float(c / 255)`
 
 `CL_UNORM_INT_101010` (10-bit unsigned integer) {rightarrow} `float`
 
-  :: normalized `float` value = `(float)c / 1023.0f`
+  :: normalized `float` value = `round_to_float(c / 1023)`
 
 `CL_UNORM_INT16` (16-bit unsigned integer) {rightarrow} `float`
 
-  :: normalized `float` value = `(float)c / 65535.0f`
+  :: normalized `float` value = `round_to_float(c / 65535)`
 
 `CL_SNORM_INT8` (8-bit signed integer) {rightarrow} `float`
 
-  :: normalized `float` value = *max*(`-1.0f`, `(float)c / 127.0f`)
+  :: normalized `float` value = `max(-1.0f, round_to_float(c / 127))`
 
 `CL_SNORM_INT16` (16-bit signed integer) {rightarrow} `float`
 
-  :: normalized `float` value = *max*(`-1.0f`, `(float)c / 32767.0f`)
+  :: normalized `float` value = `max(-1.0f, round_to_float(c / 32767))`
 
-The precision of the above conversions is \<= 1.5 ulp except for the
+The accuracy of the above conversions must be \<= 1.5 ulp except for the
 following cases.
 
 For `CL_UNORM_INT8`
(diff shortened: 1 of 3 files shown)


bashbaug:
@jlewis-austin, would you mind reviewing these changes?  Thanks!

jlewis-austin:
@bashbaug sure. I took a quick look but am out for a couple days this week. I'll try to get through it on Friday.

OpenCL_C.txt
```diff
@@ -11687,25 +11687,25 @@ These conversions are performed as follows:
 
 `CL_UNORM_INT8` (8-bit unsigned integer) {rightarrow} `float`
 
-  :: normalized `float` value = `(float)c / 255.0f`
+  :: normalized `float` value = `round_to_float(c / 255)`
```
jlewis-austin, line 11690: Do we define round_to_float anywhere? I just realized there's no dedicated section in this spec corresponding to half floats, which makes me wonder if this section was intended to be generic to cover both float and half floats generically. If so, that could be clarified in the round_to_float definition, and it would be more clear to replace instances of `float` with something more generic like "floating-point" or "floating-point value".

jlewis-austin: It looks like this is covered by https://github.com/KhronosGroup/OpenCL-Docs/pull/320/files. Only question then is a definition for round_to_float.

bashbaug: I borrowed `round_to_float` from the SPIR-V Environment spec and the OpenCL C++ kernel language spec.  I've gone back and forth whether I like it or not, but I think it's OK if you think of the division being infinitely precise.  We should say this explicitly, though.

What if we did something like the conversions from float to int, and defined:

* The exact result, which is the division with no rounding.
* The rounded result, which is the exact result subject to some implementation-defined rounding.
* That the rounded result must be within 1.5 ULP of the exact result.

Would that work?

I like your suggestion to generalize this section so it applies to floats and halfs also.  I'll see if I can fold that into some other changes I've been experimenting with, to generalize the size of each normalized channel also.

jlewis-austin: That sounds good. That leaves the CTS on the hook for computing the difference correctly, but I'm pretty sure computing the reference in double is sufficient. I don't remember how the CTS is handling this today.

bashbaug:
I think we can hold off merging this right now (there's a conflict I'd need to resolve anyhow).  I'd like to investigate the simplification I mentioned in my comment above: https://github.com/KhronosGroup/OpenCL-Docs/pull/319#discussion_r439668882

I'll convert this to a "draft" so it doesn't pop up in our lists.

Issue #183, opened 2020-01-13
Missing Spec Descriptions for CL_UNORM_SHORT_565 and CL_UNORM_SHORT_555


bashbaug:
The OpenCL C and SPIR-V environment specs are missing descriptions in many places regarding the `CL_UNORM_SHORT_565` and `CL_UNORM_SHORT_555` image channel orders.

For example:

* OpenCL C Section 8.3 Conversion Rules does not describe how `CL_UNORM_SHORT_565` or `CL_UNORM_SHORT_555` channels are converted to or from a `float`, precision requirements for the conversion, and any special conversion values.
* The descriptions of the `read_image` and `write_image` built-in functions do not describe any cases for `CL_UNORM_SHORT_565` or `CL_UNORM_SHORT_555`.

A few descriptions are also missing for `CL_UNORM_INT_101010` and `CL_UNORM_INT_101010_2`, that should be fixed at the same time.

kpet:
See point 2 on related https://github.com/KhronosGroup/OpenCL-Docs/issues/171.

b-sumner:
There's been some question about exactly which bits in the 16 bit word correspond to R, G, and B as well, so we may want to cover that too.  Similarly for 101010 and 101010_2.

StuartDBrady:
@b-sumner, that seems to be defined in the OpenCL 2.x specifications § 5.3.1.1 _Image Format Descriptor_, as follows:

> For CL_UNORM_SHORT_565, `R` is in bits `15:11`, `G` is in bits `10:5` and `B` is in bits `4:0`. For CL_UNORM_SHORT_555, bit `15` is undefined, `R` is in bits `14:10`, `G` in bits `9:5` and `B` in bits `4:0`. For CL_UNORM_INT_101010, bits `31:30` are undefined, `R` is in bits `29:20`, `G` in bits `19:10` and `B` in bits `9:0`.  For CL_UNORM_INT_101010_2, `R` is in bits `31:22`, `G` in bits `21:12`, `B` in bits `11:2` and `A` in bits `1:0`.

Are the spec changes that are needed limited to the OpenCL C spec and the SPIR-V environment spec, or are wider changes needed?

joshqti:
Can we revisit this issue?  I also noticed it is missing from the spec.

kpet:
[...]

PR #320, opened 2020-06-09
add conversions for halfs to and from normalized image channel orders

### LINKED ISSUE #193: Missing Spec Description for conversion of half to image_channel_type: CL_UNORM_INT8 / CL_UNORM_INT16 / CL_SNORM_INT8 / CL_SNORM_INT16.

bashbaug:
Fixes #193

This PR adds two sections to the `cl_khr_fp16` spec that describe how half values are converted to normalized image channel orders (for image writes) and converted from normalized image channel orders (for image reads).  These sections should match similar sections that are already in the OpenCL SPIR-V environment spec.

### DIFF
--- a/ext/cl_khr_fp16.asciidoc
+++ b/ext/cl_khr_fp16.asciidoc
@@ -1811,53 +1811,63 @@ These conversions are performed as follows:
 
 `CL_UNORM_INT8` (8-bit unsigned integer) {rightarrow} `half`
 
-  :: normalized `half` value = `round_to_half(c / 255)`
+[none]
+* normalized `half` value = `round_to_half(c / 255)`
 
 `CL_UNORM_INT_101010` (10-bit unsigned integer) {rightarrow} `half`
 
-  :: normalized `half` value = `round_to_half(c / 1023)`
+[none]
+* normalized `half` value = `round_to_half(c / 1023)`
 
 `CL_UNORM_INT16` (16-bit unsigned integer) {rightarrow} `half`
 
-  :: normalized `half` value = `round_to_half(c / 65535)`
+[none]
+* normalized `half` value = `round_to_half(c / 65535)`
 
 `CL_SNORM_INT8` (8-bit signed integer) {rightarrow} `half`
 
-  :: normalized `half` value = `max(-1.0f, round_to_half(c / 127))`
+[none]
+* normalized `half` value = `max(-1.0f, round_to_half(c / 127))`
 
 `CL_SNORM_INT16` (16-bit signed integer) {rightarrow} `half`
 
-  :: normalized `half` value = `max(-1.0f, round_to_half(c / 32767))`
+[none]
+* normalized `half` value = `max(-1.0f, round_to_half(c / 32767))`
 
 The accuracy of the above conversions must be \<= 1.5 ulp except for the
 following cases.
 
 For `CL_UNORM_INT8`
 
-  :: 0 must convert to `0.0h` and
-  :: 255 must convert to `1.0h`
+[none]
+* 0 must convert to `0.0h` and
+* 255 must convert to `1.0h`
 
 For `CL_UNORM_INT_101010`
 
-  :: 0 must convert to `0.0h` and
-  :: 1023 must convert to `1.0h`
+[none]
+* 0 must convert to `0.0h` and
+* 1023 must convert to `1.0h`
 
 For `CL_UNORM_INT16`
 
-  :: 0 must convert to `0.0h` and
-  :: 65535 must convert to `1.0h`
+[none]
+* 0 must convert to `0.0h` and
+* 65535 must convert to `1.0h`
 
 For `CL_SNORM_INT8`
 
-  :: -128 and -127 must convert to `-1.0j`,
-  :: 0 must convert to `0.0h` and
-  :: 127 must convert to `1.0h`
+[none]
+* -128 and -127 must convert to `-1.0j`,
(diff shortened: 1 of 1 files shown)


ext/cl_khr_fp16.asciidoc
```diff
@@ -1788,3 +1788,126 @@ precision floating-point values and performing the operation in single
...
+
+  :: normalized `half` value = `round_to_half(c / 1023)`
+
+`CL_UNORM_INT16` (16-bit unsigned integer) {rightarrow} `half`
+
+  :: normalized `half` value = `round_to_half(c / 65535)`
+
+`CL_SNORM_INT8` (8-bit signed integer) {rightarrow} `half`
+
+  :: normalized `half` value = `max(-1.0f, round_to_half(c / 127))`
```
alycm, line 1826:
```suggestion
  :: normalized `half` value = `max(-1.0h, round_to_half(c / 127))`
```

ext/cl_khr_fp16.asciidoc
```diff
@@ -1788,3 +1788,126 @@ precision floating-point values and performing the operation in single
...
+
+`CL_SNORM_INT16` (16-bit signed integer) {rightarrow} `half`
+
+  :: normalized `half` value = `max(-1.0f, round_to_half(c / 32767))`
```
alycm, line 1830:
```suggestion
  :: normalized `half` value = `max(-1.0h, round_to_half(c / 32767))`
```

ext/cl_khr_fp16.asciidoc
```diff
@@ -1788,3 +1788,126 @@ precision floating-point values and performing the operation in single
...
+  :: 1023 must convert to `1.0h`
+
+For `CL_UNORM_INT16`
+
+  :: 0 must convert to `0.0h` and
+  :: 65535 must convert to `1.0h`
+
+For `CL_SNORM_INT8`
+
+  :: -128 and -127 must convert to `-1.0j`,
```
alycm, line 1852:
```suggestion
  :: -128 and -127 must convert to `-1.0h`,
```

ext/cl_khr_fp16.asciidoc
```diff
@@ -1788,3 +1788,126 @@ precision floating-point values and performing the operation in single
...
+
+These conversions are performed as follows:
+
+`CL_UNORM_INT8` (8-bit unsigned integer) {rightarrow} `half`
+
+  :: normalized `half` value = `round_to_half(c / 255)`
+
+`CL_UNORM_INT_101010` (10-bit unsigned integer) {rightarrow} `half`
+
+  :: normalized `half` value = `round_to_half(c / 1023)`
```
alycm, line 1818: This is equivalent to the SPIR-V environment spec, but there the formulas use a mathematical formatting whereas here it's more C like.

Would `round_to_half((half)c / 1023.0h)` be clearer?   I don't actually know if that is actually correct though (compared to thinking of the divide as infinite-precision).

bashbaug:
Thanks for the review comments!  I'm fixing these now.

ext/cl_khr_fp16.asciidoc
```diff
@@ -1788,3 +1788,126 @@ precision floating-point values and performing the operation in single
...
+For images created with image channel data type of `CL_SNORM_INT8` and
+`CL_SNORM_INT16`, *read_imagef* will convert the channel values from an
+8-bit or 16-bit signed integer to normalized half precision floating-point
+values in the range [`-1.0h`, `1.0h`].
+
+These conversions are performed as follows:
+
+`CL_UNORM_INT8` (8-bit unsigned integer) {rightarrow} `half`
+
+  :: normalized `half` value = `round_to_half(c / 255)`
```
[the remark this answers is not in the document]
bashbaug: Yes - fixed!

ext/cl_khr_fp16.asciidoc
bashbaug: Fixed!

ext/cl_khr_fp16.asciidoc
bashbaug: Fixed!

ext/cl_khr_fp16.asciidoc
bashbaug: Fixed!

ext/cl_khr_fp16.asciidoc
bashbaug: This is the only change I haven't made, primarily because I think we should clarify that this is intended to be an infinite-precision division for all conversions, both to `half` and to `float` in the OpenCL C spec and SPIR-V Environment spec.  I'd like to make these changes as part of the resolution for #97 and #183.  Does this sound reasonable?

alycm: I didn't even write what my original issue was: read with C syntax meant I read it as an integer divide that will always be `0` or `1`.

But completely reasonable to wait until these cases are all resolved together.

Issue #246, opened 2020-03-26
Possibly missing generated refpages (enums.html & more)


TSlivede:
Hi,
I'm not sure, if this is the correct place to ask this, but it seems that https://www.khronos.org/registry/OpenCL//sdk/2.2/docs/man/html/enums.html is unreachable (presents "404… Oops") even though it is still linked in the navigation tree on the left side under "OpenCL Runtime" -> "Enumerated Types".

The corresponding source file does still seem to [exist](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/man/static/enums.txt)...

oddhack:
@outofcontrol could you attach the complete list of problem links you alluded to in email here? I'll take this, unless Ben wants it.

outofcontrol:
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/vstore_halfn.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/vstoren.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/vload_halfn.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/vload_half.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/commonMax.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/abstractDataTypes.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_map_flags.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/clCreateImage2D.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_channel_type.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/atomic_flag_test_and_set_explicit.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_sampler.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_context_info.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_queue_properties.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_kernel_sub_group_info.html
https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/cl_device_type.html
[...]

PR #321, opened 2020-06-09
Separate out 2.1 additions to device-side enqueue (#246)

lakshmih:
Change the query to a bitfield to allow for querying support for
clSetDefaultDeviceCommandQueue separately from support
for the rest of the device-side enqueue feature

### DIFF
--- a/api/appendix_e.asciidoc
+++ b/api/appendix_e.asciidoc
@@ -368,7 +368,7 @@ device:
   * {CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT} to
     determine whether a device supports the generic address space.
   * {CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES} to determine the device-side enqueue
-    capabilities of a device
+    capabilities of a device.
   * {CL_DEVICE_PIPE_SUPPORT} to determine whether a device supports
     pipe memory objects.
   * {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} to determine the
--- a/api/appendix_e.asciidoc
+++ b/api/appendix_e.asciidoc
@@ -367,8 +367,8 @@ device:
     collective functions, such as broadcasts, scans, and reductions.
   * {CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT} to
     determine whether a device supports the generic address space.
-  * {CL_DEVICE_DEVICE_ENQUEUE_SUPPORT} to determine whether a device
-    supports device queues and device-side enqueue.
+  * {CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES} to determine the device-side enqueue
+    capabilities of a device
   * {CL_DEVICE_PIPE_SUPPORT} to determine whether a device supports
     pipe memory objects.
   * {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} to determine the
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -165,7 +165,7 @@ When Device-Side Enqueue is not supported:
 
 |====
 
-When Device-Side Enqueue is supported but a replaceable On-Device Queue is not supported:
+When Device-Side Enqueue is supported but a replaceable default On-Device Queue is not supported:
 
 [cols="2,3",options="header",]
 |====
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -174,7 +174,7 @@ When Device-Side Enqueue is supported but a replaceable On-Device Queue is not s
 
 | {clGetDeviceInfo}, passing +
 {CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES}
-| May not set {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT} , indicating that _device_ does not support a replaceable default On-Device Queue.
+| May omit {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT}, indicating that _device_ does not support a replaceable default On-Device Queue.
 
 | {clSetDefaultDeviceCommandQueue}
 | Returns {CL_INVALID_OPERATION} if _device_ does not support a replaceable default On-Device Queue.
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -137,8 +137,8 @@ When Device-Side Enqueue is not supported:
 |*Behavior*
 
 | {clGetDeviceInfo}, passing +
-{CL_DEVICE_DEVICE_ENQUEUE_SUPPORT}
-| May return {CL_FALSE}, indicating that _device_ does not support Device-Side Enqueue and On-Device Queues.
+{CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES}
+| May return `0`, indicating that _device_ does not support Device-Side Enqueue and On-Device Queues.
 
 | {clGetDeviceInfo}, passing +
 {CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES}
@@ -162,6 +162,22 @@ When Device-Side Enqueue is not supported:
 
 | {clSetDefaultDeviceCommandQueue}
 | Returns {CL_INVALID_OPERATION} if _device_ does not support On-Device Queues.
+
+|====
+
+When Device-Side Enqueue is supported but a replaceable On-Device Queue is not supported:
+
+[cols="2,3",options="header",]
+|====
+|*API*
+|*Behavior*
+
+| {clGetDeviceInfo}, passing +
+{CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES}
+| May not set {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT} , indicating that _device_ does not support a replaceable default On-Device Queue.
+
+| {clSetDefaultDeviceCommandQueue}
+| Returns {CL_INVALID_OPERATION} if _device_ does not support a replaceable default On-Device Queue.
 
 |====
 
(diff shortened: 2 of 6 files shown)


lakshmih:
@bashbaug @kpet  can you please review this?

api/appendix_h.asciidoc
```diff
@@ -164,6 +164,22 @@ When Device-Side Enqueue is not supported:
...
+When Device-Side Enqueue is supported but a replaceable On-Device Queue is not supported:
+
+[cols="2,3",options="header",]
+|====
+|*API*
+|*Behavior*
+
+| {clGetDeviceInfo}, passing +
+{CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES}
+| May not set {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT} , indicating that _device_ does not support a replaceable default On-Device Queue.
```
bashbaug, line 176: I'd suggest a slight wording change here: "may omit CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT" instead of "may not set".  When I read "may not set" I initially think that it is prohibited, not that it is optional.

api/appendix_h.asciidoc
```diff
@@ -164,6 +164,22 @@ When Device-Side Enqueue is not supported:
...
+
+| {clSetDefaultDeviceCommandQueue}
+| Returns {CL_INVALID_OPERATION} if _device_ does not support a replaceable default On-Device Queue.
```
bashbaug, line 179: Consider adding this case to the list of error conditions for clSetDefaultDeviceCommandQueue.  The current error is:

> CL_​INVALID_​OPERATION if device does not support On-Device Queues.

api/opencl_platform_layer.asciidoc
```diff
@@ -1423,13 +1423,22 @@ include::{generated}/api/version-notes/CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT.a
...
+  | cl_device_device_enqueue_capabilities
+      | Describes device-side enqueue capabilities of the device.
+        This is a bit-field that describes one or more of the following
+        values:
+
+        {CL_DEVICE_QUEUE_SUPPORTED_anchor} - Device supports device-side enqueue and On-Device queues
+
+        {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT_anchor} - Device supports a replaceable default On-Device queue
+
+        If {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT} is set, {CL_DEVICE_QUEUE_SUPPORTED} should also be set.
```
bashbaug, line 1438: I think this should be stronger, and say that CL_DEVICE_QUEUE_SUPPORTED **must** also be set when CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT is set, not merely that it **should** be set.

api/appendix_h.asciidoc
lakshmih: Thanks for catching this, added to error conditions.

api/opencl_platform_layer.asciidoc
lakshmih: Done

api/appendix_h.asciidoc
lakshmih: Done

api/opencl_runtime_layer.asciidoc
```diff
@@ -235,6 +235,7 @@ Otherwise, it returns one of the following errors:
     by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
     required by the OpenCL implementation on the host.
+  * {CL_INVALID_OPERATION} if _device_ does not support a replacable default On-Device Queue.
```
bashbaug, line 238: Thanks, this error description looks good, but could you please move it textually up a few lines, so it is listed with the other errors for _device_?

One other consideration: I'm fine listing it separately, but I'd also consider combining this error description with the error description for when _device_ does not support any form of On-Device Queues, if you think that makes sense.

lakshmih: Will do, I did consider moving it up textually to be alongside similar errors but wasn't sure if newly added errors had to go at the end.
I'd rather list it separately. Should we make the special case more explicit though, like: "{CL_INVALID_OPERATION} if _device_ does not support a replaceable default On-Device Queue, even when it supports On-Device Queues. "

bashbaug: I don't think "even when..." part is necessary.

I actually think we should replace the previous error message vs. adding a new error condition since since the reason to return the error is because the device does not support a replaceable default On-Device Queue and is hence orthogonal to On-Device Queues generally.

It's also redundant, since we already require:

> If CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT is set, CL_DEVICE_QUEUE_SUPPORTED must also be set.

api/appendix_e.asciidoc
```diff
@@ -367,8 +367,8 @@ device:
     collective functions, such as broadcasts, scans, and reductions.
   * {CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT} to
     determine whether a device supports the generic address space.
-  * {CL_DEVICE_DEVICE_ENQUEUE_SUPPORT} to determine whether a device
-    supports device queues and device-side enqueue.
```
EwanC, line 371: Nitpick: dropped the `.` at end of sentence, and in `api/opencl_platform_layer.asciidoc` too

api/opencl_platform_layer.asciidoc
```diff
@@ -1444,13 +1444,22 @@ include::{generated}/api/version-notes/CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT.a
...
-      | Is {CL_TRUE} if the device supports device-side enqueue and on-device queues, and {CL_FALSE} otherwise.
+include::{generated}/api/version-notes/CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES.asciidoc[]
+  | cl_device_device_enqueue_capabilities
+      | Describes device-side enqueue capabilities of the device.
+        This is a bit-field that describes one or more of the following
+        values:
+
+        {CL_DEVICE_QUEUE_SUPPORTED_anchor} - Device supports device-side enqueue and On-Device queues
+
+        {CL_DEVICE_QUEUE_REPLACEABLE_DEFAULT_anchor} - Device supports a replaceable default On-Device queue
```
alycm, line 1457: The query is `CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES` and the bitfield type is `cl_device_device_enqueue_capabilities`, but the bitfield values are `CL_DEVICE_QUEUE_SUPPORTED` etc.

I.e. it's device-enqueue/device-enqueue/device-queue.   Should this be consistent?  Achieved by either by changing these values to the form `CL_DEVICE_ENQUEUE_SUPPORTED`, or by changing the query and type names.

Changing the query name ends up sounding very similar to `CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES`, which is maybe good because they are similar, but the distinction between capabilities being an immutable aspect of the implementation and properties being aspects that can be enabled is pretty subtle.

Is everyone is happy with the naming as-is then I'm happy to just resolve this comment.

api/appendix_h.asciidoc
```diff
@@ -165,6 +165,22 @@ When Device-Side Enqueue is not supported:
 
 |====
 
+When Device-Side Enqueue is supported but a replaceable On-Device Queue is not supported:
```
alycm, line 168:
```suggestion
When Device-Side Enqueue is supported but a replaceable default On-Device Queue is not supported:
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -228,7 +228,7 @@ Otherwise, it returns one of the following errors:
   * {CL_INVALID_CONTEXT} if _context_ is not a valid context.
   * {CL_INVALID_DEVICE} if _device_ is not a valid device or is not associated
     with _context_.
-  * {CL_INVALID_OPERATION} if _device_ does not support On-Device Queues.
+  * {CL_INVALID_OPERATION} if _device_ does not support a replaceable default On-Device Queue.
```
alycm, line 231: The new error condition is good but is it worth just listing both?   Doesn't actually change anything, just saves thinking about the transitive property that all devices that do not support on-device queues won't support also won't support replacing the default.

env/required_capabilities.asciidoc
```diff
@@ -21,7 +21,7 @@ modules that declare the following capabilities:
   * *Linkage*
   * *Vector16*
   * *DeviceEnqueue*
-    ** For OpenCL 2.0, OpenCL 2.1, OpenCL 2.2, or OpenCL 3.0 devices supporting Device-Side Enqueue (where {CL_DEVICE_DEVICE_ENQUEUE_SUPPORT} is {CL_TRUE}).
+    ** For OpenCL 2.0, OpenCL 2.1, OpenCL 2.2, or OpenCL 3.0 devices supporting Device-Side Enqueue (where {CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES} is not `0`.
```
alycm, line 24:
```suggestion
    ** For OpenCL 2.0, OpenCL 2.1, OpenCL 2.2, or OpenCL 3.0 devices supporting Device-Side Enqueue (where {CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES} is not `0`).
```

bashbaug:
I'm going to merge these changes as-is, to unblock header changes and the changes from a `cl_bool` true/false query to a `cl_bitfield` capabilities query.  We can always make changes to the names of these enums later.  Thanks!

api/opencl_runtime_layer.asciidoc
lakshmih: @bashbaug  and I went back and forth on this and decided to keep just one. I'll leave it as single condition for now, unless multiple reviewers want to list both.

alycm: Ah, sorry, I missed the earlier discussion on this point.

PR #325, opened 2020-06-12
update README and travis config for latexmath fonts

### LINKED ISSUE #324: Bad Latexmath Rendering in SPIR-V Environment Spec PDF
The latexmath in the OpenCL 3.0 provisional PDF spec appears to be rendering incorrectly.  See, for example, section 7.4.1. Conversion Rules for Normalized Integer Channel Data Types.  This problem seems to be toolchain-related since the same spec source renders fine in the published 2.2 specs.  I'm seeing similar errors when I build the Vulkan spec, also.  Note that the HTML rendering is fine.

I have a stripped down example that demonstrates the incorrect rendering.  I've tried the latest versions of asciidoctor, asciidoctor-pdf, and asciidoctor-mathematical, but they don't seem to fix the issue.

If we cannot get this issue resolved promptly, we should consider whether having the latexmath is worth the trouble, or if we should switch to monospace code font instead.

bashbaug:
Fixes #324.

Updates the README instructions to get fonts and an updated version of asciidoctor-mathmatical for proper latexmath rendering.

Also updates the travis spec build configuration file with these changes.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -245,6 +245,9 @@ Before building the OpenCL specs, you must install the following tools:
     This is for the PDF build, and it can still progress without it.
     Ghostscript is used to optimize the size of the PDF, so will be a lot
     smaller if it is included.
+  * ttf Fonts.
+    These are needed the PDF build for latexmath rendering.
+    See https://github.com/asciidoctor/asciidoctor-mathematical/blob/master/README.md#dependencies[Font Dependencies for asciidoctor-mathematical].
 
 The following Ruby Gems and platform package dependencies must also be
 installed.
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -257,7 +260,7 @@ parts you don't use) completely before trying to install.
   * Coderay (coderay, version: 1.1.1)
   * ttfunk (ttfunk, version: 1.5.1)
   * Asciidoctor PDF (asciidoctor-pdf, version: 1.5.0)
-  * Asciidoctor Mathematical (asciidoctor-mathematical, version 0.2.2)
+  * Asciidoctor Mathematical (asciidoctor-mathematical, version 0.3.2)
```
alycm, line 263: I tried this out myself, and installing `asciidoctor-mathematical` 0.3.2 pulled down `asciidoctor` 2.0.10.

> asciidoctor-mathematical-0.3.2 depends on asciidoctor (>= 2.0.0, ~> 2.0)

A few lines above we specify 1.5.8.  The diff of the raw HTML on the API spec from using 2.0.10 is small, suggesting that 2.0.10 isn't a problem for us, but if we go with it we should update the  `1.5.8` above.

bashbaug: Thanks - I went back and checked and things appear to be working for me with `asciidoctor-mathematical` 0.2.2, which doesn't require an updated version of `asciidoctor`.  Let's go with this for now, even though `asciidoctor` 2.0.10 seems to be working OK also.

An automated travis PDF build using this config can be found [here](https://bashbaug.github.io/OpenCL-Docs/pdf/OpenCL_Env.pdf), with a properly rendered Chapter 7.

PR #327, opened 2020-06-12
add missing kernel enum CLK_UNORM_INT_101010_2

### LINKED ISSUE #292: Possibly missing OpenCL C enum: CLK_UNORM_INT_101010_2
OpenCL 2.1 added the image channel order `CL_UNORM_INT_101010_2`, but because we never updated the OpenCL C spec for OpenCL 2.1 we never added the corresponding kernel enum `CLK_UNORM_INT_101010_2`.  I think we either need to add this enum to OpenCL C 3.0, or we need to define what other enum should be returned by `get_image_channel_order()` for a `CL_UNORM_INT_101010_2` image.

This is partially described by #183, but the enum isn't described specifically.

bashbaug:
Fixes #292.

This change adds the kernel enum `CLK_UNORM_INT_101010_2` to OpenCL C, which is returned by  `get_image_channel_data_type` for images that were created with image channel data types  `CL_UNORM_INT_101010_2`.  It also adds the expected SPIR-V mappings for these images.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -8906,6 +8906,7 @@ For query functions this may be `read_only`, `write_only` or `read_write`.
       `CLK_UNORM_SHORT_565` +
       `CLK_UNORM_SHORT_555` +
       `CLK_UNORM_INT_101010` +
+      `CLK_UNORM_INT_101010_2` +
       `CLK_SIGNED_INT8` +
       `CLK_SIGNED_INT16` +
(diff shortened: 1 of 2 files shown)


alycm:
I didn't merge this because @AnastasiaStulova review was requested.  Was that a specific request or just spreading out some review requests?

AnastasiaStulova:
LGTM!

PR #336, opened 2020-06-30
deprecate scalar inputs for any and all built-in functions

### LINKED ISSUE #178: Specify semantics of any() and all() for scalar inputs
The OpenCL C specification describes the `any()` and `all()` built-in functions as follows:

Function | Description
-- | --
int **any**(igentype _x_) | Returns 1 if the most significant bit in any component of _x_ is set; otherwise returns 0.
int **all**(igentype _x_) | Returns 1 if the most significant bit in all components of _x_ is set; otherwise returns 0.

This implies that the semantics of these functions have only been defined for vector types. What are the semantics for scalar types?

I would expect that `any(0) == false`, `all(0) == false`, `any(-1) == true` and `all(-1) == true)`, but it is not clear what `any(1)` and `all(1)` would evaluate to. Should these compile? (Scalar types are listed for _igentype_ in the second paragraph of _6.13.6 Relational Functions_.)

bashbaug:
This change deprecates scalar inputs for the **any** and **all** built-in functions.

Fixes #178.

### DIFF
--- a/scripts/gen_dictionaries.py
+++ b/scripts/gen_dictionaries.py
@@ -182,7 +182,7 @@ if __name__ == "__main__":
             category = type.get('category')
             if category == 'struct':
                 name = type.get('name')
-                print('found type: ' +name)
+                #print('found type: ' +name)
 
                 # Create a variant of the name that precedes underscores with
                 # "zero width" spaces.  This causes some long names to be
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -5472,9 +5472,13 @@ not a number (NaN) and the argument type is a vector.
      float is set else returns 0.
 | |
 | int *any*(igentype _x_)
+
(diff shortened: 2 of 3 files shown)


scripts/gen_dictionaries.py
```diff
@@ -182,7 +182,7 @@ def GetFooter():
             category = type.get('category')
             if category == 'struct':
                 name = type.get('name')
-                print('found type: ' +name)
+                #print('found type: ' +name)
```
StuartDBrady, line 185: This change seems unrelated.

bashbaug: Yes, this removes some debug output leftover from https://github.com/KhronosGroup/OpenCL-Docs/pull/298 that I noticed while making these changes.

StuartDBrady: Thanks. I just wanted to check that this change was intentional.

PR #338, opened 2020-07-01
Add a device query to report the latest CTS version tested

### LINKED ISSUE #304: Add a query to report the last tested version of the CTS
The OpenCL WG has agreed on the 2020/05/26 call that OpenCL 3.0 devices will be required to report the last version of the CTS that they have been tested against.

Detailed proposal to come once we've agreed the exact versioning scheme.

kpet:
Fixes #304

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1483,6 +1483,13 @@ include::{generated}/api/version-notes/CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTI
         (Refer also to {clGetKernelWorkGroupInfo} where {CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE}
         can return a different value to {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} which may be more optimal.)
 
+| {CL_DEVICE_LATEST_CTS_VERSION_PASSED_anchor}
+
+include::{generated}/api/version-notes/CL_DEVICE_LATEST_CTS_VERSION_PASSED.asciidoc[]
+  | char[]
+      | Returns the latest version of the conformance test suite that this device
+        has fully passed in accordance with the official conformance process.
+
 |====
 
 4::
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1483,9 +1483,9 @@ include::{generated}/api/version-notes/CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTI
         (Refer also to {clGetKernelWorkGroupInfo} where {CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE}
         can return a different value to {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} which may be more optimal.)
 
-| {CL_DEVICE_LATEST_CTS_VERSION_PASSED_anchor}
+| {CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED_anchor}
 
-include::{generated}/api/version-notes/CL_DEVICE_LATEST_CTS_VERSION_PASSED.asciidoc[]
+include::{generated}/api/version-notes/CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED.asciidoc[]
   | char[]
       | Returns the latest version of the conformance test suite that this device
         has fully passed in accordance with the official conformance process.
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1185,7 +1185,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x106F"        name="CL_DEVICE_OPENCL_C_FEATURES"/>
         <enum value="0x1070"        name="CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES"/>
         <enum value="0x1071"        name="CL_DEVICE_PIPE_SUPPORT"/>
-            <unused start="0x1072" end="0x107F" comment="Reserved for cl_device_info"/>
+        <enum value="0x1072"        name="CL_DEVICE_LATEST_CTS_VERSION_PASSED"/>
+            <unused start="0x1073" end="0x107F" comment="Reserved for cl_device_info"/>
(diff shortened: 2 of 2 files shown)


xml/cl.xml
```diff
@@ -1177,7 +1177,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x1067"        name="CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE"/>
         <enum value="0x1068"        name="CL_DEVICE_WORK_GROUP_COLLECTIVE_FUNCTIONS_SUPPORT"/>
         <enum value="0x1069"        name="CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT"/>
-            <unused start="0x106A" end="0x106E" comment="Reserved for upcoming KHR extension"/>
+        <enum value="0x106A"        name="CL_DEVICE_LATEST_CTS_VERSION_TESTED"/>
```
bashbaug, line 1180: I think we should use enum `0x1072` for this query, since enum `0x106A` is already assigned for an upcoming KHR extension.

api/opencl_platform_layer.asciidoc
```diff
@@ -1483,6 +1483,13 @@ include::{generated}/api/version-notes/CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTI
         (Refer also to {clGetKernelWorkGroupInfo} where {CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE}
         can return a different value to {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} which may be more optimal.)
 
+| {CL_DEVICE_LATEST_CTS_VERSION_TESTED_anchor}
+
+include::{generated}/api/version-notes/CL_DEVICE_LATEST_CTS_VERSION_TESTED.asciidoc[]
+  | char[]
+      | Returns the latest version of the conformance test suite that this device
+        has been tested against.
```
jrprice, line 1491: Maybe a little pedantic, but should we make it explicit that the device needs to have passed all of the tests in the version reported here? The phrase "tested against" doesn't necessarily imply this.

xml/cl.xml
kpet: Well, Github should really be the place for reserving enums. It's not reasonable to expect people to look at all open MRs/branches in Gitlab and scan for assigned enums IMHO.

api/opencl_platform_layer.asciidoc
kpet: Yes, I also thought we should ultimately reference the exact process here but it doesn't exist yet. I could say "successfully tested" but this could still be interpreted as "the tests ran end to end". I'll try to find something that's reasonably unambiguous.

xml/cl.xml
kpet: So, this actually was a public extension for which we hadn't reserved enums yet (I've opened https://github.com/KhronosGroup/OpenCL-Docs/pull/342). I'll change the enum.

kpet:
@bashbaug  @jrprice I believe I've addressed both comments. PTAL.

api/opencl_platform_layer.asciidoc
```diff
@@ -1483,6 +1483,13 @@ include::{generated}/api/version-notes/CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTI
         (Refer also to {clGetKernelWorkGroupInfo} where {CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE}
         can return a different value to {CL_DEVICE_PREFERRED_WORK_GROUP_SIZE_MULTIPLE} which may be more optimal.)
 
+| {CL_DEVICE_LATEST_CTS_VERSION_PASSED_anchor}
+
+include::{generated}/api/version-notes/CL_DEVICE_LATEST_CTS_VERSION_PASSED.asciidoc[]
+  | char[]
+      | Returns the latest version of the conformance test suite that this device
```
Jeremy-Kemp, line 1490: Do we need to state "latest" here (and in the enums)?

I think it should only matter about reporting the specific version that the implementation was tested against.

alycm: I think the intention was to the report "the latest version that was tested", which may not be "the latest version that exists".

I tried a few rewrites in my head, but they didn't seem better than the current wording.

Jeremy-Kemp: >I tried a few rewrites in my head, but they didn't seem better than the current wording.

Thats a good exercise and I also can't come up with a meaningful change either.

bashbaug: The analogous Vulkan physical device property is [conformanceVersion](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VkConformanceVersion), so if we wanted to be similar in OpenCL we could consider `CL_DEVICE_CONFORMANCE_VERSION`.  This also has the slight advantage of not requiring the abbreviation `CTS`, which may not be universally known.

I don't have a very strong opinion about this and I'm fine keeping `CL_DEVICE_LATEST_CTS_VERSION_PASSED`.

nikhiljnv:
> LGTM, I'm okay to not specify a format to the string, the information is for humans.

I think we at least need to agree on fixed format which must be present with additional information that may not need a format. Otherwise, writing a CTS test would be difficult to validate the contents reported by the query.
I believe we have something similar for CL_PLATFORM_VERSION which requires prefix OpenCL x.y followed by platform specific information.

api/opencl_platform_layer.asciidoc
alycm: I find `CL_DEVICE_CONFORMANCE_VERSION` easier to read, but I also don't have a strong opinion.

bashbaug:
There are two open questions in this PR.  I'd like to confirm we should merge these changes as-is, or whether we plan to address them first:

1. Are we OK with the name `CL_DEVICE_LATEST_CTS_VERSION_PASSED`?  One proposed alternative is `CL_DEVICE_CONFORMANCE_VERSION`.

2. Do we need to define the content of the string query?  If not, we cannot write a CTS test to validate the content of the query, though we can include it in the output of, say, the computeinfo CTS test for human reviewers.

bashbaug:
Latest changes LGTM. Merging as we discussed on the July 21st teleconference.

PR #346, opened 2020-07-08
add clSetContextDestructorCallback

### LINKED ISSUE #187: Need for clSetContextReleaseCallback?
When creating a `cl_context` using either `clCreateContext` or `clCreateContextFromType`, a callback can be given. The function can optionally be given a `user_data` pointer, that will be provided to the callback when it is invoked. The callback can be called several time, and potentially after the user called `clReleaseContext`.

If the `user_data` pointer is dynamically allocated, the user has no way to know when it is safe to free this pointer. A `clSetContextReleaseCallback` could be used to solve this issue. The signature would be similar to `clSetProgramReleaseCallback`.

```c
cl_int clSetContextReleaseCallback(
    cl_context context,
    void (CL_CALLBACK* pfn_notify)(cl_context context, void* user_data),
    void* user_data);
```

bashbaug:
Fixes #187.

This PR adds a new API to register a callback for a context that is called when the context is destroyed.

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -2205,3 +2205,52 @@ Otherwise, it returns one of the following errors:
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
     required by the OpenCL implementation on the host.
 --
+
+[open,refpage='clSetContextDestructorCallback',desc='Registers a destructor callback function with a context.',type='protos']
+--
+To register a callback function with a context that is called when
+the context is destroyed, call the function
+
+include::{generated}/api/protos/clSetContextDestructorCallback.txt[]
+include::{generated}/api/version-notes/clSetContextDestructorCallback.asciidoc[]
+
+  * _context_ specifies the OpenCL context to register the callback to.
+  * _pfn_notify_ is the callback function to register.
+    This callback function may be called asynchronously by the OpenCL
+    implementation.
+    It is the application's responsibility to ensure that the callback function
+    is thread-safe.
+    The parameters to this callback function are:
+  ** _context_ is the OpenCL context being deleted.
+     When the callback function is called by the implementation, this context
+     is no longer valid.
+     _context_ is only provided for reference purposes.
+  ** _user_data_ is a pointer to user-supplied data.
+  * _user_data_ will be passed as the _user_data_ argument when _pfn_notify_ is
+    called.
+    _user_data_ can be `NULL`.
+
+Each call to {clSetContextDestructorCallback} registers the specified
+callback function on a destructor callback stack associated with _context_.
+The registered callback functions are called in the reverse order in
+which they were registered.
+If a context callback function was specified when _context_ was created,
+it will not be called after any context destructor callback is called.
+Therefore, the context destructor callback provides a mechanism for an
+application to safely re-use or free any _user_data_ specified for the
+context callback function when _context_ was created.
+
+// refError
+
+{clSetContextDestructorCallback} returns {CL_SUCCESS} if the function is
+executed successfully.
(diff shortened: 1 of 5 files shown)


api/opencl_platform_layer.asciidoc
```diff
@@ -2205,3 +2205,52 @@ Otherwise, it returns one of the following errors:
...
+callback function on a destructor callback stack associated with _context_.
+The registered callback functions are called in the reverse order in
+which they were registered.
+If a context callback function was specified when _context_ was created,
+it will not be called after any context destructor callback is called.
+Therefore, the context destructor callback provides a mechanism for an
+application to safely re-use or free any _user_data_ specified for the
+context callback function when _context_ was created.
+
+// refError
```
kpet, line 2243: Did you intend to add that line?

bashbaug: Confirming: this is about:

`// refError`

If so, yes, this line is included in the description of each OpenCL API.  I believe it is used by the toolchain that generates the [OpenCL reference pages](https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/).

bashbaug:
> This also begs the (separate) question: do we want to provide an alias to `clSetProgramReleaseCallback` for consistency? I'm thinking of `clSetProgramDestructorCallback`.

Since we don't support program scope constructors and destructors in OpenCL 3.0, I think a better thing to do is to deprecate `clSetProgramReleaseCallback` and add back `clSetProgramDestructorCallback` if and only if we add functionality that requires it.

kpet:
>Since we don't support program scope constructors and destructors in OpenCL 3.0, 

We should clarify this in the spec BTW (Appendix H is pretty clear but the normative section just says "Missing before 2.2" which implies it's supported in 3.0). I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/347.

>I think a better thing to do is to deprecate clSetProgramReleaseCallback and add back clSetProgramDestructorCallback if and only if we add functionality that requires it.

Yes, I was thinking of the end state with the feature added back in, not specifically of 3.0 but I think it makes sense to deprecate in 3.0. I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/348.

keryell:
Since it is similar to some discussion we had in SYCL https://gitlab.khronos.org/sycl/Specification/-/issues/289 I wonder how this feature can be implemented safely.
For example what happens if the OpenCL runtime decide to destruct the context at exit, after `main()` exits?

bashbaug:
Interesting discussion, thanks for passing this along.  FWIW, I don't think the SYCL proposal ([link](https://github.com/codeplaysoftware/standards-proposals/blob/master/extended-context-destruction/index.md)) is exactly the same as this feature because the proposed SYCL deleter gets called *before* the SYCL context is deleted, whereas this callback gets called *after* the OpenCL context is deleted, but there's definitely a fair amount of overlap.

> For example what happens if the OpenCL runtime decide to destruct the context at exit, after `main()` exits?

I think the simplest answer is that this behavior won't be conformant.  I'm hand-waving a little, but I imagine we would have a conformance test that ensures the context destructor callback is properly called, similar to this conformance test for `clSetMemObjectDestructorCallback`:

https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/api/test_mem_objects.cpp

Does that sound reasonable?

bashbaug:
Merging as discussed on the July 16th teleconference.

keryell:
Yes, there is a difference between when a `context` is deletable and when it has been actually discarded.
I am curious about the use case.
For example you create a static global C++ object that set up some OpenCL context in the constructor.
Probably it might be too late to be useful if the context destruction call-back is called on this static object destruction...

Issue #347, opened 2020-07-09
Clarify support for program global constructors/desctructors in 3.0


kpet:
The functionality is specified as "Missing before 2.2" but the normative section of the specification says nothing about the behaviour in 3.0. Appendix H makes it clear that 3.0 doesn't support the following:
- `CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT`
- `CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT`
- `clSetProgramReleaseCallback`

This should probably be extended to a review of all behaviours described in Appendix H that are not described in the normative sections of the specification.

bashbaug:
To be clear, none of this functionality was removed in OpenCL 3.0, so it's not quite accurate to say that "3.0 doesn't support" the queries or the APIs.  It is fair to say that the the queries and the callback API are not very useful in OpenCL 3.0 though, given that OpenCL 3.0 does not support program scope constructors and destructors.

Do you have a proposal for what you would like to clarify?  If deprecate the queries and the callback API (#348) I think that would improve clarity?

alycm:
Agreed that none of these things are "removed", and the main body descriptions are currently technically correct, but certainly give no indication that only one of the described paths will ever apply to OpenCL 3.0 devices.

Given program scope variables are hard "not supported" in 3.0 rather than "optional", deprecating all three of these would certainly clearly mark them out.  I don't think they need a stronger "removed" wording because they are still present, they're just hard-coded to the "not supported" path as far as the spec is concerned.

PR #351, opened 2020-07-14
deprecate APIs and queries for program init and cleanup kernels

### LINKED ISSUE #348: Deprecation of clSetProgramReleaseCallback in 3.0
The API call was added in 2.2 and is not supported in 3.0. We're likely to want to rename it to `clSetProgramDesctructorCallback` when we reintroduced the functionaltiy (see #346). Should we deprecate `clSetProgramReleaseCallback` in 3.0?

bashbaug:
This PR describes the changes needed to deprecate the APIs and queries for Program Initialization and Cleanup Kernels, also known as Program Scope Constructors and Destructors.

The "deprecated by" text will be inserted automatically into the OpenCL specs with the changes to the XML grammar.

edit:

This PR fixes #348 

A slightly different change is needed to address #347.  I will create a separate PR for it shortly.

### DIFF
--- a/api/appendix_e.asciidoc
+++ b/api/appendix_e.asciidoc
@@ -406,6 +406,16 @@ when creating buffers, images, pipes, samplers, and command queues:
   * {CL_SAMPLER_PROPERTIES}
   * {CL_QUEUE_PROPERTIES_ARRAY}
 
+// GitHub issue #348
+Program Initialization and Clean-Up kernels are not supported in OpenCL
+3.0 due to implementation complexity and lack of demand.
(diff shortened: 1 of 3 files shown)


bashbaug:
I went back to the recording from Thursday's call to be sure I didn't mis-remember and confirmed that we did agree to deprecate these APIs and queries.  I think this PR can be merged anytime.  I've also updated the PR description to indicate that merging these changes will fix #348 

I will create a separate PR that describes the behavior of these APIs and queries for an OpenCL 3.0 device, which will address #347.

PR #352, opened 2020-07-15
Introduce dedicated cl_properties type

### LINKED ISSUE #323: Type inconsistency between cl_mem_properties and cl_pipe_properties
`cl_pipe_properties` is typedef'ed to `intptr_t` whose size can vary, `cl_mem_properties` to `cl_bitfield` that has a fixed size. Given that pipes are memory objects, this is a bit confusing and may increase the complexity of some implementations.

Looking around, we seem to have a mix of `intptr_t` and `cl_bitfield` for properties. Is there a good reason for this? Should we change `cl_mem_properties`? What should the policy be going forward?

kpet:
Fixes #323

### DIFF
--- a/ext/cl_khr_create_command_queue.asciidoc
+++ b/ext/cl_khr_create_command_queue.asciidoc
@@ -52,7 +52,7 @@ cl_command_queue clCreateCommandQueueWithPropertiesKHR(
 
 [source,c]
 ----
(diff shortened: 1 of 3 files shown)


kpet:
@bashbaug I've also modified the cl_intel_create_buffer_with_properties extension, assuming you'd want to use the new type.

If people are happy with this change, I'll push the matching headers change.

Issue #82, opened 2019-05-28
Use of extensions in the kernel language


AnastasiaStulova:
I would like to summarize the intended use of extension (after my discussion with @bashbaug) and if possible adjust the implementation.

**A.** The way to check that the extension is being supported is by predefined extension macro i.e.

```
#if defined(cl_khr_subgroups)
size_t i = get_sub_group_local_id();
#endif
```

**B.** Some extensions might require extra compiler support or alter traditional compilation phase. When different compilation functionality is required it can be enabled using pragma. I.e.

```
#pragma OPENCL EXTENSION cl_khr_fp64 : enable
cos(0.1); //use cos overload with double.
#pragma OPENCL EXTENSION cl_khr_fp64 : disable
cos(0.1); //use cos overload with float.
```

**Observation**: the use of **get_sub_group_local_id** doesn't require any compilation support and can be handled as regular include and therefore no extension pragma is needed to use the function. The same applies to _cos_. Only the use of double literal requires enabling pragma in the above examples.

This doesn't align with upstream implementation in Clang. The implementation requires extension pragma for some extensions that are not part of core support in corresponding spec version i.e.  pragma will be required for _subgroups_ up to OpenCL v2.0. The implementation is quite inconsistent though. For example some extensions from ARM and Intel require pragma but extensions from AMD don't seem to be requiring it.

**Actions:**
**1.** Gather the list of extensions that require pragma
**2.** Simplify compilation by not requiring pragma where it is not necessary.

Backwards compatibility: Old code will continue to compile. Pragma will be simply ignored by the new rules.

bashbaug:
This sounds reasonable to me - I think I hinted at something similar here https://github.com/KhronosGroup/OpenCL-Docs/issues/19#issuecomment-436463265.

> Backwards compatibility: Old code will continue to compile.

Is there also a risk that new code will fail on older drivers.  Presumably some compilers currently require enabling pragmas even if others do not, so perhaps we should have a single control to enable "simpler compilation"?

AnastasiaStulova:
Yes old drivers will continue requesting the extension to be enabled.

kpet:
As discussed in the 2020/05/22 tooling call, we probably want to resolve this for CL3.0.

AnastasiaStulova:
I looked at Clang upstream and those are estensions that are actually used in compilation:
- _cl_khr_fp16_ enables an extra fp type 'half'
[...]

PR #355, opened 2020-07-20
[OpenCL C 3.0] Clarify use of extension pragma (#82).

AnastasiaStulova:
This change adds clarification for the compiler directive
that enables and disables the extensions.

Summary:
- Add high-level description of OpenCL specific pragma
  directives into the core spec.
- Clarify the use of extension pragma in the extension spec.
- Clarify the need of extension pragma directive in cl_khr_fp64
  and cl_khr_fp16.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -3180,6 +3180,19 @@ following forms whose meanings are described elsewhere:
 #pragma OPENCL EXTENSION all : behavior
 ----------
 
+`FP_CONTRACT` directive can be used to specify contraction for some floating
+point operations (more details can be found in <<floating-point-macros-and-pragmas,
+the floating-point macros and pragmas section>>).
+
+`EXTENSION` directive is used to enable logic in the compiler that deviates
+from the standard behavior described in this document. Extensions are not
+guaranteed to be supported by all implementations and their presence can be checked
+by availability of specific predefined macros. More information can be found in
+<<opencl-extension-spec,section 9.1 in the OpenCL Extension Specification>>.
+Pragma directives are only needed for extensions that require changes in the
+language such that the source code is interpreted differently during the
+compilation.
+
 The following predefined macro names are available.
 
 `+__FILE__+` ::
@@ -3719,8 +3732,8 @@ compile-time constant expression.
 The attribute syntax can be extended for standard language extensions and
 vendor specific extensions.
 Any extensions should follow the naming conventions outlined in the
-introduction to <<opencl-extension-spec,section 9 in the OpenCL 2.0
-Extension Specification>>.
+introduction to <<opencl-extension-spec,section 9 in the OpenCL Extension
+Specification>>.
 
 Attributes are intended as useful hints to the compiler.
 It is our intention that a particular implementation of OpenCL be free to
--- a/ext/introduction.asciidoc
+++ b/ext/introduction.asciidoc
@@ -121,7 +121,16 @@ The initial state of the compiler is as if the directive
 ----
 
 was issued, telling the compiler that all error and warning reporting must
-be done according to this specification, ignoring any extensions.
+be done according to the core specification, ignoring any extensions.
+
+Pragma directives are only required for extensions that change the compiler
+behavior i.e. the source code is interpreted differently if the pragma
+directive is used. Each extension should specify whether it adds the
+directive or not. However, if it is not explicitly specified it is assumed
+that the pragma directive is optional for the extension.  As unknown
+directives are ignored by the compiler the use of pragma with extensions
+that do not support them will not cause the compilation to fail but might
+result in a diagnostic.
 
 Every extension which affects the OpenCL language semantics, syntax or adds
 built-in functions to the language must create a preprocessor `#define` that
--- a/ext/cl_khr_fp16.asciidoc
+++ b/ext/cl_khr_fp16.asciidoc
@@ -9,6 +9,15 @@ This section describes the *cl_khr_fp16* extension.
 This extension adds support for half scalar and vector types as built-in
 types that can be used for arithmetic operations, conversions etc.
 
+This extension modifies behavior of OpenCL C and it has to be
+activated/deactivated using pragma `EXTENSION` directive.
+
+[source,c]
+----
+#pragma OPENCL EXTENSION cl_khr_fp16 : enable
+#pragma OPENCL EXTENSION cl_khr_fp16 : disable
+----
+
 === General information
 
 ==== Version history
--- a/ext/cl_khr_fp64.asciidoc
+++ b/ext/cl_khr_fp64.asciidoc
@@ -8,6 +8,15 @@
 This section describes the *cl_khr_fp64* extension.
 This extension became an optional core feature in OpenCL 1.2.
 
+This extension modifies behavior of OpenCL C and it has to be
+activated/deactivated using pragma `EXTENSION` directive.
+
+[source,c]
+----
+#pragma OPENCL EXTENSION cl_khr_fp64 : enable
+#pragma OPENCL EXTENSION cl_khr_fp64 : disable
+----
+
 === General information
 
 ==== Version history


ext/introduction.asciidoc
```diff
@@ -121,7 +121,16 @@ The initial state of the compiler is as if the directive
...
 
 was issued, telling the compiler that all error and warning reporting must
-be done according to this specification, ignoring any extensions.
+be done according to the core specification, ignoring any extensions.
+
+Pragma directives are only required for extensions that change the compiler
+behavior i.e. the source code is interpreted differently if the pragma
+directive is used. Each extension should specify whether it adds the
+directive or not. However, if it is not explicitly specified it is assumed
+that the pragma directive is optional for the extension.  As unknown
```
bashbaug, line 130: This PR updated the descriptions of `cl_khr_fp16` and `cl_khr_fp64` to indicate that they explicitly require pragma directives to enable the extensions.  Are there other extensions that also require pragma directives, even if they are vendor extensions?  If so, to the best of our ability I think we need to update all of the extensions that require a pragma, or we run the risk of breaking shipping implementations.

For example, it looks like the Intel extension `cl_intel_device_side_avc_motion_estimation` requires a pragma (rightly or wrongly): https://godbolt.org/z/T4TGYv

Alternatively, we could specify that an extension pragma is required by default unless it is explicitly specified to be optional.  This is more conservative, and I suspect that fewer extensions will require a pragma than extensions where the pragma is optional or not required, but this is also less likely to break shipping implementations.

AnastasiaStulova: > Are there other extensions that also require pragma directives, even if they are vendor extensions? If so, to the best of our ability I think we need to update all of the extensions that require a pragma, or we run the risk of breaking shipping implementations.

Sure, I am happy to update those if we categorize them first.

> Alternatively, we could specify that an extension pragma is required by default unless it is explicitly specified to be optional. This is more conservative, and I suspect that fewer extensions will require a pragma than extensions where the pragma is optional or not required, but this is also less likely to break shipping implementations.

Sure. We can also change the default behavior. From the tooling implementation I would prefer to keep the default behavior to not require the pragma but accept it because:
1. Most of extensions won't need pragma in practice so not needing the pragma will likely be a common case.
2. We can simply implementation of many extensions straight away.

But it certainly has an implication on backward compatibility so if some vendors feel there is an issue to update we can fall back to the conservative approach.

MathiasMagnus:
Coming from an end-user perspective trying to do the minimal amount of thinking that results in portable code, if there is an accompanying pragma to any extension I'm using, I'll put it there, whether it's mandated or not. That is the defensive behavior. If some 1.2, 2.x extensions mandated it, but 3.0 extensions allow omission, I'll still add the pragma cause then I don't have to go back and check when the pragma was introduced.

If the spec will allow omission of new extensions going forward, due to mighty useful non-language altering 1.x extensions such as printf still having an associated pragma, people wanting to write portable code will likely still default to the no-brainer policy of stamping out the pragma anyway.

AnastasiaStulova:
@MathiasMagnus If we accept the pragma optionally by default it will cover your use case isn't it? But also allows more flexibility for implementations or developers that prefer to be more specific. As a matter of fact unknown pragmas are always just ignored by the compiler so feel free to add them as you feel.

FYI the suggestion is not to make the extensions behave differently in various versions of OpenCL C. The idea is to fix the behaviour for each of them across all OpenCL C versions. I guess the main benefit is to simplify implementations of future extensions but we  might be able to do something better for some existing ones too.

As a matter of fact clang has extensions that don't even have kernel language changes. Not clear why? I don't think we can scale if we add every single extensions into clang. We have quite a number of them.

MathiasMagnus:
I think it will cover it, yes.

bashbaug:
I found an odd case where enabling the int64 atomic extensions is required to use the 64-bit C11 atomic functions while investigating a separate issue: https://godbolt.org/z/Kb3Pxf

My main reason for pointing this out is to say that specifying extensions pragmas to be optional after the fact is dangerous and runs the risk of breaking shipping implementations in unpredictable ways.  I'd feel much more comfortable requiring a pragma by default, at least for shipping extensions, and then explicitly stating when a pragma is optional for newer extensions.

To continue @MathiasMagnus 's comment though, I wonder if there is a better way:  What if a newer compiler supported a mechanism to enable all (or at least most?) extensions in one shot, without listing them individually?

One way to do this is to support:

```c
#pragma OPENCL EXTENSION all : enable
```

This is explicitly **disallowed** today, though disabling "all" extensions is allowed.

Since enabling "all" extensions is also debatably dangerous, we could instead define a subset of extensions - the ones that don't impact compiler internals - and support enabling them in one shot.  I haven't thought of the right noun for this, but the basic idea is something like:

```c
#pragma OPENCL EXTENSION basic_subset : enable
```

The key is to add a new feature to simplify enabling extensions that could apply uniformly to new extensions and old extensions, vs. specifying rules for each extension.  I _think_ this would preserve compatibility with shipping kernels and implementations, while providing a simpler option for newer kernels targeting newer compilers.

What do you think?

AnastasiaStulova:
> I found an odd case where enabling the int64 atomic extensions is required to use the 64-bit C11 atomic functions while investigating a separate issue: https://godbolt.org/z/Kb3Pxf

This is just one example out of many. I don't know what adding extra types have to do with the pragma directives. We could easily make them available if extensions are supported. And their presence can be checked using extension macros. It is very sad that with time we have added so many extensions with pragmas that do nothing but complicate developer's life and add extra code in the implementation. :(

> My main reason for pointing this out is to say that specifying extensions pragmas to be optional after the fact is dangerous and runs the risk of breaking shipping implementations in unpredictable ways. I'd feel much more comfortable requiring a pragma by default, at least for shipping extensions, and then explicitly stating when a pragma is optional for newer extensions.

Yes, this should be more conservative i.e. in case implementations don't provide the pragma it would just be ignored at the price of an extra warning. However, I think we should encourage to specify the use of pragma explicitly and especially for the future extensions. Alternatively, we can omit the default behavior and just rely on each extension to explicitly specify the pragma use. If nothing is specified about the pragma it will just comply to old behavior. I would still go through all the published non-vendor extensions and attempt to define the behavior explicitly.

> Since enabling "all" extensions is also debatably dangerous, we could instead define a subset of extensions - the ones that don't impact compiler internals - and support enabling them in one shot. 

I presume this has been discussed already?

> The key is to add a new feature to simplify enabling extensions that could apply uniformly to new extensions and old extensions, vs. specifying rules for each extension. I think this would preserve compatibility with shipping kernels and implementations, while providing a simpler option for newer kernels targeting newer compilers.
>What do you think?

Interesting, I am keen to explore this idea further. Do we want to specify this list or leave it defined by implementation to allow the inclusion of vendor extensions too? We could also think of a combined approach. I guess this would be an additional utility though? I would still attempt to specify the behavior per extension too.

AnastasiaStulova:
I have looked at this topic for a bit longer and I  think we should require to describe how the pragma changes the compiler state otherwise it is a black magic for the developers and tooling so it doesn't really help to just state whether or not the pragma is needed.

If we look at other pragmas i.e. FP_CONTRACT, it is explicitly explained what it actually does. For the extensions however, I can only find some information for `cl_khr_fp16`. I believe that the lack of information has led to misunderstanding and misinterpretations that has resulted in inadequate implementation that I would like to correct as much as possible.

I feel that many extensions have been implemented incorrectly wrt pragma. For example, we not only add pragma for nearly every extension we also
guard the use of identifiers by the pragma i.e. code using types or functions from the extensions would only be compiled if pragma enable was added. However, I don't see how it is doing anything but bringing extra complexity and inconvenience.
Perhaps, the confusion was due to some early extensions for example cl_khr_fp64 was used to alter the state of parser that would map double from invalid reserved identifier into a valid one or it would change the floating-point literal into a double type constant instead of float. Although the use of double type could still be accepted without the need of pragma.

Currently, there are plenty of extensions that add new types or functions before parsing the source code from extensions supported by the target. However, this is done during the compiler setup, before parsing. When pragma is actually parsed nothing changes for such extensions because the types have already been added previously. Generally, C/C++ doesn't allow to disable the identifiers. So we wouldn't have a way to load and unload them dynamically w/o significant changes in the architecture of the compiler. The only thing we do with pragma for those extensions is to accept/reject the code that uses the identifiers i.e. types or functions when pragma enabled/disable.

Earlier example illustrates what I am describing here:
> I found an odd case where enabling the int64 atomic extensions is required to use the 64-bit C11 atomic functions while investigating a separate issue: https://godbolt.org/z/Kb3Pxf

I don't see a need for pragma in those use cases. And if nobody else can highlight where this could be useful the right action for upstream implementation would be to just remove the need for pragmas and also the extensions that don't need the compiler support from clang. Note that this will largely not affect backward compatibility of existing code, but an extra warning for unknown pragmas might appear in some source code. There has never been clear description of how and where the pragma is used so perhaps a warning won't be surprising behavior after all.

ext/introduction.asciidoc
AnastasiaStulova: @bashbaug I have updated this PR without specifying the default behavior at all but however mandating each extension to provide the description of the pragma use.

The reason why I have changed it is that I personally think it is not sufficient to say whether the pragma is needed or not but it should also be explained what changes it makes. This is needed both for the compiler and application developers.

I have also added an explanation of pragma use in **cl_khr_fp64**. I have not added pragma to any other extension as I still doubt it is needed anywhere else. Note that I don't think guarding the use of types or functions that are available if an extension is supported by a pragma is useful and I have therefore excluded those (see my earlier comment).

OpenCL_C.txt
```diff
@@ -3180,6 +3180,20 @@ following forms whose meanings are described elsewhere:
...
+`EXTENSION` directive is used to enable logic in the compiler that deviates
+from the standard behavior described in this document. Extensions are not
+guaranteed to be supported by all implementations and their presence can be
+checked by availability of specific predefined macros. More information can be
+found in <<opencl-extension-spec,section 1.2 in the OpenCL Extension
+Specification>>. Pragma directives are only needed for extensions that require
+changes in the language such that the source code is interpreted differently
+with pragma and without. Each extensions should specify whether it adds support
+for pragma and what behavior will be changed due to its use in the source.
+
```
etomzak, line 3196: Could this text be further clarified with the cases where pragma directives have no effect? A vendor may have platform/runtime extensions, and I assume the intent is that pragma directives cannot and should not be used to enable or disable those.

As a contrived example, I assume `pragma OPENCL EXTENSION cl_khr_icd : disable` is nonsensical, but it seems that the spec does actually allow for someone to put that in an OpenCL-C kernel.

etomzak:
Should there also be some clarification in here about the interaction between pragma directives and preprocessor macros? For example, AIUI:

```c
#pragma OPENCL EXTENSION cl_khr_fp64 : disable
#ifdef cl_khr_fp64
// Code here is actually compiled
#endif
```

I.e., I think the intent is that preprocessor macros communicate to the OpenCL-C kernel what extensions are available, and kernels use pragmas to indicate which extensions they want to use. What might not be immediately obvious to a developer is that the preprocessor runs before pragmas are evaluated, so a `pragma ... disable` doesn't undef the extension macro.

OpenCL_C.txt
AnastasiaStulova: > Could this text be further clarified with the cases where pragma directives have no effect? A vendor may have platform/runtime extensions, and I assume the intent is that pragma directives cannot and should not be used to enable or disable those.

I would say that there shouldn't be any pragma with no effect at all. Becasue it is simply redundant. And therefore the current clarification says that the spec should explain what the pragma chages/modifies from the core specification. 

We could of course add something like that the pragma that doesn't provide any clarification is not guaranteed to be accepted. The important is that I would prefer to avoid saying that the pragma can alway be accepted but with no effect because it mean it has to be implemented in tooling  without any benefit for the developers.

> As a contrived example, I assume pragma OPENCL EXTENSION cl_khr_icd : disable is nonsensical, but it seems that the spec does actually allow for someone to put that in an OpenCL-C kernel.

FYI I looked into this in more detail and the only conclusion I have made is that sadly it is just a result of copy-paste. I have provided more details in https://github.com/KhronosGroup/OpenCL-Docs/issues/82#issuecomment-724703376 if you are interested. I hope that the pragma from those extensions can simply be removed.

AnastasiaStulova:
I have made some investigation and I realized that the majority of extensions will have a section with the pragma in the same exact format regardless of whether they even add any change to OpenCL C. However, neither individual extensions nor extensions spec or language spec explain whether and how the pragma is to be used. Moreover the extensions spec and language spec doesn't contain the pragmas for any extension, so only individual extension specs do. That means that neither upstream implementation in Clang nor existing kernels using pragmas can be conformant. My guess is that the pragma in the specification is a result of copy-paste. 
These are the only extensions that somehow refer to the potential use of pragmas: cl_khr_fp16, cl_khr_int64_base_atomics, cl_khr_int64_extended_atomics. Those are the wording in the spec.

>     Unless the cl_khr_fp16 extension is supported and has been enabled.
> 
>     The half scalar and vector types can only be used if the cl_khr_fp16 extension is supported andhas been enabled. The double scalar and vector types can only be used if double precision is supported.
> 
>     The atomic_long and atomic_ulong types are supported if the cl_khr_int64_base_atomics andcl_khr_int64_extended_atomics extensions are supported and have been enabled.
> 
>     The atomic_double type is only supported if double precision is supported and the cl_khr_int64_base_atomics and cl_khr_int64_extended_atomics extensions are supported andhave been enabled.
> 
>     If the device address space is 64-bits, the data types atomic_intptr_t, atomic_uintptr_t,atomic_size_t and atomic_ptrdiff_t are supported if the cl_khr_int64_base_atomics andcl_khr_int64_extended_atomics extensions are supported and have been enabled.

The problem I have with those is that I don't understand what exactly should happen when the pragma is enabled and disabled. After several discussions with the developers I realised that there are multiple interpretations, including the one implemented in clang that in my opinion doesn't help but only makes things complicated. In addition the upstream implementation is inconsistent i.e. (1) pragma is required for some but not the others and (2)
 the same extension can be implemented differently in two default headers.

To resolve this I suggest the following steps:
1. Clarify general use of extension pragma to avoid future mistakes and confusions: https://github.com/KhronosGroup/OpenCL-Docs/pull/355
1. Identify in what extensions the pragma is needed and add clarification to each of those i.e. what happens when pragma is disabled/enabled.
2. Add clarifications to the existing extensions that don't need pragma but list it in the extension spec that the pragma is only accepted for backward compatibility reasons but doesn't change any functionality. I.e. new kernels won't need to add it but the existing kernels will continue to compile without the warnings about the unknown pragmas.
3. Modify upstream implementation according to the clarified spec.

> Should there also be some clarification in here about the interaction between pragma directives and preprocessor macros? For example, AIUI:
> 
> ```c
> #pragma OPENCL EXTENSION cl_khr_fp64 : disable
> #ifdef cl_khr_fp64
> // Code here is actually compiled
> #endif
> ```
> 
> I.e., I think the intent is that preprocessor macros communicate to the OpenCL-C kernel what extensions are available, and kernels use pragmas to indicate which extensions they want to use. What might not be immediately obvious to a developer is that the preprocessor runs before pragmas are evaluated, so a `pragma ... disable` doesn't undef the extension macro.

Ok, right now the clarification says that each pragma should explain what it changes in the standard behavior. So it is implying there is no change by default i.e. it doesn't affect any macros. Although I see some value in being explicit with respect to some common cases. I will try to refine the wording to explain this explicitly. Thanks for the input!

OpenCL_C.txt
AnastasiaStulova: I have added the following:
> If the behavior of an extension pragma is unspecified the implementations are not
required to accept it. Moreover, there is no default behavior for the extension
pragma e.g. it does not define or undefine an associated predefined extension
macro.

Does it seem reasonable?

AnastasiaStulova: Btw this clarification has been moved into ext/introduction.asciidoc

etomzak: It's a lot clearer now. Thanks!

OpenCL_C.txt
```diff
@@ -3180,6 +3180,20 @@ following forms whose meanings are described elsewhere:
...
+point operations (more details can be found in <<floating-point-macros-and-pragmas,
+the floating-point macros and pragmas section>>).
+
+`EXTENSION` directive is used to extend the language with functionality that
+deviates from the standard behavior described in this document. Extension pragma
+directives are only needed for extensions that require changes in the language
+such that the source code is interpreted differently when the pragma is used.
+Each extension should specify whether it adds the pragma and what behavior
+will be changed due to its use in the source. If the behavior of an extension
+pragma is unspecified the implementations are not required to accept it.
```
bashbaug, line 3193: This feels a little pedantic, but is "accepting" a pragma well-defined?  I don't see "accepting" a pragma described explicitly in the C99 spec, for example.  Would it be more accurate to say that "if the behavior of an extension pragma is unspecified then implementations may ignore it"?

OpenCL_C.txt
```diff
@@ -3180,6 +3180,20 @@ following forms whose meanings are described elsewhere:
...
+More information can be found in <<opencl-extension-spec,section 1.2 in
+the OpenCL Extension Specification>>.
```
bashbaug, line 3195: Minor: I can't comment on the exact line, but since these pragmas are describe briefly now the preceding paragraph should remove "whose meanings are described elsewhere" (line 3170).

ext/introduction.asciidoc
```diff
@@ -121,7 +121,18 @@ The initial state of the compiler is as if the directive
...
+Extension pragma directives are optional and they are only required for
+extensions that change the compiler behavior i.e. the source code is interpreted
+differently when the pragma directive is used. Each extension should specify
+whether it adds the directive or not.  There is no default behavior for
+the extension pragma directive that would apply to extensions universally e.g.
+it does not define or undefine an associated predefined extension
+macro; it does not add or remove function or type declarations. Implementations
+are not required to accept extension pragmas if their behavior is not specified
+explicitly. The use of extension pragma with unknown behavior does not cause
+the compilation to fail but might result in a diagnostic.
```
bashbaug, line 135: This description is good but it is subtly different than the description in the OpenCL C spec.  Would it be better to keep them more in sync?  Or, should the OpenCL C spec description be shortened further and refer to this description in the extension spec for more detail?

I especially like the way that this description clarifies that unknown extension pragmas will not cause the compilation to fail but might result in a diagnostic.

bashbaug (commented):
Thanks, these latest changes look very good and we could have avoided a lot of confusion if we adopted this policy from the beginning.

I've left a few specific comments on the changes themselves and I have a few general questions:

1. Should `cl_khr_fp16` also require an extension pragma, specifically to enable half-precision literals such as `1.0h`?  See, for example: https://godbolt.org/z/brMWTn

2. How confident are we that no other implementations require other extension pragmas, either rightly or wrongly?  I'm still worried that these changes will cause shipping implementations to start rejecting kernels that are valid according to the spec after these changes, but that were debatably invalid beforehand because they are missing extension pragmas.

    * See previous comment here: https://github.com/KhronosGroup/OpenCL-Docs/pull/355#discussion_r457707521
    * Also see previous comment here: https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-662260546

For (2), do we know which extensions currently require an extension pragma in Clang?  Is it tractable to update the descriptions of these extensions to say something to the effect of: These extensions shouldn't need an extension pragma, and newer implementations may not require an extension pragma, but some implementations do require an extension pragma so it should be included for maximum portability?

AnastasiaStulova:
> Thanks, these latest changes look very good and we could have avoided a lot of confusion if we adopted this policy from the beginning.
> 
> I've left a few specific comments on the changes themselves and I have a few general questions:
> 
>     1. Should `cl_khr_fp16` also require an extension pragma, specifically to enable half-precision literals such as `1.0h`?  See, for example: https://godbolt.org/z/brMWTn
> 

We could do. For now I only wanted to specify the pragma where we know that it is needed. For the half literal, I don't really understand what benefit does the pragma bring. If `cl_khr_fp16` is supported `1.0h` will always be parsed as a half literal. The only difference that with pragma enabled it is parsed without an error but with pragma disabled it is parsed with an error. I find this extra step of enabling not only useless but also complicated. If the type is supported it should just be possible to use it without any pragma. If it is not supported then the error is given regardless of whether pragma is enabled or not. This is how extensions normally work. 

>     2. How confident are we that no other implementations require other extension pragmas, either rightly or wrongly?  I'm still worried that these changes will cause shipping implementations to start rejecting kernels that are valid according to the spec after these changes, but that were debatably invalid beforehand because they are missing extension pragmas.
>        
>        * See previous comment here: [#355 (comment)](https://github.com/KhronosGroup/OpenCL-Docs/pull/355#discussion_r457707521)
>        * Also see previous comment here: [#355 (comment)](https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-662260546)
> 
> 
> For (2), do we know which extensions currently require an extension pragma in Clang? Is it tractable to update the descriptions of these extensions to say something to the effect of: These extensions shouldn't need an extension pragma, and newer implementations may not require an extension pragma, but some implementations do require an extension pragma so it should be included for maximum portability?

Ok, right now my intention is not to fix any default behavior because either way we fix it we can't have any prove that no vendor is using the pragma differently. So my intention was to make the unspecified pragma an implementation-defined so to say.  This is exactly what it is right now. However, perhaps my current wording doesn't reflect this well as it only states that the implementations are allowed to ignore pragmas:

`Implementations are not required to accept extension pragmas if their behavior is not specified explicitly. `

Do you think we should add something like:

`Implementations are allowed to require the pragma even if its behavior if not specified.`?

ext/introduction.asciidoc
bashbaug: One other observation: the table above (line 91) says:

> Report an error on the #pragma OPENCL EXTENSION if the extension_name is not supported, or if all is specified.

Is this still correct and should an error be reported?  Or, is an unsupported extension considered to be unknown, meaning it may result in a diagnostic but will not fail compilation?

bashbaug:
I don't want to feature-creep this PR, but I think the observations in this comment are good and we should fix them, perhaps by refactoring the intro section in the extensions spec: https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-721245003

> I think the intent is that preprocessor macros communicate to the OpenCL-C kernel what extensions are available, and kernels use pragmas to indicate which extensions they want to use. What might not be immediately obvious to a developer is that the preprocessor runs before pragmas are evaluated, so a pragma ... disable doesn't undef the extension macro.

@etomzak would you mind creating an issue to track this so it doesn't slip through the cracks?  Thanks!

AnastasiaStulova:
> I don't want to feature-creep this PR, but I think the observations in this comment are good and we should fix them, perhaps by refactoring the intro section in the extensions spec: [#355 (comment)](https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-721245003)
> 
> > I think the intent is that preprocessor macros communicate to the OpenCL-C kernel what extensions are available, and kernels use pragmas to indicate which extensions they want to use. What might not be immediately obvious to a developer is that the preprocessor runs before pragmas are evaluated, so a pragma ... disable doesn't undef the extension macro.
> 
> @etomzak would you mind creating an issue to track this so it doesn't slip through the cracks? Thanks!

Btw I have covered this in my change to the extension spec:

> There is no default behavior for the extension pragma directive that would apply to extensions universally e.g. it does not define or undefine an associated predefined extension macro; it does not add or remove function or type declarations.

Or do you think we need to add something more?

ext/introduction.asciidoc
AnastasiaStulova: The way I understood this table is that it applies to extensions where the pragma `enable`/`disable` is needed i.e. actually doing something. If this is the case then if the extension is not supported but the pragma is attempting to enable it the error is sensible. However, I am not sure whether this is achievable to implement because compilers can't possibly know all extensions that exist and we don't want them to know all of them. So if there is let's say an extension that is implemented by only one vendor and not added into any other compiler then the other compilers wouldn't know about such extension and therefore wouldn't report that it is not supported. They would only be able to report that the extension is unknown. It seems that the definition of unknown and unsupported extension here is somewhat vague.

Perhaps, it is reasonable to change this to say that a diagnostic will occur w/o detailing which exactly diagnostic?

bashbaug:
> Btw I have covered this in my change to the extension spec:

> Or do you think we need to add something more?

I think we're a lot closer but I think we could describe the intent behind the extension defines, probably by slightly refactoring the description in section 1.2 of the extension spec.  It would make more sense to me at least to describe the extension #defines first, since they're always present, and then to describe the extension #pragmas next, for example.

etomzak:
> I think we're a lot closer but I think we could describe the intent behind the extension defines, probably by slightly refactoring the description in section 1.2 of the extension spec. It would make more sense to me at least to describe the extension #defines first, since they're always present, and then to describe the extension #pragmas next, for example.

I've opened #514 to track this.

bashbaug:
Hello, could you please take another look at my comments here https://github.com/KhronosGroup/OpenCL-Docs/pull/355#pullrequestreview-541524654 and ensure they're addressed?

I'm especially concerned about:

> Extension pragma directives are optional and they are only required for
extensions that change the compiler behavior i.e. the source code is interpreted
differently when the pragma directive is used. Each extension should specify
whether it adds the directive or not.

My previous comment describes three extensions that currently require a pragma to function correctly by all Clang-based compilers, but this PR does not update any of these extensions to say that a pragma is required.

Even with some recent cleanup there are a lot of extension pragmas that Clang currently accepts ([link](https://github.com/llvm/llvm-project/blob/a84599f177a67d4a8c1c30ccd96c99fa40af75f7/clang/include/clang/Basic/OpenCLExtensions.def)).  How many of these extensions do we need to update to say that a pragma is required?

AnastasiaStulova:
> Hello, could you please take another look at my comments here [#355 (review)](https://github.com/KhronosGroup/OpenCL-Docs/pull/355#pullrequestreview-541524654) and ensure they're addressed?
> 
> I'm especially concerned about:
> 
> > Extension pragma directives are optional and they are only required for
> > extensions that change the compiler behavior i.e. the source code is interpreted
> > differently when the pragma directive is used. Each extension should specify
> > whether it adds the directive or not.
> 
> My previous comment describes three extensions that currently require a pragma to function correctly by all Clang-based compilers, but this PR does not update any of these extensions to say that a pragma is required.
> 
> Even with some recent cleanup there are a lot of extension pragmas that Clang currently accepts ([link](https://github.com/llvm/llvm-project/blob/a84599f177a67d4a8c1c30ccd96c99fa40af75f7/clang/include/clang/Basic/OpenCLExtensions.def)). How many of these extensions do we need to update to say that a pragma is required?

I believe this is covered in my earlier reply https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-736621940

alycm:
I was surprised to see a pragma for fp64 but not fp16 in the current iteration of this PR.  I've read the extensive discussion above, and I'm not sure I have anything truly new to add.  The rationale for keeping pragma's to a minimum seems to come down to: 

> Note that I don't think guarding the use of types or functions that are available if an extension is supported by a pragma is useful and I have therefore excluded those (see my earlier comment).

From a language design perspective I see the case for this.  However, this change retroactively applies this definition to all extensions, implying a potential difference between current and future implementations.  The version of the Intel GPU OpenCL driver that my laptop happens to have installed errors if trying to use `cl_khr_fp16` functionality without the pragma enable.  This PR says that pragma is optional, but at least one shipped implementation treats it as mandatory.

I'm not sure what the complete set of extensions are that falls into this category, but this thread seems to at least document fp16 and 64-bit atomics as being in this situation.

AnastasiaStulova:
> I was surprised to see a pragma for fp64 but not fp16 in the current iteration of this PR. I've read the extensive discussion above, and I'm not sure I have anything truly new to add. The rationale for keeping pragma's to a minimum seems to come down to:
> 
> > Note that I don't think guarding the use of types or functions that are available if an extension is supported by a pragma is useful and I have therefore excluded those (see my earlier comment).
> 
> From a language design perspective I see the case for this. However, this change retroactively applies this definition to all extensions, implying a potential difference between current and future implementations. 

Just to point out the intension for this PR is to document the current situation i.e. use of pragma is implementation-defined for most of extensions because the spec has never documented their use. I appreciate perhaps I've failed to deliver this message. Therefore I was thinking whether adding the following statement can help.

 > Ok, right now my intention is not to fix any default behavior because either way we fix it we can't have any prove that no vendor is using the pragma differently. So my intention was to make the unspecified pragma an implementation-defined so to say. This is exactly what it is right now. However, perhaps my current wording doesn't reflect this well as it only states that the implementations are allowed to ignore pragmas:

> `Implementations are not required to accept extension pragmas if their behavior is not specified explicitly.`

> Do you think we should add something like:

> `Implementations are allowed to require the pragma even if its behavior if not specified.`?

AnastasiaStulova:
I am trying to understand what is blocking this PR or general progress on #82 at the moment? What would be helpful if someone provides any feedback of what is to be changed and how? I will do my best to make it happen.

Just FYI perhaps I haven't made it clear - there are significant improvements in tooling that are blocked on this that the community can benefit including the implementation of OpenCL 3.0.

bashbaug:
> I am trying to understand what is blocking this PR or general progress on #82 at the moment? What would be helpful if someone provides any feedback of what is to be changed and how? I will do my best to make it happen.

Hi @AnastasiaStulova , my biggest concern is that kernels written against a version of the spec including these changes may fail against conformant implementations that were developed against a version of the spec without these changes.  Specifically, our Clang-based implementation currently requires a pragma to enable an extension in at least three cases that are not described in this PR (and there may be more):

* [this](https://godbolt.org/z/Kb3Pxf) example for OpenCL C 2.0 64-bit atomics.
* [this](https://godbolt.org/z/brMWTn) example for `cl_khr_fp16`
* [this](https://godbolt.org/z/T4TGYv) example for `cl_intel_device_side_avc_motion_estimation`

I very much agree that the current policies for extension pragmas are unclear, confusing, and/or complicated for tooling.  The changes in this PR are a huge improvement, we just need to be sure we don't break shipping implementations while we're at it.

AnastasiaStulova:
> I very much agree that the current policies for extension pragmas are unclear, confusing, and/or complicated for tooling. The changes in this PR are a huge improvement, we just need to be sure we don't break shipping implementations while we're at it.

I am with you on that and this is why I have completely changed this PR to document that the behavior is implementation-defined. Do you agree with this? It just what it is now already, we can't fix the behavior based on what clang does, because we don't know what other implementations are doing or have been doing.

If the wording on the PR doesn't reflect the intention I am happy to change it and I have made the following suggestion.

  >  Ok, right now my intention is not to fix any default behavior because either way we fix it we can't have any prove that no vendor is using the pragma differently. So my intention was to make the unspecified pragma an implementation-defined so to say. This is exactly what it is right now. However, perhaps my current wording doesn't reflect this well as it only states that the implementations are allowed to ignore pragmas:
> `Implementations are not required to accept extension pragmas if their behavior is not specified explicitly.`
  >  Do you think we should add something like:
> `Implementations are allowed to require the pragma even if its behavior if not specified.`

Does it make sense?

If it makes things simpler I suggest to drop `fp64` from this PR because it seems to confuse people. It was added originally with the initial wording and I kept it mainly because I know that the pragma is useful. But I am ok to leave it out and follow up on individual extensions in the following PRs.

roma2:
> > Кстати, я рассмотрел это в своем изменении спецификации расширения:
> 
> > Или вы думаете, что нам нужно добавить что-то еще?
> 
> Я думаю, что мы намного ближе, но я думаю, что мы могли бы описать цель, стоящую за определениями расширения, возможно, немного изменив описание в разделе 1.2 спецификации расширения. Для меня было бы разумнее, по крайней мере, сначала описать расширение #defines, поскольку они всегда присутствуют, а затем, например, описать расширение #pragmas.



> > Кстати, я рассмотрел это в своем изменении спецификации расширения:
> 
> > Или вы думаете, что нам нужно добавить что-то еще?
> 
> Я думаю, что мы намного ближе, но я думаю, что мы могли бы описать цель, стоящую за определениями расширения, возможно, немного изменив описание в разделе 1.2 спецификации расширения. Для меня было бы разумнее, по крайней мере, сначала описать расширение #defines, поскольку они всегда присутствуют, а затем, например, описать расширение #pragmas.

https://github.com/KhronosGroup/OpenCL-Docs/pull/355#issuecomment-738299222[https://github.com/jfrog/jfrog-cli-plugins/pull/15#issue-538004878](Mou-Ikkai:imgbot)

roma2:
Duplicate of #

bashbaug:
Since this PR came up again in today's teleconference: If we want to get this over the finish line (and it still looks like most of the changes are really good, and exactly what we need), I think we need to add similar text that this PR adds for cl_khr_fp64 to the list of extensions described here:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/82#issuecomment-635467096
* (also here: https://github.com/KhronosGroup/OpenCL-Docs/issues/82#issuecomment-785496025)

Rightly or wrongly, this is the set of extensions that we know will require an extension pragma on shipping implementations.

If we want to differentiate between the set of extensions that can cause a diagnostic with an unrecognized pragma vs. the set of extensions that cannot (described here: https://github.com/KhronosGroup/OpenCL-Docs/issues/82#issuecomment-786411529), then we'll need to update a few more extensions to say this.

cycheng:
Hello, are we going to land this PR??

svenvh:
I'm interested in seeing this progressed too.  There is still a bit of work to do: the PR needs to be rebased and updated for the new way extensions are organized in this repo now.  We may also want to work on https://github.com/KhronosGroup/OpenCL-CTS/issues/886 first.

Issue #330, opened 2020-06-16
Is Description of OpenCL C "Features" Clear Enough?


alycm:
This was internal issue 218.

Is the generic introduction of what the new OpenCL C features and associated macros are sufficient?  Does it make it clear what a feature is?  Is there any confusion compared to extension pragmas (i.e. extensions are enabled by the programmer, features are enabled or not by the compiler).

alycm:
Comment from @AnastasiaStulova that having a defined "name" per feature may help to reference them in the text: https://github.com/KhronosGroup/OpenCL-Docs/pull/358#discussion_r459071696

AnastasiaStulova:
From OpenCL Tooling call today:
- We should explain the behavior that is represented by features and extensions (example - double)
- We should also explain difference between core otional extensions and extensions

AnastasiaStulova:
I think this was closed accidentally.

bashbaug:
> I think this was closed accidentally.

Yes, I agree, it closed automatically because it was linked with #372.  I think #372 should have been linked to #370 instead?  I'll fix the link, and close #370 manually.

AnastasiaStulova:
PR: https://github.com/KhronosGroup/OpenCL-Docs/pull/387

alycm:
https://github.com/KhronosGroup/OpenCL-Docs/pull/387 already addressed this issue and was merged, I'll close this issue.

PR #358, opened 2020-07-20
Allow C11 Atomics to Be Used Without Generic Address Space

### LINKED ISSUE #331: Allow C11 Atomics to Be Used Without Generic Address Space
This was internal issue 230.

The legacy OpenCL 1.x atomics only work with named address spaces, the OpenCL 2.x atomics (C11 style) all use generic address space.

In OpenCL 3.0 provisional the C11 style atomics thus got tied to the generic address space feature, but that is undesirable.  It should be possible to support C11 atomic syntax without supporting generic address spaces.

Thus named address spaces should be supported for C11 style atomic functions, specifically the `global` and `local` address spaces.  Overloaded versions of the relevant atomic functions should be added.

### LINKED ISSUE #357: Atomic Functions without scope aren't guarded by __opencl_c_atomic_scope_device
Currently the explicit atomic functions are not guarded by feature macros (pending changes w.r.t. generics), but the specification says

> The functions that do not have `memory_scope` argument have the same semantics as the corresponding functions with the `memory_scope` argument set to `memory_scope_device`.

As` memory_scope_device` is not mandatory I think that that version of the builtin should be guarded by `__opencl_c_atomic_scope_device`.

A sentence such as the following to each of the explicit functions that take an order but not a scope should do it:

> Requires OpenCL C 2.0 or the `__opencl_c_atomic_scope_device` feature macro.

alycm:
Fixes #331
Fixes #357

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -6532,7 +6532,7 @@ void atomic_store(volatile __local A *object, C desired)
 // macros.
 void atomic_store(volatile A *object, C desired)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 void atomic_store_explicit(volatile __global A *object,
                            C desired,
                            memory_order order)
@@ -6540,7 +6540,8 @@ void atomic_store_explicit(volatile __local A *object,
                            C desired,
                            memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 void atomic_store_explicit(volatile A *object,
                            C desired,
                            memory_order order)
@@ -6597,13 +6598,14 @@ C atomic_load(volatile __local A *object)
 // macros.
 C atomic_load(volatile A *object)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 C atomic_load_explicit(volatile __global A *object,
                        memory_order order)
 C atomic_load_explicit(volatile __local A *object,
                        memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 C atomic_load_explicit(volatile A *object,
                        memory_order order)
 
@@ -6655,7 +6657,7 @@ C atomic_exchange(volatile __local A *object, C desired)
 // macros.
 C atomic_exchange(volatile A *object, C desired)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 C atomic_exchange_explicit(volatile __global A *object,
                            C desired,
                            memory_order order)
@@ -6663,7 +6665,8 @@ C atomic_exchange_explicit(volatile __local A *object,
                            C desired,
                            memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 C atomic_exchange_explicit(volatile A *object,
                            C desired,
                            memory_order order)
@@ -7060,7 +7063,7 @@ C atomic_fetch_key(volatile __local A *object, M operand)
 // macros.
 C atomic_fetch_key(volatile A *object, M operand)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 C atomic_fetch_key_explicit(volatile __global A *object,
                             M operand,
                             memory_order order)
@@ -7068,7 +7071,8 @@ C atomic_fetch_key_explicit(volatile __local A *object,
                             M operand,
                             memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 C atomic_fetch_key_explicit(volatile A *object,
                             M operand,
                             memory_order order)
@@ -7164,7 +7168,7 @@ bool atomic_flag_test_and_set(
 bool atomic_flag_test_and_set(
     volatile atomic_flag *object)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 bool atomic_flag_test_and_set_explicit(
     volatile __global atomic_flag *object,
     memory_order order)
@@ -7172,7 +7176,8 @@ bool atomic_flag_test_and_set_explicit(
     volatile __local atomic_flag *object,
     memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 bool atomic_flag_test_and_set_explicit(
     volatile atomic_flag *object,
     memory_order order)
@@ -7229,7 +7234,7 @@ void atomic_flag_clear(volatile __local atomic_flag *object)
 // macros.
 void atomic_flag_clear(volatile atomic_flag *object)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
 void atomic_flag_clear_explicit(
     volatile __global atomic_flag *object,
     memory_order order)
@@ -7237,7 +7242,8 @@ void atomic_flag_clear_explicit(
     volatile __local atomic_flag *object,
     memory_order order)
 
-// Requires OpenCL C 2.0 or the __opencl_c_generic_address_space feature macro.
+// Requires OpenCL C 2.0 or both the __opencl_c_generic_address_space and
+// __opencl_c_atomic_scope_device feature macros.
 void atomic_flag_clear_explicit(
     volatile atomic_flag *object,
     memory_order order)
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -6522,7 +6522,7 @@ This section specifies each general kind.
 --
 [source,c]
 ----------
-// Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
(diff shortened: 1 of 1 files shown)


alycm:
This PR is obviously verbose, I'll investigate ways of expressing the same things more cleanly, but that can wait for a future PR.  This allows implementers to start etc.

- [ ] @alycm File a ticket about using a `genaddrspace` or similar to reduce verbosity.

OpenCL_C.txt
```diff
@@ -6303,6 +6303,11 @@ pointed to by obj to the value value.
 
 [source,c]
 ----------
+// Requires OpenCL C 2.0 or newer.
```
bashbaug, line 6306: Should this be "OpenCL C **2.0** or newer" or "OpenCL C **3.0** or newer"?

I'm not sure if it's observable, but the function overloads for the named (non-generic) address spaces were not in previous OpenCL C specifications.

This same comment applies elsewhere in this PR, also.

alycm: Yes, you're correct.  It should say 3.0.  I'll fix that.

alycm: Fixed.

alycm:
I've now pushed an additional to this PR that address https://github.com/KhronosGroup/OpenCL-Docs/issues/357 so that it can all be reviewed together.

OpenCL_C.txt
```diff
@@ -6550,11 +6590,34 @@ For the explicit variants, memory order and scope enumerations must respect the
 ----------
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
bashbaug, line 6591: Similar to above:

```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6550,11 +6590,34 @@ For the explicit variants, memory order and scope enumerations must respect the
...
 // and __opencl_c_atomic_scope_device feature macros.
+C atomic_load(volatile __global A *object)
+C atomic_load(volatile __local A *object)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
+// __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature
+// macros.
 C atomic_load(volatile A *object)
 
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
```
bashbaug, line 6601:
```suggestion
// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device feature macro.
```

OpenCL_C.txt
```diff
@@ -6582,12 +6649,39 @@ For the explicit variants, memory order and scope enumerations must respect the
 ----------
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
bashbaug, line 6650: Similar to above:

```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6582,12 +6649,39 @@ For the explicit variants, memory order and scope enumerations must respect the
...
 // and __opencl_c_atomic_scope_device feature macros.
+C atomic_exchange(volatile __global A *object, C desired)
+C atomic_exchange(volatile __local A *object, C desired)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
+// __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature
+// macros.
 C atomic_exchange(volatile A *object, C desired)
 
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
```
bashbaug, line 6660:
```suggestion
// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device feature macro.
```

OpenCL_C.txt
```diff
@@ -6618,17 +6716,123 @@ For the explicit variants, memory order and scope enumerations must respect the
 ----------
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
bashbaug, line 6717: Similar to above:

```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6741,12 +7055,39 @@ For *atomic_fetch* and modify functions with *key* = `or`, `xor`, `and`,
...
```
bashbaug, line 7056: Similar to above:

```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6741,12 +7055,39 @@ For *atomic_fetch* and modify functions with *key* = `or`, `xor`, `and`,
...
 // and __opencl_c_atomic_scope_device feature macros.
+C atomic_fetch_key(volatile __global A *object, M operand)
+C atomic_fetch_key(volatile __local A *object, M operand)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
+// __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature
+// macros.
 C atomic_fetch_key(volatile A *object, M operand)
 
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
```
bashbaug, line 7066:
```suggestion
// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device feature macro.
```

OpenCL_C.txt
```diff
@@ -6812,13 +7157,42 @@ global atomic_flag guard = ATOMIC_FLAG_INIT;
 ----------
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
bashbaug, line 7158:
```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6812,13 +7157,42 @@ global atomic_flag guard = ATOMIC_FLAG_INIT;
...
+bool atomic_flag_test_and_set(
+    volatile __local atomic_flag *object)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
+// __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature
+// macros.
 bool atomic_flag_test_and_set(
     volatile atomic_flag *object)
 
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
```
bashbaug, line 7171:
```suggestion
// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device feature macro.
```

OpenCL_C.txt
```diff
@@ -6848,12 +7226,39 @@ For the explicit variants, memory order and scope enumerations must respect the
 ----------
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
bashbaug, line 7227:
```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6848,12 +7226,39 @@ For the explicit variants, memory order and scope enumerations must respect the
...
 // and __opencl_c_atomic_scope_device feature macros.
+void atomic_flag_clear(volatile __global atomic_flag *object)
+void atomic_flag_clear(volatile __local atomic_flag *object)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
+// __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature
+// macros.
 void atomic_flag_clear(volatile atomic_flag *object)
 
+// Requires OpenCL C 3.0 and the __opencl_c_atomic_scope_device feature macro.
```
bashbaug, line 7237:
```suggestion
// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device feature macro.
```

OpenCL_C.txt
```diff
@@ -6639,17 +6843,123 @@ bool atomic_compare_exchange_strong_explicit(
 
 // Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
```
alycm, line 6844:
```suggestion
// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
```

OpenCL_C.txt
```diff
@@ -6513,14 +6522,41 @@ This section specifies each general kind.
 --
 [source,c]
 ----------
-// Requires OpenCL C 2.0, or both the __opencl_c_atomic_order_seq_cst
+// Requires OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst
 // and __opencl_c_atomic_scope_device feature macros.
+void atomic_store(volatile __global A *object, C desired)
+void atomic_store(volatile __local A *object, C desired)
+
+// Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space,
```
AnastasiaStulova, line 6530: btw I am not sure this wording is correct but I think we use it everywhere across:

> Requires OpenCL C 2.0, or all of the __opencl_c_generic_address_space, __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device feature macros.

I guess what we mean here is that we require OpenCL C v2.0 or the following features to be supported. However, in contrast to extensions we never gave name to features so there is no way to enumerate them unambiguously now. I think however this is just a wording issue that can be fixed later?

alycm: > I guess what we mean here is that we require OpenCL C v2.0 or the following features to be supported i.e. require feature macro doesn't really explain precisely what is needed.

Yes, exactly.  The wording is deliberately terse because there are a lot of cases.

> I presume if feature macros are defined by the user then most this functionality will be available but this is not the case generally.

Now I'm not sure I understand your comment.  The implementation sets the feature macros, not the user.   I.e. this builtin will only be present for implementations that also set the listed feature macros.

> However, in contrast to extensions, we never gave names to features so there is no way to enumerate them unambiguously now. I think this is just a wording issue that can be fixed later?

There is a table at the start of the document, these macro names are all present there.

alycm: There is also still this ticket about seeing if features can be made clearer (but without actually changing their meaning etc): https://github.com/KhronosGroup/OpenCL-Docs/issues/330

It's not done yet but it's in the 3.0 project.

AnastasiaStulova: Yes I think it makes sense to cover this in https://github.com/KhronosGroup/OpenCL-Docs/issues/330.

So what I actually meant is that for extensions we used to define a name that was the same as extension macro normally (see OpenCL Extensions section 1.1).

Then we could write something like:
> The half scalar and vector types can only be used if the cl_khr_fp16 extension is supported andhas  been  enabled. 

`cl_khr_fp16` was referring to the name of the extension.

Currently we can not say "pipe can only be used if the __opencl_c_pipes feature is supported" because `__opencl_c_pipes` is a macro technically and not a feature. But describing that a type requires feature macro is not entirely correct or at least imprecise.

However, let's leave this for now and address this in #330.

bashbaug:
Merging since we have two approvals, as we discussed in the July 22nd teleconference.

PR #364, opened 2020-07-22
OpenCL C features macros must expand to 1 in OpenCL C 3.0

### LINKED ISSUE #311: Do OpenCL C feature macros have specified values
This was internal issue 226.

Currently feature macros have no specified values, options:

1. Leave it unspecified, implementations may or may not give the macros values.
2. Explicitly specify that they have no value, only their presence is important.
3. Explicitly specify some universal arbitrary value, such as 1.
4. Explicitly specify some value per macro, such as YYYYMM for the date introduced (like C++).
5. Explicitly specify some value per macro, such the version of OpenCL that introduced the feature (following CL_MAKE_VERSION convention)

Discussion on the internal issue leaned towards (5), or (1) or (2).

bashbaug:
This PR specifies that in OpenCL C 3.0 feature macros must expand to `1` in OpenCL C 3.0 if the feature macro is defined.

I've also included a line that some feature macros may expand to different values in the future, but if they do the value of the feature macro must compare greater than the prior value of the feature macro.  This would enable switching to an OpenCL C version-based feature macro, or a value based on the date of the feature (similar to feature macros in C and C++), if desired.

Fixes #311.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -126,6 +126,10 @@ OpenCL C compilers that define the feature macro `+__opencl_c_pipes+` must also 
 
 |====
 
+In OpenCL C 3.0, feature macros must expand to the value `1` if the feature macro is defined by the OpenCL C compiler.
+A feature macro must not be defined if the feature is not supported by the OpenCL C compiler.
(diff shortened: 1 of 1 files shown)


bashbaug:
Merging as agreed in the July 23rd teleconference.

PR #368, opened 2020-07-22
Clarify strictness of legacy atomics

### LINKED ISSUE #353: OpenCL C 3.0 old atomic scope and ordering
In atomics prior to OpenCL 2.0 there was no definition of memory ordering and scope and therefore old atomics behaved in implementation defined manner wrt to these.

OpenCL 3.0 provides explicit behavior for old atomics wrt memory scope and ordering (see s6.13.11).
> These  were equivalent       to       atomic       operations       with       memory_order_relaxed       consistency,       and memory_scope_work_group scope.

Explicitly defining this might affect backward compatibility although technically it is less conservative and shouldn't invalidate existing  implementations it might not be a helpful behavior moving forward. The default implementations often assumed sequential consistency i.e. `atomic_add` in _**libclc**_ is implemented using `__sync_fetch_and_add` that isn't C11 builtin but clang maps it to sequential consistency equivalent of C11 atomic operation. In C11/OpenCL C 2.0 the default behavior is sequential consistency and device scope. Old applications that were developed might no longer run correctly for the new implementations or if old implementations upgrade the behavior to comply to the new standard. As a result, it might force vendors to maintain old flavor of old atomics or even two flavours if they are to be strictly compliant to the spec.

Old atomics even though deprecated are still found useful to the developers because they allow mixing regular operations on types with atomic operations more intuitively https://github.com/KhronosGroup/OpenCL-Docs/issues/66. Atomic types from C11 don't support this in an easy way (there use even with the regular operator still translate to atomic operations).

Could we change this behavior back to implementation-defined? We could add a query mechanisms to either language or API to allow developer to know scope/ordering explicitly if it is important enough.

alycm:
Fixes #353

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -7294,6 +7294,12 @@ only in the case of **atomic_xchg** additionally single precision floating-point
 These were equivalent to atomic operations with `memory_order_relaxed`
 consistency, and `memory_scope_work_group` scope.
 
+NOTE: Some implementations may implement legacy atomics with a stricter memory
+consistency order than `memory_order_relaxed` or a broader scope than
+`memory_scope_work_group`.
+This is because all the stricter orders and broader scopes fully satisfy the
+semantics of the minimum requirements.
+
 // Copied from table 6.19 in OpenCL 1.2 spec
 [[table-legacy-atomic-functions]]
 .Legacy Atomic Functions


bashbaug:
Merging as discussed in the July 23rd teleconference.

PR #370, opened 2020-07-23
Add uniform_work_group_size kernel attribute

### LINKED ISSUE #267: reqd_work_group_size and non-uniform NDRanges
The OpenCL C 2.0 specification currently reads (§6.7.2):

>The optional __attribute__((reqd_work_group_size(X, Y, Z))) is the work-group size that must be
used as the local_work_size argument to clEnqueueNDRangeKernel. This allows the compiler to
optimize the generated code appropriately for this kernel.

The only thing the compiler can know is that the work group size will be in the [{1,1,1},reqd_work_group_size] range when non-uniform NDRanges are in use which seems to defeat the purpose of having `reqd_work_group_size`.

FWIW, the non_uniform_work_group CTS tests do exercise the case where the attibute is specified but the NDRange is such that some work groups are smaller than the value specified with `reqd_work_group_size`.

Should we require NDRange uniformity when `reqd_work_group_size` is used instead?

kpet:
Fixes #267


api/opencl_runtime_layer.asciidoc
```diff
@@ -8198,6 +8198,10 @@ Otherwise, it returns one of the following errors:
   * {CL_INVALID_WORK_GROUP_SIZE} if _local_work_size_ is specified and is not
     consistent with the required number of sub-groups for _kernel_ in the
     program source.
+  * {CL_INVALID_WORK_GROUP_SIZE} if _global_work_size_ is not a multiple of
```
jrprice, line 8201: Isn't this scenario already covered by the "{CL_INVALID_WORK_GROUP_SIZE} if the work-group size must be uniform and..." bullet two below this one?

kpet: Arguably it is but I think it's useful to specify exact error conditions individually for at least the following reasons:
1. This makes it much easier to tie an error to something the application has done. How do you know when the "work-group size must be uniform"?
1. This is a new failure mode that can only happen in OpenCL 3.0, which I think is generally helpful to document. A 1.2 application developer can rule 3.0-only errors with a quick look at the spec. Granted, in this particular case, the buttet point you mentioned makes this moot.
1. If/when we develop a validation layer, having all error cases being documented separately makes for easier number / automatic processing.
1. This makes it easier to write negative tests.

bashbaug:
We discussed these changes in the July 23rd teleconference.  The main concern was that there does not appear to be a way to represent this attribute in SPIR-V, meaning that applications could see different behavior (including error behavior) compiling through SPIR-V vs. OpenCL C.

bashbaug:
Separately, I also wonder if we should add a kernel query similar to `CL_​KERNEL_​COMPILE_​WORK_​GROUP_​SIZE` so host code can determine whether the kernel requires uniform work groups or supports non-uniform work groups.  I don't think this is strictly necessary, since we don't have a similar query for the `-cl-uniform-work-group-size` build option, though I do think it could be useful.

bashbaug:
Moving off of the OpenCL 3 board, as discussed in the July 28th teleconference.

PR #372, opened 2020-07-23
Add three feature macros for features that were already optional

### LINKED ISSUE #330: Is Description of OpenCL C "Features" Clear Enough?

alycm:
* `__opencl_c_images`
* `__opencl_c_fp64`
* `__opencl_c_int64`

Don't precisely follow the existing "Required OpenCL C 2.0 or ..." language as
all of these features were already optional in OpenCL C 2.0.

Add a section to appendix H even though this doesn't change optionality, all
the other feature macros are summarised there.

Fixes #330

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -87,6 +87,7 @@ The following table describes OpenCL C 3.0 feature macros and their meaning:
 | `+__opencl_c_3d_image_writes+`
 | The OpenCL C compiler supports built-in functions for writing to 3D image objects.
 
+OpenCL C compilers that define the feature macro `+__opencl_c_3d_image_writes+` must also define the feature macro `+__opencl_c_images+`.
 | `+__opencl_c_atomic_order_acq_rel+`
 | The OpenCL C compiler supports enumerations and built-in functions for atomic operations with acquire and release memory consistency orders.
 
@@ -107,6 +108,16 @@ OpenCL C compilers that define the feature macro `+__opencl_c_device_enqueue+` m
 | `+__opencl_c_generic_address_space+`
 | The OpenCL C compiler supports the unnamed generic address space.
 
+| `+__opencl_c_fp64+`
+| The OpenCL C compiler supports types and built-in functions with 64-bit floating point types.
+
+| `+__opencl_c_images+`
+| The OpenCL C compiler supports types and built-in functions for images.
+
+| `+__opencl_c_int64+`
+| The OpenCL C compiler supports types and built-in functions with 64-bit integers.
+
+OpenCL C compilers for FULL profile devices or devices with 64-bit pointers must always define the `+__opencl_c_int64+` feature macro.
 | `+__opencl_c_pipes+`
 | The OpenCL C compiler supports the pipe modifier and built-in functions to read and write from a pipe.
 
@@ -118,6 +129,7 @@ OpenCL C compilers that define the feature macro `+__opencl_c_pipes+` must also 
 | `+__opencl_c_read_write_images+`
 | The OpenCL C compiler supports reading from and writing to the same image object in a kernel.
 
+OpenCL C compilers that define the feature macro `+__opencl_c_read_write_images+` must also define the feature macro `+__opencl_c_images+`.
 | `+__opencl_c_subgroups+`
 | The OpenCL C compiler supports built-in functions operating on sub-groupings of work-items.
 
@@ -169,9 +181,9 @@ The following table describes the list of built-in scalar data types.
     | A signed two's complement 32-bit integer.
 | `unsigned int`, `uint`
     | An unsigned 32-bit integer.
-| `long`
+| `long`^2x^
     | A signed two's complement 64-bit integer.
-| `unsigned long`, `ulong`
+| `unsigned long`, `ulong`^2x^
     | An unsigned 64-bit integer.
 | `float`
     | A 32-bit floating-point.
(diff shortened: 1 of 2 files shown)


alycm:
Some general comments:

* Obviously I still need to roll through the footnote renumberings, but they go up to 74, I'll do that in a separate PR after this merges.
* I actually prefer `__opencl_c_double` and `__opencl_c_long` for reading, but the names that I went with match `cl_khr_fp64` and `cles_khr_int64`.
* Arguably I could have annotated far fewer things, as these types were already optional, with less description, and doesn't seem to have been a problem.   Definitely open to feedback on that point.
* I didn't note the feature macro for gentype's that contain `long` or `double`.  I could, it just seemed pretty verbose once I extended the descriptions.

OpenCL_C.txt
```diff
@@ -4457,31 +4502,44 @@ all arguments and the return type, unless otherwise specified.
...
+  feature macro: +
+
+  float__n__ **lgamma_r**(float__n__ _x_, int__n__ *_signp_) +
+  float **lgamma_r**(float _x_, int *_signp_)
+    | Log gamma function.
+      Returns the natural logarithm of the absolute value of the gamma
+      function.
+      The sign of the gamma function is returned in the _signp_ argument of
+      *lgamma_r*.
+| double__n__ **lgamma_r**(double__n__ _x_, {global} int__n__ *_signp_) +
```
EwanC, line 4522: Does these `double lgamma_r`  definitions need  a `Requires support for doubles, e.g. with the +__opencl_c_fp64+ feature macro. ` comment too

OpenCL_C.txt
```diff
@@ -352,19 +375,30 @@ The following table describes the list of built-in vector data types.
...
     | A vector of _n_ 32-bit floating-point values.
 | `double__n__`^5^
     | A vector of _n_ 64-bit floating-point values.
 |====
 
+// TODO include this footnote in the renumbering.
+[5x] The `long` and `ulong` vector types are optional types for
+EMBEDDED profile devices that are supported if the value of the
+<<opencl-device-queries, `CL_​DEVICE_​EXTENSIONS` device query>> contains
+`+cles_khr_int64+`
```
EwanC, line 392: full stop at end of sentence

OpenCL_C.txt
```diff
@@ -207,9 +219,20 @@ The following table describes the list of built-in scalar data types.
 [1] When any scalar value is converted to `bool`, the result is 0 if the
 value compares equal to 0; otherwise, the result is 1.
 
+// TODO: Renumber the footnotes from here onwards.
+[2x] The `long`, `unsigned long` and `ulong` scalar types are optional types
+for EMBEDDED profile devices that are supported if the value of the
+<<opencl-device-queries, `CL_​DEVICE_​EXTENSIONS` device query>> contains
+`+cles_khr_int64+`
```
EwanC, line 226: full stop at end of sentence

OpenCL_C.txt
alycm: So it already has this.

I'm a bit inconsistent in how I do this, but not without reason.   Rows were some built-ins use doubles get a comment next to the applicable functions in the left column, rows were all the built-ins use doubles get a comment in the right column.   So here it's in the right column, but only because I split `lgamma_r` into two rows here.  It makes more sense in the rendered form.

alycm:
> LGTM, btw do we also need to update this:
> 
> > The      image2d_t,      image3d_t,      image2d_array_t,      image1d_t,      image1d_buffer_t,image1d_array_t,  image2d_depth_t,  image2d_array_depth_t  and  sampler_t  types  areonly    defined    if    the    device    supports    images,    i.e.    the    value    of    the CL_DEVICE_IMAGE_SUPPORT device query) is CL_TRUE.
> 
> I guess all those types should only be available if `__opencl_c_images` is defined?

I think this is covered by the next line:

> If this is the case then an OpenCL C 3.0 compiler must also define the `__opencl_c_images` feature macro.

AnastasiaStulova:
> > LGTM, btw do we also need to update this:
> > > The      image2d_t,      image3d_t,      image2d_array_t,      image1d_t,      image1d_buffer_t,image1d_array_t,  image2d_depth_t,  image2d_array_depth_t  and  sampler_t  types  areonly    defined    if    the    device    supports    images,    i.e.    the    value    of    the CL_DEVICE_IMAGE_SUPPORT device query) is CL_TRUE.
> > 
> > 
> > I guess all those types should only be available if `__opencl_c_images` is defined?
> 
> I think this is covered by the next line:
> 
> > If this is the case then an OpenCL C 3.0 compiler must also define the `__opencl_c_images` feature macro.

Right, slightly different format but the same meaning. So should be fine indded.

alycm:
> This mostly looks very good and I think we should merge these changes as-is.
> 
> My one cosmetic comment is that there are quite a few changes in this PR, for fp64 specifically (you mentioned this in your review request also). I wonder if there's a way to reduce the verbosity somewhat, without losing clarity. One thought is: since we've guarded the `double` types with the `__opencl_c_fp64` feature macro, is that sufficient to cover all of the built-in functions that accept `double` types also?

Yes, that is essentially the situation that the specification is in currently (i.e. without this PR).

> If this isn't quite verbose enough, we could add text in the description of each category of affected built-in functions, before the tables. The main goal would be to avoid needing to annotate each individual built-in function.

Yes, maybe.  I'm not even sure if that is necessary.   If I did I'd add a paragraph noting the exclusion from `gentype` (which I haven't currently done) and that all explicitly typed use of `double` or `gentyped` is also excluded.   And equivalently for `long`.

I can probably get that done today, but happy for this PR to merge first to get the functionality locked in.

bashbaug:
Merging so implementation can begin.  We can tidy up the verbosity in a follow-on PR.

PR #378, opened 2020-07-24
Remove extraneous wording for fp64 vload/vstore alignment

StuartDBrady:
This also clarifies the wording slightly for fp16.


StuartDBrady:
> I think this is clearer and it's consistent with similar text for the `cl_khr_fp16` extension, so we could merge it as-is.
> 
> I think the rationale behind the extraneous text though is that the extension is adding `double` to the generic type name `gentype`, so the descriptions of these functions should be generic as well. Would it be better to drop the extraneous text but add a few words so it's clear the description only applies to the types added by the extension? I'm thinking of something like:
> 
> > If `gentype` is `double`, the read address computed as (_p_ + (_offset * n_)) must be 64-bit aligned.
> 
> (With similar changes to the `cl_khr_fp16` extension.)
> 
> I'll hold off merging these changes for a couple days for additional discussion.

Good point. I've made the changes you've suggested, leaving **vstore_half** and **vstore_halfn** as they were, because these only ever take `half *`.

PR #383, opened 2020-07-27
Use consistent naming for the OpenCL SPIR-V Environment Specification

StuartDBrady:
In some places, this was referred to as the "SPIR-V OpenCL Environment Specification" or just the "OpenCL Environment Specification" even though the document's title is the "OpenCL SPIR-V Environment Specification". Also, the file names for the specification did not include "SPIR-V".

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -85,7 +85,7 @@ for each Specification:
 | cxxhtml   | HTML outputs for C++ Specification
 | cxxpdf    | PDF outputs for C++ Specification
 |           |
-| env       | HTML and PDF outputs for Environment Specification
+| env       | HTML and PDF outputs for SPIR-V Environment Specification
 | envhtml   | HTML outputs for Environment Specification
 | envpdf    | PDF outputs for Environment Specification
 |           |
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -4556,7 +4556,7 @@ all arguments and the return type, unless otherwise specified.
 | gentype *mad*(gentype _a_, gentype _b_, gentype _c_)
     | *mad* computes _a_ * _b_ + _c_.
     The function may compute _a_ * _b_ + _c_ with reduced accuracy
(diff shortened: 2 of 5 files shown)


README.adoc
```diff
@@ -85,7 +89,7 @@ for each Specification:
 | cxxhtml   | HTML outputs for C++ Specification
 | cxxpdf    | PDF outputs for C++ Specification
 |           |
-| env       | HTML and PDF outputs for Environment Specification
+| env       | HTML and PDF outputs for SPIR-V Environment Specification
 | envhtml   | HTML outputs for Environment Specification
 | envpdf    | PDF outputs for Environment Specification
```
alycm, line 94: If we're going for 100% consistency might as well change these two as well.

StuartDBrady: Can do. In the repository structure section above, abbreviated names of the specifications are used after the first mention of a specification, dropping "OpenCL" in subsequent mentions. I thought it might be preferable to do the same here, and drop "OpenCL SPIR-V" since it is clear what the specific environment specification is on these two lines. Given this explanation, do you still think it would be better to be unambiguous in all places, or is conciseness preferable?

alycm: The current concise approach is fine, no action is required for this comment, I'll resolve it.

StuartDBrady:
I would say that the file naming _is_ important, but that it is not of utmost importance. It will only become harder to rename this later on. Could we rename this now, when it is still not too painful to do so?

bashbaug:
Can we separate out the filename changes into a different issue or PR?  I'd like to merge the rest of these changes ASAP while the file renaming is still under discussion.  Thanks!

bashbaug:
Discussed in the August 18th teleconference and agreed to keep the same filenames we are using currently.  I will merge this change as soon as the renaming is reverted - thanks!

alycm (approved):
LGTM, thanks!

Issue #354, opened 2020-07-16
Implementing restriction on relaxed memory scope atomic use 


AnastasiaStulova:
The following restriction is hard to implement in tools as it adds extra complexity for specializing one function otherwise we could just disable the relaxed versions of scopes completely that is a straight forward task.

>Using  memory_order_acquire  with  any  built-in  atomic  function  except  atomic_work_item_fencerequires OpenCL C 2.0 or the __opencl_c_atomic_order_acq_rel feature macro.•  Using  memory_order_release  with  any  built-in  atomic  function  except  atomic_work_item_fencerequires OpenCL C 2.0 or the __opencl_c_atomic_order_acq_rel feature macro.•  Using  memory_order_acq_rel  with  any  built-in  atomic  function  except  atomic_work_item_fencerequires OpenCL C 2.0 or the __opencl_c_atomic_order_acq_rel feature macro

Before diving into any suggestion for its alternative ways to specify this it would be good to understand the motivation for special case of this function.

FYI, it was highlighted that there is the following similar restriction in OpenCL 2.0 for this function that has never been implemented in the upstream clang:

> The atomic_work_item_fence(CLK_IMAGE_MEM_FENCE) built-in function can be used to make sure that
> sampler-less   writes   are   visible   to   later   reads   by   the   same   work-item.   Only   a   scope   of memory_scope_work_item  and  an  order  of  memory_order_acq_rel  is  valid  for  atomic_work_item_fencewhen  passed  the  CLK_IMAGE_MEM_FENCE  flag.

bashbaug:
I believe the way we got here was by looking at OpenCL 1.2 device capabilities:

* For the OpenCL 1.2 atomic functions (e.g. `atomic_add`), there is no expectation that they enforce memory consistency, so they are "relaxed" atomics.

* To enforce memory consistency, OpenCL 1.2 has fence functions (e.g. `mem_fence`).  While the definition of these functions is somewhat ambiguous, they appear to describe "acquire-release" fences.

So, in order to describe OpenCL 1.2 device capabilities using the C11 / OpenCL C 2.0 atomic functions, we need to support no more than relaxed memory orders for most atomic functions, but we need to support the acquire and release memory orders for `atomic_work_item_fence`.  This is how we ended up with the wording referenced above and the special case for `atomic_work_item_fence`.

If we want to eliminate the special case and simplify tooling (and maybe the spec also), here are a few options:

[...]

PR #384, opened 2020-07-27
unify atomic acquire and release capabilities for ops and fences

bashbaug:
This is a draft PR for discussion.  It describes an alternative way of expressing atomic capabilities and feature macros that is simpler to understand and implement, but is not as expressive as the current method of expressing atomic capabilities.  The alternative was first discussed in an issue here: https://github.com/KhronosGroup/OpenCL-Docs/issues/354#issuecomment-662768559.

In short, this PR unifies the atomic acquire and release capabilities for atomic operations (e.g. `atomic_fetch_add_explicit` and atomic fences (e.g. `atomic_work_item_fence`).  Without this change, the atomic acquire and release memory orders could be supported for fences only but not other atomic operations, to match the semantics of the OpenCL 1.2 atomic operations and memory fences.  With this change, an implementation coming from OpenCL 1.2 will either be unable to express the OpenCL 1.2 memory fences via the C11-style atomics (if it does not support the acquire and release memory orders), or will need to add support for acquire and release memory orders for atomic operations (if it does support the acquire and release memory orders).

Note that this PR does NOT change the availability of the OpenCL 1.2 memory fence functions `mem_fence`, `read_mem_fence`, and `write_mem_fence`, and these functions will remain required even if an implementation does not support the explicit acquire and release memory orders.

Please also see the comment in the PR itself regarding the host API queries for atomic memory capabilities.

### DIFF
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -89,7 +89,6 @@ Note that a device that provides the same level of capabilities as an OpenCL 2.x
 | May return:
 
 {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
-{CL_DEVICE_ATOMIC_ORDER_ACQ_REL} \| +
 {CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP}
 
 indicating that _device_ does not support the full memory consistency model for atomic fence operations.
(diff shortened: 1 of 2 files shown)


api/appendix_h.asciidoc
```diff
@@ -89,7 +89,6 @@ Note that a device that provides the same level of capabilities as an OpenCL 2.x
 | May return:
 
 {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
```
bashbaug, line 91: We have a couple of options for this `CL_DEVICE_ATOMIC_FENCE_CAPABILITIES` query:

1. We could remove it completely, if it's redundant with `CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES`.
2. We could keep it as defined in this PR and remove the requirement to support `CL_DEVICE_ATOMIC_ORDER_ACQ_REL`.
3. We could revert this change, and continue to require `CL_DEVICE_ATOMIC_ORDER_ACQ_REL`, since this is the behavior of the OpenCL 1.2 `mem_fence` built-ins even if they don't use the acquire and release memory orders explicitly.

Note that we have a separate query for the OpenCL C features.  We should also define (or at least understand) the relationship between these atomic queries and the `__opencl_c_atomic_order_acq_rel` feature macro.

alycm: Thinking on option (1), it seems the cleanest.  The main case that it doesn't support is hardware with strict fences but no way to support strict memory operations (or vice-versa, but that would be more surprising).  I.e. the existing OpenCL 1.2 `mem_fence` problem, or a stronger version of it.

The OpenCL C 3.0 already uses this language for the relevant feature macros (emphasis mine):

> Indicating atomic operations **and** fences support" 

Theoretically other languages, e.g. IL, may not pair these.

AnastasiaStulova:
Looks neater and will minimize programmer errors due to incorrect use of atomics that can't be detected at compile time.

bashbaug:
We discussed this change in the August 4th teleconference and decided to keep the spec as-is, so I am closing this PR without merging.

PR #385, opened 2020-07-27
Reduce the verbosity of int64 and fp64 feature macros

alycm:
Undo some of the wording from https://github.com/KhronosGroup/OpenCL-Docs/pull/372 and replace it with simpler footnotes.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -5119,9 +5119,6 @@ indicated by the presence of the `+__opencl_c_int64+` feature macro.
   ulong__n__ *upsample*(uint__n__ _hi_, uint__n__ _lo_)
     | _result_[i] = ((long)_hi_[i] << 32) \| _lo_[i] +
       _result_[i] = ((ulong)_hi_[i] << 32) \| _lo_[i]
-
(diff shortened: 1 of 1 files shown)


alycm:
I've filed https://github.com/KhronosGroup/OpenCL-Docs/issues/386 to address the footnote numbering issue.

StuartDBrady:
This seems like an improvement.

Later, we can perhaps add some wording for functions that operate on scalar doubles (and so are not covered by the footnotes for gentype and gentyped). I think the intent is clear without this, though.

(The type promotion rules could be a potential loophole, but the wording in _§ 6.13 Builtin Functions_ stating that "implicit scalar widening shall not occur" covers most of this, and although I can't see where it is stated that as_double() is forbidden without the fp64 feature, but if that is a problem, it is not new.)

alycm:
> Just one small comment: it looks like there is still a bit of extra verbosity for __opencl_c_int64 in the description of the upsample function.

Oops, deleted that line now.

> Later, we can perhaps add some wording for functions that operate on scalar doubles (and so are not covered by the footnotes for gentype and gentyped). I think the intent is clear without this, though.

I started covering that, but it just became more an more words.   Maybe something more in the initial tables that define the types would be enough.

PR #388, opened 2020-07-28
Delete some zero-width spaces added by mistake

alycm:
I copy-pasted from rendered HTML spec and accidentally included some
zero-width spaces in the raw text, delete those.


alycm:
@StuartDBrady You're correct, I've changed to commit and description.

Issue #391, opened 2020-07-29
CL_COMMAND_SVM_MIGRATE_MEM not documented in normative 3.0 specification


kpet:
`CL_COMMAND_SVM_MIGRATE_MEM` is only mentioned once in Appendix E. It should be mentioned in the description for `CL_EVENT_COMMAND_TYPE`. Also we should say what command type is returned for `clEnqueueSVMMigrateMem` prior to 3.0. One option could be to say it's implementation-defined, another to require implementations report `CL_COMMAND_MIGRATE_MEM_OBJECTS`.

bashbaug:
I believe `CL_COMMAND_SVM_MIGRATE_MEM` is described in the normative spec, at least as well as the other `CL_EVENT_COMMAND_TYPEs`.  See the bottom of the table cell, [here](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_EVENT_COMMAND_TYPE).

Admittedly this isn't a very good definition of _any_ of the command types.  It's also a very long table cell, which messes up the PDF rendering (it's possible this is why you didn't see it previously), so I've been thinking of ways to refactor it.  One idea I had was to add a separate table below this one that describes the `CL_DEVICE_COMMAND_TYPE` and the API call that generates it.  Does that sound reasonable?

As for pre-OpenCL 3.0 behavior, I believe our currently shipping, pre-OpenCL 3.0 implementations currently return `CL_COMMAND_MIGRATE_MEM_OBJECTS` for this query, but I think it's best to say implementation-defined before OpenCL 3.0.

kpet:
Agree it's there in the source.

>It's also a very long table cell, which messes up the PDF rendering (it's possible this is why you didn't see it previously)

Yes, that's what's happened. The table line doesn't fit on a page and was cut short in the PDF.

> I believe our currently shipping, pre-OpenCL 3.0 implementations currently return CL_COMMAND_MIGRATE_MEM_OBJECTS for this query, but I think it's best to say implementation-defined before OpenCL 3.0.

I think that's what we do as well IIRC. Agree that implementation-defined is the safest but I still find it tempting to mandate `CL_COMMAND_MIGRATE_MEM_OBJECTS`. I don't think anyone would have done something else here really.

PR #394, opened 2020-08-03
add a separate table for event command types

bashbaug:
This is proposed fix for #391.

This PR adds a separate table for event command types, after the table of supported _param_names_ for `clGetEventInfo`.  Moving this information to a separate table shortens a long table cell that was broken in the PDF rendering that was causing `CL_COMMAND_SVM_MIGRATE_MEM` to be missing in the PDF spec.  It also clarifies which event command types are returned by which APIs, which was missing in the current specifications, and is a much better linking target for these event command types.

Open questions:

* What should the event command type for `clEnqueueSVMMigrateMem` be prior to OpenCL 3.0?  In this PR, it's described as "implementation-defined", which is safest, but perhaps we could do a little better.

* I've removed (commented out) the event command types `CL_COMMAND_ACQUIRE_GL_OBJECTS` and `CL_COMMAND_RELEASE_GL_OBJECTS` from this table.  They were in this table previously, but they are returned for extension APIs and not core APIs.  Is this the right thing to do?  If so, I should add these event command types to the proper extension spec.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -8706,61 +8706,8 @@ include::{generated}/api/version-notes/CL_EVENT_CONTEXT.asciidoc[]
 
 include::{generated}/api/version-notes/CL_EVENT_COMMAND_TYPE.asciidoc[]
   | cl_command_type
-      | Return the command type associated with _event_.
-        Can be one of the following values:
-
-// OpenCL 1.0
-        {CL_COMMAND_NDRANGE_KERNEL_anchor} +
-        {CL_COMMAND_TASK_anchor} +
-        {CL_COMMAND_NATIVE_KERNEL_anchor} +
-        {CL_COMMAND_READ_BUFFER_anchor} +
-        {CL_COMMAND_WRITE_BUFFER_anchor} +
-        {CL_COMMAND_COPY_BUFFER_anchor} +
-        {CL_COMMAND_READ_IMAGE_anchor} +
-        {CL_COMMAND_WRITE_IMAGE_anchor} +
-        {CL_COMMAND_COPY_IMAGE_anchor} +
-        {CL_COMMAND_COPY_BUFFER_TO_IMAGE_anchor} +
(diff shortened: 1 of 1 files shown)


bashbaug:
I created https://github.com/KhronosGroup/OpenCL-Docs/issues/400 so we don't lose track of the event command type for extension enqueue APIs.

alycm:
> What should the event command type for clEnqueueSVMMigrateMem be prior to OpenCL 3.0? In this PR, it's described as "implementation-defined", which is safest, but perhaps we could do a little better.

> As previously discussed on the issue I'm fine with this but I'm tempted to suggest we mandate CL_COMMAND_MIGRATE_MEM_OBJECTS. It's hard to imagine an implementation having done something else that's justifiable/reasonable.

The only OpenCL 2.1 implementations listed on the [conformant products page](https://www.khronos.org/conformance/adopters/conformant-products/opencl) are from Intel and Arm, both have stated they currently return `CL_COMMAND_MIGRATE_MEM_OBJECTS`.  Of course, there could be other in-progress implementations.

The main advantage of "implementation defined" is that from the perspective of an OpenCL programmer just reading the spec it seems less confusing to transition from "implementation defined" to "specific" than an arbitrary change of command type.  But doesn't help with knowing what code to check for.

> I've removed (commented out) the event command types CL_COMMAND_ACQUIRE_GL_OBJECTS and CL_COMMAND_RELEASE_GL_OBJECTS from this table. They were in this table previously, but they are returned for extension APIs and not core APIs. Is this the right thing to do? If so, I should add these event command types to the proper extension spec.

Moving them to the extension specification seems correct, i.e. https://github.com/KhronosGroup/OpenCL-Docs/issues/400 is fine.

bashbaug:
I just checked and our CPU OpenCL 2.1 implementation currently does NOT return `CL_COMMAND_MIGRATE_MEM_OBJECTS`, unfortunately.  Therefore, I think we should merge this change as-is, where the returned event command type is implementation-defined prior to OpenCL 3.0.

alycm:
Okay, if shipping 2.1 implementations are not consistent in their behaviour I agree that "implementation defined" is the best we can do for pre-3.0.

PR #395, opened 2020-08-04
unify the descriptions for event for all enqueue APIs

### LINKED ISSUE #312: Clarification on event creation
In several places there is text likes (event for clEnqueueNDRangeKernel)

> **event returns an event object that identifies this particular kernel-instance.** Event objects are unique and can be used to identify a particular kernel-instance later on. If event is NULL, no event will be created for this kernel-instance and therefore it will not be possible for the application to query or queue a wait for this particular kernel-instance. If the event_wait_list and the event arguments are not NULL, the event argument should not refer to an element of the event_wait_list array.

There is no exception for event creation if the clEnqueueNDRangeKernel fails.  I think this clarification should be added.

> event returns an event object that identifies this particular kernel-instance. Event objects are unique and can be used to identify a particular kernel-instance later on. If event is NULL **or the return code is not CL_SUCCESS**, no event will be created for this kernel-instance and therefore it will not be possible for the application to query or queue a wait for this particular kernel-instance. If the event_wait_list and the event arguments are not NULL, the event argument should not refer to an element of the event_wait_list array.

I believe the same problem exists in all APIs that return an event.

bashbaug:
Fixes #312.

* Clarify that events are not created if the enqueue fails.
* Clarify that the event cannot also be in the event wait list.
* Use the same text consistently for all event descriptions.

The exact spec text is:

>   * _event_ returns an event object that identifies this command and can be used to query or queue a wait for this command to complete. If _event_ is `NULL` or the enqueue is unsuccessful, no event will be created and therefore it will not be possible to query the status of this command or to wait for this command to complete. If _event_wait_list_ and _event_ are not `NULL`, _event_ must not refer to an element of the _event_wait_list_ array.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -816,15 +816,13 @@ include::{generated}/api/version-notes/clEnqueueWriteBuffer.asciidoc[]
     must be the same.
     The memory associated with _event_wait_list_ can be reused or freed after
     the function returns.
-  * _event_ returns an event object that identifies this particular read / write
-    command and can be used to query or queue a wait for this particular command
-    to complete.
-    _event_ can be `NULL` in which case it will not be possible for the
-    application to query the status of this command or queue a wait for this
-    command to complete.
-    If the _event_wait_list_ and the _event_ arguments are not `NULL`, the
-    _event_ argument should not refer to an element of the _event_wait_list_
-    array.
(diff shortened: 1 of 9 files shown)


bashbaug:
Ping @raunraun, since you are the submitter of #312.  Thanks!

raunraun:
LGTM

PR #398, opened 2020-08-05
clarify the values returned by properties queries

### LINKED ISSUE #317: Values Returned by Properties Array Queries
This was internal issue 217.

>> I think you don't strictly need to store the entire list as passed by the application. Nothing clearly calls for the properties to be returned in the same order. You could recompose the list in clGet*Info from internal state.
>> I think we should decide now whether we want the order to ever be significant or the value returned to be the exact list as passed by the user or the implementation's interpretation of that list and make this 100% explicit.

> In addition to the order of queried properties I think we also need to decide what to do for properties that were not specified by the application.  Specifically, is it OK to include default values in the queried properties array?  If so, note that the size of the queried properties array may be larger than the size of the properties array passed by the application.

And a comment in response.

>> Specifically, is it OK to include default values in the queried properties array?

> I would lean toward yes.
> I think these queries are more useful if they return the implementation's interpretation of user-supplied properties and internal defaults. It allows applications (probably more libraries / utilities / tools) to check that the implementation has interpreted the input correctly. This means that we need to make ordering insignificant but I think the opposite would likely result in confusion or dependency issues anyway.

bashbaug:
Fixes #317.

Clarifies that the values returned by the new OpenCL 3.0 properties queries must exactly match the values passed when the object was created.  The properties cannot be in a different order, and the properties cannot include other (default) properties.

I have not changed the specified behavior for `NULL` properties in this PR but I will add this if desired - see the last question in https://github.com/KhronosGroup/OpenCL-Docs/issues/317#issuecomment-668730289.

tagging @kerilk  @kpet

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -2178,17 +2178,17 @@ include::{generated}/api/version-notes/CL_CONTEXT_PROPERTIES.asciidoc[]
         {clCreateContextFromType}.
 
         If the _properties_ argument specified in {clCreateContext} or
-        {clCreateContextFromType} used to create _context_ is not `NULL`, the
-        implementation must return the values specified in the properties
-        argument.
+        {clCreateContextFromType} used to create _context_ was not `NULL`,
+        the implementation must return the values specified in the
+        properties argument in the same order and without including
+        additional properties.
 
         If the _properties_ argument specified in {clCreateContext} or
-        {clCreateContextFromType} used to create _context_ is `NULL`, the
-        implementation may return either a _param_value_size_ret_ of 0
-        i.e. there is no context property value to be returned or can return
-        a context property value of 0 (where 0 is used to terminate the
-        context properties list) in the memory that _param_value_ points
-        to.
+        {clCreateContextFromType} used to create _context_ was `NULL`, the
+        implementation may return either a
+        _param_value_size_ret_ of 0 (i.e. there are no properties to be
+        returned), or the implementation may return a property value of 0
+        (where 0 is used to terminate the properties list).
 |====
 
 11::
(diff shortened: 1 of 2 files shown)


api/opencl_platform_layer.asciidoc
```diff
@@ -2178,17 +2178,17 @@ include::{generated}/api/version-notes/CL_CONTEXT_PROPERTIES.asciidoc[]
...
-        implementation may return either a _param_value_size_ret_ of 0
-        i.e. there is no context property value to be returned or can return
-        a context property value of 0 (where 0 is used to terminate the
-        context properties list) in the memory that _param_value_ points
-        to.
+        {clCreateContextFromType} used to create _context_ was `NULL`, the
+        implementation may return either a
+        _param_value_size_ret_ of 0 (i.e. there are no properties to be
+        returned), or the implementation may return a property value of 0
+        (where 0 is used to terminate the properties list).
```
alycm, line 2191: I think this is the last open question referenced in the description.

It seems slightly complicated to to have two possible behaviours in this case.  I don't really have a preference between them, if forced to a side I'd go with the `0` size to match the `NULL` input.

bashbaug: Discussed in the August 11th teleconference and agreed to return a size of zero for a `NULL` input.  I will update this PR to reflect this.

bashbaug:
I've updated this PR with the decision from the August 11th teleconference regarding `NULL` properties.  The updated text (example below for contexts, but similar text exists for other objects also) is:

> If the _properties_ argument specified in **clCreateContext** or        **clCreateContextFromType** used to create _context_ was `NULL`, the        implementation must return _param_value_size_ret_ equal to 0,        indicating that there are no properties to be returned.

api/opencl_platform_layer.asciidoc
bashbaug: Fixed!

alycm:
Textually looks fine, and conceptually was approved on the call.  Merging now.

PR #406, opened 2020-08-14
generate different zwsp for HTML and PDF docs

### LINKED ISSUE #393: Zero width spaces break search in HTML version of specs in Firefox
Searching for enum names in the HTML version of the specification is awkward due to the zero width spaces.  E.g. searching for `CL_DEVICE_NAME` fails due to the zero width spaces

However, search in the HTML version in Chrome is fine.  Search in the PDF version in the viewer I tried is fine.  In fact, even search in the PDF version in Firefox's embedded PDF view is fine!

There is a Firefox bug about this: https://bugzilla.mozilla.org/show_bug.cgi?id=389651

Replacing the zero width spaces with a `<wbr>` tags in the HTML output works in both Firefox and Chrome:

```sh
sed -i out/html/OpenCL_API.html -e 's/&#8203;/<wbr>/g'
```
However, adjusting the script that inserts `&#8203` into the Asciidoctor to use `<wbr>` instead breaks PDF rendering.

bashbaug:
Fixes #393 

This is a small modification to the dictionary generation script to generate different "zero width space" asciidoctor attributes for the HTML spec (`<wbr>`) and the PDF spec (`&#8203;`).  Using the different character for HTML specs enables searching and easier copying and pasting for at least some browsers.

There should otherwise be no visible changes to the HTML spec output, and no changes at all to the PDF spec output.

### DIFF
--- a/scripts/gen_dictionaries.py
+++ b/scripts/gen_dictionaries.py
@@ -78,18 +78,13 @@ if __name__ == "__main__":
             name = api.get('name')
             #print('found api: ' + name)
 
-            # Create a variant of the name that precedes underscores with
-            # "zero width" spaces.  This causes some long names to be
-            # broken at more intuitive places.
-            sName = name.replace("_", "_&#8203;")
-
             # Example with link:
             #
             # // clEnqueueNDRangeKernel
             # :clEnqueueNDRangeKernel_label: pass:q[*clEnqueueNDRangeKernel*]
             # :clEnqueueNDRangeKernel: <<clEnqueueNDRangeKernel,{clEnqueueNDRangeKernel_label}>>
             linkFile.write('// ' + name + '\n')
-            linkFile.write(':' + name + '_label: pass:q[*' + sName + '*]\n')
+            linkFile.write(':' + name + '_label: pass:q[*' + name + '*]\n')
             linkFile.write(':' + name + ': <<' + name + ',{' + name + '_label}>>\n')
             linkFile.write('\n')
 
@@ -98,7 +93,7 @@ if __name__ == "__main__":
             # // clEnqueueNDRangeKernel
             # :clEnqueueNDRangeKernel: pass:q[*clEnqueueNDRangeKernel*]
             nolinkFile.write('// ' + name + '\n')
-            nolinkFile.write(':' + name + ': pass:q[*' + sName + '*]\n')
+            nolinkFile.write(':' + name + ': pass:q[*' + name + '*]\n')
             nolinkFile.write('\n')
 
             numberOfFuncs = numberOfFuncs + 1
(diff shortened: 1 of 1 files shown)


scripts/gen_dictionaries.py
```diff
@@ -154,7 +145,12 @@ def GetFooter():
                 #:CL_MEM_READ_ONLY: <<CL_MEM_READ_ONLY,{CL_MEM_READ_ONLY_label}>>
                 #:CL_MEM_READ_ONLY_anchor: [[CL_MEM_READ_ONLY]]{CL_MEM_READ_ONLY}
                 linkFile.write('// ' + name + '\n')
-                linkFile.write(':' + name + '_label: pass:q[`' + sName + '`]\n')
+                linkFile.write('ifdef::basebackend-html[]\n')
+                linkFile.write(':' + name + '_label: pass:q[`' + htmlName + '`]\n')
+                linkFile.write('endif::basebackend-html[]\n')
+                linkFile.write('ifdef::backend-pdf[]\n')
```
StuartDBrady, line 151: Should this be ifndef::basebackend-html instead? Asciidoctor has backends besides just the HTML and PDF ones, and with the preprocessing above, nothing would be output for those backends, should we choose to use them. Replacing with &#8203; for all backends except the HTML one would at least be in keeping with what we already had.

scripts/gen_dictionaries.py
```diff
@@ -165,7 +161,12 @@ def GetFooter():
                 #:CL_MEM_READ_ONLY: pass:q[`CL_MEM_READ_ONLY`]
                 #:CL_MEM_READ_ONLY_anchor: {CL_MEM_READ_ONLY}
                 nolinkFile.write('// ' + name + '\n')
-                nolinkFile.write(':' + name + ': pass:q[`' + sName + '`]\n')
+                nolinkFile.write('ifdef::basebackend-html[]\n')
+                nolinkFile.write(':' + name + ': pass:q[`' + htmlName + '`]\n')
+                nolinkFile.write('endif::basebackend-html[]\n')
+                nolinkFile.write('ifdef::backend-pdf[]\n')
```
StuartDBrady, line 167: As above.

scripts/gen_dictionaries.py
bashbaug: Yes, good point.  As I mentioned on #393 this didn't quite work out when I tried it though, and I think I figured out why.  In short, the basebackend for the PDF converter is also HTML, which is a little confusing.  See:

https://github.com/asciidoctor/asciidoctor/issues/2479

I've changed the script to instead look for the HTML5 **backend**, which I think has the desired behavior:

```asciidoc
// CL_CHAR_BIT
ifdef::backend-html5[]
:CL_CHAR_BIT_label: pass:q[`CL_<wbr>CHAR_<wbr>BIT`]
endif::[]
ifndef::backend-html5[]
:CL_CHAR_BIT_label: pass:q[`CL_&#8203;CHAR_&#8203;BIT`]
endif::[]
```

This should use the `<wbr>` syntax for our current HTML output, but will use `&#8203;` for everything else.  Does this seem better?  Thanks!

StuartDBrady: Looks good.

StuartDBrady: Does that mean that in this earlier version, when using the PDF backend, _both_ htmlName _and_ otherName were getting written? I.e. the recent change fixing things for other backends, but _also_ fixes the output for the PDF backend?

bashbaug: > Does that mean that in this earlier version, when using the PDF backend, _both_ htmlName _and_ otherName were getting written?

Yes, I believe so.  In the previous version, the attribute was first defined to use `<wbr>` because the PDF backend uses the HTML basebackend, but then it was re-defined to use `&#8203;` for the PDF backend.  This means that it worked for both the HTML and PDF backends, but a bit confusingly, and if the two conditions were switched it would break unexpectedly.

The updated version is much more robust.  Thanks again!

PR #409, opened 2020-08-18
clarify behavior for CL_QUEUE_SIZE for non-device queues

### LINKED ISSUE #402: Querying CL_QUEUE_SIZE for non-Device Queues
The current description for `clGetCommandQueueInfo(CL_QUEUE_SIZE)` is:

> Return the currently specified size for the device command-queue. This query is only supported for device command queues.

It's not clear what "this query is only supported for device command queues" means though.  Specifically:

* Is it an error to perform this query on a non-device queue?  If so, what should the error condition be?
* Or, does the query return a known quantity (such as `0`) for a non-device queue?
* Or, something else entirely?

FWIW, the current text in [Appendix H for Device-Side Enqueue](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#_device_side_enqueue) says that calling:

> clGetCommandQueueInfo, passing CL_​QUEUE_​SIZE... Returns 0 if the device associated with command_queue does not support On-Device Queues.

Depending how we resolve this issue we may need to update this text as well.

bashbaug:
Fixes #402

This PR clarifies that querying `CL_QUEUE_SIZE` requires that _command_queue_ be a device command-queue.  Otherwise, it is considered an invalid command queue and `clGetCommandQueueInfo` will return `CL_INVALID_COMMAND_QUEUE`.

Also updates Appendix H to describe that this query will return an error rather than returning `0` when an implementation does not support device-side enqueue and on-device queues.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -383,8 +383,10 @@ include::{generated}/api/version-notes/CL_QUEUE_PROPERTIES_ARRAY.asciidoc[]
 
 include::{generated}/api/version-notes/CL_QUEUE_SIZE.asciidoc[]
   | cl_uint
-      | Return the currently specified size for the device command-queue.
-        This query is only supported for device command queues.
+      | Return the size of the device command-queue.
+        To be considered valid for this query, _command_queue_ must be a
+        device command-queue.
+
 | {CL_QUEUE_DEVICE_DEFAULT_anchor}
 
 include::{generated}/api/version-notes/CL_QUEUE_DEVICE_DEFAULT.asciidoc[]
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -406,7 +406,8 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_COMMAND_QUEUE} if _command_queue_ is not a valid
-    command-queue.
+    command-queue, or if _command_queue_ is not a valid command-queue
+    for _param_name_.
   * {CL_INVALID_VALUE} if _param_name_ is not one of the supported values or
     if size in bytes specified by _param_value_size_ is < size of return
     type as specified in the <<command-queue-param-table,Command Queue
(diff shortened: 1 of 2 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -383,8 +383,10 @@ include::{generated}/api/version-notes/CL_QUEUE_PROPERTIES_ARRAY.asciidoc[]
 
 include::{generated}/api/version-notes/CL_QUEUE_SIZE.asciidoc[]
   | cl_uint
-      | Return the currently specified size for the device command-queue.
-        This query is only supported for device command queues.
+      | Return the size of the device command-queue.
+        To be considered valid for this query, _command_queue_ must be a
+        device command-queue.
```
kpet, line 388: I think we should add an entry for the error case below so we keep them all in the same place. Suggest something like:

>{CL_INVALID_COMMAND_QUEUE} if _param_name_ is {CL_QUEUE_SIZE} and  _command_queue_ is not a device command-queue.

bashbaug: I thought about this, but I'm wary of replicating information in the error descriptions, and especially listing a bunch of specific enums that could easily grow out-of-date.  Instead, I thought it was better to describe that the only valid command queues for this query are device command queues, and then the invalid command queue error falls out naturally for host command queues.

If that's a little too clever, how would you feel about a more general error condition, something like:

> `CL_INVALID_COMMAND_QUEUE` if _command_queue_ is invalid for _param_name_.

Thanks!

kpet: I understand the logic but I think it was too indirect. Agree that we should avoid duplication as much as possible but at the same time it is very useful to have all the error cases in one place when writing negative tests or layers/tools. I like your proposal as it achieves both: the list of error conditions is exhaustive _and_ we don't duplicate the details.

bashbaug: Sounds good, I've updated the list of error conditions also.  The error code description for `CL_INVALID_COMMAND_QUEUE` is now:

>   * `CL_INVALID_COMMAND_QUEUE` if _command_queue_ is not a valid command-queue, or if _command_queue_ is not a valid command-queue for _param_name_.

kpet: I'd prefer seeing the new case as a separate bullet point but I'm fine with what you've done.

kpet (approved):

alycm (approved):
LGTM.

PR #410, opened 2020-08-19
clarify KHR extension requirements for OpenCL 3.0 devices

bashbaug:
This PR adds the KHR extension requirements for OpenCL 3.0 devices discussed in the August 18th teleconference.

Specifically, when the optional OpenCL 3.0 features 3D image writes, depth images, or creating a 2D image from a buffer are supported by an OpenCL 3.0 device, the equivalent KHR extensions cl_khr_3d_image_writes, cl_khr_depth_images, or cl_khr_2d_image_from_buffer must also be returned in the device extension string.

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1167,20 +1167,19 @@ include::{generated}/api/version-notes/CL_DEVICE_EXTENSIONS.asciidoc[]
         *cl_khr_local_int32_extended_atomics*
 
         Additionally, the following Khronos extension names must be returned
-        by all devices that support OpenCL 1.2 when the optional feature is
-        supported:
+        by all devices that support OpenCL 1.2 when and only when the optional
+        feature is supported:
 
-        *cl_khr_fp64* (if double precision is supported)
+        *cl_khr_fp64*
 
         Additionally, the following Khronos extension names must be returned
         by all devices that support OpenCL 2.0, OpenCL 2.1, or OpenCL 2.2.
         For devices that support OpenCL 3.0, these extension names must only
-        be returned when the optional feature is supported:
+        be returned when and only when the optional feature is supported:
 
-        *cl_khr_3d_image_writes* (if 3D image writes are supported) +
-        *cl_khr_depth_images* (if depth images are supported) +
-        *cl_khr_image2d_from_buffer* (if creating 2D images from a buffer is
-        supported)
+        *cl_khr_3d_image_writes* +
+        *cl_khr_depth_images* +
+        *cl_khr_image2d_from_buffer*
 
         Please refer to the OpenCL Extension Specification or vendor
         provided documentation for a detailed description of these extensions.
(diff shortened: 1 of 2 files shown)


api/opencl_platform_layer.asciidoc
```diff
@@ -1167,18 +1167,20 @@ include::{generated}/api/version-notes/CL_DEVICE_EXTENSIONS.asciidoc[]
...
+        supported:
 
         *cl_khr_fp64* (if double precision is supported)
 
         Additionally, the following Khronos extension names must be returned
         by all devices that support OpenCL 2.0, OpenCL 2.1, or OpenCL 2.2.
-        Note that these Khronos extension names are *not* required for
-        devices supporting OpenCL 3.0:
+        For devices that support OpenCL 3.0, these extension names must only
+        be returned when the optional feature is supported:
```
kpet, line 1178: Being pedantic: maybe we want something along the lines of "when and only when".

api/opencl_platform_layer.asciidoc
```diff
@@ -1167,18 +1167,20 @@ include::{generated}/api/version-notes/CL_DEVICE_EXTENSIONS.asciidoc[]
...
 
-        *cl_khr_3d_image_writes* +
-        *cl_khr_depth_images* +
-        *cl_khr_image2d_from_buffer*
+        *cl_khr_3d_image_writes* (if 3D image writes are supported) +
+        *cl_khr_depth_images* (if depth images are supported) +
```
kpet, line 1181: Do we want to define what support for depth images mean? I note there's a bit of a contrast between what we say in Appendix H for 3d image writes and depth images :). Neither of which are normative BTW.

kpet (commented):
Agree with the intent, a couple pedantry comments.

api/opencl_platform_layer.asciidoc
bashbaug: Good clarification, fixed.

api/opencl_platform_layer.asciidoc
bashbaug: I updated Appendix H so it is consistent for 3D Image Writes and Depth Images (and sRGB images) now.

I didn't precisely define what is meant by "support for depth images" though.  If we to define this I'm open to it, but I don't think it's strictly necessary, and I don't think it belongs here in the description for `CL_DEVICE_EXTENSIONS` anyhow.

api/opencl_platform_layer.asciidoc
kpet: Looks good.

api/opencl_platform_layer.asciidoc
kpet: The change is definitely an upgrade but the next question becomes: what does it mean for the feature to be supported with no formats returned :).

bashbaug: I know the appendix is non-normative, but we do cover this case there (e.g. "Will not return any image formats with `image_channel_order` equal to `CL_DEPTH` if no devices in _context_ support Depth Images."), and we'll have a CTS test that tests for strange combinations like reporting the extension but no image formats.  I hope we're OK :)

alycm (approved):
LGTM

api/opencl_platform_layer.asciidoc
alycm: @kpet Is this sufficient for you?

alycm:
I think the last conversation thread is resolved, so I'm merging this.

PR #411, opened 2020-08-19
Reserve 16 enums for IMG.

Jeremy-Kemp:
Actual enum values will be filled in at a later date.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1719,8 +1719,12 @@ server's OpenCL/api-docs repository.
             <unused start="0x4213" end="0x421F"/>
     </enums>
 
-    <enums start="0x4220" end="0x4FFF" name="enums.4220" comment="Reserved for vendor extensions. Allocate in groups of 16.">
-            <unused start="0x420F" end="0x4FFF"/>
+    <enums start="0x4220" end="0x422F" name="enums.4220" comment="IMG">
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -1719,8 +1719,12 @@ server's OpenCL/api-docs repository.
             <unused start="0x4213" end="0x421F"/>
     </enums>
 
-    <enums start="0x4220" end="0x4FFF" name="enums.4220" comment="Reserved for vendor extensions. Allocate in groups of 16.">
-            <unused start="0x420F" end="0x4FFF"/>
+    <enums start="0x4220" end="0x4230" name="enums.4220" vendor="IMG">
```
alycm, line 1722: This reserves 17 values but reserving is done in blocks of 16.   If you actually need 17 values then reserve a block of size 32.

alycm (changes requested):
Reserve 1 fewer or 15 more values.

xml/cl.xml
Jeremy-Kemp: Woops, thats embarrassing. Should be fixed now.

alycm: Thanks!

alycm (approved):
LGTM, I'll just merge this immediately.

PR #413, opened 2020-08-20
Minor Updates to the Event Command Type Table

bashbaug:
This PR is optional and has a few minor updates to the event command type table added in #394:

* It adds a recommended event command type that should be returned for `clEnqueueSVMMigrateMem` pre-OpenCL 3.0, specifically `CL_COMMAND_MIGRATE_MEM_OBJECTS`, even though other implementation-defined event command types may also be returned.

Also, two editorial updates:

* It swaps the columns of the table, which I think is a little easier to follow.

* It adjusts the widths of the columns of the table to give a little more space to the event command types and a little less to the API names.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -8693,71 +8693,86 @@ include::{generated}/api/version-notes/CL_EVENT_REFERENCE_COUNT.asciidoc[]
 
 [[event-command-type-table]]
 .List of supported event command types
-[width="100%",cols="1,1",options="header"]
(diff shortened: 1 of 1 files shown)


kpet (approved):
+1 for the `CL_COMMAND_MIGRATE_MEM_OBJECTS` change. The rest looks fine but I haven't reviewed the details or looked at a built spec.

alycm (approved):
LGTM.  Using should/may language for the `CL_COMMAND_MIGRATE_MEM_OBJECTS` error code makes sense.

I rendered the table and it looks as expected.

PR #415, opened 2020-08-20
reserve cl_mem_flags bit for upcoming Intel extension

bashbaug:
Please reserve one `cl_mem_flags` bit for an upcoming Intel extension.

In this PR I am reserving bit 20 because we have a POC that is currently using this bit and this will save us some thrash, but if it is important to use the next bit (23?) to avoid gaps, please us know ASAP and we will switch.  Thanks!


alycm (approved):
LGTM, I see no reason to require the bit to be adjacent to others.

I won't merge immediately in case someone else has a reason.

alycm:
No one has said anything about the exact bit, so I'll just merge this now.

PR #418, opened 2020-08-25
use CL_INVALID_OPERATION as the error code for unsupported features

### LINKED ISSUE #408: Error Codes for Unsupported Functionality
When an optional feature is not supported in OpenCL 3.0 we currently use two classes of error codes.  See, for example, the error codes returned by pipe functions when pipes are not supported ([spec link](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#_pipes_2)):

* If a suitable error code did NOT exist in a prior version of OpenCL then the unsupported API returns `CL_INVALID_OPERATION`.  This is the case for `clCreatePipe`.
* If a suitable error code DID exist in a prior version of OpenCL then the unsupported API will return that error code instead.  This is the case for `clGetPipeInfo`, which returns `CL_INVALID_MEM_OBJECT`, since it is impossible to pass in a valid pipe object if pipes are not supported.

This issue is regarding the second item: Is it better to reuse existing error codes when a feature is not supported, limiting the error cases for e.g. `clGetPipeInfo`, or should we use a different error code instead to more clearly identify that the feature isn't supported?

bashbaug:
Fixes #408.

This PR updates three APIs related to optional OpenCL 3.0 features to use CL_INVALID_OPERATION as the error code when the feature is not supported vs. relying on implied error behavior from earlier versions of OpenCL.

The one case where a different error code may be returned is for clGetPipeInfo, since ICD implementations cannot create a valid
pipe memory object to query and get the ICD dispatch table from.  In these cases, the ICD loader will return CL_INVALID_MEM_OBJECT before calling into the ICD implementation.  Note that non-ICD implementations may return CL_INVALID_OPERATION unconditionally.

### DIFF
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -211,7 +211,7 @@ When Pipes are not supported:
 | Returns {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
 
 | {clGetPipeInfo}
-| Returns either {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object, or {CL_INVALID_OPERATION} if Pipes are not supported.
+| Returns {CL_INVALID_OPERATION} if no devices in the context associated with _pipe_ support Pipes.
 
 |====
 
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -211,8 +211,7 @@ When Pipes are not supported:
 | Returns {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
 
 | {clGetPipeInfo}
-| Returns {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object.
-// Note: for {CL_PIPE_PACKET_SIZE} and {CL_PIPE_MAX_PACKETS}.
+| Returns either {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object, or {CL_INVALID_OPERATION} if Pipes are not supported.
 
 |====
 
@@ -371,10 +370,10 @@ When Intermediate Language Programs are not supported:
 | Returns an empty buffer (such as _param_value_size_ret_ equal to `0`) if no devices in the context associated with _program_ support Intermediate Language Programs.
 
 | {clCreateProgramWithIL}
-| Returns {CL_INVALID_VALUE} if no devices in _context_ support Intermediate Language Programs.
+| Returns {CL_INVALID_OPERATION} if no devices in _context_ support Intermediate Language Programs.
 
 | {clSetProgramSpecializationConstant}
-| Returns {CL_INVALID_PROGRAM}, since _program_ cannot have been created from an Intermediate Language.
+| Returns {CL_INVALID_OPERATION} if no devices associated with _program_ support Intermediate Language Programs.
 
 |====
 
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -211,7 +211,7 @@ When Pipes are not supported:
 | Returns {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
 
 | {clGetPipeInfo}
-| Returns {CL_INVALID_OPERATION} if no devices in the context associated with _pipe_ support Pipes.
+| Returns {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object.
 
 |====
 
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -3649,8 +3649,6 @@ include::{generated}/api/version-notes/clGetPipeInfo.asciidoc[]
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_MEM_OBJECT} if _pipe_ is a not a valid pipe object.
-  * {CL_INVALID_OPERATION} if no devices in the context associated with _pipe_
-    support Pipes.
   * {CL_INVALID_VALUE} if _param_name_ is not valid, or if size in bytes
     specified by _param_value_size_ is < size of return type as described in
     the <<pipe-info-table,Pipe Object Queries>> table and _param_value_ is
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -3591,7 +3591,7 @@ Otherwise, it returns a `NULL` value with one of the following error values
 returned in _errcode_ret_:
 
   * {CL_INVALID_CONTEXT} if _context_ is not a valid context.
-  * {CL_INVALID_OPERATION} if no devices in _context_ support pipes.
+  * {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
   * {CL_INVALID_VALUE} if values specified in _flags_ are not as defined
     above.
   * {CL_INVALID_VALUE} if _properties_ is not `NULL`.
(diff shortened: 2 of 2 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -3591,7 +3591,7 @@ Otherwise, it returns a `NULL` value with one of the following error values
 returned in _errcode_ret_:
 
   * {CL_INVALID_CONTEXT} if _context_ is not a valid context.
-  * {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
+  * {CL_INVALID_OPERATION} if no devices in _context_ support pipes.
```
Kerilk, line 3594: You changed capitalization here but not above

api/appendix_h.asciidoc
```diff
@@ -211,8 +211,7 @@ When Pipes are not supported:
 | Returns {CL_INVALID_OPERATION} if no devices in _context_ support Pipes.
 
 | {clGetPipeInfo}
-| Returns {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object.
-// Note: for {CL_PIPE_PACKET_SIZE} and {CL_PIPE_MAX_PACKETS}.
+| Returns either {CL_INVALID_MEM_OBJECT} since _pipe_ cannot be a valid pipe object, or {CL_INVALID_OPERATION} if Pipes are not supported.
```
Kerilk, line 214: You did not change capitalization for Pipes here but did below

api/opencl_runtime_layer.asciidoc
bashbaug: Thanks - you are very observant!

I initially thought that the "feature names" were not capitalized in the main spec but were capitalized in the appendix, but it looks like there isn't a consistent capitalization policy.  For this PR, I've tried to be as consistent as possible with Pipes and Intermediate Language Programs, since both appear to be consistently capitalized, but to really "fix" this we'd want to take a pass through the specs to apply a consistent policy.  I'm willing to do this but I'd like to do it as a separate PR.

api/appendix_h.asciidoc
bashbaug: See above - thanks again!

api/appendix_h.asciidoc
```diff
@@ -371,10 +370,10 @@ When Intermediate Language Programs are not supported:
 | Returns an empty buffer (such as _param_value_size_ret_ equal to `0`) if no devices in the context associated with _program_ support Intermediate Language Programs.
 
 | {clCreateProgramWithIL}
-| Returns {CL_INVALID_VALUE} if no devices in _context_ support Intermediate Language Programs.
+| Returns {CL_INVALID_OPERATION} if no devices in _context_ support Intermediate Language Programs.
 
 | {clSetProgramSpecializationConstant}
-| Returns {CL_INVALID_PROGRAM}, since _program_ cannot have been created from an Intermediate Language.
+| Returns {CL_INVALID_OPERATION} if no devices associated with _program_ support Intermediate Language Programs.
```
alycm, line 376: This PR treats `clGetPipeInfo` and `clSetProgramSpecializationConstant` differently, in the former that the first has two possible return codes and the latter has one.  But they seem similar to me.

Taking possible uses of `clGetPipeInfo` on an implementation that doesn't support pipes:

```c
cl_mem bad_pipe = clCreatePipe(...);
cl_int err1 = clGetPipeInfo(bad_pipe, ...);

// ICD loader case: err1 is definitely CL_INVALID_MEM_OBJECT.
// Non-loader case: err1 is probably CL_INVALID_OPERATION.

cl_mem not_pipe = clCreateBufferWithProperties(...);
cl_int err2 = clGetPipeInfo(not_pipe, ...);

// ICD loader case: Uses 'not_pipe' to reach driver, err2 is probably CL_INVALID_OPERATION.
// Non-loader case: err2 is probably CL_INVALID_OPERATION.
```

So now, equivalently for `clSetProgramSpecializationConstant` on implementations that do not support IL.

```c
cl_program bad_program = clCreateProgramWithIL(...);
cl_int err3 = clSetProgramSpecializationConstant(bad_program, ...);

// ICD loader case: err3 is definitely CL_INVALID_PROGRAM.
// Non-loader case: err3 is probably CL_INVALID_OPERATION.

cl_program not_il = clCreateProgramWithSource(...);
cl_int err4 = clSetProgramSpecializationConstant(not_il, ...);

// ICD loader case: Uses 'not_il' to reach driver, err4 is probably CL_INVALID_OPERATION.
// Non-loader case: err4 is probably CL_INVALID_OPERATION.
```

Both `clGetPipeInfo` and `clSetProgramSpecializationConstant` seem equivalent in the sense that given an OpenCL program that is valid on an implementation that supports the relevant feature, the same program will result in `CL_INVALID_THING` from the ICD loader when backed by an implementation that does not support the relevant feature.

And both allow reaching the implementation anyway via OpenCL programs that will fail on any implementation, due to the multiple ways of getting a `cl_mem` or a `cl_program`.

So should both `clGetPipeInfo` and `clSetProgramSpecializationConstant` should have a similar error code pattern?  Such as:

```suggestion
| Returns either {CL_INVALID_PROGRAM} since _program_ cannot be a valid IL program, or {CL_INVALID_OPERATION} if no devices associated with _program_ support Intermediate Language Programs.
```

Or does the multiple-devices-in-context case make the difference?

alycm (commented):
LGTM in general, but a very long question.

api/appendix_h.asciidoc
bashbaug: Ah, this is a very good question.  Thank you for the very thorough write-up!

In short: I think you're right and I think we should remove the `CL_INVALID_MEM_OBJECT` condition from Appendix H.

I had forgotten that a pipe is still a `cl_mem` like any other, so even if you cannot create a `cl_mem` object that is an actual pipe when the implementation doesn't support pipes, you can create a non-pipe `cl_mem` object as shown in your `not_pipe` case, and get to an ICD implementation's `clGetPipeInfo` that way.

This does land us in the situation I was describing offline, where there are two error conditions but we are mandating that one of the two error conditions should be reported:

1. CL_INVALID_MEM_OBJECT if pipe is a not a valid pipe object. (existing)
2. CL_INVALID_OPERATION if pipes are not supported. (new in this PR)

I think that's OK though and it'll be helpful to return `CL_INVALID_OPERATION` consistently.  We will have a CTS test that checks for this behavior.

If this is OK, my action items are:

* Update this PR to remove `CL_INVALID_MEM_OBJECT` from the list of errors in Appendix H.
* I'll also update the error condition for `CL_INVALID_OPERATION` to be a bit more descriptive, so it'll say:
> * `CL_INVALID_OPERATION` if no devices in the context associated with _pipe_ support pipes.
* Update the CTS test to check for `CL_INVALID_MEM_OBJECT` specifically, by using the `not_pipe` case described above.

alycm: > I think that's OK though and it'll be helpful to return CL_INVALID_OPERATION consistently.

Yes, that sounds okay.

> We will have a CTS test that checks for this behavior.

Based on the main body of specification text saying in a footnote that "The OpenCL specification does not describe the order of precedence for error codes returned by API calls.", technically either error code is still valid, unless appendix H overrides that.  

However, the CTS test sounds fine and in the unlikely event that an implementation fails it then the test can be changed.

> Update the CTS test to check for CL_INVALID_MEM_OBJECT specifically, by using the not_pipe case described above.

Should this say `CL_INVALID_OPERATION`?

bashbaug: Great, I've made these changes.

> Should this say `CL_INVALID_OPERATION`?

Yes, it should, thanks!

alycm: Thanks!

alycm (approved):
LGTM!

alycm:
@Kerilk This PR has changed a bit since you reviewed it, are you okay with the updated version?

Jeremy-Kemp (approved):
LGTM. As already stated, KhronosGroup/OpenCL-CTS#912 assumes this change.

bashbaug:
Converting to a draft, we will make one small change to `clGetPipeInfo` as discussed in the September 8th teleconference.

bashbaug:
This is ready for final review and merging.

alycm:
LGTM, changes are consistent with discussion on call 2020-09-08, merging.

PR #421, opened 2020-08-27
publish the cl_khr_device_uuid extension spec

bashbaug:
Publishing as agreed on the OpenCL working group mailing list.

### DIFF
--- a/ext/cl_khr_device_uuid.asciidoc
+++ b/ext/cl_khr_device_uuid.asciidoc
@@ -2,101 +2,48 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
-= cl_khr_device_uuid
+[[cl_khr_device_uuid]]
+== Unique Device Identifiers
 
-// This section needs to be after the document title.
-:doctype: book
-:toc2:
-:toc: left
-:encoding: utf-8
-:lang: en
+This section describes the *cl_khr_device_uuid* extension.
 
-:blank: pass:[ +]
+This extension adds the ability to query a universally unique identifier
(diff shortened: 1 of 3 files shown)


nikhiljnv (commented):
Thanks Ben !! Looks good to me !!

nikhiljnv (approved):
Approving.

ext/cl_khr_device_uuid.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+// #define CL_DEVICE_LUID_KHR          0x106D
+// #define CL_DEVICE_NODE_MASK_KHR     0x106E
+// ----
+// 
+// Constants describing the size of the driver and device UUIDs, and the device LUID:
+// 
+// [source,c]
+// ----
+// #define CL_UUID_SIZE_KHR            16
+// #define CL_LUID_SIZE_KHR            8
```
alycm, line 43: These constants aren't in the XML so the headers will be the only written source for their values (once they're added there).

The `cl_khr_device_uuid` entry could have a `<require comment="Constants">` section.

alycm (commented):
LGTM, just one small XML point.

ext/cl_khr_device_uuid.asciidoc
bashbaug: The XML change is pretty straightforward - I'll add that and include it in this PR.

Do we want to describe the **values** of these constants here in the spec though, or only in the XML and headers?  We do include the values of several other constants in the spec, but they're mostly numerical limits in the half-precision extension.  We don't include the values of other enumerants (though we have talked about adding them).

bashbaug:
Discussed in the September 1st teleconference.  We will make the XML changes in a subsequent PR.

PR #424, opened 2020-08-28
Describe how new profiles may affect backwards compatibility

jrprice:
This adds a paragraph to the backwards compatibility section documenting the fact that minor versions of the standard may introduce new profiles that make existing capabilities optional, and includes a recommendation that applications check the profile supported by the device they are running on.

Discussed in internal issue !254 and on the 2020-08-25 teleconference.

### DIFF
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -2105,18 +2105,16 @@ Whenever an OpenCL C feature is optional in the new version of the OpenCL C prog
 If a feature macro is defined then the feature is supported by the OpenCL C compiler, otherwise the optional feature is not supported.
 
 In order to allow future versions of OpenCL to support new types of
-devices, some capabilities that are currently required of all OpenCL
-devices may become optional in future versions of OpenCL.
-When a required capability becomes optional, all existing profiles
-will continue to require that it be supported, thereby guaranteeing
-that applications targeting any of these profiles will continue to run
-with subsequent minor releases of OpenCL.
-New profiles that are introduced in or after the version of the
-standard that made the capability optional will not be required to
-support the capability.
+devices, minor releases of OpenCL may add new profiles where some
+features that are currently required for all OpenCL devices become
+optional.
+All features that are required for an OpenCL profile will also be
+required for that profile in subsequent minor releases of OpenCL,
+thereby guaranteeing backwards compatibility for applications
+targeting specific profiles.
 It is therefore strongly recommended that applications
 <<CL_DEVICE_PROFILE,query the profile>> supported by the OpenCL device
-they are running on, in order to remain robust to future changes.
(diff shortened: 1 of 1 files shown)


api/opencl_architecture.asciidoc
```diff
@@ -2104,6 +2104,20 @@ The new version of OpenCL C must be explicitly requested via the `-cl-std=` buil
 Whenever an OpenCL C feature is optional in the new version of the OpenCL C programming language, it will be paired with a feature macro, such as `+__opencl_c_feature_name+`, and a corresponding API query.
 If a feature macro is defined then the feature is supported by the OpenCL C compiler, otherwise the optional feature is not supported.
 
+In order to allow future versions of OpenCL to support new types of
+devices, some capabilities that are currently required of all OpenCL
```
alycm, line 2108: The paragraphs above this one use the term "feature" so I'd be inclined to continue using that.   Unless you're trying to stay away from the specific features made optional in OpenCL 3.0?

alycm (approved):
LGTM.

api/opencl_architecture.asciidoc
```diff
@@ -2104,6 +2104,20 @@ The new version of OpenCL C must be explicitly requested via the `-cl-std=` buil
...
+devices may become optional in future versions of OpenCL.
+When a required capability becomes optional, all existing profiles
+will continue to require that it be supported, thereby guaranteeing
+that applications targeting any of these profiles will continue to run
+with subsequent minor releases of OpenCL.
+New profiles that are introduced in or after the version of the
+standard that made the capability optional will not be required to
+support the capability.
```
bashbaug, line 2116: This is fairly nitpicky, but:

To me, the key takeaways from this paragraph should be:

* Within a profile, no features or capabilities that are currently required will become optional, except perhaps in a major API revision (and even then, this will hopefully be rare).
* New profiles may be added in minor API revisions with different required features or capabilities to enable new classes of devices.

Right now, the first sentence doesn't convey either of these key takeaways, and the next two sentences are necessary to clarify when a required capability may become optional.  I'd consider swapping the first two sentences so there's less room for possible confusion.  How about something like:

> All features that are required for an OpenCL profile will also be required in subsequent minor releases of OpenCL, thereby guaranteeing backwards compatibility for applications targeting specific profiles.  However, in order to allow future versions of OpenCL to support new devices, minor releases of OpenCL may add new profiles where some features that are currently required for all OpenCL profiles becomes optional.  It is therefore strongly recommended that applications query the profile supported by the OpenCL device they are running on in order to remain robust to future changes.

This is just a suggestion and these changes can go in as-is if needed.

jrprice:
I've updated the proposed text based on feedback, PTAL.

kpet (approved):
LGTM!

bashbaug:
This change has three approvals - merging.

PR #427, opened 2020-09-01
clarify the host_ptr usage for images with USE_HOST_PTR

### LINKED ISSUE #417: clCreateImage USE_HOST_PTR clarification
Between OpenCL 2.1 and 2.2 we added the following to clCreateImage:

> host_ptr is a pointer to the image data that may already be allocated by the application. **It is only used  to  initialize  the  image,  and  can  be  freed  after  the  call  to  clCreateImage.**

This seems to imply that USE_HOST_PTR is identical to COPY_HOST_PTR.  Is this intended?

bashbaug:
Fixes #417.

Removes the following sentence from the description of _host_ptr_ for the image creation functions such as `clCreateImage`:

> It is only used to initialize the image, and can be freed after the call to clCreateImage.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -1638,8 +1638,6 @@ include::{generated}/api/version-notes/clCreateImageWithProperties.asciidoc[]
     description of the image descriptor.
   * _host_ptr_ is a pointer to the image data that may already be allocated by
     the application.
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM.

kpet (approved):

Jeremy-Kemp (approved):
LGTM as well.

PR #428, opened 2020-09-01
add event command types for extension APIs

### LINKED ISSUE #400: Extension Enqueue APIs and Event Command Types
See: https://github.com/KhronosGroup/OpenCL-Docs/pull/394

We should check all of the KHR extensions that add new enqueue APIs and document their event command type.

This is certainly needed for the event command types CL_COMMAND_ACQUIRE_GL_OBJECTS and CL_COMMAND_RELEASE_GL_OBJECTS since they were removed from the core spec by #394, but we should check other extensions too while we're at it.

bashbaug:
Fixes #400.

This PR adds tables describing the event command types mapping for extensions that add new enqueue APIs.

### DIFF
--- a/ext/cl_khr_gl_sharing__memobjs.asciidoc
+++ b/ext/cl_khr_gl_sharing__memobjs.asciidoc
@@ -751,3 +751,28 @@ behavior.
 Similarly, attempting to access a shared CL/GL object from OpenCL before it
 has been acquired by the OpenCL command queue, or after it has been
 released, will result in undefined behavior.
+
+[[cl_khr_gl_sharing__memobjs-event-command-types]]
+==== Event Command Types for Sharing memory objects that map to GL objects
+
+The following table describes the event command types for the OpenCL commands
+to acquire and release OpenCL memory objects that have been created from
+OpenCL objects:
+
+.List of supported event command types
+[width="100%",cols="2,3",options="header"]
+|====
+| *Events Created By*
+| *Event Command Type*
+
+| {clEnqueueAcquireGLObjects}
+| {CL_COMMAND_ACQUIRE_GL_OBJECTS_anchor}
+
+include::{generated}/api/version-notes/CL_COMMAND_ACQUIRE_GL_OBJECTS.asciidoc[]
+
+| {clEnqueueReleaseGLObjects}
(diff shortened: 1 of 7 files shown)


scripts/gen_dictionaries.py
```diff
@@ -127,7 +127,9 @@ def GetFooter():
     for enums in spec.findall('enums'):
         # Skip Vendor Extension Enums
         vendor = enums.get('vendor')
-        if not vendor or vendor == 'Khronos' or vendor == 'Multiple':
+        name = enums.get('name')    # special-case: enum block with KHR enums assigned to vendor
+        include_anyway = name == 'enums.4010'
```
Jeremy-Kemp, line 131: No issue with this, but does this just happen when an extension goes from being a Vendor extension to a KHR?

Jeremy-Kemp (approved):
LGTM!

scripts/gen_dictionaries.py
bashbaug: I filed https://github.com/KhronosGroup/OpenCL-Docs/issues/431 so we don't lose track of this.  I agree this currently solution is a bit of a hack.

I'll also see if the script could also just generate dictionaries for all enums, if there is no downside to doing this.

ext/cl_khr_gl_sharing__memobjs.asciidoc
```diff
@@ -751,3 +751,28 @@ behavior.
 Similarly, attempting to access a shared CL/GL object from OpenCL before it
 has been acquired by the OpenCL command queue, or after it has been
 released, will result in undefined behavior.
+
+[[cl_khr_gl_sharing__memobjs-event-command-types]]
+==== Event Command Types for Sharing memory objects that map to GL objects
+
+The following table describes the event command types for the OpenCL commands
+to acquire and release OpenCL memory objects that have been created from
+OpenCL objects:
```
alycm, line 760:
```suggestion
OpenGL objects:
```

alycm (approved):
LGTM, just one tiny typo.

ext/cl_khr_gl_sharing__memobjs.asciidoc
bashbaug: Thanks - fixed!

PR #430, opened 2020-09-02
Correct `atomic_flag_test_and_set_explicit` `__local` Overload

FranklandJack:
* Use `__local` rather than `__global` in the second overload of the
`atomic_flag_test_and_set_explicit`. Currently both declarations use
`__global` and are identical.


alycm (approved):
Thanks, LGTM!

Jeremy-Kemp (approved):

PR #439, opened 2020-09-08
Clarify when CL_IMAGE_FORMAT_NOT_SUPPORTED is reported on image creation

kpet:


bashbaug (approved):
LGTM, this is a definite improvement and matches my understanding when this error should be returned.  I'll leave this open for another day or so just in case someone else has a different opinion.

alycm (approved):
LGTM.

bashbaug:
No reason to wait any longer I think.  Merging!

PR #440, opened 2020-09-09
Fix a typo: workgroup -> work-group

bader:


bashbaug (approved):
LGTM - thanks!

I'd eventually like to go through the entire spec to be sure we use the text "work-item" and "work-group" (and maybe "sub-group"?) consistently, since there are other places where we use alternate spacing or hyphenation, but this PR is definitely an improvement.

bashbaug:
No reason to wait to merge this I think, since this is purely a formatting change.

PR #442, opened 2020-09-10
fix two issues with the asciidoctor attributes

bashbaug:
Fixes for two small asciidoctor attribute-related issues:

1. One of the links was to clCreateProgramWithBinaries, but the actual API name is clCreateProgramWithBinary.
2. The XML file was missing an entry for the cl_uchar type.


alycm (approved):
LGTM, merging immediately as "obvious".

PR #444, opened 2020-09-15
add a TYPE suffix to all asciidoctor type attributes

### LINKED ISSUE #443: handle case-insensitive asciidoctor attributes
Unfortunately, asciidoctor attributes are currently case-insensitive.  This means that the asciidoctor enum attribute for `CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES` is indistinguishable for the asciidoctor type attribute for `cl_device_device_enqueue_capabilities`.  When both attributes are indistinguishable, it looks like the last one "wins", and as a result there are several places in the spec where the type name is used in place of the enum name.

There is an Asciidoctor issue open to fix this, but it's been open for a while without any updates.

Options to fix this issue in the OpenCL specs:

1. Add a prefix or suffix to one class of attributes to disambiguate, probably the type name since there are fewer of them.
2. Special-case this particular attribute and try to avoid duplicates in the future.

I suppose we could also rename either the type or the enum to disambiguate - is there any other reason to prefer unique case-insensitive names? - but that feels a bit like the tail wagging the dog.

I'll probably bite the bullet and go with (1) if it doesn't look too bad.

bashbaug:
Fixes #443.

Because asciidoctor attributes are not currently case-sensitive, the TYPE suffix is needed to disambiguate between an API type and an enum with the same name.

### DIFF
--- a/api/appendix_a.asciidoc
+++ b/api/appendix_a.asciidoc
@@ -53,15 +53,15 @@ being used by another command-queue are undefined.
 == Multiple Host Threads
 
 All OpenCL API calls are thread-safe footnote:[{fn-thread-safe}] except those
-that modify the state of {cl_kernel} objects: {clSetKernelArg},
+that modify the state of {cl_kernel_TYPE} objects: {clSetKernelArg},
 {clSetKernelArgSVMPointer}, {clSetKernelExecInfo} and {clCloneKernel}.
 {clSetKernelArg}, {clSetKernelArgSVMPointer}, {clSetKernelExecInfo} and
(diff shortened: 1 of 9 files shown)


bashbaug:
FWIW, I found another case that will be fixed by this PR: Without this change there is an ambiguity between the `CL_FLOAT` image channel data type and the `cl_float` type.

alycm (approved):
LGTM.   I hadn't realized that the actual wrong text was ending up in the rendered output.

The changes are all very procedural, this doesn't need to wait for extra reviews, I'll merge immediately.

PR #447, opened 2020-09-18
add CL_KERNEL_COMPILE_NUM_SUB_GROUPS to Appendix H

### LINKED ISSUE #445: Clarification requested on handling of clGetKernelSubGroupInfo parameter CL_KERNEL_COMPILE_NUM_SUBGROUPS in 3.0 spec
Table 34 of the OpenCL 3.0 API spec has the following reference to CL_KERNEL_COMPILE_NUM_SUBGROUPS

CL_KERNEL_COMPILE_NUM_SUB_GROUPS
Missing before version 2.1. Also see extension cl_khr_subgroups.	ignored	size_t
Returns the number of sub-groups per work-group specified in the kernel source or IL. If the sub-group count is not specified then 0 is returned.

However we could not find the corresponding token for source or IL in the OpenCL_C spec or OpenCL_Env spec. The closest reference we could find was "cl::required_num_sub_groups" in
https://www.khronos.org/registry/OpenCL/specs/2.2/pdf/OpenCL_Cxx.pdf
Should CL_KERNEL_COMPILE_NUM_SUBGROUPS be considered optional for implementations that support OpenCL C only.

bashbaug:
Fixes #445.

Since there is no way to specify this attribute for programs created from source, the query will always return zero if the device does not support intermediate language programs.

### DIFF
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -397,6 +397,10 @@ When Intermediate Language Programs are not supported:
 | {clSetProgramSpecializationConstant}
 | Returns {CL_INVALID_OPERATION} if no devices associated with _program_ support Intermediate Language Programs.
 
+| {clGetKernelSubGroupInfo}, passing +
+{CL_KERNEL_COMPILE_NUM_SUB_GROUPS}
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM, but @bcalidas should review as well.

bcalidas (approved):
Looks good

PR #448, opened 2020-09-21
Allow use of CL_DEVICE_COMPILER_AVAILABLE for IL

### LINKED ISSUE #438: Allow clBuildProgram to return CL_COMPILER_NOT_AVAILABLE for IL programs
Currently `clBuildProgram` has an error condition:

> CL_COMPILER_NOT_AVAILABLE if program is created with clCreateProgramWithSource and a compiler is not available i.e. CL_DEVICE_COMPILER_AVAILABLE specified in the Device Queries table is set to CL_FALSE.

I think this should be:

> CL_COMPILER_NOT_AVAILABLE if program is created with clCreateProgramWithSource **or clCreateProgramWithIL** and a compiler is not available i.e. CL_DEVICE_COMPILER_AVAILABLE specified in the Device Queries table is set to CL_FALSE.

Emphasis just for clarity here, not what would go in the specification.

I don't know if any existing 2.1 implementations support a compiler-less configuration, but if they do I don't see what else they could do for the `clCreateProgramWithIL` case.

alycm:
Currently how the use of IL programs should behave when
`CL_DEVICE_COMPILER_AVAILABLE` is `CL_FALSE` is underspecified.  This change
allows `CL_DEVICE_COMPILER_AVAILABLE` to be returned by entry points that
require use of a compiler.

`clBuildProgram` could already return `CL_DEVICE_COMPILER_AVAILABLE` for OpenCL
source, so extend that to IL.  `clCompileProgram` can already return
`CL_DEVICE_COMPILER_AVAILABLE` for any input type.

I've also added the `CL_DEVICE_COMPILER_AVAILABLE` to
`clSetProgramSpecializationConstant` because for it to be able to check the
validity of the provided `spec_id` it must be able to parse the IL program.

Fixes #438

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -5716,6 +5716,10 @@ Otherwise, it returns one of the following errors:
     language does not support specialization constants.
   * {CL_INVALID_OPERATION} if no devices associated with _program_ support
     Intermediate Language Programs.
+  * {CL_COMPILER_NOT_AVAILABLE} if _program_ is created with
+    {clCreateProgramWithIL} and a compiler is not
+    available, i.e. {CL_DEVICE_COMPILER_AVAILABLE} specified in the
+    <<device-queries-table,Device Queries>> table is set to {CL_FALSE}.
   * {CL_INVALID_SPEC_ID} if _spec_id_ is not a valid specialization constant
     identifier.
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Makes sense to me.  Changes are exactly as described in the PR description.

kpet (approved):

PR #449, opened 2020-09-21
Program scope constructor/destructor queries require a built program

### LINKED ISSUE #436: Program scope constructor/destructor clGetProgramInfo queries don't require a built program
While looking at https://github.com/KhronosGroup/OpenCL-Docs/issues/435 I realized that a similar issue was introduced in OpenCL 2.2, but due to the lack of conformant 2.2 implementations could still be fixed.  This is not a critical point as the entire feature behind these queries is both deprecated and optional in OpenCL 3.0, but I'll put it in the 3.0 project for now but can move to maintenance if it's distracting.

The `clGetProgramInfo` `CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT` and `CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT` queries currently don't require a program to be built before being queried, but it seems unreasonable to expect an implementation to know the correct response before then.

Other queries in the same table already account for this, the descriptions for `CL_PROGRAM_NUM_KERNELS` and `CL_PROGRAM_KERNEL_NAMES` both say:

> This information is only available after a successful program executable has been built for at least one device in the list of devices associated with program.

With the following error condition:

> `CL_INVALID_PROGRAM_EXECUTABLE` if param_name is `CL_PROGRAM_NUM_KERNELS` or `CL_PROGRAM_KERNEL_NAMES` and a successful program executable has not been built for at least one device in the list of devices associated with program.

Should `CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT` and `CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT` have been the same?

alycm:
It isn't possible to know what to return for
`CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT` or
`CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT` passed to `clGetProgramInfo` without
examining the program.   So require the program to be built first.

Fixes #436

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -6691,6 +6691,9 @@ include::{generated}/api/version-notes/CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT.asc
       | This indicates that the _program_ object contains non-trivial
         constructor(s) that will be executed by runtime before any kernel
         from the program is executed.
+        This information is only available after a successful program
+        executable has been built for at least one device in the list of
+        devices associated with _program_.
 
         Querying {CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT} may unconditionally
         return {CL_FALSE} if no devices associated with _program_ support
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM

bashbaug:
@kpet you might want to take a quick look at this change.

kpet (approved):
LGTM. Makes sense.

bashbaug:
Great.  Since this has two +1s now, no reason to wait to merge any longer.

PR #457, opened 2020-09-23
Add an introductory section on Unified OpenCL C spec

alycm:
Add this separately to get the section numbers in order.

Also renumber references to sections of OpenCL in the API specification,
mostly in the change descriptions.  Do not renumber references to older
versions of the specification when noting that a feature was deprecated.


bashbaug (approved):
This looks good to me.

I'll plan to merge first thing tomorrow morning if there are no additional comments.

PR #458, opened 2020-09-24
Correct version that CL_DEPTH became core

### LINKED ISSUE #284: Wrong version listed for the introduction of CL_DEPTH
In the current specification, `CL_DEPTH` is marked as “Missing before 1.2” however as far as I can tell, depth images made it into core only in 2.0 and not 1.2.
From the OpenCL API Specification, under “Summary of changes from OpenCL 1.2 to OpenCL 2.0”:

> Images support for 2D image from buffer, depth images and sRGB images.

Looking at the OpenCL Extensions Specification, cl_khr_depth_images is also listed as having been integrated into 2.0, not 1.2.

alycm:
Also correct other XML issues with cl_khr_depth_images and
cl_khr_gl_depth_images, but the other changes have no visible affect on the
specification.

Fixes #284

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -4092,13 +4092,6 @@ server's OpenCL/api-docs repository.
             <enum name="CL_MIGRATE_MEM_OBJECT_HOST"/>
             <enum name="CL_MIGRATE_MEM_OBJECT_CONTENT_UNDEFINED"/>
         </require>
-        <require comment="cl_channel_order">
-            <enum name="CL_DEPTH"/>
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Changes look good to me.

In case it helps other reviewers, the enums for `CL_DEPTH_STENCIL` and `CL_UNORM_INT24` already exist in a block for `cl_khr_gl_depth_images`, which makes sense since they are not part of a core API.

I think the headers have a similar bug.  I'll file an issue to correct this.

PR #453, opened 2020-09-23
Unify the OpenCL C spec

alycm:
* Note which version of OpenCL C is required for features added after 1.0.
* Re-adding descriptions of deprecated functionality.
  * A new 6.15.9 subsection describing `mem_fence`, `read_mem_fence`, `write_mem_fence`.
  * Restrictions that were crossed out have version annotations, the large of which is around byte addressable memory.
  * The `__ROUNDING_MODE__` macro.

Steps to take this PR out of draft:

  - [x] Rework "unified spec" introduction section. Moved to https://github.com/KhronosGroup/OpenCL-Docs/pull/457.
  - [x] Link existing use of "required" or "deprecated" to introduction where appropriate.
  - [x] Eliminate all TODO notes.
  - [x] Check the HTML and PDF rendering of all changes.
  - [x] Work through all the "2.0 or feature" and "feature alone" notes to add OpenCL 3.0 requirements as appropriate.

Fixes #315 #452

PR #464, opened 2020-09-28
Document nosvm attribute with other variables attributes

kpet:
... instead of documenting it in the section for function attributes.

Change-Id: I03626ac7f2f52e7bfdb8803bca2bca2a1a418517


bashbaug (approved):
LGTM - agree this is not a function attribute.

Note there is a likely interaction with #453 since it enhanced the `nosvm` description also.

alycm (approved):
LGTM.

PR #467, opened 2020-09-29
Address review feedback from Unifying OpenCL C specification

alycm:
The original feedback was

> The "Features" chapter intro still talks about OpenCL C 3.0 specifically. Not a big issue, but everywhere else uses "OpenCL C 3.0 or newer" consistently.
>
> A few functions have existing text saying "For OpenCL C 2.0 or with the `__opencl_c_generic_address_space feature`" and could be updated for consistency.
>
> There is a line about "and the OpenCL C 3.0 compiler must define the `__opencl_c_images` feature macro" that would be worth taking another look at, especially since the optional support was indicated by `__IMAGE_SUPPORT__` pre-OpenCL C 3.0.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -58,7 +58,7 @@ consistent with previous versions of the OpenCL and OpenCL C Programming
 Language specifications.
 --
 
-This section describes the OpenCL C 3.0 programming language.
+This section describes the OpenCL C programming language.
 The OpenCL C programming language may be used to write kernels that execute
 on an OpenCL device.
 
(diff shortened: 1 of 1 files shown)


alycm:
> There is a line about "and the OpenCL C 3.0 compiler must define the __opencl_c_images feature macro" that would be worth taking another look at, especially since the optional support was indicated by __IMAGE_SUPPORT__ pre-OpenCL C 3.0.

For this point in particular I've addressed this @bashbaug, but I don't actually know if my change addresses your original concern.

bashbaug (approved):
Thanks!  This addresses the remainder of my comments.

PR #468, opened 2020-10-02
memory_scope_all_devices is supported only for OpenCL C 3.0 or newer

azabazno:
This is probably a fix. 

Generally  memory_scope_all_devices requires support for OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_all_devices feature. But here the spec says that memory_scope_all_devices can be supported for OpenCL C 2.0 which is erroneous.


alycm (approved):
Thank you!  Yes, this fixes a spec bug.

I'll just merge this immediately as the change is small.

alycm:
Ah, actually, needs the CLA first.

azabazno:
Thanks! CLA is signed.

PR #476, opened 2020-10-13
fix the calculation in "mapping work-items onto an NDRange"

### LINKED ISSUE #475: Typo error in "3.2.1. Mapping work-items onto an NDRange" in OpenCL_API 3.0 spec
"(wx, wy) = ( (gx sx Fx) / Sx, (gy sy Fy) / Sy )" should be changed to "(wx, wy) = ( (gx - Fx) / Sx, (gy - Fy) / Sy )" in below section
3.2.1. Mapping work-items onto an NDRange
Given a global ID and the work-group size, the work-group ID for a work-item is computed as:
(wx, wy) = ( (gx sx Fx) / Sx, (gy sy Fy) / Sy )

bashbaug:
Fixes #475

It looks like several symbols were omitted from the calculations in "Mapping work-items onto an NDRange".  This PR fixes the calculations so they are consistent with the OpenCL 1.2 specs.

### DIFF
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -368,7 +368,7 @@ work-group (S~x~,S~y~) and the local ID (s~x~, s~y~) inside the work-group
 such that
 
 [none]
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM, I checked against the 1.2 spec.  Merging immediately as this has already been open a few weeks.

PR #477, opened 2020-10-15
Clarify usage of feature test macros for pre-OpenCL C 3.0

azabazno:
We discussed it with @bashbaug.

So basically this is just a clarification of this IMPORTANT note in order to allow compilers for earlier OpenCL C versions to define feature macros. For example when compiling for OpenCL C 2.0 compiler could define all the features and use macros internally in header. Furthermore, I don't see any stronger concerns if compiler for OpenCL C 2.0 will define (for example) `__opencl_c_pipes` macro because pipes are core OpenCL C 2.0 feature.  The way I see it: the spec strictly prohibits to do so now without allowing this implementation defined behavior.

We can adjust the wording and the place where to put it in the spec.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -119,7 +119,8 @@ by the presence of special predefined macros.
 === Features
 
 IMPORTANT: Feature test macros <<unified-spec, require>> support for OpenCL C
-3.0 or newer.
+3.0 or newer. This means that applications which target previous OpenCL C versions
+shouldn't rely on the presence of feature test macros.
 
 Optional language features are described in this document. They are optional
 from OpenCL C 3.0 onwards and therefore are not supported by all
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -119,8 +119,7 @@ by the presence of special predefined macros.
 === Features
 
 IMPORTANT: Feature test macros <<unified-spec, require>> support for OpenCL C
-3.0 or newer. This means that applications which target previous OpenCL C versions
-shouldn't rely on the presence of feature test macros.
+3.0 or newer.
 
 Optional language features are described in this document. They are optional
 from OpenCL C 3.0 onwards and therefore are not supported by all
@@ -223,6 +222,13 @@ defined if the feature is not supported by the OpenCL C compiler. A feature
 macro may expand to a different value in the future, but if this occurs the
 value of the feature macro must compare greater than the prior value of the
 feature macro.
+
+As specified in <<C99-spec,section 7.1.3 of the C99 Specification>> double
+underscore identifiers are reserved for any use and therefore compilers
+for earlier OpenCL C versions are allowed to define feature test macros
+but they are not required doing so. This means that applications which
+target earlier OpenCL C versions shouldn't rely on the presence of
+feature test macros unless compiler explicitly documents their availability.
 
 [[extensions]]
 === Extensions
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -224,11 +224,12 @@ value of the feature macro must compare greater than the prior value of the
 feature macro.
 
 As specified in <<C99-spec,section 7.1.3 of the C99 Specification>> double
-underscore identifiers are reserved for any use and therefore compilers
+underscore identifiers are reserved and therefore implementations
 for earlier OpenCL C versions are allowed to define feature test macros
 but they are not required doing so. This means that applications which
-target earlier OpenCL C versions shouldn't rely on the presence of
-feature test macros unless compiler explicitly documents their availability.
+target earlier OpenCL C versions should not rely on the presence of
+feature test macros because there is no guarantee that feature test macros
+will be defined and indicate support of functionallity.
 
 [[extensions]]
 === Extensions
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -226,10 +226,11 @@ feature macro.
 As specified in <<C99-spec,section 7.1.3 of the C99 Specification>> double
 underscore identifiers are reserved and therefore implementations
 for earlier OpenCL C versions are allowed to define feature test macros
-but they are not required doing so. This means that applications which
+but they are not required to do so. This means that applications which
 target earlier OpenCL C versions should not rely on the presence of
 feature test macros because there is no guarantee that feature test macros
-will be defined and indicate support of functionallity.
+will be defined and that if defined they will indicate the presence of the 
+corresponding optional functionality.
 
 [[extensions]]
 === Extensions


OpenCL_C.txt
```diff
@@ -119,7 +119,8 @@ by the presence of special predefined macros.
 === Features
 
 IMPORTANT: Feature test macros <<unified-spec, require>> support for OpenCL C
-3.0 or newer.
+3.0 or newer. This means that applications which target previous OpenCL C versions
```
AnastasiaStulova, line 122: Generally, double underscore prefixed identifiers are reserved in C, and OpenCL doesn't restrict this. This means there is no harm for implementations to define such macros in earlier OpenCL C versions i.e. there shouldn't be name clusing with the application code.

So I agree with your approach and I think it will make things easier in clang if we add them unconditionally. 

Regarding the phrasing in the spec perhaps we can refer to C99 section 7.1.3  "Reserved identifiers" paragraph 1

` All identifiers that begin with an underscore and either an uppercase letter or anotherunderscore are always reserved for any use.`

stating that implementations might provide (but don't have to!) definitions of these macros in earlier OpenCL C versions?

azabazno: > All identifiers that begin with an underscore and either an uppercase letter or anotherunderscore are always reserved for any use

Well, looking from this point of view we are definitely fine to define feature test macros whenever we want and especially for pre-OpenCL C 3.0. I'm just wondering if special adjustment needed for OpenCL C spec because of this:

> _IMPORTANT_: Feature test macros support _for OpenCL C 3.0 or newer_.

AnastasiaStulova: >    IMPORTANT: Feature test macros require support for OpenCL C 3.0 or newer.

Ok, this leaves the behavior of earlier versions unspecified.

The way I see OpenCL C spec is that it intends to document the difference to C99 used as a base. That implies whenever OpenCL doesn't specify certain behavior C99 rules apply.

If we consider the above then we probably don't need to add anything strictly speaking. However, I am very much in favor of self-contained specs because they reduce the pain of reply on separate documents and they also tend to be more consistent. So I find it helpful where possible to provide extra clarifications if they can be made concisely. The issues you have highlighted seems to be a good candidate. However, I think it would help to reference C99 to make it clear we are not reinventing the wheel and also to point out where more information is to be found.

One thing we should consider though that the macros with the above names could `theoretically` be already used by the implementations for different purposes from those documented in the spec. It makes the matter a bit more complicated but however this is highly unlikely.

azabazno: >"Feature test macros are only available in OpenCL C 3.0 or newer".

We discussed it with @bashbaug and its purpose was as exactly as I wrought down in this patch.

>The way I see OpenCL C spec in general is that it intends to document the difference to C99 used as a base. That implies whenever OpenCL doesn't specify certain behavior C99 rules apply.

Yes, but here spec specifies this behavior.

> However, I am very much in favor of self-contained specs because they reduce the pain of relying on separate documents and they also tend to be more consistent

Yeah, I would be glad to write some more clarifications, something like: _compilers for earlier OpenCL C versions are allowed to define feature test macros, but they are not required doing so_, if spec maintainers will find it valuable. I personally don't think it will hurt in any cases except that there are some other causes of putting this IMPORTANT note in the spec.

alycm:
> The way I see it: the spec strictly prohibits to do so now without allowing this implementation defined behavior.

I probably wrote the line in question, it certainly wasn't my intent to ban OpenCL C 1.x or 2.0 compilers from using such macros.

azabazno:
> > The way I see it: the spec strictly prohibits to do so now without allowing this implementation defined behavior.
> 
> I probably wrote the line in question, it certainly wasn't my intent to ban OpenCL C 1.x or 2.0 compilers from using such macros.

I see. So can we proceed with this? Should I add explicit wording about compilers for OpenCL C 1.x and 2.0, or we should refer to the fact that all double under-scored identifiers are reserved?

AnastasiaStulova:
> > > The way I see it: the spec strictly prohibits to do so now without allowing this implementation defined behavior.
> > 
> > 
> > I probably wrote the line in question, it certainly wasn't my intent to ban OpenCL C 1.x or 2.0 compilers from using such macros.
> 
> I see. So can we proceed with this? Should I add explicit wording about compilers for OpenCL C 1.x and 2.0, or we should refer to the fact that all double under-scored identifiers are reserved?

I think it makes sense to quote C99 regarding reserved underscored identifiers and highlight that therefore those macros can be available in earlier OpenCL C versions but I would also point out that they could also be used for different purposes in earlier OpenCL C versions i.e. in the unlikely case - some toolchains already introduced such macros for something else. I think we can't force updating earlier implementation if they were already using these identifiers for different purposes.

AnastasiaStulova:
Thanks for working on this! I have just added a few small comments, but otherwise, it looks great!

bcalidas:
Qualcomm has the following question - 

" This means that applications which
target earlier OpenCL C versions shouldn't rely on the presence of
feature test macros unless the compiler explicitly documents their availability."

What would be the mechanism for the compiler to document the macro availability.  Would that be implementation specific documentation.

AnastasiaStulova:
> Qualcomm has the following question -
> 
> " This means that applications which
> target earlier OpenCL C versions shouldn't rely on the presence of
> feature test macros unless the compiler explicitly documents their availability."
> 
> What would be the mechanism for the compiler to document the macro availability. Would that be implementation specific documentation.

Yes, but I don't think vendors have been doing this so I am suggesting changing the wording slightly so it doesn't strictly say that it has to be documented even if it should be encouraged as a good practice.

azabazno:
Thanks for feedback! I've recently updated the patch. I've removed that call for docs correction so the change seems implementation-independent for me now.

OpenCL_C.txt
```diff
@@ -223,6 +223,14 @@ macro may expand to a different value in the future, but if this occurs the
 value of the feature macro must compare greater than the prior value of the
 feature macro.
 
+As specified in <<C99-spec,section 7.1.3 of the C99 Specification>> double
+underscore identifiers are reserved and therefore implementations
+for earlier OpenCL C versions are allowed to define feature test macros
+but they are not required doing so. This means that applications which
```
bashbaug, line 229:
```suggestion
but they are not required to do so. This means that applications which
```

bashbaug (commented):
One minor grammar fix otherwise I think this is ready to go?

OpenCL_C.txt
```diff
@@ -223,6 +223,14 @@ macro may expand to a different value in the future, but if this occurs the
...
+target earlier OpenCL C versions should not rely on the presence of
+feature test macros because there is no guarantee that feature test macros
+will be defined and indicate support of functionallity.
```
AnastasiaStulova, line 232:
```suggestion
will be defined and that if defined they will indicate the presence of the corresponding optional functionality.
```

AnastasiaStulova:
Apart from two small corrections from @bashbaug and me, it looks good!

azabazno:
Anastasia, Ben, thanks for corrections. Updated the patch.

AnastasiaStulova (approved):
LGTM! Thanks

alycm (approved):
LGTM, this is the third review so I'll merge.

PR #481, opened 2020-11-05
Tighten vec_step builtin specification

svenvh:
* Specify that the vec_step argument is not evaluated, so any side
   effects of the argument expression will not have an effect.  This
   makes vec_step follow the evaluation behavior of the sizeof
   operator.  Clang currently already assumes that the vec_step
   argument is not evaluated.

 * Replace "pure type" by "type name", as the spec does not define
   what a "pure type" is and it intends to refer to the type name.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -7648,14 +7648,15 @@ integer data types.
   int *vec_step*(_type_)
     | The *vec_step* built-in function takes a built-in scalar or vector
       data type argument and returns an integer value representing the
-      number of elements in the scalar or vector.
+      number of elements in the scalar or vector.  The argument is not
+      evaluated.
 
       For all scalar types, *vec_step* returns 1.
 
       The *vec_step* built-in functions that take a 3-component vector
       return 4.
 
-      *vec_step* may also take a pure type as an argument, e.g.
+      *vec_step* may also take a type name as an argument, e.g.
       *vec_step*(float2)
 
       <<unified-spec, Requires>> support for OpenCL C 1.1 or newer.


bashbaug (approved):
This is reasonable to me.

Do we need a CTS test to ensure the argument is not evaluated?

StuartDBrady:
A related question: can a vec_step() value be used as a compile-time constant, e.g. as the length of an array (as per `sizeof()` and `alignof()`)? Clang seems to permit using it this way.

kpet:
Having a CTS test seems like a good idea. @svenvh Any chance you could contribute a simple test into the `vectors` suite?

@StuartDBrady Makes sense to me.

alycm (approved):
LGTM.  Clearly all Clang based implementations already behave this way.

However I agree with the CTS test requirement, this would be an easy point to miss if implementing an OpenCL C parser.

alycm:
As a side-note, although Clang does not evaluate the argument to vec_step, it also does not warn about it.  Clang does produce such a warning for `sizeof`.

```quote
clang-12: warning: argument unused during compilation: '--gcc-toolchain=/opt/compiler-explorer/gcc-snapshot' [-Wunused-command-line-argument]
<source>:4:19: warning: expression with side effects has no effect in an unevaluated context [-Wunevaluated-expression]
    int x = sizeof(in[gid++]);
```

svenvh:
> As a side-note, although Clang does not evaluate the argument to vec_step, it also does not warn about it. Clang does produce such a warning for `sizeof`.

Yes I'm planning to fix that, but I first wanted to have it written down explicitly in the spec.

> @svenvh Any chance you could contribute a simple test into the vectors suite?

I agree adding a CTS test would be good.  I'm not that familiar with the CTS code base, so if anyone wants to pick it up before I find time to dig into it, feel free to do so.

AnastasiaStulova:
It would be good to merge this change asap.

> 
>     @svenvh Any chance you could contribute a simple test into the vectors suite?
> 
> I agree adding a CTS test would be good. I'm not that familiar with the CTS code base, so if anyone wants to pick it up before I find time to dig into it, feel free to do so.

I suggest to create an issue in CTS repo and go ahead with the merge of this PR.

PR #486, opened 2020-11-09
Publish cl_img_cached_allocations extension specification.

Jeremy-Kemp:


Oblomov:
Out of curiosity, what would be the difference between using this extension and using the standard `CL_MEM_ALLOC_HOST_PTR` flag?

bashbaug (approved):
LGTM - this matches what is in the XML file and in the .txt extension spec in the registry.

Note, to update the spec on the OpenCL registry you will want to create a PR with the HTML output in the OpenCL-Registry repo (and probably update the .txt spec to signpost to the HTML spec, also).

Jeremy-Kemp:
@Oblomov this extension was intended as a performance hint for certain scenarios; typically for devices that have older drivers. Using a modern driver may not yield a measurable difference in performance.

@bashbaug Thanks. I'm going through and refreshing our old extensions and also upstreaming as-yet unpublished extensions so was planning on doing the registry changes in one go once they're all done. (Its certainly my intention to update the registry as well though).

bashbaug:
> planning on doing the registry changes in one go once they're all done.

Sounds good, makes sense.

I'll plan to merge all of these then.  @oblomov, would you mind moving any questions to a separate issue?  I don't want to lose track of them, but I don't want to hold up this PR either.  Thanks!

PR #487, opened 2020-11-10
Publish cl_img_use_gralloc_ptr extension specification.

Jeremy-Kemp:
Updated the XML to change CL_INVALID_GRALLOC_OBJECT_IMG from being unused to used.

### DIFF
--- a/extensions/cl_img_use_gralloc_ptr.asciidoc
+++ b/extensions/cl_img_use_gralloc_ptr.asciidoc
@@ -1,0 +1,272 @@
+
+:data-uri:
+:icons: font
+include::../config/attribs.txt[]
+:source-highlighter: coderay
+
+= cl_img_use_gralloc_ptr
+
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
LGTM!

bashbaug:
Note, this isn't anything catastrophic and it's probably too late to change it, but in the future it would be best to use negative numbers for error codes:

```xml
        <enum value="0x40D4"        name="CL_GRALLOC_RESOURCE_NOT_ACQUIRED_IMG"/>
        <enum value="0x40D5"        name="CL_INVALID_GRALLOC_OBJECT_IMG"/>
```

Jeremy-Kemp:
I also spotted this while converting to asciidoc.

Unfortunately I think it *is* too late as the extension has been out in the wild for quite some time but I agree it would have been more suitable for it to be negative.

Issue #496, opened 2020-11-11
Argument Names for clEnqueueReadBufferRect and clEnqueueWriteBufferRect


bashbaug:
Found while making changes for #478.

The names of the arguments for `clEnqueueReadBufferRect` and `clEnqueueWriteBufferRect` are inconsistent between the OpenCL headers and the XML file (used to generate the function signatures in the spec) and the narrative spec text.

In the headers and the XML file the function arguments are named `host_offset` and `buffer_offset`, but in the narrative spec text the function arguments are `host_origin` and `buffer_origin`.  I think the narrative spec text is correct.  Note that `origin` is consistent with `clEnqueueCopyBufferRect`, the various image APIs, and the signatures of these functions in the OpenCL 1.2 spec.

If so, the fix is to update the headers and the XML file.

bashbaug:
Since the XML changes are merged and the headers PR has been created I am closing this issue.

PR #497, opened 2020-11-11
clarify that a NULL origin or region is a CL_INVALID_VALUE

### LINKED ISSUE #478: Unclear requirements for NULL region/offset/origin arguments in various functions
There are a couple of functions which have `region`, `offset` and `origin` array arguments, but never specify if those can be NULL or not. For `region` it would make sense to return `CL_INVALID_VALUE` if it is `NULL`, but for the others the runtime could assume `0` for all elements.

List of function affected:
* clEnqueueReadBufferRect
* clEnqueueWriteBufferRect
* clEnqueueCopyBufferRect
* clEnqueueReadImage
* clEnqueueWriteImage
* clEnqueueFillImage
* clEnqueueCopyImage
* clEnqueueCopyImageToBuffer
* clEnqueueCopyBufferToImage
* clEnqueueMapImage

bashbaug:
Fixes #478.

This is also a partial fix for #496, but we will need a similar fix in the OpenCL headers.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -988,10 +988,10 @@ Otherwise, they return one of the following errors:
     _buffer_ are not the same or if the context associated with
     _command_queue_ and events in _event_wait_list_ are not the same.
   * {CL_INVALID_MEM_OBJECT} if _buffer_ is not a valid buffer object.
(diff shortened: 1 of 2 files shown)


kpet (approved):
LGTM (and a nice tidy up of related bits and pieces).

alycm (approved):
LGTM!  This is a good set of clarifications.   Technically the additional `CL_INVALID_VALUE` cases are changes to behaviour, but what is written is the only way to implement this while being consistent with the rest of the specification.

PR #498, opened 2020-11-19
Add deploy build artifacts on releases to Travis

AnastasiaStulova:
Setup upload of build artifacts to Trevis. In the current change only pdfs will be uploaded to tags along with archives of sources. See example in: https://github.com/AnastasiaStulova/OpenCL-Docs/releases/tag/testci6.

@bashbaug do you think we should upload htmls too or anything else?

@Khronoswebmaster I would need help with setting up api_key and also activating Travis for OpenCL-Docs repo.

Btw once the Travis is activated the build will run for all the commits to the master. I assume this is desirable behavior though, otherwise, we can look into disabling this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,12 @@ before_install:
 script:
   - git describe --tags --dirty
   - make api c env ext cxx4opencl
+
+deploy:
+  provider: releases
+  api_key: "FIXME - ADD GITHUB OAUTH TOKEN"
+  file_glob: true
+  file: out/pdf/*
+  skip_cleanup: true
+  on:
+    tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,10 @@ script:
 
 deploy:
   provider: releases
-  api_key: "FIXME - ADD GITHUB OAUTH TOKEN"
+  api_key: $GH_TOKEN
   file_glob: true
   file: out/pdf/*
   skip_cleanup: true
   on:
     tags: true
+  edge: true


AnastasiaStulova:
> @Khronoswebmaster I would need help with setting up api_key and also activating Travis for OpenCL-Docs repo.
> 
> Btw once the Travis is activated the build will run for all the commits to the master. I assume this is desirable behavior though, otherwise, we can look into disabling this.
> 

Please ignore my comment regarding the Travis setup as it is already working.

outofcontrol:
> > @Khronoswebmaster I would need help with setting up api_key and also activating Travis for OpenCL-Docs repo.
> > Btw once the Travis is activated the build will run for all the commits to the master. I assume this is desirable behavior though, otherwise, we can look into disabling this.
> 
> Please ignore my comment regarding the Travis setup as it is already working.

All Khronos GitHub repositories now come with Travis-CI enabled at travis-ci.com.

.travis.yml
```diff
@@ -19,3 +19,12 @@ before_install:
 script:
   - git describe --tags --dirty
   - make api c env ext cxx4opencl
+
+deploy:
+  provider: releases
+  api_key: "FIXME - ADD GITHUB OAUTH TOKEN"
```
bashbaug, line 25: I think this needs to be setup before we can merge these changes.

Syntactically I have an example using this `api_key` here (I used `github_token` - I forget why):

https://github.com/bashbaug/OpenCL-Docs/blob/travis-ci/.travis.yml#L42

I think there was some setup needed on the GitHub or Travis side to make this work also though - is that what you were referring to in your PR description?

bashbaug (changes requested):
This is great!  Thanks for getting this going.

* I can't think of a reason not to deploy the HTML output also.  I've found that I refer to the HTML output more frequently than the PDF output, personally, and we're building the HTML output anyhow.

* Building on each commit is the desired behavior I think, and we've been doing this a while now.

.travis.yml
AnastasiaStulova: Yes! I would prefer to avoid using my personal token,  as the token apply to all repositories the user has write access to, and since I have such permission to quite many open source projects I would prefer to avoid using my personal token.

In this tutorial, they recommend creating a machine user: https://docs.travis-ci.com/user/deployment/releases/. I could create one but I was thinking to check whether we already have one for the repositories hosted at Khronos.

AnastasiaStulova (commented):
> > > @Khronoswebmaster I would need help with setting up api_key and also activating Travis for OpenCL-Docs repo.
> > > Btw once the Travis is activated the build will run for all the commits to the master. I assume this is desirable behavior though, otherwise, we can look into disabling this.
> > 
> > 
> > Please ignore my comment regarding the Travis setup as it is already working.
> 
> All Khronos GitHub repositories now come with Travis-CI enabled at travis-ci.com.

Yes indeed. Sorry I got confused by looking at the fork of the repo. Btw do we have a machine user already created for the repositories hosted at Khronos, that we can use to set up uploading of artifacts to releases? I would prefer to avoid using my personal account for this. See inline comment on .travis.yml.

AnastasiaStulova:
> This is great! Thanks for getting this going.
> 
>     * I can't think of a reason not to deploy the HTML output also.  I've found that I refer to the HTML output more frequently than the PDF output, personally, and we're building the HTML output anyhow.

Ok, I have tried the HTML uploads and it looks like this: https://github.com/AnastasiaStulova/OpenCL-Docs/releases/tag/testci9

It would have been cleaner if we could have subfolders but I think it is not supported by Github. Here is the error message I got when testing with the folder uploads:

https://travis-ci.com/github/AnastasiaStulova/OpenCL-Docs/builds/203566503

`out/html/ is not a regular file. Skipping.`

However, we don't have that many files so perhaps it's ok without the folders? If you agree I will update the PR with the upload of HTMLs.

bashbaug:
> Ok, I have tried the HTML uploads and it looks like this: https://github.com/AnastasiaStulova/OpenCL-Docs/releases/tag/testci9

Thanks, this is helpful.

I still think it's worth releasing the HTML output since it's basically "free", but it's not quite as useful as it could be since you cannot browse directly to the HTML output and instead need to download it first.  If we'd rather release just the PDF versions instead to reduce clutter I'd be fine with that too.

AnastasiaStulova:
> > Ok, I have tried the HTML uploads and it looks like this: https://github.com/AnastasiaStulova/OpenCL-Docs/releases/tag/testci9
> 
> Thanks, this is helpful.
> 
> I still think it's worth releasing the HTML output since it's basically "free", but it's not quite as useful as it could be since you cannot browse directly to the HTML output and instead need to download it first. If we'd rather release just the PDF versions instead to reduce clutter I'd be fine with that too.

I don't have a strong preference. I think having less attachments is better if they don't add significant value. One advantage of HTMLs I can see that they are easier to view i.e. they only need browsers but at the same time PDF viewers have become pretty standard too. So perhaps it's not such a big value after all?

.travis.yml
AnastasiaStulova: Thanks to @Khronoswebmaster for setting up the machine user. 

@bashbaug do you add the token as a secret in your repository setting? I think I prefer your approach, for my own testing I have added the encrypted token in Travis file directly: https://github.com/AnastasiaStulova/OpenCL-Docs/commit/0a78370e3ee3c10eb6e61d00f1c494378eb9e2b2. But it doesn't look that neat.

AnastasiaStulova: Repo secrets seems to be only available for the Github Actions, so I have added this via Travis settings.

AnastasiaStulova:
@bashbaug this change is now updated with a token and ready to be merged. I haven't included the HTMLs yet, but if you prefer to have them in I can do this asap.

bashbaug (approved):
Merging as discussed in the December 1st teleconference.

PR #499, opened 2020-11-19
Changed version of C++ for OpenCL doc

AnastasiaStulova:
C++ for OpenCL doc doesn't take a version from the repo tag.
Its versioning will be provided manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,16 +69,19 @@ SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
 
-ATTRIBOPTS   = -a revnumber="$(SPECREVISION)" \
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM.

Out of curiosity, how are you planning to set the version for the C++ for OpenCL doc?  Will you embed the version in the asciidoc source itself, pass it as a different command-line option, or something else entirely?

It's possible there is a simpler solution depending on the versioning mechanism.

AnastasiaStulova:
> LGTM.
> 
> Out of curiosity, how are you planning to set the version for the C++ for OpenCL doc? Will you embed the version in the asciidoc source itself, pass it as a different command-line option, or something else entirely?
> 

For the time being, it will be in the asciidoc source indeed. 

> It's possible there is a simpler solution depending on the versioning mechanism.

PR #502, opened 2020-11-23
[C++ for OpenCL] Allow NULL to be implementation defined macro.

AnastasiaStulova:
This is to align C++ for OpenCL with C++17.

### DIFF
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -95,6 +95,9 @@ void foo(){
 }
 ----------
 
+To improve code portability and compatibility, implementations are
+encouraged to define `NULL` as an alias to pointer literal `nullptr`.
+
 ===== Use of restrict
 
 {cpp}17 does not support `restrict` and therefore {cpp} for OpenCL
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -139,9 +139,8 @@ Variadic macros can be used normally as per {cpp}17 `[cpp.replace]`.
 
 ===== Predefined macros
 
-The predefined macros `+__OPENCL_C_VERSION__+` and `NULL` (see also
-<<null_literal, _Null literal_>>), described in `OpenCL C v2.0 s6.10`,
-are not supported.
+The macro `+__OPENCL_C_VERSION__+` described in `OpenCL C v2.0 s6.10`,
+is not defined.
 
 The following new predefined macros are added in {cpp} for OpenCL:
 
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -67,29 +67,29 @@ void foo(){
 ----------
 
 [[null_literal]]
-===== Null literal
+===== Null pointer constant
 
-In C and OpenCL C the null constant is defined using other
+In C and OpenCL C the null pointer constant is defined using other
 language features as it is not represented explicitly i.e.
 commonly it is defined as
 
 [source,c]
 ----------
-#define NULL (void*)0
+#define NULL ((void*)0)
 ----------
 
-In {cpp} there is an explicit builtin literal `nullptr` that should be used
-instead ({cpp}17 `[lex.nullptr]`).
+In {cpp}17 there is an explicit builtin pointer literal `nullptr` that should
+be used instead ({cpp}17 `[lex.nullptr]`).
 
-Null pointer definition in {cpp} for OpenCL follows {cpp}
-`[support.types.nullptr]` where `NULL` is an implementation defined macro and it
+`NULL` macro definition in {cpp} for OpenCL follows {cpp}17
+`[support.types.nullptr]` where it is an implementation defined macro and it
 is not guaranteed to be the same as in OpenCL C. Reusing the definition of
 `NULL` from OpenCL C does not guarantee that any code with NULL is legal in
 {cpp} for OpenCL even if it is legal in OpenCL C.
 
 [source,c]
 ----------
-#define NULL (void*)0
+#define NULL ((void*)0)
 void foo(){
  int *ptr = NULL; // invalid initialization of int* with void*
 }
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -69,7 +69,7 @@ void foo(){
 [[null_literal]]
 ===== Null literal
 
-In C and OpenCL C the null literal is defined using other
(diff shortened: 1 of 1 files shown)


cxx4opencl/diff2openclc.txt
```diff
@@ -78,19 +78,18 @@ commonly it is defined as
...
-instead. However as a workaround to avoid large modifications
-`NULL` can also be defined/aliased to `nullptr` in custom
-headers or using command line flag `-D`. It is not recommended
-to reuse the C definition of `NULL` in {cpp} for OpenCL as it may
-cause compilation failures in cases that work for C.
+Null pointer definition in {cpp} for OpenCL follows {cpp}
+`[support.types.nullptr]` where `NULL` is an implementation defined macro and it
+is not guaranteed to be the same as in OpenCL C. Reusing the definition of
+`NULL` from OpenCL C does not guarantee that any code with NULL is legal in
+{cpp} for OpenCL even if it is legal in OpenCL C.
```
StuartDBrady, line 88: If implementations may define NULL as 0 in C++ for OpenCL, this carries the same downsides as the earlier definition of NULL as 0 in C++.  (Note: defining it as `(void *)0` would carry the same C++-specific downsides that prevented C++ from defining it in that way and necessitated adding of the `nullptr_t` type.)  I believe part of the motivation of allowing NULL to be defined as nullptr since C++11 is to allow future versions of C++ to define NULL as nullptr for all implementations, whilst retaining compatibility with C.  Note that C++11 allowed implementations to continue to define NULL as 0 for compatibility with older implementations but C++ for OpenCL has no old implementations to maintain compatibility with.

StuartDBrady (changes requested):

cxx4opencl/diff2openclc.txt
AnastasiaStulova: > Note that C++11 allowed implementations to continue to define NULL as 0 for compatibility with older implementations but C++ for OpenCL has no old implementations to maintain compatibility with.

I think we can't confidently say so. For example, someone could port the legacy C++ code to OpenCL. Something with metaprogramming that works on regular targets might be ok for OpenCL too - type traits is one example but there is, of course, C++17 compliant implementations on it. Of course it is less practical for OpenCL than it was for C++.

FYI there is a similar discussion here: https://bugs.llvm.org/show_bug.cgi?id=48098

cxx4opencl/diff2openclc.txt
```diff
@@ -69,7 +69,7 @@ void foo(){
 [[null_literal]]
 ===== Null literal
 
-In C and OpenCL C the null literal is defined using other
+In C and OpenCL C the null constant is defined using other
```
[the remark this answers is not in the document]
AnastasiaStulova: Agree!

cxx4opencl/diff2openclc.txt
```diff
@@ -78,19 +78,18 @@ commonly it is defined as
 #define NULL (void*)0
```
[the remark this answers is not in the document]
AnastasiaStulova: Fix in the latest commit!

mantognini:
I cannot comment inline for some reason so here it is:
```
===== Predefined macros

The predefined macros `+__OPENCL_C_VERSION__+` and `NULL` (see also
<<null_literal, _Null literal_>>), described in `OpenCL C v2.0 s6.10`,
are not supported.
```

This sentence should be updated given that `NULL` is now supported (for some definition of supported). I suppose `NULL` can simply be removed here?

StuartDBrady:
> I cannot comment inline for some reason so here it is:
> 
> ```
> ===== Predefined macros
> 
> The predefined macros `+__OPENCL_C_VERSION__+` and `NULL` (see also
> <<null_literal, _Null literal_>>), described in `OpenCL C v2.0 s6.10`,
> are not supported.
> ```
> 
> This sentence should be updated given that `NULL` is now supported (for some definition of supported). I suppose `NULL` can simply be removed here?

Regardless of _how_ `NULL` is defined, _if_ it is defined then you're right, this would also need updating.

It would  seem GitHub's UI does not allow comments to be added inline for lines that are not part of the change and not part of the immediate surrounding context.

AnastasiaStulova:
> I cannot comment inline for some reason so here it is:
> 
> ```
> ===== Predefined macros
> 
> The predefined macros `+__OPENCL_C_VERSION__+` and `NULL` (see also
> <<null_literal, _Null literal_>>), described in `OpenCL C v2.0 s6.10`,
> are not supported.
> ```
> 
> This sentence should be updated given that `NULL` is now supported (for some definition of supported). I suppose `NULL` can simply be removed here?

Oh, of course - updated! Thanks!

cxx4opencl/diff2openclc.txt
StuartDBrady: > I think we can't confidently say so. For example, someone could port the legacy C++ code to OpenCL. Something with metaprogramming that works on regular targets might be ok for OpenCL too - type traits is one example but there is, of course, C++17 compliant implementations on it. Of course it is less practical for OpenCL than it was for C++.

Although absolute compatibility with C++ would be ideal, it seems to me that type safety is also a concern, and we should not let any one concern become an overriding concern in all situations.

If we only had to consider code ported from C++, I would be more inclined to suggest either leaving `NULL` undefined, or mandating that it only be defined as `nullptr`, as it is not entirely unreasonable to expect some effort from developers to ensure their C++ code still works when compiled as C++ for OpenCL.

Since we also need to consider compatibility with OpenCL C code, and not all implementations of OpenCL C necessarily define `NULL` as `((void *)0)`, there would be benefit in allowing the definition of `NULL` as 0 in C++ for OpenCL, in cases where the OpenCL C implementation is known to do the same and this has been relied upon. The [original proposal for nullptr in C++](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2431.pdf) (N2431) states:

> We do not propose to define the standard library macro NULL to nullptr. We considered that and liked the idea, but the EWG opinion was that it would break too much code, even though in many cases that would be code that deserved to be broken.

Unfortunately, a similar requirement applies here in terms of compatibility with existing OpenCL C implementations. I think the proposed wording that "implementations are _encouraged_ to define `NULL` as an alias to pointer literal `nullptr`" (_emphasis_ mine) is sufficient. There is a [C++23 proposal prohibiting the use of NULL under `epoch 2023`](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2048r0.html) (P2048R0), but it seems too soon to deprecate this in C++ for OpenCL from the very start.

StuartDBrady (approved):

PR #503, opened 2020-11-25
Reserved a block of error codes for upcoming IMG extensions.

Jeremy-Kemp:


alycm (approved):
LGTM, will merge immediately.

xml/cl.xml
```diff
@@ -558,8 +558,12 @@ server's OpenCL/api-docs repository.
         <enum value="-1121"         name="CL_CONTEXT_TERMINATED_KHR"/>
     </enums>
 
-    <enums start="-1122" end="-9999" name="ErrorCodes.future" vendor="Khronos" comment="RESERVED FOR FUTURE ALLOCATIONS BY KHRONOS">
-            <unused start="-1122" end="-9999"/>
+    <enums start="-1122" end="-1137" name="ErrorCodes.1122" vendor="IMG" comment="Reserved for Imagination extensions">
```
alycm, line 561: Actually, error codes don't seem to be allocated in blocks of 16.  No problem with an allocation of 16, there are 1000's of code left, but you want 16?

Jeremy-Kemp: Its more than we need with what is currently planned, but if there aren't any objections then can we keep it as-is?

alycm: Alright, no problem.

PR #505, opened 2020-11-30
Added missing XML for clEnqueueGenerateMipmapIMG.

Jeremy-Kemp:
Spotted in KhronosGroup/OpenCL-Headers#135


bashbaug (approved):
This looks good but there's one more missing part that should be added to the XML file for this extension for the new type, probably at the top of the XML block:

```xml
<require>
    <type name="cl_mipmap_filter_mode_img"/>
</require>
```

I'm only finding these issues with my extension header generator so don't feel bad about missing them.  Since I think I'm going to need a few other XML file changes before the extension header generator is ready to go, I'll just merge this PR as-is and make any other minor XML corrections like this one separately.

PR #506, opened 2020-11-30
Update cl.xml for new vendor IDs in vk.xml

franz:
This depends on PR to [add](https://github.com/KhronosGroup/Vulkan-Headers/pull/169) PoCL vendor ID in Vulkan headers.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1089,7 +1089,9 @@ server's OpenCL/api-docs repository.
         </comment>
         <unused start="0x10000" end="0x10003" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
         <enum value="0x10004" name="CL_KHRONOS_VENDOR_ID_CODEPLAY" comment="Codeplay Software Ltd."/>
-        <unused start="0x10005" end="0x1FFFF" comment="See vk.xml for the next available Khronos vendor ID"/>
+        <unused start="0x10005" end="0x10005" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
+        <enum value="0x10006" name="CL_KHRONOS_VENDOR_ID_POCL" comment="PoCL (http://portablecl.org/)"/>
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -1087,9 +1087,8 @@ server's OpenCL/api-docs repository.
             document under the section "Registering a Vendor ID with Khronos".
             Only once the ID has been reserved should a new enum entry be added here.
         </comment>
-        <unused start="0x10000" end="0x10003" comment="These Khronos vendor IDs are already assigned in vk.xml"/>
-        <enum value="0x10004" name="CL_KHRONOS_VENDOR_ID_CODEPLAY" comment="Codeplay Software Ltd."/>
```
EwanC, line 1091: We're using this vendor ID, looks like it's getting removed. Should PoCL be getting added rather than removing Codeplay?

EwanC (changes requested):

bashbaug (changes requested):
I think you'll want a Vulkan-Docs PR first, to reserve the vendor ID enum.  See, for example, https://github.com/KhronosGroup/Vulkan-Docs/pull/1168, which reserved the Codeplay enum.

Note that this should be a Vulkan-Docs PR and not a Vulkan-Headers PR.

We should merge this PR after the Vulkan-Docs PR is merged.  Thanks!

xml/cl.xml
EwanC: Looks good now, thanks

alycm (commented):
LGTM, needs the VK XML PR to merge first to reserve the value: https://github.com/KhronosGroup/Vulkan-Docs/pull/1411

Both PRs just need the CLA to be agreed.

EwanC (approved):
Accompanying https://github.com/KhronosGroup/Vulkan-Docs/pull/1411 is merged now

bashbaug (approved):
LGTM

Issue #504, opened 2020-11-26
Clarify use of bool*, size_t*, ptrdiff_t*, (u)intptr_t* as kernel arguments


svenvh:
Paragraph 6.11.m of the spec says:

> Arguments to kernel functions in a program cannot be declared with the built-in scalar types bool, size_t, ptrdiff_t, intptr_t, and uintptr_t or a struct and/or union that contain fields declared to be one of these built-in scalar types. 

That is clear and unambiguous.  The following is not allowed:
```
kernel void foo(size_t s) {}
```
And the following is not allowed either:
```
struct bar {
  size_t s;
};
kernel void foobar(struct bar x) {}
```

The same paragraph then continues with the following:

> The size in bytes of these types are implementation-defined and in addition can also be different for the OpenCL device and the host processor making it difficult to allocate buffer objects to be passed as arguments to a kernel declared as pointer to these types.

The sentence only says that it is "difficult to allocate ...", but does not explicitly forbid anything.  What message is this sentence meant to convey?

Is it purely advisory and is the following allowed?
```
kernel void foo(global size_t *ptr_s) {}
```

Or is the intention of this sentence to forbid the use of pointers to implementation-defined integer types as kernel arguments?

bashbaug:
IMO this restriction is very odd, except for perhaps `bool`: while it's true that the size of the built-in scalar types `size_t`, etc. are_implementation-defined, the size is discoverable via the `CL_DEVICE_ADDRESS_ BITS` query, so it's certainly possible (albeit not always trivial) to use `size_t`, etc. as a kernel argument, or a struct with a `size_t` field as a kernel argument.  All that being said, the spec is clear that this is not allowed, and Clang does appear to be enforcing this restriction.

Note that Clang allows kernel arguments that are pointers to these types though, so there's at least one interpretation that the "difficult to allocate..." phrase is purely advisory.

https://godbolt.org/z/T5sed6

Would it make sense to remove this entire sentence?  I don't think the advisory note is adding much of value.  I also think we should consider relaxing this restriction in the future.

svenvh:
> Would it make sense to remove this entire sentence? I don't think the advisory note is adding much of value. I also think we should consider relaxing this restriction in the future.

That sounds reasonable to me.  I have opened #508 for this.

bashbaug:
This issue is going to close automatically when I merge #508.  I think that's good, but if there is still something missing please feel free to re-open it - thanks!

PR #508, opened 2020-12-02
Remove phrase about kernel argument pointer types

svenvh:
This phrase seems to be purely advisory and is not an explicit
restriction.  The Clang implementation has always been accepting
pointer types for the affected types.

Fixes https://github.com/KhronosGroup/OpenCL-Docs/issues/504

Signed-off-by: Sven van Haastregt <sven.vanhaastregt@arm.com>


bashbaug (approved):
This makes sense and looks good to me.  I'd like to wait for working group signoff (or at least another reviewer) before merging, though.

alycm (approved):
LGTM, I agree that this sentence doesn't add anything concrete to the previous one.

PR #509, opened 2020-12-02
[C++ for OpenCL] Lift restriction on reference members w local objects

AnastasiaStulova:
`__local` addr space variables can't be initialized and therefore it was thought that such objects shouldn't have members that are references. This seems unnecessary because user defined constructor/assignments can be provided to work around reference initialization.

See example: https://godbolt.org/z/5Ee6Ps

**Note** that if user default special members are not provided the diagnostic will be given about their absence when objects in local address space are initialized in assignment operations. Therefore, invalid behavior will be rejected anyway.

### DIFF
--- a/cxx4opencl/address_spaces.txt
+++ b/cxx4opencl/address_spaces.txt
@@ -343,10 +343,8 @@ their declaration. Developers are free to define a special member function
 that can initialize local address space objects after their declaration. Any
 default values provided for the initialization of members in a class
 declaration are ignored when creating the local address space objects.
-Classes that are instantiated with local address space objects can not
-have references as their members. The same restriction applies recursively
-to the members and base classes. Destructors of local address space objects
-are not invoked automatically. They can be called manually if required.
+Destructors of local address space objects are not invoked automatically.
+They can be called manually if required.
 
 User defined constructors are not allowed to construct objects in `__constant`
 address space. Such objects can be initialized using literals and
--- a/cxx4opencl/address_spaces.txt
+++ b/cxx4opencl/address_spaces.txt
@@ -344,7 +344,7 @@ that can initialize local address space objects after their declaration. Any
 default values provided for the initialization of members in a class
 declaration are ignored when creating the local address space objects.
 Destructors of local address space objects are not invoked automatically.
-They can be called manually if required.
+They can be called manually.
 
 User defined constructors are not allowed to construct objects in `__constant`
 address space. Such objects can be initialized using literals and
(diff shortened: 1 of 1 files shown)


mantognini (commented):
It seems fair to remove this restriction. A genuine question though: given that the workaround to use reference is quite involve (with placement new and explicit calls to destructors), I wonder if the spec should clarify (with a high level description) how reference member can be initialised? Or do we consider this to be out of the scope of the spec?

AnastasiaStulova:
> It seems fair to remove this restriction. A genuine question though: given that the workaround to use reference is quite involved (with placement new and explicit calls to destructors), I wonder if the spec should clarify (with a high level description) how reference member can be initialised? Or do we consider this to be out of the scope of the spec?

Even though this is a C++ specific issue I think it is important enough to be mentioned. I am not sure though how many details we should provide since there isn't anything OpenCL specific. Right now I have updated the document to highlight this. I find it difficult to explain how to workaround the issue concisely. If you think it's necessary perhaps it's better to add an example instead? Otherwise, many good explanations can be found on stackoverflow, but they all use examples.

mantognini:
> I find it difficult to explain how to workaround the issue concisely. If you think it's necessary perhaps it's better to add an example instead? Otherwise, many good explanations can be found on stackoverflow, but they all use examples.

It's definitely not trivial to explain. Perhaps the current wording is good and we should let the users rely on existing material instead of bloating this section with potentially more confusing explanation. At least, your update makes it clear some non-conventional solution is needed.

I suppose, if there is a demand for it, we can revisit this later without needing to bump the version.

cxx4opencl/address_spaces.txt
```diff
@@ -342,11 +342,11 @@ created in the local address space have undefined state at the point of
 their declaration. Developers are free to define a special member function
 that can initialize local address space objects after their declaration. Any
 default values provided for the initialization of members in a class
-declaration are ignored when creating the local address space objects.
-Classes that are instantiated with local address space objects can not
-have references as their members. The same restriction applies recursively
-to the members and base classes. Destructors of local address space objects
-are not invoked automatically. They can be called manually if required.
+declaration are ignored when creating the local address space objects. Since
+the initialization is performed after the variable declaration special handling
```
svenvh, line 346:
```suggestion
the initialization is performed after the variable declaration, special handling
```

cxx4opencl/address_spaces.txt
```diff
@@ -342,11 +342,11 @@ created in the local address space have undefined state at the point of
...
+is required for classes with data members that are references as their values
```
svenvh, line 347:
```suggestion
is required for classes with data members that are references because their values
```

svenvh (approved):

PR #511, opened 2020-12-02
[C++ for OpenCL] Added references to execution and memory models

AnastasiaStulova:
Improved introduction by highlighting the fundamental difference in the execution and memory model between OpenCL and C/C++.

### DIFF
--- a/cxx4opencl/intro.txt
+++ b/cxx4opencl/intro.txt
@@ -7,7 +7,7 @@
 
 This language is built on top of OpenCL C v2.0 and {cpp}17 enabling
 most of regular {cpp} features in OpenCL kernel code. Most functionality
-from OpenCL C and {cpp} is inherited. Since both OpenCL C and {cpp} are
+from {cpp} and OpenCL C is inherited. Since both OpenCL C and {cpp} are
 derived from C and moreover {cpp} is backward compatible with C, the main
 design principle of {cpp} for OpenCL is to reapply existing OpenCL concepts
 to {cpp}. Therefore, it is important to refer to OpenCL v2.0 section 3.2
--- a/cxx4opencl/intro.txt
+++ b/cxx4opencl/intro.txt
@@ -8,10 +8,10 @@
 This language is built on top of OpenCL C v2.0 and {cpp}17 enabling
 most of regular {cpp} features in OpenCL kernel code. Most functionality
 from {cpp} and OpenCL C is inherited. Since both OpenCL C and {cpp} are
-derived from C and moreover {cpp} is backward compatible with C, the main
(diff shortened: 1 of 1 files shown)


cxx4opencl/intro.txt
```diff
@@ -7,7 +7,12 @@
 
 This language is built on top of OpenCL C v2.0 and {cpp}17 enabling
 most of regular {cpp} features in OpenCL kernel code. Most functionality
-from OpenCL C and {cpp} is inherited.
+from OpenCL C and {cpp} is inherited. Since both OpenCL C and {cpp} are
```
svenvh, line 10: "OpenCL C and C++" could be interpreted as "OpenCL C and OpenCL C++", even though you do not intend this.

I'd suggest to either swap the order everywhere ("C++ and OpenCL") or explicitly mention C++17 to avoid the ambiguity.

cxx4opencl/intro.txt
```diff
@@ -7,7 +7,12 @@
...
+derived from C and moreover {cpp} is backward compatible to C, the main
```
svenvh, line 11:
```suggestion
derived from C and moreover {cpp} is backward compatible with C, the main
```

svenvh:
```suggestion
derived from C and moreover {cpp} is almost fully backward compatible to C, the main
```

cxx4opencl/intro.txt
```diff
@@ -7,7 +7,12 @@
...
+design principle of {cpp} for OpenCL is to reapply existing OpenCL concepts
+to {cpp}. Therefore, it is important to refer to OpenCL v2.0 section 3.2
+and section 3.3 detailing fundamental differences of OpenCL execution and
+memory models from the convention C and {cpp} view.
```
svenvh, line 15:
```suggestion
memory models from the conventional C and {cpp} view.
```

cxx4opencl/intro.txt
AnastasiaStulova: Good thinking. I have swapped them.

cxx4opencl/intro.txt
svenvh: You did not integrate the second suggestion.
(perhaps I should have merged them into one)

AnastasiaStulova: Ah strange! Not sure what went wrong...

cxx4opencl/intro.txt
AnastasiaStulova: I have applied it manually now. I think I have forgotten about this somehow.

svenvh (approved):

PR #515, opened 2020-12-07
[OpenCL C] lift restriction on l-value vec literal

AnastasiaStulova:
OpenCL C has disallowed vector literals to be l-values. However,
similar functionality in C99 is allowed for the compound literals.

This therefore prevented implementing the restriction in upstream
tooling, therefore it is decided to change to an implementation
defined behavior instead of restricting the functionality.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -717,7 +717,7 @@ Vector literals can be used to create vectors from a list of scalars,
 vectors or a mixture thereof.
 A vector literal can be used either as a vector initializer or as a primary
 expression.
-A vector literal cannot be used as an l-value.
+Whether a vector literal can be used as an l-value is implementation-defined.
 
 A vector literal is written as a parenthesized vector type followed by a
 parenthesized comma delimited list of parameters.


kpet (approved):
LGTM

AnastasiaStulova:
I would like to be more helpful to the developers and allow using l-value vector literals with newer OpenCL C versions but I am not sure if this is doable for OpenCL C 3.0? If not I am happy to go ahead with the current wording in this PR for now and create an issue to improve this in the next versions of the language...

bashbaug (approved):
LGTM.

keryell (approved):

alycm (approved):
LGTM.

alycm:
> I would like to be more helpful to the developers and allow using l-value vector literals with newer OpenCL C versions but I am not sure if this is doable for OpenCL C 3.0? If not I am happy to go ahead with the current wording in this PR for now and create an issue to improve this in the next versions of the language...

I think that recording an issue for changing this in a future version is the best that we can do currently.

PR #521, opened 2020-12-11
Proposed formal extension definition for Layers

Kerilk:
This is the description of the layers extension to the loader.
Edit: I added definitions for APIs and enums to the xml file, I think those can help the discussion.

### DIFF
--- a/ext/cl_khr_layers.asciidoc
+++ b/ext/cl_khr_layers.asciidoc
@@ -103,7 +103,7 @@ cl_int clGetLayerInfo(cl_layer_info  param_name,
 
   * _param_name_ is an enumeration constant that identifies the platform
 information being queried. It can be one of the following values as
-specified in the <<layer-queires-table, Layer Queries>> table.
+specified in the <<layer-queries-table, Layer Queries>> table.
   * _param_value_size_ specifies the size in bytes of memory pointed to by
     _param_value_.
     This size in bytes must be ≥ to the size of return type specified in the
@@ -179,4 +179,3 @@ https://github.com/KhronosGroup/OpenCL-Headers
 The header file *CL/cl_icd.h* defines the OpenCL dispatch table.
 The header file *CL/cl_icd_layer.h* defines the necessary types
 and API entry points.
-
--- a/ext/cl_khr_layers.asciidoc
+++ b/ext/cl_khr_layers.asciidoc
@@ -59,7 +59,7 @@ Accepted as _param_name_ to the function *clGetLayerInfo*:
 
 [source,c]
 ----
-#define CL_LAYER_API_VERSION        0xDEADBEEF
+#define CL_LAYER_API_VERSION        0x4240
 ----
 
 
--- a/ext/cl_khr_layers.asciidoc
+++ b/ext/cl_khr_layers.asciidoc
@@ -106,7 +106,7 @@ information being queried. It can be one of the following values as
 specified in the <<layer-queires-table, Layer Queries>> table.
   * _param_value_size_ specifies the size in bytes of memory pointed to by
     _param_value_.
-    This size in bytes must be {geq} size of return type specified in the
+    This size in bytes must be ≥ to the size of return type specified in the
     <<layer-queries-table, Layer Queries>> table.
   * _param_value_ is a pointer to memory location where appropriate values for a
     given _param_name_, as specified in the <<layer-queries-table, Layer
@@ -116,7 +116,7 @@ specified in the <<layer-queires-table, Layer Queries>> table.
     queried by _param_name_.
     If _param_value_size_ret_ is `NULL`, it is ignored.
 
-The information that can be queried using {clGetLayerInfo} is specified
+The information that can be queried using *clGetLayerInfo* is specified
 in the <<layer-queries-table, Layer Queries>> table.
 
 [[layer-queries-table]]
@@ -124,14 +124,14 @@ in the <<layer-queries-table, Layer Queries>> table.
 [width="100%",cols="<34%,<33%,<33%",options="header"]
 |====
 | Layer Info | Return Type | Description
-| {CL_LAYER_API_VERSION} | cl_layer_api_version | OpenCL Layer API version
+| *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
 |====
 
-{clGetLayerInfo} returns {CL_SUCCESS} if the function is executed
+*clGetLayerInfo* returns *CL_SUCCESS* if the function is executed
 successfully.
 Otherwise, it returns one of the following errors.
 
-  * {CL_INVALID_VALUE} if _param_name_ is not one of the supported values or
+  * *CL_INVALID_VALUE* if _param_name_ is not one of the supported values or
     if size in bytes specified by _param_value_size_ is < size of return
     type as specified in the <<layer-queries-table, Layer Queries>> table,
     and _param_value_ is not a `NULL` value.
@@ -159,11 +159,11 @@ cl_int clInitLayer(cl_uint                         num_entries,
     table can contain `NULL` value entries, and the loader is responsible for
     providing the missing entries using the entries from the target dispatch.
 
-{clInitLayer} returns {CL_SUCCESS} if the function is executed
+*clInitLayer* returns *CL_SUCCESS* if the function is executed
 successfully.
 Otherwise, it returns one of the following errors.
 
-  * {CL_INVALID_VALUE} if _num_entries_ is sufficient for the layer or if
+  * *CL_INVALID_VALUE* if _num_entries_ is sufficient for the layer or if
     _target_dispatch_ is a `NULL` value, or _num_entries_ret_ is a `NULL`
     value, or _layer_dispatch_ret_ is a `NULL` value.
 
--- a/ext/cl_khr_layers.asciidoc
+++ b/ext/cl_khr_layers.asciidoc
@@ -1,0 +1,183 @@
+// Copyright 2017-2020 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+[[cl_khr_layers-opencl]]
+== Installable OpenCL Loader Layers
+
+[[cl_khr_layers-overview]]
+=== Overview
+
+This section describes *cl_khr_layers* loader extension which defines
+a simple mechanism through which the Khronos OpenCL installable client
+driver loader (ICD Loader) may expose load multiple separate installable
+interception layers (Layer ICDs) for OpenCL.
+
+=== General information
+
+==== Version history
+
+[cols="1,1,3",options="header",]
+|====
+| *Date*     | *Version* | *Description*
+| 2020-11-04 | 1.0.0     | First assigned version.
+|====
+
+[[cl_khr_layers-new-procedures-and-functions]]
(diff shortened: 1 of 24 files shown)


xml/cl.xml
```diff
@@ -3462,6 +3469,20 @@ server's OpenCL/api-docs repository.
...
+            <proto><type>cl_int</type>                                  <name>clGetLayerInfo</name></proto>
+            <param><type>size_t</type>                                  <name>param_value_size</name></param>
+            <param><type>cl_layer_info</type>                           <name>param_name</name></param>
+            <param><type>void</type>*                                   <name>param_value</name></param>
+            <param><type>size_t</type>*                                 <name>param_value_size_ret</name></param>
+        </command>
+        <command>
+            <proto><type>cl_int</type>                                  <name>clInitLayer</name></proto>
+            <param><type>cl_uint</type>                                 <name>num_entries</name></param>
+            <param>const<type>struct _cl_icd_dispatch</type>*           <name>target_dispatch</name></param>
```
bashbaug, line 3482: It looks like other APIs include a space after `const` by convention, so I'd recommend adding one here also.

This will only matter when we start to generate extension headers, but hopefully that will be soon.

bashbaug: Also, this is purely cosmetic, but should we be using the typedef'd name `cl_icd_dispatch` instead of `struct _cl_icd_dispatch`?

ext/cl_khr_layers.asciidoc
```diff
@@ -0,0 +1,183 @@
...
+    value, or _layer_dispatch_ret_ is a `NULL` value.
+
+[[cl_khr_layer-source-code]]
+=== Source Code
+
+The official source for the ICD loader is available on github, at:
+
+https://github.com/KhronosGroup/OpenCL-ICD-Loader
+The header file *include/cl_icd_layer.h* defines the necessary types
+and API entry points.
```
bashbaug, line 177: I think we'll want to eventually include the layer APIs and defines in the official API headers.  Do you think this makes sense, or should we be sign-posting to the ICD loader headers instead?

My first reaction is to include them in the extension headers.  I think that'd make sense, since this is a "loader extension".  I'm open to other alternatives though.

bashbaug (approved):
LGTM.  I think we should merge this as-is while we fix any remaining minor issues.

The only item we may want to consider before merging is: currently this extension is not included into the main extension spec so it won't be in the published HTML or PDF specs.  At some point should it be included into the published specs also?  I can help with that if needed, just want to be sure it's not being left out on purpose.

xml/cl.xml
Kerilk: > It looks like other APIs include a space after `const` by convention, so I'd recommend adding one here also.
> 
> This will only matter when we start to generate extension headers, but hopefully that will be soon.

Fixed.

Kerilk: > Also, this is purely cosmetic, but should we be using the typedef'd name `cl_icd_dispatch` instead of `struct _cl_icd_dispatch`?

This I think touches the real issues: `cl_icd.h` is not included by `opencl.h` or extensions headers. `cl_icd_layer.h` pulls `cl_icd.h` to get the definition. I think this prevents including the extension definitions into normal extension headers. I think the header should go along with `cl_icd.h` in the `CL` directory.

The question that is raised now is do we want those files generated alongside the others, especially `cl_icd.h`, and if yes, how much information do we need to add to the xml to be able to generate `cl_icd.h`. Or do we only include the type defined in 'cl_icd.h` in the xml, wich would allow us to use the typedef'd `cl_icd_dispatch`.

What are your views on the problem?

ext/cl_khr_layers.asciidoc
Kerilk: > I think we'll want to eventually include the layer APIs and defines in the official API headers. Do you think this makes sense, or should we be sign-posting to the ICD loader headers instead?
> 
> My first reaction is to include them in the extension headers. I think that'd make sense, since this is a "loader extension". I'm open to other alternatives though.

see above

Kerilk:
> The only item we may want to consider before merging is: currently this extension is not included into the main extension spec so it won't be in the published HTML or PDF specs. At some point should it be included into the published specs also? I can help with that if needed, just want to be sure it's not being left out on purpose.

That would be great, it wasn't left out on purpose, I naively assumed it would be picked up "somehow". I'll try adding it to the doc building.

Kerilk:
> > The only item we may want to consider before merging is: currently this extension is not included into the main extension spec so it won't be in the published HTML or PDF specs. At some point should it be included into the published specs also? I can help with that if needed, just want to be sure it's not being left out on purpose.
> 
> That would be great, it wasn't left out on purpose, I naively assumed it would be picked up "somehow". I'll try adding it to the doc building.

On second thoughts maybe doing that in a different pull-request is wiser, there are lots of decisions to make regarding where to include the extension in the documentation.

xml/cl.xml
bashbaug: `cl_icd.h` is already in the API headers ([link](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_icd.h)), so we could include it from either the extensions header `cl_ext.h` or a new layer-specific header.

I think it would be preferable to put the layer stuff in the same API headers vs. adding a dependency on another repo.  I don't have a strong opinion whether the layer stuff goes into the same extension header or in a layer-specific header, though.

Kerilk: To make things clear, I agree with you we should put it in the same directory, and I think, to avoid including 'cl_icd.h' everywhere it should go in its own file.

Kerilk: > `cl_icd.h` is already in the API headers ([link](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_icd.h)), so we could include it from either the extensions header `cl_ext.h` or a new layer-specific header.

Indeed, but it is not in the xml file. How should I introduce it? How much of it should be introduced?

alycm:
Repeating my comment from call 2021-01-26:  The khr part of the name `cl_khr_layers` makes this sound like a ratified extension, but it isn't (that in itself is fine).

An obvious option is `cl_ext_layers`, which avoids the `khr` part, though currently is just used for multi-vendor non-Khronos extensions.

As an ICD loader extension is fundamentally different to implementers (i.e. they don't need to implement it), `cl_icd_layers` may also work.  I'm not sure how much this distinction matters to OpenCL users though.

ext/cl_khr_layers.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+discard the layer.
+
+[source,c]
+----
+cl_int clGetLayerInfo(cl_layer_info  param_name,
+                      size_t         param_value_size,
+                      void          *param_value,
+                      size_t        *param_value_size_ret);
+
+cl_int clInitLayer(cl_uint                 num_entries,
```
alycm, line 40: Should these definitions include `CL_API_ENTRY` and `CL_API_CALL`?  I'm not entirely sure if it is required, the pending header PR sets them there.

These entry points, however, differ from all others because end users will write implementations of them rather than just implementers.

ext/cl_khr_layers.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+                   const cl_icd_dispatch **layer_dispatch);
+----
+
+[[cl_khr_layers-new-api-types]]
+=== New API Types
+
+[source]
+----
+typedef cl_uint cl_layer_info;
+typedef cl_uint cl_layer_api_version;
```
alycm, line 52: Not sure technical changes are appropriate now, but should this typedef be removed and instances of `cl_layer_api_version` be replaced with `cl_version`?

It wasn't added until OpenCL 3.0, but as layers only work with the latest ICD loader and headers I'm not sure that matters.

Then values would be produced via `CL_MAKE_VERSION_KHR(1, 0, 0)`, which would produce a different number than 100.

I don't feel strongly about this, especially given the code already merged into the ICD loader.  It is completely fine to resolve this comment and leave things as is if you think that's best.

ext/cl_khr_layers.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+----
+cl_int clGetLayerInfo(cl_layer_info  param_name,
+                      size_t         param_value_size,
+                      void          *param_value,
+                      size_t        *param_value_size_ret);
+----
+
+  * _param_name_ is an enumeration constant that identifies the platform
+information being queried. It can be one of the following values as
+specified in the <<layer-queires-table, Layer Queries>> table.
```
alycm, line 106:
```suggestion
specified in the <<layer-queries-table, Layer Queries>> table.
```

ext/cl_khr_layers.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+    layer will be returning.
+  * _layer_dispatch_ret_ return the address of the layer dispatch table. The
+    table can contain `NULL` value entries, and the loader is responsible for
+    providing the missing entries using the entries from the target dispatch.
+
+*clInitLayer* returns *CL_SUCCESS* if the function is executed
+successfully.
+Otherwise, it returns one of the following errors.
+
+  * *CL_INVALID_VALUE* if _num_entries_ is sufficient for the layer or if
```
alycm, line 166:
```suggestion
  * *CL_INVALID_VALUE* if _num_entries_ is insufficient for the layer or if
```

I assume?

alycm (approved):
LGTM!  A couple of questions.

Kerilk:
Thanks for the great review.

> Repeating my comment from call 2021-01-26: The khr part of the name `cl_khr_layers` makes this sound like a ratified extension, but it isn't (that in itself is fine).
> 
> An obvious option is `cl_ext_layers`, which avoids the `khr` part, though currently is just used for multi-vendor non-Khronos extensions.
> 
> As an ICD loader extension is fundamentally different to implementers (i.e. they don't need to implement it), `cl_icd_layers` may also work. I'm not sure how much this distinction matters to OpenCL users though.

Personally I do prefer  `cl_icd_layers` rather than an `ext` based name.

ext/cl_khr_layers.asciidoc
Kerilk: I don't have a strong opinion on the subject. This is confined to the loader, and I don't expect the API to incrementally change much, so the simple API mechanism should suffice, but if the majority want to change it I wont fight it.

ext/cl_khr_layers.asciidoc
Kerilk: Thanks, nice catch.

ext/cl_khr_layers.asciidoc
Kerilk: You are correct.

ext/cl_khr_layers.asciidoc
Kerilk: I thought that the documentation part had to omit them. I could be wrong though.

Kerilk:
In order for the pull request to be able to be merged today, I renamed cl_khr_layers to cl_icd_layers. We can still change it if we want something else.

ext/cl_khr_layers.asciidoc
Kerilk: I checked other extension file defining API entry points and the `CL_API_ENTRY` `CL_API_CALL` were omitted, even if they defined entry points that had to be implemented (`clIcdGetPlatformIDsKHR`). So I would leave them out for now. We can always change it should we deem it better.

bashbaug:
Merging as discussed in the February 2 teleconference.

PR #522, opened 2020-12-14
clarify __OPENCL_VERSION__

### LINKED ISSUE #471: Specification of __OPENCL_VERSION__ macro
The unified OpenCL C specification states that the predefined `__OPENCL_VERSION__` macro:

> Substitutes an integer reflecting the version number of the OpenCL supported by the OpenCL device. The version of OpenCL described in this document will have `__OPENCL_VERSION__` substitute the integer 300.

These two statements are contradictory.

For an OpenCL 2.1 implementation, the integer suggested by the former statement would be 210, and not 300. For future OpenCL versions, one would expect an `__OPENCL_VERSION__` greater than 300, even if the OpenCL C language specification is not updated.

Note that there are OpenCL 2.1 implementations that have passed conformance that defien `__OPENCL_VERSION__` as 200, and not 210.

Note also issue #249, which raises the issue of the lack of `CL_VERSION_2_1` and `CL_VERSION_2_2` macros in the OpenCL C specification. It is reasonable that there are not provided by OpenCL 2.0 implementations, but it is unclear if these should be defined for more recent versions of OpenCL.

bashbaug:
Fixes #471.

Updated text:

> `__OPENCL_VERSION__`
>
> For OpenCL devices with OpenCL version less than or equal to OpenCL 2.0, substitutes an integer reflecting the OpenCL version supported by the device. This predefined macro is deprecated by OpenCL 2.1. For OpenCL devices with OpenCL version greater than OpenCL 2.0, it must be defined but may substitute any implementation-defined integer value greater than OpenCL 2.0.

I also added as a footnote that this macro must be defined but may substitute any implementation-defined integer value when compiling offline.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -3178,10 +3178,13 @@ The following predefined macro names are available.
     source line (an integer constant).
 
 `+__OPENCL_VERSION__+` ::
-    Substitutes an integer reflecting the version number of the OpenCL
-    supported by the OpenCL device.
-    The version of OpenCL described in this document will have
-    `+__OPENCL_VERSION__+` substitute the integer 300.
+    For OpenCL devices with OpenCL version less than or equal to OpenCL 2.0,
+    substitutes an integer value reflecting the OpenCL version supported by the
+    device.
+    This predefined macro is <<unified-spec, deprecated by>> OpenCL 2.1.
+    For OpenCL devices with OpenCL version greater than OpenCL 2.0, it must be
+    defined but may substitute any implementation-defined integer value greater
+    than OpenCL 2.0. footnote:[{fn-OPENCL_VERSION}]
 
 `CL_VERSION_1_0` ::
     Substitutes the integer 100 reflecting the OpenCL 1.0 version.
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -3184,7 +3184,7 @@ The following predefined macro names are available.
     This predefined macro is <<unified-spec, deprecated by>> OpenCL 2.1.
     For OpenCL devices with OpenCL version greater than OpenCL 2.0, it must be
     defined but may substitute any implementation-defined integer value greater
-    than OpenCL 2.0. footnote:[{fn-OPENCL_VERSION}]
+    than 200, reflecting OpenCL 2.0. footnote:[{fn-OPENCL_VERSION}]
 
 `CL_VERSION_1_0` ::
     Substitutes the integer 100 reflecting the OpenCL 1.0 version.
--- a/c/footnotes.asciidoc
+++ b/c/footnotes.asciidoc
@@ -193,7 +193,7 @@ I.e. the _global_work_size_ values specified to *clEnqueueNDRangeKernel* are not
 ]
 
 :fn-OPENCL_VERSION: pass:n[ \
-When OpenCL C is compiled offline, `+__OPENCL_VERSION__+` should be defined and may substitute any implementation-defined integer value. \
+When OpenCL C is compiled offline, `+__OPENCL_VERSION__+` may be defined and may substitute any implementation-defined integer value. \
 ]
 
 :fn-pointer-invalid-value-indirection: pass:n[ \
--- a/c/footnotes.asciidoc
+++ b/c/footnotes.asciidoc
@@ -192,6 +192,10 @@ Here `TYPE_MIN` and `TYPE_MIN_EXP` should be substituted by constants appropriat
 I.e. the _global_work_size_ values specified to *clEnqueueNDRangeKernel* are not evenly divisible by the _local_work_size_ values for each dimension. \
 ]
 
+:fn-OPENCL_VERSION: pass:n[ \
+When OpenCL C is compiled offline, `+__OPENCL_VERSION__+` must be defined but may substitute any implementation-defined integer value. \
+]
+
 :fn-pointer-invalid-value-indirection: pass:n[ \
 Among the invalid values for dereferencing a pointer by the unary *+*+* operator are a null pointer, an address inappropriately aligned for the type of object pointed to, and the address of an object after the end of its lifetime. \
 If *+*P+* is an l-value and *T* is the name of an object pointer type, *+*(T)P+* is an l-value that has a type compatible with that to which *T* points. \
--- a/c/footnotes.asciidoc
+++ b/c/footnotes.asciidoc
@@ -193,7 +193,7 @@ I.e. the _global_work_size_ values specified to *clEnqueueNDRangeKernel* are not
 ]
 
 :fn-OPENCL_VERSION: pass:n[ \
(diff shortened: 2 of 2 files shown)


OpenCL_C.txt
```diff
@@ -3178,10 +3178,13 @@ The following predefined macro names are available.
...
-    supported by the OpenCL device.
-    The version of OpenCL described in this document will have
-    `+__OPENCL_VERSION__+` substitute the integer 300.
+    For OpenCL devices with OpenCL version less than or equal to OpenCL 2.0,
+    substitutes an integer value reflecting the OpenCL version supported by the
+    device.
+    This predefined macro is <<unified-spec, deprecated by>> OpenCL 2.1.
+    For OpenCL devices with OpenCL version greater than OpenCL 2.0, it must be
+    defined but may substitute any implementation-defined integer value greater
+    than OpenCL 2.0. footnote:[{fn-OPENCL_VERSION}]
```
kpet, line 3187: We should say how the version is encoded, i.e `Mmp` with `M` the major version, `m` the minor version and `p` the patch version.

Being pedantic "OpenCL 2.0" is not a value :).

kpet (changes requested):

AnastasiaStulova:
> Fixes #471.
> 
> Updated text:
> 
> > `__OPENCL_VERSION__`
> > For OpenCL devices with OpenCL version less than or equal to OpenCL 2.0, substitutes an integer reflecting the OpenCL version supported by the device. This predefined macro is deprecated by OpenCL 2.1. For OpenCL devices with OpenCL version greater than OpenCL 2.0, it must be defined but may substitute any implementation-defined integer value greater than OpenCL 2.0.
> 

 Looking good! Thanks!

> I also added as a footnote that this macro must be defined but may substitute any implementation-defined integer value when compiling offline.

I am trying to think why requiring to define the macro is useful? Would it be better to just make the macro implementation-defined for offline compilations? This will make clang conformant, especially releases up to clang 12 that we can't change.

kpet:
>I am trying to think why requiring to define the macro is useful?

One argument in favour that was discussed in the teleconference was that there may be code using this to alter the behaviour for OpenCL.

```
#ifdef __OPENCL_VERSION__
// OpenCL path
#else
// Alternative
#endif
```

alycm:
> > I am trying to think why requiring to define the macro is useful?
> 
> One argument in favour that was discussed in the teleconference was that there may be code using this to alter the behaviour for OpenCL.
> 
> ```
> #ifdef __OPENCL_VERSION__
> // OpenCL path
> #else
> // Alternative
> #endif
> ```

If you search GitHub you definitely find this pattern, and I've used it myself if the past.  Logically `__OPENCL_C_VERSION__` may make more sense, but for this pattern the two macros work as well as each other.  Keeping it defined but implementation defined value means that vendor-agnostic offline compilers (e.g. SPIR-V) can both support this pattern while also not having to worry about what the number should be.

alycm (approved):
LGTM.

AnastasiaStulova:
Didn't we discuss that we should deprecate this macro from OpenCL 3.0 or will it happen afterward? I am just worried that more code will be written using it. `__OPENCL_C_VERSION__` should be good enough to check the compilation for OpenCL.

AnastasiaStulova:
> > > I am trying to think why requiring to define the macro is useful?
> > 
> > 
> > One argument in favour that was discussed in the teleconference was that there may be code using this to alter the behaviour for OpenCL.
> > ```
> > #ifdef __OPENCL_VERSION__
> > // OpenCL path
> > #else
> > // Alternative
> > #endif
> > ```
> 
> If you search GitHub you definitely find this pattern, and I've used it myself if the past. 

I can only see the occurrences in OpenCL-Docs.

> Logically `__OPENCL_C_VERSION__` may make more sense, but for this pattern the two macros work as well as each other. Keeping it defined but implementation defined value means that vendor-agnostic offline compilers (e.g. SPIR-V) can both support this pattern while also not having to worry about what the number should be.

Ok, we can't modify that for any releases of the upstream tools, which are quite many. That's why by making it implementation-defined we will be friendlier to upstream tools and also discourage the use of this macro in the future, which where we are heading to as far as I understand?

alycm:
> Didn't we discuss that we should deprecate this macro from OpenCL 3.0 or will it happen afterward? I am just worried that more code will be written using it. `__OPENCL_C_VERSION__` should be good enough to check the compilation for OpenCL.

This PR has language that says

> This predefined macro is <<unified-spec, deprecated by>> OpenCL 2.1.

AnastasiaStulova:
> > Didn't we discuss that we should deprecate this macro from OpenCL 3.0 or will it happen afterward? I am just worried that more code will be written using it. `__OPENCL_C_VERSION__` should be good enough to check the compilation for OpenCL.
> 
> This PR has language that says
> 
> > This predefined macro is <<unified-spec, deprecated by>> OpenCL 2.1.

I guess this implies the same for offline compilation? It just typically not bind to any OpenCL version.

OpenCL_C.txt
bashbaug: Yes, I thought about defining the encoding also while making this PR, but I didn't want to change too much, since the encoding is also not precisely defined for `__OPENCL_C_VERSION__`.

For this PR, how about changing "any implementation-defined integer value grater than OpenCL 2.0" to "any implementation-defined integer value greater than `CL_VERSION_2_0`"?  That's still not as precise as I would like it to be but I think it's better?

I'll make this change.

bashbaug: I changed this to: "any implementation-defined integer value greater than 200, reflecting OpenCL 2.0".  Let me know what you think - thanks!

kpet (approved):
Works for me. Let's not spend too many cycles on that.

AnastasiaStulova (changes requested):
For offline compilation, the only route to avoid invalidating existing implementation is to change to fully implementation-defined behavior.

bashbaug:
> For offline compilation, the only route to avoid invalidating existing implementation is to change to fully implementation-defined behavior.

I've changed the offline compiler footnote to: "When OpenCL C is compiled offline, `__OPENCL_VERSION__` should be defined and may substitute any implementation-defined integer value."

Note **should** instead of **must**.

This is probably the best we can do anyhow, since we do not conformance test offline compiler toolchains.

AnastasiaStulova:
> > For offline compilation, the only route to avoid invalidating existing implementation is to change to fully implementation-defined behavior.
> 
> I've changed the offline compiler footnote to: "When OpenCL C is compiled offline, `__OPENCL_VERSION__` should be defined and may substitute any implementation-defined integer value."
> 
> Note **should** instead of **must**.
> 
> This is probably the best we can do anyhow, since we do not conformance test offline compiler toolchains.

Yes, this is better. Thanks! I just want to prevent bugs submitted to open source tool repositories that we can never close because we can't undo the past.

AnastasiaStulova (approved):

bashbaug:

Issue #500, opened 2020-11-23
Clarify the meaning of (optional) core extension


AnastasiaStulova:
OpenCL C 3.0 spec has section 6.2 where features and extensions are described however OpenCL has and will probably still have core and optional core features. It would be good to add a clear definition regarding those and differences between those.

AnastasiaStulova:
After discussion on OpenCL WG teleconference today, it feels to me that the majority of extensions that migrated into features have exactly the functionality of features described in OpenCL C 3.0 spec section 6.2. FYI if an extension had a pragma its use has no effect (neither when it's disabled or enabled) in the language versions where this extension has become a feature (optional or not). @bashbaug could I double-check this with you please?

FYI the references to features in OpenCL C 2.0 are as follows

> Double precision floating-point is an optional feature.

And in the extension spec it has:

> The functionality previously described by cl_khr_byte_addressable_store is now part of the
core feature set. 

So we used to have:
- **optional core features** that are mostly equivalent to OpenCL C 3.0 feature definition. Except for pragma.
- **core features** that are different to other OpenCL C 3.0 features because they are non-optional i.e. every device needs to support those starting from the version it becomes core.

In OpenCL 3.0 extension spec we still have sections in **Appendix A** where we list core features but however, we never describe what they are or at least explain that they are not to be confused with feature definition in the section 6.2, so there is still some work to be done for clarifying the features.

bashbaug:
Sorry for the slow reply, I'm still catching up after the US thanksgiving holidays.

> could I double-check this with you please?

Yes, I think this sounds right.  The way I think about it is:

* An extension is an optional feature that is described in an extension spec.  It is not described in the core spec because it "extends" the core spec.  When an extension extends OpenCL C it will have an extension define and it may have an extension pragma.

* Core features are described in the core spec.  A core feature may be required for all conformant implementations (e.g. integer arithmetic is required for all OpenCL devices), or it may be optional for some implementations (e.g. double-precision arithmetic).  When a core feature is optional in OpenCL C it will have a feature test macro, but it will never have a pragma - at least so far, all optional core features are unconditionally enabled.

* If an feature began as an extension before becoming an optional core feature (e.g. double-precision) then it may have a feature test macro, an extension define, and an extension pragma, but the extension pragma is irrelevant since an optional core feature is always enabled and cannot be disabled.

[...]

PR #523, opened 2020-12-14
clarify relationship between optional core features and extensions

bashbaug:
See #500.

This PR attempts to clarify the relationship between optional core language features and language extensions.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -121,8 +121,8 @@ by the presence of special predefined macros.
 IMPORTANT: Feature test macros <<unified-spec, require>> support for OpenCL C
 3.0 or newer.
 
-Optional language features are described in this document. They are optional
-from OpenCL C 3.0 onwards and therefore are not supported by all
+Optional core language features are described in this document. They are
+optional from OpenCL C 3.0 onwards and therefore are not supported by all
 implementations. When an OpenCL C 3.0 optional feature is supported, an
 associated __feature test macro__ will be predefined.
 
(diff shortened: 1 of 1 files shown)


AnastasiaStulova:
This looks good. Thanks. I think we should also clarify:
- Difference to **non-optional** core feature (especially those that used to be extensions).
- Pragma use for non-optional core/optional core features. This might need to be done in the extension spec though?

alycm (approved):
LGTM.

bashbaug:
> * Difference to **non-optional** core feature (especially those that used to be extensions).

I think this is covered by the "Requires support for OpenCL C major.minor or newer..." text in the [Unified Specification](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#unified-spec) section?

> * Pragma use for non-optional core/optional core features. This might need to be done in the extension spec though?

Yeah, exactly, my current thinking is to document usage of extension pragmas in the extensions spec, especially since it's a bit in flux (see #355).  This PR already refers to the extension spec for details.  Sound good?

AnastasiaStulova:
> > * Difference to **non-optional** core feature (especially those that used to be extensions).
> 
> I think this is covered by the "Requires support for OpenCL C major.minor or newer..." text in the [Unified Specification](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#unified-spec) section?
> 

I was thinking whether we should explain if feature/extension macros are available or not for non-optional features? Honestly, I actually don't know the answer myself. Although providing behavior now might affect conformant implementations? We could also create a follow-up issue for this.

> > * Pragma use for non-optional core/optional core features. This might need to be done in the extension spec though?
> 
> Yeah, exactly, my current thinking is to document usage of extension pragmas in the extensions spec, especially since it's a bit in flux (see #355). This PR already refers to the extension spec for details. Sound good?

Ok, we can document this in a separate PR, could we make a note somewhere? Maybe #82 is a good place?

bashbaug:

PR #525, opened 2020-12-14
clarify subgroup barrier behavior in non-uniform control flow

### LINKED ISSUE #492: SPIR-V Subgroup Barriers and Non-Uniform Control Flow
The OpenCL C spec is clear that a sub-group barrier can only be used in uniform control flow:

> For these functions, if any work-item in a subgroup encounters a sub_group_barrier, the barrier must be encountered by all work-items in the subgroup before any are allowed to continue execution beyond the barrier.

But, this behavior is not enforced in the SPIR-V spec, and instead the SPIR-V spec leaves this behavior to the client API [link](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpControlBarrier):

> When Execution is Subgroup or Invocation, the behavior of this instruction in non-uniform control flow is defined by the client API.

This means that we should explicitly state in the OpenCL SPIR-V environment spec that behavior is undefined for OpControlBarrier where the scope for execution is Subgroup unless all invocations execute the same dynamic instance of the instruction.

bashbaug:
Fixes #492.

Clarifies in that behavior is undefined in an OpenCL environment for a subgroup barrier unless all work-items in the subgroup execute the same dynamic instance of the barrier.  This brings the OpenCL SPIR-V Environment spec in line with the OpenCL C spec.

### DIFF
--- a/env/validation_rules.asciidoc
+++ b/env/validation_rules.asciidoc
@@ -117,6 +117,11 @@ And, the memory-order constraint in _Memory Semantics_ must be one of:
        supporting {CL_DEVICE_ATOMIC_ORDER_SEQ_CST} in
        {CL_DEVICE_ATOMIC_FENCE_CAPABILITIES}.
 
+In all OpenCL environments, for the *Barrier Instruction* *OpControlBarrier*,
+when the _Scope_ for _Execution_ is *Subgroup*, behavior is undefined unless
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM.

bashbaug:

Issue #279, opened 2020-05-08
SPIR-V Environment: Type of Built-In Variables


bashbaug:
PR #278 clarifies that all built-in variables must be in the input storage class, but does not describe what the type of each built-in variable must be.  We should describe the supported types for each built-in variable.

Some questions to answer:

* For built-in vectors that can be vectors for multi-dimensional ND-ranges, are scalars valid if the ND-range is known to be 1D?

* I'm pretty sure all of the built-in variables are currently integers, but we need to define how big the integers representing each built-in variable may be.  Specifically, are there cases where a built-in variable must be 64-bits?  Are there cases where a built-in variable may be 16-bits?

pierremoreau:
> Specifically, are there cases where a built-in variable must be 64-bits?

Yes, all the built-ins that are of type `size_t` such as `get_global_id()` & co, since (regarding `size_t` and some other types)

> They are 32-bit types if the value of the CL_DEVICE_ADDRESS_BITS device query is 32-bits, and 64-bit types if the value of the query is 64-bits.

Note that `get_work_dim()` does return an `uint`, so on 64-bit devices you can end up with built-ins of both 32- and 64-bits sizes.

bashbaug:
Summary Table:

| OpenCL C Built-in | SPIR-V Built-In | OpenCL C Return Type | 
| - | - | - |
| get_num_groups(dim) | NumWorkgroups | size_t | 
| get_local_size(dim) | WorkgroupSize | size_t |
| get_group_id(dim) | WorkgroupId | size_t |
| get_num_groups(dim) | NumWorkgroups | size_t |
| get_local_size(dim) | WorkgroupSize | size_t |
| get_group_id(dim) | WorkgroupId | size_t |
| get_local_id(dim) | LocalInvocationId | size_t |
| get_global_id(dim) | GlobalInvocationId | size_t |
| get_local_linear_id | LocalInvocationIndex | size_t |
| get_work_dim(dim) | WorkDim | uint |
| get_global_size(dim) | GlobalSize | size_t |
| get_enqueued_local_size(dim) | EnqueuedWorkgroupSize | size_t |
| get_global_offset(dim) | GlobalOffset | size_t |
| get_global_linear_id | GlobalLinearId | size_t |
| get_sub_group_size | SubgroupSize | uint |
| get_max_sub_group_size | SubgroupMaxSize | uint |
| get_num_sub_groups | NumSubgroups | uint |
[...]

PR #526, opened 2020-12-15
add required type for SPIR-V built-in variables

bashbaug:
This is a possible fix for #279.

This PR essentially adds an Asciidoctor version of the table described in https://github.com/KhronosGroup/OpenCL-Docs/issues/279#issuecomment-633175823, with a few minor changes:

* For all of the OpenCL built-in functions that accept a `dim` argument, I've required that the SPIR-V variables must be a vector with three components.  This matches the behavior of the SPIR-V LLVM Translator, but is not strictly required, as described in the issue.

* For all of the OpenCL built-in functions that return a `size_t`, I've required that the SPIR-V variables also be a `size_t`, which means that they must be 64-bit integers for 64-bit pointers.  This also matches the behavior of the SPIR-V LLVM Translator, but is also not strictly required, as described in the issue.

I've left the column showing the associated OpenCL C built-in function for now since I think it's useful, but it is informational and non-normative and could be removed.

### DIFF
--- a/env/common_properties.asciidoc
+++ b/env/common_properties.asciidoc
@@ -400,49 +400,13 @@ In this table, `size_t` is used as a generic type to represent:
   * *OpTypeInt* with _Width_ equal to 32 if the _Addressing Model_ declared in *OpMemoryModel* is *Physical32*.
   * *OpTypeInt* with _Width_ equal to 64 if the _Addressing Model_ declared in *OpMemoryModel* is *Physical64*.
 
-The mapping from an OpenCL C built-in function is informational and non-normative.
+The mapping from an OpenCL C built-in function to the SPIR-V *BuiltIn* is informational and non-normative.
 
 [cols="2,2,3",options="header"]
 |====
 |*OpenCL C Function*
 |*SPIR-V BuiltIn*
 |*Required SPIR-V Type*
-
-| `get_num_groups`
- | *NumWorkgroups*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_local_size`
- | *WorkgroupSize*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_group_id`
- | *WorkgroupId*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_num_groups`
- | *NumWorkgroups*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_local_size`
- | *WorkgroupSize*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_group_id`
- | *WorkgroupId*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_local_id`
- | *LocalInvocationId*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_global_id`
- | *GlobalInvocationId*
-  | *OpTypeVector* of 3 components of `size_t`
-
-| `get_local_linear_id`
- | *LocalInvocationIndex*
-  | `size_t`
 
 | `get_work_dim`
  | *WorkDim*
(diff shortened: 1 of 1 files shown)


env/common_properties.asciidoc
```diff
@@ -393,3 +393,120 @@ additional types in an entry point's parameter list.
...
+
+| `get_sub_group_id`
+ | *SubgroupId*
+  | *OpTypeInt* with _Width_ equal to 32
+
+| `get_sub_group_local_id`
+ | *SubgroupLocalInvocationId*
+  | *OpTypeInt* with _Width_ equal to 32
+
+// TODO: Ensure this is documented as part of the extension!
```
alycm, line 491: It is, though as a paragraph rather than a table.

alycm: Hmm, I wrote the above a month ago but I can't find what made me say it.  The extension specification describes the non-normative mapping between OpenCL C and SPIR-V, but not the type of the SPIR-V builtin variables.

So this could be uncommented now?

alycm (approved):
LGTM in general.

env/common_properties.asciidoc
```diff
@@ -393,3 +393,120 @@ additional types in an entry point's parameter list.
...
+
+| `get_local_size`
+ | *WorkgroupSize*
+  | *OpTypeVector* of 3 components of `size_t`
+
+| `get_group_id`
+ | *WorkgroupId*
+  | *OpTypeVector* of 3 components of `size_t`
+
+| `get_num_groups`
```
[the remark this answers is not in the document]
bashbaug: Fixed - thanks!  I'm not sure how that happened but they were duplicated in my original markdown table also.

I've removed all of the duplicates (there were several others also) and I've verified that all of the built-in functions in section [6.15.1. Work-Item Functions](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#work-item-functions) are in the table.

env/common_properties.asciidoc
bashbaug: They're listed at the very bottom of [this section](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_cl_khr_subgroup_ballot) in text.

We need to decide if this table should only contain "core" BuiltIns or if it should contain extension BuiltIns also.  If it should only contain core BuiltIns then I think it should be merged as-is (I could delete the comments).  If it should contain extension BuiltIns also, are any others missing?

alycm: > They're listed at the very bottom of this section in text.

Ah, thanks.

> We need to decide if this table should only contain "core" BuiltIns or if it should contain extension BuiltIns also. If it should only contain core BuiltIns then I think it should be merged as-is (I could delete the comments). If it should contain extension BuiltIns also, are any others missing?

I think it would be okay to include them in this table, other extension-only information, such as Float16 ULPs, exist alongside non-extension information.  But if that meant deleting the mentioned paragraph in the `cl_khr_subgroup_ballot` section to avoid duplication then probably best to not add to the table, but keep all the information together in one section.

alycm: Merge this as is?

> If it should contain extension BuiltIns also, are any others missing?

From reading the [builtin variables table](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_builtin_a_builtin), no, I do not believe any more apply to OpenCL.

bashbaug:
Replying to the [comment above](https://github.com/KhronosGroup/OpenCL-Docs/pull/526#discussion_r589842289), yes, I think this should be merged as-is.  I think it properly documents the current behavior of the SPIR-V LLVM Translator so it's a definite improvement.

We can continue to discuss further enhancements in #279, or decide that these changes are sufficient and close it with no additional action required.  Thanks!

Issue #147, opened 2019-10-23
Alignment of Types for SPIR-V Kernels


bashbaug:
The OpenCL C spec has a section describing "Alignment of Types":

https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_C.html#alignment-of-types

This section describes assumptions compilers can make for OpenCL C kernels regarding alignments of loads and stores and alignments of kernel arguments.

It's possible I'm just looking in the wrong places, but I don't see similar text describing similar assumptions compilers can make for SPIR-V kernels, even though though I think this is intended and desirable.

Notes:

* I think this text probably belongs in the SPIR-V environment spec.
* There is an [Alignment Decoration](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Decoration) in the SPIR-V spec, but I think this is intended to be used for specific alignments (e.g. `__attribute__((aligned(16)))`) and not to communicate default alignment assumptions.

kpet:
I was about to file the same issue before I found this one. Agree this needs doing.

bashbaug:
Fixed by #527.

PR #527, opened 2020-12-15
add required alignment of types to the OpenCL SPIR-V Environment spec

bashbaug:
This is a possible fix for #147.

It adds a section to the OpenCL SPIR-V Environment specification describing the required alignment of types, similar to the [related section in the OpenCL C spec](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#alignment-of-types).

This PR also fixes a minor formatting issue in the OpenCL C spec I found while researching these changes.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -426,7 +426,7 @@ write the `half` scalar or vector value to memory.
 --
 
 The `char`, `unsigned char`, `short`, `unsigned short`, `int`, `unsigned int`,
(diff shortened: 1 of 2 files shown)


kpet (approved):
LGTM.

I wondered for a while whether we'd want to call out vload/vstore explicitly but I convinced myself that it was not necessary (the pointer they take is to the scalar type, not the vector type).

bashbaug:

PR #529, opened 2020-12-16
Extended Subgroups

bashbaug:
This PR adds a family of extensions that provide extended subgroup functionality. The extensions in this family are:

* `cl_khr_subgroup_extended_types`
* `cl_khr_subgroup_non_uniform_vote`
* `cl_khr_subgroup_ballot`
* `cl_khr_subgroup_non_uniform_arithmetic`
* `cl_khr_subgroup_shuffle`
* `cl_khr_subgroup_shuffle_relative`
* `cl_khr_subgroup_clustered_reduce`

The functionality added by these extensions includes:

* Additional data type support for subgroup broadcast, scan, and reduction functions;
* The ability to elect a single work item from a subgroup to perform a task;
* The ability to hold votes among work items in a subgroup;
* The ability to collect and operate on ballots from work items in the subgroup;
* The ability to use some subgroup functions, such as any, all, broadcasts, scans, and reductions within non-uniform flow control;
* Additional scan and reduction operators;
* Additional ways to exchange data among work items in a subgroup;
* Clustered reductions, that operate on a subset of work items in the subgroup.

### DIFF
--- a/ext/cl_khr_subgroup_extensions.asciidoc
+++ b/ext/cl_khr_subgroup_extensions.asciidoc
@@ -254,7 +254,8 @@ gentype sub_group_non_uniform_broadcast(
     uint index )
 ----
 | Returns _value_ for the work item with subgroup local ID equal to _index_.
-The value of _index_ must be equivalent for all active work items in the subgroup.
+
+Behavior is undefined when the value of _index_ is not equivalent for all active work items in the subgroup.
 
 The return value is undefined if the work item with subgroup local ID equal to _index_ is inactive or if _index_ is greater than or equal to the size of the subgroup.
 
@@ -280,8 +281,9 @@ int sub_group_inverse_ballot(
     uint4 value )
 ----
 | Returns the predicate value for this work item in the subgroup from the bitfield _value_ representing predicate values from all work items in the subgroup.
-The bitfield _value_ must be equivalent for all active work items in the subgroup.
 The predicate return value will be non-zero if the bit in the bitfield _value_ for this work item is set, and zero otherwise.
+
+Behavior is undefined when _value_ is not equivalent for all active work items in the subgroup.
 
 This is a specialized function that may perform better than the equivalent `sub_group_ballot_bit_extract` on some implementations.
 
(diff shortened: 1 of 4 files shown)


ext/cl_khr_subgroup_extensions.asciidoc
```diff
@@ -0,0 +1,1072 @@
...
+    gentype value )
+gentype sub_group_non_uniform_scan_exclusive_mul(
+    gentype value )
+----
+| Returns the result of an exclusive scan operation, which is the summation, multiplication, minimum, or maximum of _value_ for all active work items in the subgroup with a subgroup local ID less than this work item's subgroup local ID.
+
+If there is no active work item in the subgroup with a subgroup local ID less than this work item's subgroup local ID then an identity value `I` is returned.
+For *add*, the identity value is `0`.
+For *min*, the identity value is the largest representable value for integer types, or `+INF` for floating point types.
+For *max*, the identity value is the smallest representable value for integer types, or `-INF` for floating point types.
```
alycm, line 440: Not sure if it's just me but I read "smallest representable value" as magnitude, which would be `0` for signed or unsigned types.  Maybe "minimum" is less ambiguous?

```suggestion
For *max*, the identity value is the minimum representable value for integer types, or `-INF` for floating point types.
```

alycm: Can always come back and change this later though.

alycm (approved):
LGTM.

Hopefully we can get the `&#8203;` all automatically generated via the dictionaries, but not today.

bashbaug:
> Hopefully we can get the `&#8203;` all automatically generated via the dictionaries, but not today.

Yeah - I think we should be able to have a text file of "other symbols" that aren't in the XML file, for things like built-in function names, feature macro names, etc, then we can generate a dictionary of attributes from the text file similarly.

With the hard-coded `&#8203` characters this is going to have the same copy-paste and searching issue in some browsers we solved previously, but I think we can live with this temporarily.

I've committed all of your other suggestions.  Thanks!

PR #532, opened 2020-12-18
OpenCL C spec editorial fixes

bashbaug:
Editorial fixes for the OpenCL C spec.  There are no functional changes.

I'm not super happy with the way the Asciidoctor attributes for the OpenCL C features worked out, especially how the attribute names don't have the double-underscore prefix, but Asciidoctor seemed to be tripping over the attributes when they had the prefix.  I think this is workable for now, and Table 8. Address space behavior flows much better with the zero-width spaces, but this is definitely an area to revisit.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -34,6 +34,9 @@ Khronos{R} OpenCL Working Group
 // Various special / math symbols. This is easier to edit with than Unicode.
 include::config/attribs.txt[]
 
+// Feature Dictionary
+include::c/feature-dictionary.asciidoc[]
+
 // External Footnotes
 include::c/footnotes.asciidoc[]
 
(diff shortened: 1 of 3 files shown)


Oblomov:
I think I've seen other reports in AsciiDoc about the handling of double-underscores. I'll try to give a look at it to see if I can solve the thing in a satisfactory way.

OpenCL_C.txt
```diff
@@ -7868,14 +7877,15 @@ semantics of the minimum requirements.
 The OpenCL C programming language implements the following additional
 built-in vector functions.
 We use the generic type name `gentype__n__` (or `gentype__m__`) to indicate the
```
AnastasiaStulova, line 7879: Btw is `__m__` explained somewhere?

OpenCL_C.txt
```diff
@@ -2047,17 +2050,17 @@ types.
...
 
-OpenCL has hierarchical memory architecture represented by address spaces
-defined in <<embedded-c-spec, section 5 of Embedded C Specification>>. It
-extends C syntax to allow an address space name as a valid type qualifier
-(<<embedded-c-spec, section 5.1.2 of Embedded C>>).
-OpenCL implements the following disjoint named address spaces with the spelling:
+OpenCL C has a hierarchical memory architecture represented by address spaces, as
+defined in section 5 of the <<embedded-c-spec, Embedded C Specification>>. It
+extends the C syntax to allow an address space name as a valid type qualifier
+(section 5.1.2 of the <<embedded-c-spec, Embedded C Specification>>).
```
AnastasiaStulova, line 2056:
```suggestion
(section 5.1.2 of <<embedded-c-spec, the Embedded C Specification>>).
```

OpenCL_C.txt
```diff
@@ -2047,17 +2050,17 @@ types.
 [open,refpage='addressSpaceQualifiers',desc='Address Space Qualifiers',type='freeform',spec='clang',anchor='address-space-qualifiers',xrefs='constant genericAddressSpace global local private']
 --
 
-OpenCL has hierarchical memory architecture represented by address spaces
-defined in <<embedded-c-spec, section 5 of Embedded C Specification>>. It
-extends C syntax to allow an address space name as a valid type qualifier
-(<<embedded-c-spec, section 5.1.2 of Embedded C>>).
-OpenCL implements the following disjoint named address spaces with the spelling:
+OpenCL C has a hierarchical memory architecture represented by address spaces, as
+defined in section 5 of the <<embedded-c-spec, Embedded C Specification>>. It
```
AnastasiaStulova, line 2054:
```suggestion
defined in section 5 of <<embedded-c-spec, the Embedded C Specification>>. It
```

OpenCL_C.txt
```diff
@@ -2208,9 +2207,9 @@ only for the lifetime of the work-group executing the kernel.
 --
 
 The `+__constant+` or `constant` address space name is used to describe
-variables accessible globally (declared in program scope or inside functions
-with `static`/`extern` storage class specifier) as read-only variables.
-These read-only variables can be accessed by all work-items of different
+read-only variables that are accessible globally. They may
+be declared in program scope or inside functions with a `static` or `extern`
```
AnastasiaStulova, line 2211:
```suggestion
be declared in program scope or in the outermost kernel scope or inside
```

OpenCL_C.txt
```diff
@@ -2208,9 +2207,9 @@ only for the lifetime of the work-group executing the kernel.
...
+storage class specifier, and can be accessed by all work-items of different
```
AnastasiaStulova, line 2212:
```suggestion
 functions with a `static` or `extern` storage class specifier. Such variables
```

OpenCL_C.txt
```diff
@@ -2208,9 +2207,9 @@ only for the lifetime of the work-group executing the kernel.
...
 kernels during their execution.
```
AnastasiaStulova, line 2213:
```suggestion
 can be accessed by all work-items or by different kernels during the program execution.
```

AnastasiaStulova: I realized that the original wording might have been confusing since it might have implied that the function scope variables can be accessed by different kernels which is not possible.

AnastasiaStulova (commented):
Very nice improvements! Thanks... just a few small corrections/comments.

OpenCL_C.txt
bashbaug: I don't think it is precisely defined, but I think it should have the same possible values of `__n__`, correct?

If so, I could change from:

> _n_ is 2, 4, 8, or 16.

to:

> _n_ or __m__ is 2, 4, 8, or 16.

It may be easier to make this change in a subsequent PR though.

OpenCL_C.txt
```diff
@@ -2302,44 +2298,45 @@ kernel void foo(int a)
...
+        // Example of variable in __local address space but not
         // declared at __kernel function scope.
-        local float c; // not allowed.
+        local float c;  // Error.
     }
 }
 ----------
 
-Program scope variables or variables with `extern`/`static` storage class
+Program scope variables or variables with a `extern` or `static` storage class
```
alycm, line 2325:
```suggestion
Program scope variables or variables with an `extern` or `static` storage class
```

bashbaug:
> EDIT: I think I'm missing something, but most double-underscore usages are already properly escaped, and render correctly, that I can see (I did spot a couple that can be fixed, and I will provide a PR for them momentarily). Which attribute are missing the `__`?

Sorry bout that, I should have been clearer.

The rendering is all OK in the spec **outputs**, but I had to remove the double-underscore prefix from the attribute names in the spec **sources**, which is a touch confusing.  Example:

```
OpenCL C compilers that define the feature macro {opencl_c_3d_image_writes}
must also define the feature macro {opencl_c_images}.
```

I would have rather written this as:

```
OpenCL C compilers that define the feature macro {__opencl_c_3d_image_writes}
must also define the feature macro {__opencl_c_images}.
```

This _mostly_ worked, but some of the attributes didn't get substituted properly with the double-underscore prefix, and I couldn't figure out why it worked in some cases and not in others, even for the same feature attribute.

As a related aside, I really wanted to use the `+TEXT+` syntax in the attribute name:

```
:opencl_c_3d_image_writes: pass:q[`+__opencl_c_<wbr>3d_<wbr>image_<wbr>writes+`]
```

 vs. `\` escaping the attribute name:

```
:opencl_c_3d_image_writes: pass:q[`\__opencl_c_<wbr>3d_<wbr>image_<wbr>writes`]
```

But this also didn't work, and I ended up with `+` in the rendered text.

Note, I didn't test the latest versions of Asciidoctor, so it's possible some of this behavior is a bug that has since been fixed.

alycm (approved):
LGTM.

OpenCL_C.txt
AnastasiaStulova: sure!

Oblomov:
@bashbaug I see, thanks for the clarification. My understanding is that double-underscores usually fail when more than one are in the same line of text.. Anyway, I have some commits that fix up some of the uses for existing macros and tries to uniformize some of the usages, I'll set it up in a PR for review.

PR #533, opened 2020-12-19
Tuning to AsciiDoctor macro usage in the OpenCL C spec

Oblomov:
This is a follow-up to some comments from #532. It fixes one actual case where the lack of escaping caused misrendering, and then introduces a few extra macros and a more consistent usage of these macros throughout the document. source.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -93,14 +93,14 @@ what versions of OpenCL C specify that feature.
   ** In some instances the variation of "For OpenCL C _major.minor_ or newer"
      is used, it has the identical meaning.
   * Requires support for OpenCL C 2.0, or OpenCL C 3.0 or newer and the
-    `+__opencl_c_feature_name+` feature:
+    {opencl_c_feature_name} feature:
     Features that were introduced in OpenCL C 2.0 as mandatory, but made
     <<optional-functionality, optional>> in OpenCL C 3.0.
     Compilers for versions of OpenCL C 1.2 or below will not provide these
     features, compilers for OpenCL C 2.0 will provide these features,
     compilers for OpenCL C 3.0 or newer may provide these features.
   * Requires support for OpenCL C 3.0 or newer and the
-    `+__opencl_c_feature_name+` feature: <<optional-functionality,
+    {opencl_c_feature_name} feature: <<optional-functionality,
     Optional>> features that were introduced in OpenCL C 3.0.
     Compilers for an earlier version of OpenCL C will not provide these
     features, compilers for OpenCL C 3.0 or newer may provide these features.
@@ -131,7 +131,7 @@ associated __feature test macro__ will be predefined.
 
 The following table describes OpenCL C 3.0 or newer features and their
 meaning. The naming convention for the feature macros is
-`+__opencl_c_<name>+`.
+{opencl_c_feature_name}.
 
 Feature macro identifiers are used as names of features in this document.
 
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -2325,9 +2325,9 @@ kernel void my_func(...)
 Program scope variables or variables with a `extern` or `static` storage class
 specifier:
 
-  * Must be qualified by `__constant` in OpenCL C prior to 2.0 or OpenCL C 3.0
+  * Must be qualified by `{constant}` in OpenCL C prior to 2.0 or OpenCL C 3.0
     without {opencl_c_program_scope_global_variables} feature.
-  * Can be qualified by either `__constant` or `__global` for OpenCL C 2.0 or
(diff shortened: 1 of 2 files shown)


c/feature-dictionary.asciidoc
```diff
@@ -2,6 +2,14 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
+// “generic” opencl_c_feature_name
+ifdef::backend-html5[]
+:opencl_c_feature_name: pass:q[`\__opencl_c_<wbr><em>feature_<wbr>name</em>`]
```
bashbaug, line 7: I don't think we want to include the `<em>` and `</em>` tags here, since it causes the HTML output to be _italicized_.

bashbaug (changes requested):
Thank you for your contribution and my sincere apologies for not getting back to you sooner - I missed this PR initially over the holidays.

In addition to the specific comment below I'm also seeing a few places where `pass[__xxx]` appear in the spec after these changes.  They can be identified by diff'ing the HTML specs before and after these changes.  Could you please investigate why this is occurring and fix this also?  Thanks!

c/feature-dictionary.asciidoc
Oblomov: That was actually the objective, although I should have been more explicit about the it in the commit message 8-)

Since  `feature_name` is not actually the string that appears in the feature macro, but a placeholder, I believe it would be better to make it typographically distinct.

I'm actually only now noticing that further down `__opencl_c_<name>` is used instead, using `<>` to mark the replaceable part. I think we should uniformize these. Shall I go with the `<>` idea?

Oblomov:
No problem for the delay. And sorry for missing the excess of passthrough, I'll look at the locations where they appear and solve them.

OpenCL_C.txt
```diff
@@ -22,10 +22,15 @@ Khronos{R} OpenCL Working Group
 
 :cpp: pass:[C++]
 :cpp11: pass:[C++11]
+:kernel: pass[__kernel]
+:read_only: pass[__read_only]
+:write_only: pass[__write_only]
+:read_write: pass[__read_write]
```
alycm, line 28:
```suggestion
:kernel: pass:[__kernel]
:read_only: pass:[__read_only]
:write_only: pass:[__write_only]
:read_write: pass:[__read_write]
```
These missing `:` are the reason that the raw pass text is in the rendered specification.

c/feature-dictionary.asciidoc
alycm: Using `<>` is more distinctive to me than italics, so I'd unify on `__opencl_c_<name>`.

alycm (changes requested):
LGTM, needs the pass text in rendered output fixed of course.

Oblomov:
Thanks for the suggestions. I've pushed a refresh of the branch on top of the current master, with the missing colons inserted, and angle brackets around <feature_name>. I've also turned one explicit use of `opencl_c_<name>` to the feature name macro.

The HTML output seems correct now (I can't get the PDF to build, due to some missing prawn/templates dependency, will look into this).

c/feature-dictionary.asciidoc
```diff
@@ -2,6 +2,14 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
+// “generic” opencl_c_feature_name
+ifdef::backend-html5[]
+:opencl_c_feature_name: pass:q[`\__opencl_c_<wbr>&lt;feature_<wbr>name&gt;</em>`]
+endif::[]
+ifndef::backend-html5[]
+:opencl_c_feature_name: pass:q[`\__opencl_c_&#8203;<feature_&#8203;name>`]
```
alycm, line 10: I'm seeing the follow error on lines where this substitutes for PDF generation.

> asciidoctor: ERROR: failed to parse formatted text:

Side-note: Travis actually saw this error too, but `asciidoctor` still has a `0` exitcode and produces a PDF.  So the build passes.

Copying the HTML approach more closely seems to work:

```suggestion
:opencl_c_feature_name: pass:q[`\__opencl_c_&#8203;&lt;feature_&#8203;name&gt;`]
```

Oblomov: Oops, sorry. I can't build the PDF documentation so I missed it, and Travis not complaining didn't help 8-). I've pushed a rebased branch with the wrapper macro fixed (hopefully!)

alycm: Thanks, no more build problems now.

alycm (approved):
LGTM.

alycm:
@bashbaug I believe your requested changes have been addressed.

bashbaug (approved):
Thanks, all of my feedback has been addressed.  This is a nice improvement!

PR #537, opened 2021-01-07
Specification for cl_arm_scheduling_controls v0.2.0

kpet:

### DIFF
--- a/extensions/cl_arm_scheduling_controls.asciidoc
+++ b/extensions/cl_arm_scheduling_controls.asciidoc
@@ -29,7 +29,7 @@ Shipping
 == Version
 
 Built On: {docdate} +
-Version: 0.1.0
+Version: 0.2.0
 
 == Dependencies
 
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
Mostly LGTM.

Do we need to add the bit for `CL_DEVICE_SCHEDULING_DEFERRED_FLUSH_ARM` to the XML file also?

kpet:
Thanks for the review!

Yes, as well as other bit definitions for the `CL_DEVICE_SCHEDULING_CONTROLS_CAPABILITIES_ARM` query. There doesn't seem to be a precedent for anonymous bitmasks (the query returns `cl_bitfield` as opposed to a custom type). I'll fix this in the next version of the specification (probably by introducing an alias for `cl_bitfield`). It doesn't make any practical difference at present as the bit definitions aren't typed but it would certainly make it easier to improve type safety and automatic header generation in the future.

Merging as is for now.

Issue #540, opened 2021-01-13
tidy up clEnqueueNDRangeKernel error descriptions


bashbaug:
The error descriptions for `clEnqueueNDRangeKernel` are a little messy and it would be nice to tidy them up.

Some things to fix:

* The error descriptions are not in the canonical order.
* Several error descriptions refer to a "queue" instead of "command_queue".
* Several error descriptions can be shortened or otherwise rewritten to increase clarify.

PR #542, opened 2021-01-15
Remove erroneous condition for CL_INVALID_KERNEL_ARGS

jrprice:
This should be an error at program build time, not during kernel
enqueue.


bashbaug (approved):
LGTM.  I think additional work is needed in this area (see #540), but this change is a step in the right direction.

I'll leave this open for another day or so before merging, in case anyone else wants to review.  Thanks!

kpet (approved):
Thanks!

alycm (approved):
LGTM, merging.

PR #550, opened 2021-01-25
Improved description for pipe specifier.

AnastasiaStulova:
Pipe specifies a very distinct type from its
element and therefore a better wording describing
it should be `'type specifier'`.

Note that pipe can not be a type qualifier because
in C-based languages, a type qualifier is an optional
property that doesn't modify the representation of the
type (i.e. casting away qualifiers is valid in certain
circumstances). OpenCL spec never allowed to cast away
pipe and doing so won't result in anything useful
unless there is some knowledge of its implementation.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -191,7 +191,7 @@ OpenCL C compilers for FULL profile devices or devices with 64-bit pointers
 must always define the {opencl_c_int64} feature macro.
 
 | {opencl_c_pipes}
-| The OpenCL C compiler supports the pipe modifier and built-in functions
+| The OpenCL C compiler supports the pipe specifier and built-in functions
 to read and write from a pipe.
 
 OpenCL C compilers that define the feature macro {opencl_c_pipes} must
@@ -10070,14 +10070,14 @@ pipe user_type_t pipeB; // a pipe with user_type_t packets
 ----------
 
 The `read_only` (or `+__read_only+`) and `write_only` (or `+__write_only+`)
-qualifiers must be used with the `pipe` qualifier when a pipe is a parameter
+qualifiers must be used with the `pipe` specifier when a pipe is a parameter
 of a kernel or of a user-defined function to identify if a pipe can be read
 from or written to by a kernel and its callees and enqueued child kernels.
 If no qualifier is specified, `read_only` is assumed.
 
 A kernel cannot read from and write to the same pipe object.
 Using the `read_write` (or `+__read_write+`) qualifier with the `pipe`
-qualifier is a compilation error.
+specifier is a compilation error.
 
 In the following example
 
@@ -10102,7 +10102,7 @@ The macro `CLK_NULL_RESERVE_ID` refers to an invalid reservation ID.
   * Pipes can only be passed as arguments to a function (including kernel
     functions).
     The <<operators,C operators>> cannot be used with variables declared
-    with the pipe qualifier.
+    with the pipe specifier.
   * The `pipe` specifier cannot be used with variables declared inside a
     kernel, a structure or union field, a pointer type, an array, global
     variables declared in program scope or the return type of a function.


AnastasiaStulova:
In clang `pipe` is implemented similarly to `_Atomic` that is also defined as a **_type specifier_** in C11 spec.

AnastasiaStulova:
I also think that the description of `pipe` should be moved elsewhere because in this section we are supposed to describe the functions only. If it's relevant I would suggest doing in a separate PR.

OpenCL_C.txt
```diff
@@ -10052,8 +10052,8 @@ NOTE: The functionality described in this section <<unified-spec, requires>>
 support for OpenCL C 2.0, or OpenCL C 3.0 or newer and the {opencl_c_pipes} feature.
 
 A pipe is identified by specifying the `pipe` keyword with a type.
-The data type specifies the size of each packet in the pipe.
-The `pipe` keyword is a type modifier.
+The data type specifies the type of each element in the pipe.
```
kpet, line 10055: I'm wondering whether we want to keep something along the lines of "specifies the type, **and thus size**" or if it's "obvious".

kpet (approved):
LGTM as is.

Agree it's odd that we don't describe pipes outside of 6.15 (Built-in functions). We should probably touch on it in 6.3 (Supported Data Types). Agree that it's best tackled as a separate PR.

bashbaug (commented):
Generally LGTM - "specifier" does seem to be more correct than "modifier", though I've always felt that the pipe type is very strange.

I think we should update the description in the feature macro table to be consistent though.  It currently says:

> The OpenCL C compiler supports the pipe **modifier** and built-in functions to read and write from a pipe.

Maybe change to:

> The OpenCL C compiler supports the pipe **type specifier** and built-in functions to read and write from a pipe.

I didn't find any other places where "modifier" should be changed to "specifier" but it's probably worth a quick grep through the spec to confirm.

OpenCL_C.txt
AnastasiaStulova: I find it implicit since every type has its size.

AnastasiaStulova:
> Generally LGTM - "specifier" does seem to be more correct than "modifier", though I've always felt that the pipe type is very strange.
> 
> I think we should update the description in the feature macro table to be consistent though. It currently says:
> 
> > The OpenCL C compiler supports the pipe **modifier** and built-in functions to read and write from a pipe.
> 
> Maybe change to:
> 
> > The OpenCL C compiler supports the pipe **type specifier** and built-in functions to read and write from a pipe.
> 
> I didn't find any other places where "modifier" should be changed to "specifier" but it's probably worth a quick grep through the spec to confirm.

We didn't use the modifier but we used qualifier in various places. I have fixed them now.

AnastasiaStulova:
> LGTM as is.
> 
> Agree it's odd that we don't describe pipes outside of 6.15 (Built-in functions). We should probably touch on it in 6.3 (Supported Data Types). Agree that it's best tackled as a separate PR.



> LGTM as is.
> 
> Agree it's odd that we don't describe pipes outside of 6.15 (Built-in functions). We should probably touch on it in 6.3 (Supported Data Types). Agree that it's best tackled as a separate PR.

I have created an issues to track this: https://github.com/KhronosGroup/OpenCL-Docs/issues/551

bashbaug (approved):

OpenCL_C.txt
StuartDBrady: Yeah, it does seem implicit. Does it need to be explicit, though? The API spec talks about packet sizes, and I don't see anything saying that the read and write ends have to use identical data types, but I would expect the packet size would need to match.

The concern would be that someone might assume it's valid to write int4 values to one end and read scalar ints from the other, but as far as I can see, this would be invalid.

StuartDBrady: The OpenCL C specification consistently uses the term "packet" when referring to pipes. It may be worth discussing whether "element" would be clearer, it would make sense to use a single term consistently. Perhaps the term "element" was originally avoided to prevent confusion between "pipe elements" and "vector elements".

bashbaug: FWIW, the OpenCL API spec also refers almost exclusively to pipe "packets" and not pipe "elements".

The OpenCL C spec seems to hint that they are equivalent:

> When it is applied to another type T, the result is a pipe type whose elements (or packets) are of type T.

We're getting a bit out in the weeds here from the purpose of this PR, but I also wonder how the `pipe_packet_size` specified when the pipe is created relates to the OpenCL C pipe type.  Does the size have to match?  Or, it OK if the pipe packet size is bigger than the OpenCL C pipe type?  What happens if the pipe packet size is smaller than the OpenCL C pipe type?

AnastasiaStulova: As for OpenCL C I would prefer to use `element` as it is more common concept in C-based lnguages.

>We're getting a bit out in the weeds here from the purpose of this PR, but I also wonder how the pipe_packet_size specified when the pipe is created relates to the OpenCL C pipe type. Does the size have to match? Or, it OK if the pipe packet size is bigger than the OpenCL C pipe type? What happens if the pipe packet size is smaller than the OpenCL C pipe type?

My understanding is that pipe type itself is opaque and its size is implementation defined and allowed to be different from its element/packet type. In clang it is just a pointer type to a struct that comntains implementation of pipe data structure.

AnastasiaStulova: I can see thare are a lot more we can clarify for pipes - should we create separate issues/PRs?

PR #552, opened 2021-01-26
cl_ext_float_atomics

bashbaug:
This PR is for a multi-vendor (EXT) extension for floating-point atomics.

It is currently a draft PR because there are a few issues to resolve before it should be considered "done", but it is ready for review.  I am hoping we can close the remaining issues promptly, at which time I will remove the draft status and bump the version to `1.0.0`.

The most important areas where I would like feedback are:

- [x] Do the OpenCL C function names and types added by this extension need a prefix?  Past extensions have prefixed function names and types for vendor-specific extensions, but not for standard (KHR) extensions.  I don't think we have had an EXT extension that extensions OpenCL C before.
- [x] What should the convention be for OpenCL C feature names added by extensions?  This extension uses `__opencl_c_ext_feature_name`, which seems reasonable for an EXT extension, but I can switch to a different convention if desired.
- [x] Are there any general comments or requested changes regarding the use of OpenCL C features and feature macros to guard the optional parts of this extension, since this is the first extension that has added OpenCL C features?
- [x] Is there any need or desire to support floating-point atomics using the pre-OpenCL C 2.0 legacy atomic syntax?  Note, I think this could be added later as a follow-on extension or a newer version of this extension, if desired.

Other minor TODOs:

- [x] Confirm the correct copyright for an EXT extension - should this be the Khronos copyright or somethign different?
- [x] Assign enums in the XML file for the new API queries.  This can come out of an Intel block or a shared block.
- [x] Final editorial pass to fix any awkward wrapping, especially for long feature names or function prototypes.
- [x] Summarize any of the decisions above in the extension template.

### DIFF
--- a/extensions/cl_ext_float_atomics.asciidoc
+++ b/extensions/cl_ext_float_atomics.asciidoc
@@ -320,11 +320,11 @@ C atomic_fetch_max_explicit(volatile __local A *object, M operand,
 
 // In addition to the requirements described in the OpenCL C 3.0 specification,
 // requires the __opencl_c_ext_fp16_global_atomic_min_max feature
-// or the __opencl_c_ext_fp16_local_atomic_min_mas feature (for atomic_half),
+// or the __opencl_c_ext_fp16_local_atomic_min_max feature (for atomic_half),
 // requires the __opencl_c_ext_fp32_global_atomic_min_max feature
-// or the __opencl_c_ext_fp32_local_atomic_min_mas feature (for atomic_float), or
+// or the __opencl_c_ext_fp32_local_atomic_min_max feature (for atomic_float), or
 // requires the __opencl_c_ext_fp64_global_atomic_min_max feature
-// or the __opencl_c_ext_fp64_local_atomic_min_mas feature (for atomic_double).
+// or the __opencl_c_ext_fp64_local_atomic_min_max feature (for atomic_double).
 C atomic_fetch_min(volatile A *object, M operand)
 C atomic_fetch_max(volatile A *object, M operand)
 C atomic_fetch_min_explicit(volatile A *object, M operand, memory_order order)
@@ -589,12 +589,12 @@ C atomic_fetch_max_explicit(volatile __local A *object, M operand,
     memory_order order, memory_scope scope)
 
 // In addition to the requirements described in the OpenCL C 3.0 specification,
-// requires the __opencl_c_ext_fp16_global_atomic_min_mas feature
-// or the __opencl_c_ext_fp16_local_atomic_min_mas feature (for atomic_half),
-// requires the __opencl_c_ext_fp32_global_atomic_min_mas feature
-// or the __opencl_c_ext_fp32_local_atomic_min_mas feature (for atomic_float), or
-// requires the __opencl_c_ext_fp64_global_atomic_min_mas feature
-// or the __opencl_c_ext_fp64_local_atomic_min_mas feature (for atomic_double).
+// requires the __opencl_c_ext_fp16_global_atomic_min_max feature
+// or the __opencl_c_ext_fp16_local_atomic_min_max feature (for atomic_half),
+// requires the __opencl_c_ext_fp32_global_atomic_min_max feature
+// or the __opencl_c_ext_fp32_local_atomic_min_max feature (for atomic_float), or
+// requires the __opencl_c_ext_fp64_global_atomic_min_max feature
+// or the __opencl_c_ext_fp64_local_atomic_min_max feature (for atomic_double).
 C atomic_fetch_min(volatile A *object, M operand)
 C atomic_fetch_max(volatile A *object, M operand)
 C atomic_fetch_min_explicit(volatile A *object, M operand, memory_order order)
--- a/extensions/cl_ext_float_atomics.asciidoc
+++ b/extensions/cl_ext_float_atomics.asciidoc
@@ -3,6 +3,7 @@
 // http://creativecommons.org/licenses/by/4.0/
 
 :data-uri:
+:sectanchors:
 :icons: font
 :source-highlighter: coderay
 
--- a/extensions/cl_ext_float_atomics.asciidoc
+++ b/extensions/cl_ext_float_atomics.asciidoc
@@ -642,7 +642,7 @@ If the OpenCL environment supports the extension `cl_ext_float_atomics` and the 
   * When {CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT} includes `CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT` and `CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT`, and the *GenericPointer* capability is supported and declared, the _Pointer_ operand may be a pointer to the *Generic* _Storage Class_.
 
 If the OpenCL environment supports the extension `cl_ext_float_atomics` and the {CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT}, or {CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT} bitfields include `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extension `SPV_EXT_shader_atomic_float_add`.
-If the OpenCL environment supports the extension `cl_ext_float_atomics` and the {CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT} bitfield includes `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extension `SPV_EXT_shader_atomic_float16_add`.
+If the OpenCL environment supports the extension `cl_ext_float_atomics` and the {CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT} bitfield includes `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extensions `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float16_add`.
 Additionally:
 
   * When {CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT} includes `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, the *AtomicFloat32AddEXT* capability must be supported.
--- a/extensions/cl_ext_float_atomics.asciidoc
+++ b/extensions/cl_ext_float_atomics.asciidoc
@@ -1,0 +1,607 @@
+// Copyright 2018-2021 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+:data-uri:
+:icons: font
+:source-highlighter: coderay
+
+= cl_ext_float_atomics
+:R: pass:q,r[^(R)^]
+Khronos{R} OpenCL Working Group
+
+== Name Strings
+
+`cl_ext_float_atomics`
+
+== Contact
+
+Please see the *Issues* list in the Khronos *OpenCL-Docs* repository: +
+https://github.com/KhronosGroup/OpenCL-Docs
+
+== Contributors
+
+// spell-checker: disable
+Ben Ashbaugh, Intel +
+Alex Paige, Intel +
+Lukasz Towarek, Intel
+// spell-checker: enable
+
+== Notice
+
+Copyright TBD
+
+== Status
+
+Working Draft - [red]*DO NOT SHIP*
+
+== Version
+
+Built On: {docdate} +
+Revision: 0.9.0
+
+== Dependencies
+
+This extension is written against the OpenCL API Specification, OpenCL C Specification, and OpenCL SPIR-V Environment Specification Versions 3.0.5.
+
+The functionality added by this extension uses the OpenCL C 2.0 atomic syntax and hence requires OpenCL 2.0 or newer.
+
+This extension interacts with `cl_khr_fp16` by optionally adding the ability to atomically operate on 16-bit floating point values in memory.
+
+This extension depends on `SPV_EXT_shader_float_atomic_add` and `SPV_EXT_shader_atomic_float_min_max` for implementations that support SPIR-V and floating-point atomic add, min, or max operations.
+
+== Overview
+
+This extension enables programmers to perform atomic operations on floating-point numbers in memory.
+An OpenCL device supporting this extension may support atomic operations on 16-bit half-precision floating-point values (`fp16`), 32-bit single-precision floating-point values (`fp32`), or 64-bit double-precision floating-point values (`fp64`).
+For these types, an OpenCL device may support basic atomic operations (load, store, and exchange), atomic addition and subtraction, and atomic min and max.
+The floating-point numbers may be in global or local memory.
+
+== New API Functions
+
+None.
+
+== New API Enums
+
+Accepted value for the _param_name_ parameter to *clGetDeviceInfo* to query the floating-point atomic capabilities of an OpenCL device:
+
+[source]
+----
+#define CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT 0xXXXX
+#define CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT 0xXXXX
+#define CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT   0xXXXX
+----
+
+Bitfield type describing the floating-point atomic capabilities of an OpenCL device.
+Subsequent versions of this extension may add additional floating-point atomic capabilities:
+
+[source]
+----
+typedef cl_bitfield         cl_device_fp_atomic_capabilities;
+
+#define CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT       (1 << 0)
+#define CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT              (1 << 1)
+#define CL_DEVICE_GLOBAL_FP_ATOMIC_MIN_MAX_EXT          (1 << 2)
+
+/* bits 3 - 15 are currently unused */
+
+#define CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT        (1 << 16)
+#define CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT               (1 << 17)
+#define CL_DEVICE_LOCAL_FP_ATOMIC_MIN_MAX_EXT           (1 << 18)
+
+/* bits 19 and beyond are currently unused */
+----
+
+== New OpenCL C Feature Names
+
+[source]
+----
+__opencl_c_ext_fp16_global_atomic_load_store
+__opencl_c_ext_fp16_local_atomic_load_store
+__opencl_c_ext_fp16_global_atomic_add
+__opencl_c_ext_fp32_global_atomic_add
+__opencl_c_ext_fp64_global_atomic_add
+__opencl_c_ext_fp16_local_atomic_add
+__opencl_c_ext_fp32_local_atomic_add
+__opencl_c_ext_fp64_local_atomic_add
+__opencl_c_ext_fp16_global_atomic_min_max
+__opencl_c_ext_fp32_global_atomic_min_max
+__opencl_c_ext_fp64_global_atomic_min_max
+__opencl_c_ext_fp16_local_atomic_min_max
+__opencl_c_ext_fp32_local_atomic_min_max
+__opencl_c_ext_fp64_local_atomic_min_max
+----
+
+== New OpenCL C Types
+
+[source]
+----
+atomic_half
+----
+
+== New OpenCL C Functions
+
+Add support for `atomic_half` for the following functions:
+
+[source]
+----
+// atomic_store:
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_global_atomic_load_store feature.
+void atomic_store(volatile __global A *object, C desired)
+void atomic_store_explicit(volatile __global A *object, C desired, memory_order order)
+void atomic_store_explicit(volatile __global A *object, C desired, memory_order order, memory_scope scope)
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_local_atomic_load_store feature.
+void atomic_store(volatile __local A *object, C desired)
+void atomic_store_explicit(volatile __local A *object, C desired, memory_order order)
+void atomic_store_explicit(volatile __local A *object, C desired, memory_order order, memory_scope scope)
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_global_atomic_load_store feature
+// or the __opencl_c_ext_fp16_local_atomic_load_store feature.
+void atomic_store(volatile A *object, C desired)
+void atomic_store_explicit(volatile A *object, C desired, memory_order order)
+void atomic_store_explicit(volatile A *object, C desired, memory_order order, memory_scope scope)
+
+// atomic_load:
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_global_atomic_load_store feature.
+C atomic_load(volatile __global A *object)
+C atomic_load_explicit(volatile __global A *object, memory_order order)
+C atomic_load_explicit(volatile __global A *object, memory_order order, memory_scope scope)
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_local_atomic_load_store feature.
+C atomic_load(volatile __local A *object)
+C atomic_load_explicit(volatile __local A *object, memory_order order)
+C atomic_load_explicit(volatile __local A *object, memory_order order, memory_scope scope)
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_global_atomic_load_store feature
+// or the __opencl_c_ext_fp16_local_atomic_load_store feature.
+C atomic_load(volatile A *object)
+C atomic_load_explicit(volatile A *object, memory_order order)
+C atomic_load_explicit(volatile A *object, memory_order order, memory_scope scope)
+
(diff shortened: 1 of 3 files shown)


b-sumner:
Given the variety of current hardware support for floating point atomics, this EXT may see more implementations if some relaxation can be permitted with rounding mode, and subnormal handling.   But minimally, this spec does need to say something about rounding modes, subnormal handling, and Inf and NaN handling.

bashbaug:
> But minimally, this spec does need to say something about rounding modes, subnormal handling, and Inf and NaN handling.

Thanks, this should absolutely be included.

> Given the variety of current hardware support for floating point atomics, this EXT may see more implementations if some relaxation can be permitted with rounding mode, and subnormal handling.

It'd be great to understand if any vendors would like to implement this extension that have specific floating-point needs so they can be incorporated.  Note that there are several special-cases already described in the corresponding SPIR-V extensions, e.g. for [float atomic min and max](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_shader_atomic_float_min_max.html).

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,608 @@
...
+
+== Dependencies
+
+This extension is written against the OpenCL API Specification, OpenCL C Specification, and OpenCL SPIR-V Environment Specification Versions 3.0.5.
+
+The functionality added by this extension uses the OpenCL C 2.0 atomic syntax and hence requires OpenCL 2.0 or newer.
+
+This extension interacts with `cl_khr_fp16` by optionally adding the ability to atomically operate on 16-bit floating point values in memory.
+
+This extension depends on `SPV_EXT_shader_float_atomic_add` and `SPV_EXT_shader_atomic_float_min_max` for implementations that support SPIR-V and floating-point atomic add, min, or max operations.
```
zhang3663Q, line 52: Is it too late to make these two SPIR-V extension naming consistent?

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,608 @@
...
+  * For the *Atomic Instructions* *OpAtomicFMinEXT* and *OpAtomicFMaxEXT* added by this extension:
+   ** When `CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES`, `CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES`, or `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_GLOBAL_FP_ATOMIC_MIN_MAX_EXT`, the _Pointer_ operand may be a pointer to the *CrossWorkGroup* _Storage Class_.
+   ** When `CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES`, `CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES`, or `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_LOCAL_FP_ATOMIC_MIN_MAX_EXT`, the _Pointer_ operand may be a pointer to the *Workgroup* _Storage Class_.
+//   ** When `CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES`, `CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES`, or `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_GLOBAL_FP_ATOMIC_MIN_MAX_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_MIN_MAX_EXT`, and the *GenericPointer* capability is supported and declared, the _Pointer_ operand may be a pointer to the *Generic* _Storage Class_.
+
+--
+
+== Issues
```
zhang3663Q, line 563: Should we clarify min/max behavior regarding to qNaN, sNaN, NaN, +0.0, and -0.0?

extensions/cl_ext_float_atomics.asciidoc
bashbaug: This is a typo in the spec - the extension for float atomic add is [SPV_EXT_shader_atomic_float_add](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_shader_atomic_float_add.html), which is consistent with [SPV_EXT_shader_atomic_float_min_max](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_shader_atomic_float_min_max.html).  I'll fix this - thanks!

extensions/cl_ext_float_atomics.asciidoc
bashbaug: Yes, I will add this, see the related comment above: https://github.com/KhronosGroup/OpenCL-Docs/pull/552#issuecomment-767850862.  I plan to carry the same min/max behavior from the SPIR-V spec into this OpenCL spec, unless there is a compelling reason to deviate.

I'd also like to document any special behavior for floating-point atomic add, if you have any input.  Thanks!

extensions/cl_ext_float_atomics.asciidoc
bashbaug: I fixed the typo in the latest version.

extensions/cl_ext_float_atomics.asciidoc
bashbaug: I brought over the behavior from the SPIR-V spec in the latest version.

bashbaug:
A few thoughts regarding floating-point special-cases before writing anything formal:

* I think we should keep the same rounding behavior as standard floating-point add operations unless there's a very strong reason to deviate.  So, we'd just refer to the existing section for [rounding modes](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_rounding_modes), if anything.
* It would be nice to support most INFs and NaNs similar to standard floating-point add operations.  Are any special cases required, similar to what we have documented for min and max?
* Our implementation should be able to support denorms and +/-0 similar to standard floating-point add operations, but I think the extension would still be useful if we relaxed these (or specified behavior as implementation-defined).  We already say that denorms may be flushed to zero for texture sampling, for example.

Does this seem reasonable?  Would any potential implementers prefer even more relaxed behavior?

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,641 @@
...
+| Device Info | Return Type | Description
+| `CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT`
+  `CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT`
+  `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT`
+  | `cl_device_fp_atomic_capabilities`
+      | Describes the floating-point atomic operations supported by the device.
+        This is a bit-field that describes a combination of the following values:
+
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT` - Can perform floating-point load, store, and exchange atomic operations in global memory. +
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` - Can perform floating-point addition and subtraction atomic operations in global memory.
```
StuartDBrady, line 292: It looks to me as though there is a missing "+" at the end of this line.

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,641 @@
...
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_MIN_MAX_EXT` - Can perform floating-point min and max atomic operations in global memory.
+
+        `CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT` - Can perform floating-point load, store, and exchange atomic operations in local memory. +
+        `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT` - Can perform floating-point addition and subtraction atomic operations in local memory.
```
StuartDBrady, line 296: Here, too, a missing "+" at the end.

StuartDBrady (changes requested):
Is `atomic_compare_exchange` (etc.) deliberately excluded for `atomic_half` (despite being already available for `atomic_float`)?  If so, why is that?

bashbaug:
Discussion in the SPIR-V call on February 10th: We should document if/how build flags (such as `-cl-no-signed-zeros` and`-cl-finite-math-only`) affect floating-point atomic operations.

Kerilk:
I will try to reflect the discussion I started during WG meeting of May 25:

I think we have an opportunity to offer broader float atomic support in general through the use of integer atomics and compiler support. The idea would be to offer default support for all floating point atomics operations through the usual compare and exchange loop, while still offering the opportunity to transparently replace supported variants with hardware accelerated versions.

This would avoid people re-implementing those atomic operation in their code (as I have don myself in the past), and may allow potentially better implementations even without hardware acceleration.

I found this discussion that may be relevant:
https://reviews.llvm.org/D71726

extensions/cl_ext_float_atomics.asciidoc
bashbaug: Fixed - thanks!

extensions/cl_ext_float_atomics.asciidoc
bashbaug: Fixed - thanks!

bashbaug:
> Is `atomic_compare_exchange` (etc.) deliberately excluded for `atomic_half` (despite being already available for `atomic_float`)? If so, why is that?

No strong reason, mostly:

* We haven't had requests for a bitwise fp16 compare-and-exchange (yet).
* The [corresponding Vulkan extension](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceShaderAtomicFloatFeaturesEXT.html) only supports load, store, and exchange.

If there is desire to support compare-and-exchange I think it would be as easy as adding a new atomic capability bit for it.  We could do this in this initial version or in a subsequent version.  Would a bitwise fp16 compare and exchange be interesting for 
 you?

bashbaug:
The latest version documents interactions with compiler options affecting floating-point behavior.  Specifically, these floating-point atomic operations may be affected by `-cl-no-signed-zeros`, `-cl-denorms-are-zero`, and `-cl-finite-math-only`.

bashbaug:
Regarding @kerilk's [comment above](https://github.com/KhronosGroup/OpenCL-Docs/pull/552#issuecomment-848078965):

I think this is advocating for some kind of generalized floating-point atomic library that:

1. Could use "native" floating-point atomic support when available, such as the atomic support provided by this extension, but...
2. Would fall back to an emulated version using compare-and-exchange loops or locks otherwise, when no native support is available.

Does this sound correct?  If so, I certainly agree this could be useful.  The trick will be figuring out how to deliver this kind of library.  We've had similar discussions in the past regarding math libraries.

Do you think this is a separate (complementary) discussion to this extension spec, or do you think it could affect the extension spec directly?

Tagging @AnastasiaStulova also, since this may make for an interesting tooling topic.

Kerilk:
@bashbaug You are correct.

I have the same interrogation, it would certainly:
 - be broader, I think @kpet mentioned we could define the operations using a meta description like [atomic_fetch_key](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_fetch_key) or generate the specification
 - simplify `CL_DEVICE_{SINGLE,DOUBLE,HALF}_FP_ATOMIC_CAPABILITIES` as all operations would be supported, and maybe add a `CL_DEVICE_{SINGLE,DOUBLE,HALF}_FP_ATOMIC_NATIVE_CAPABILITIES` that could be more refined
 - The default mode of operation would be those of the non atomic operations under the current optimization flags (NaN, denorms etc...), this may be a minefield though.
That's all I can think of, but I may have missed a few. All in all it would not be fundamentally different from what you propose here I think.

bashbaug:
I believe I have integrated all of the feedback from recent teleconferences, I've updated the extension template, and I've resolved the issues in the extension spec, so I've marked this PR as ready for review.

There are a few remaining (minor) tasks:

* We still need to assign enums to this extension.  I'll assign them out of an Intel block shortly unless there's a request to allocate from a different enum block instead.
* I still need to do a final editorial pass through the extension spec.  I'll complete this shortly.
* @StuartDBrady can you please confirm that your feedback has been incorporated?
* @Kerilk can you please confirm that we should not hold up this extension based on the discussion above and that the extension will help with the creation of a floating-point atomic library?

Thanks!

bashbaug:
I've assigned enums and completed the final editorial pass.  If there are no new bugs or last minute suggestions I think this is "done".

StuartDBrady:
The question regarding `atomic_compare_exchange` no longer seems to apply, as I can no longer see any changes relating to this function in this PR.  The missing `+` characters seem to have been added.  It seems my feedback has been incorporated.

I don't think I stated this here before, but I have some amount of uneasiness as to what the principles are behind atomic floating point operations - I'd be very keen to hear the opinions of experts in numerical analysis, or to see relevant published material regarding this.

Aside from concerns as to how these functions might be _used_ (especially considering floating point accuracy, handling of NaNs, best practices for using fetches to build higher level floating point atomics), the changes themselves seem okay to me.

Kerilk:
@bashbaug I don't have a proposal that would be mature enough to be making a timely contribution, so I don't want to hold this extension that I know some people need. This being said a "software"/library based first implementation would maybe alleviate some of @StuartDBrady concerns, at least ensuring each floating point operation is realized consistently with regard to the non-atomic operations. The scheduling of those atomic operations, and the impact on the numerical precision of the result, would still be an open question.

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,734 @@
...
+C atomic_fetch_min_explicit(volatile __local A *object, M operand, memory_order order)
+C atomic_fetch_max_explicit(volatile __local A *object, M operand, memory_order order)
+C atomic_fetch_min_explicit(volatile __local A *object, M operand,
+    memory_order order, memory_scope scope)
+C atomic_fetch_max_explicit(volatile __local A *object, M operand,
+    memory_order order, memory_scope scope)
+
+// In addition to the requirements described in the OpenCL C 3.0 specification,
+// requires the __opencl_c_ext_fp16_global_atomic_min_max feature
+// or the __opencl_c_ext_fp16_local_atomic_min_mas feature (for atomic_half),
```
svenvh, line 323:
```suggestion
// or the __opencl_c_ext_fp16_local_atomic_min_max feature (for atomic_half),
```

This `_mas` -> `_max` typo occurs in 8 more places further down.

haonanya:
Hi, @bashbaug.
Clang community is slowly transitioning from old and enormously huge opencl-c.h header to tablegen-generated version of it. Supporting cl_ext_float_atomics needs to add the new builtins to OpenCLBuiltins.td now. 
The issue is that the generic address space builtins are enabled by one of multiple feature macros, which is something that is currently not supported by the OpenCLBuiltins.td handling. Could we provide a dedicated feature macro for generic?
What is your opinion?
+@svenvh 
Thanks very much.

extensions/cl_ext_float_atomics.asciidoc
bashbaug: Thank you, I think I've fixed all of these (my grep for `mas` didn't find anything).

I didn't rev the document version since this is just a typo fix.

bashbaug:
> Could we provide a dedicated feature macro for generic?

I'm absolutely open to suggestions for improvement here.  Note that we already have a feature macro for the generic address space - `__opencl_c_generic_address_space`.  My intent is that this would influence which of these built-ins are available - this is one of "the requirements described in the OpenCL 3.0 specification".

Example:

```c
// In addition to the requirements described in the OpenCL C 3.0 specification,
// requires the __opencl_c_ext_fp16_global_atomic_load_store feature
// or the __opencl_c_ext_fp16_local_atomic_load_store feature.
C atomic_load(volatile A *object)
C atomic_load_explicit(volatile A *object, memory_order order)
C atomic_load_explicit(volatile A *object,
    memory_order order, memory_scope scope)
```

Is this sufficient?  Or is the request to have an explicit `__opencl_c_ext_fp16_generic_atomic_load_store` feature in addition to the current "global" and "local" feature macros?

@AnastasiaStulova is this a topic we discuss this in an upcoming tooling call?

Thanks!

AnastasiaStulova:
> > Could we provide a dedicated feature macro for generic?
> 
> I'm absolutely open to suggestions for improvement here. Note that we already have a feature macro for the generic address space - `__opencl_c_generic_address_space`. My intent is that this would influence which of these built-ins are available - this is one of "the requirements described in the OpenCL 3.0 specification".
> 
> Example:
> 
> ```c
> // In addition to the requirements described in the OpenCL C 3.0 specification,
> // requires the __opencl_c_ext_fp16_global_atomic_load_store feature
> // or the __opencl_c_ext_fp16_local_atomic_load_store feature.
> C atomic_load(volatile A *object)
> C atomic_load_explicit(volatile A *object, memory_order order)
> C atomic_load_explicit(volatile A *object,
>     memory_order order, memory_scope scope)
> ```
> 
> Is this sufficient? Or is the request to have an explicit `__opencl_c_ext_fp16_generic_atomic_load_store` feature in addition to the current "global" and "local" feature macros?
> 
> @AnastasiaStulova is this a topic we should discuss in an upcoming tooling call?
> 

Sure, I have added it to the agenda for today's call.

svenvh:
> > Could we provide a dedicated feature macro for generic?
> 
> I'm absolutely open to suggestions for improvement here. Note that we already have a feature macro for the generic address space - `__opencl_c_generic_address_space`. My intent is that this would influence which of these built-ins are available - this is one of "the requirements described in the OpenCL 3.0 specification".
> 
> [...]
>
> Is this sufficient? Or is the request to have an explicit `__opencl_c_ext_fp16_generic_atomic_load_store` feature in addition to the current "global" and "local" feature macros?

With the current proposal, I believe the full guard for those generic builtins would be something like
```
#if cl_ext_float_atomics && __opencl_c_generic_address_space && \
        (__opencl_c_ext_fp16_global_atomic_load_store || __opencl_c_ext_fp16_local_atomic_load_store)
```

As far as I can see, this would be the first ever set of builtins that have to be guarded by a "conjunction of disjunctions" instead of a plain conjunction.  Supporting that in Clang's tablegen-driven OpenCL builtin handling would require additional complexity.  So having an explicit `__opencl_c_ext_fp16_generic_atomic_load_store` feature macro would simplify tooling.

Now Clang could define the generic feature macros itself, and that's the kind of work-around I envision to avoid overcomplicating the tablegen machinery:
```
#define __opencl_c_ext_fp16_generic_atomic_load_store \
        __opencl_c_generic_address_space && \
        (__opencl_c_ext_fp16_global_atomic_load_store || __opencl_c_ext_fp16_local_atomic_load_store)
#if cl_ext_float_atomics && __opencl_c_ext_fp16_generic_atomic_load_store
```
But if we end up defining these generic feature macros in Clang anyway, we might just as well add them to the spec?

bashbaug:
> As far as I can see, this would be the first ever set of builtins that have to be guarded by a "conjunction of disjunctions" instead of a plain conjunction.

Thinking out loud: perhaps this is a bug and it should be a pure conjunction instead:

```c
#if cl_ext_float_atomics && __opencl_c_generic_address_space && \
        __opencl_c_ext_fp16_global_atomic_load_store && __opencl_c_ext_fp16_local_atomic_load_store
```

In other words, the generic address space built-ins are only supported if the generic address space AND the global atomics AND the local atomics are supported.  Thoughts?  Would any implementers want to support one of the named address spaces, but not the other, and still support the generic address space built-ins?

The main reason I can think of why we shouldn't add the generic feature macro in the spec is to constrain the set of options that are available.  I suppose we could still do this in the spec and CTS tests (via requirements like "if feature X and/or feature Y are available then feature Z must be available also), but it'd be nice to avoid this complexity if possible.

bashbaug:
Discussed in the August 19th tooling teleconference.  Changing from an OR to and AND for the generic address space built-ins will work for tools (Clang) and we believe it will work for implementers as well.

Action items:

* Implementers: Please confirm that this behavior is OK!  To recap, the affected scenario is where the generic address space is supported and either the global or local floating-point atomics are supported, but not both.  Before this change the generic address space built-ins would have been supported in this scenario, but after this change the generic address space built-ins will require both the global and local capabilities.
* Me: Update the spec to change the OR to AND for the generic addres space built-in functions.
* Me: Clarify when a generic address space pointer is accepted at the SPIR-V level (this is currently missing, or at least ambiguous).

bashbaug:
I think I've made the generic address space updates as described above with https://github.com/KhronosGroup/OpenCL-Docs/pull/552/commits/feb11cb84ceec1cf8bde3a573944a1f631bafe5d.  Please take a look - thanks!

haonanya:
> * Implementers: Please confirm that this behavior is OK!  To recap, the affected scenario is where the generic address space is supported and either the global or local floating-point atomics are supported, but not both.  Before this change the generic address space built-ins would have been supported in this scenario, but after this change the generic address space built-ins will require both the global and local capabilities.

This looks good, thanks very much.

alycm (approved):
LGTM, this all seems very well thought through.

extensions/cl_extension_template.asciidoc
```diff
@@ -461,6 +492,7 @@ best not to renumber issues, either.
 | 0.4.0   | 2019-10-25 | Kévin Petit  | Added conformance tests section.
 | 0.5.0   | 2020-02-13 | Kévin Petit  | Syntax fixes + added SPIR-V environment section.
 | 0.6.0   | 2020-04-20 | Alastair Murray | Use naming conventions in the new type example.
+| 0.7.0   | 2021-08-09 | Ben Ashbaugh | Added OpenCL C feature names section
```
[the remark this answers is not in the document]
bashbaug: Thanks, I updated this, and bumped the modification date to today so things stay in chronological order.

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,729 @@
...
+  * When `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT`, the _Pointer_ operand may be a pointer to the *CrossWorkGroup* _Storage Class_.
+  * When `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT`, the _Pointer_ operand may be a pointer to the *Workgroup* _Storage Class_.
+  * When `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES` includes `CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT` and `CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT`, and the *GenericPointer* capability is supported and declared, the _Pointer_ operand may be a pointer to the *Generic* _Storage Class_.
+
+If the OpenCL environment supports the extension `cl_ext_float_atomics` and the {CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT}, {CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT}, or {CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT} bitfields include `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extensions `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float16_add`.
```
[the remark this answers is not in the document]
bashbaug: No, this wasn't the intent, good catch.

I've broken this into two statements now.  I believe this is correct:

> If the OpenCL environment supports the extension `cl_ext_float_atomics` and the `CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT`, or `CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT` bitfields include `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extension `SPV_EXT_shader_atomic_float_add`.
> If the OpenCL environment supports the extension `cl_ext_float_atomics` and the `CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT` bitfield includes `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT`, then the environment must accept modules that declare use of the extensions `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float16_add`.
> Additionally...

It'd probably be OK to say that the module must accept the `OpExtension` without the related `OpCapability` but that would be a little weird and unnecessary.

Note that `SPV_EXT_shader_atomic_float16_add` extends and requires `SPV_EXT_shader_atomic_float_add` so `SPV_EXT_shader_atomic_float_add` must also be accepted for fp16 atomic adds.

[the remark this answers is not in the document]
alycm: > I've broken this into two statements now. I believe this is correct:

Yes, that looks good.

> Note that SPV_EXT_shader_atomic_float16_add extends and requires SPV_EXT_shader_atomic_float_add so SPV_EXT_shader_atomic_float_add must also be accepted for fp16 atomic adds.

Yes, I see, that's consistent with the SPIR-V extensions.

extensions/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,729 @@
...
+      | Describes the floating-point atomic operations supported by the device.
+        This is a bit-field that describes a combination of the following values:
+
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_LOAD_STORE_EXT` - Can perform floating-point load, store, and exchange atomic operations in global memory. +
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` - Can perform floating-point addition and subtraction atomic operations in global memory. +
+        `CL_DEVICE_GLOBAL_FP_ATOMIC_MIN_MAX_EXT` - Can perform floating-point min and max atomic operations in global memory.
+
+        `CL_DEVICE_LOCAL_FP_ATOMIC_LOAD_STORE_EXT` - Can perform floating-point load, store, and exchange atomic operations in local memory. +
+        `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT` - Can perform floating-point addition and subtraction atomic operations in local memory. +
+        `CL_DEVICE_LOCAL_FP_ATOMIC_MIN_MAX_EXT` - Can perform floating-point min and max atomic operations in local memory.
```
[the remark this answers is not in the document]
alycm: Good point, I didn't intend for all queries to be be `0` at once.  Your suggested sentence makes sense.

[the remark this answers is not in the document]
bashbaug: After thinking about this a bit more I've taken your suggestion as-is.  I think it would be a little strange to support the extension and return zero for all of the queries but perhaps there's a reason for doing so.  Regardless, it seems like an extreme corner case we're unlikely to encounter in practice.

alycm:
I'm going to merge this now.

I believe all concrete review feedback has been addressed.  This has been discussed on working group and tooling calls, and it appears that at least SPIR-V work has already started.  The last comment that wasn't from myself or the author was four months ago.

There are some open discussion points on FP atomics in general, but as far as I understand they discuss issues which would equally apply to `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float_min_max`.

PR #555, opened 2021-02-09
Specification for cl_arm_scheduling_controls v0.3.0

kpet:
- Add register allocation control
- Introduce dedicated type for capabilities and add all definitions to XML

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -217,6 +217,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_uint</type>          <name>cl_layer_info</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_layer_api_version</name>;</type>
         <type category="define">typedef struct _cl_icd_dispatch       <name>cl_icd_dispatch</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_scheduling_controls_capabilities_arm</name>;</type>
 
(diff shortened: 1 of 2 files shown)


xml/cl.xml
```diff
@@ -5831,6 +5842,16 @@ server's OpenCL/api-docs repository.
             </require>
         </extension>
         <extension name="cl_arm_scheduling_controls" supported="opencl">
+            <require comment="Types">
```
bashbaug, line 5845: I think `CL_DEVICE_SUPPORTED_REGISTER_ALLOCATIONS_ARM` should be included somewhere here, perhaps in the `cl_device_info` block?

We can merge these changes as-is since this won't break anything right now, but it will be needed when we start generating extension headers (hopefully soon!).

bashbaug (approved):
Just one comment, feel free to merge as-is if this is time-critical.  Thanks!

xml/cl.xml
bashbaug: It looks like we're also missing `CL_QUEUE_DEFERRED_FLUSH_ARM`.

kpet: Good catch! I've now added them both.

Another advantage of generation is that it will provide us with stronger guarantees on the correctness/completeness of the XML.

bashbaug: Yeah, I've been (manually) generating the headers each time to ensure the XML and header changes are consistent.  It's good motivation to get the header generation across the finish line so I can stop doing this.  😃

PR #558, opened 2021-02-09
Clarify semantics for CL_KERNEL_ARG_TYPE_NAME

StuartDBrady:
This changes the specification of CL_KERNEL_ARG_TYPE_NAME in two respects:
* It specifies substitutions for signed integer scalar types matching the already specified substitutions for unsigned scalar types. (For `short`, `int`, and `long` the newly specified substitutions match the existing behavior of Clang. For `char`, the behavior of Clang was inconsistent to this, which will be addressed by https://reviews.llvm.org/D96161.)
* It specifies further canonicalization that had been omitted from the specification, for subscript-based pointer syntax and superfluous type specifiers and parentheses. (This canonicalization is consistent with the existing behavior of Clang.)

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -7892,9 +7892,16 @@ include::{generated}/api/version-notes/CL_KERNEL_ARG_TYPE_NAME.asciidoc[]
         _arg_index_.
         The type name returned will be the argument type name as it was
         declared with any whitespace removed.
-        If argument type name is an unsigned scalar type (i.e. unsigned
-        char, unsigned short, unsigned int, unsigned long), uchar, ushort,
-        uint and ulong will be returned.
+        If the argument type name includes an unsigned scalar type (e.g.
+        `unsigned char`, `unsigned short`, `unsigned int`, `unsigned long`),
+        `uchar`, `ushort`, `uint` and `ulong` will be substituted.
+        If the argument type name includes a signed scalar type with a
+        `signed` type specifier (e.g. `signed char`, `signed short`,
+        `signed int`, `signed long`), `char`, `short`, `int` and `long`
+        will be substituted, with all other type specifiers removed.
+        Any subscript denoting a pointer type (e.g. `[]`) will be replaced
+        with `*`, including subscripts containing an unused integer value.
+        Any superfluous parentheses will be removed.
         The argument type name returned does not include any type
         qualifiers.
 | {CL_KERNEL_ARG_TYPE_QUALIFIER_anchor}


alycm:
I did a tiny bit of checking with two implementations, including Intel GPU.   Both implementations return "int" for "signed int", but both return "signed char" for "signed char".  Based on your comment these shipping implementations are matching Clang's behaviour as it was before the linked patch.

I think that in C99 `char`, `signed char` and `unsigned char` are three distinct types.   `char` being special is this regard, as for `short`, `int` and `long` there are only two distinct types.  Though I can't find explicit wording and maybe I'm getting mixed up with C++.  If I'm correct then OpenCL C doesn't override that, though unlike C99 `char` is explicitly signed.  Is that the reason for Clang's behaviour pre-patch?

Regardless, the `char` part of this change is not consistent with shipping implementations.  So if we're clarifying this we need to be certain, as it effectively means they have a bug.

bashbaug:
I don't have a very strong opinion regarding this change, although I do observe that this feature has been around since OpenCL 1.2 so I would prefer not to update the spec in a way that could break (potentially very many) shipping implementations.

I'm especially nervous since it sounds like this spec change is inconsistent with current versions of Clang.  Could we perhaps weaken the statement from a "must" to a "should" so shipping implementations aren't broken after this update?

> If the argument type name includes a signed scalar type with a `signed` type specifier (e.g. `signed char`, `signed short`, `signed int`, `signed long`), `char`, `short`, `int` and `long` **should** be substituted (... but may include the signed specifier?) ~~with all other type specifiers removed.~~

I'm open to other options.

StuartDBrady:
> I did a tiny bit of checking with two implementations, including Intel GPU. Both implementations return "int" for "signed int", but both return "signed char" for "signed char". Based on your comment these shipping implementations are matching Clang's behaviour as it was before the linked patch.

The same is true for Arm's implementation.

> I think that in C99 `char`, `signed char` and `unsigned char` are three distinct types. `char` being special is this regard, as for `short`, `int` and `long` there are only two distinct types. Though I can't find explicit wording and maybe I'm getting mixed up with C++. If I'm correct then OpenCL C doesn't override that, though unlike C99 `char` is explicitly signed. Is that the reason for Clang's behaviour pre-patch?

C99 § 6.2.5 Types

> 15. The three types **char**, **signed char**, and **unsigned char** are collectively called the _character types_. The implementation shall define **char** to have the same range, representation, and behavior as either **signed char** or **unsigned char**. <sup>35)</sup>

> 35) [...] Irrespective of the choice made, **char** is a separate type from the other two and is not compatible with either.

C++ appears to maintain this same distinction. As such, C/C++ frontends are required to treat these as three distinct types.

There is no mention of the `signed` qualifier in the OpenCL C specification whatsoever. However, there is nothing forbidding its use. Although one would expect `signed char` and `char` to be compatible types in OpenCL C, Clang treats them as distinct types, even for OpenCL C.

> Regardless, the `char` part of this change is not consistent with shipping implementations. So if we're clarifying this we need to be certain, as it effectively means they have a bug.

They do have a bug: the specification states that the type name returned will have "any whitespace removed", and `signed char` contains whitespace.

alycm:
Thank you for finding the quotes @StuartDBrady, it was that crucial second part that I couldn't find.

> There is no mention of the signed qualifier in the OpenCL C specification whatsoever. However, there is nothing forbidding its use. Although one would expect signed char and char to be compatible types in OpenCL C, Clang treats them as distinct types, even for OpenCL C.

Though I 100% realize that these rules for `char` aren't necessary for OpenCL C, nothing in the specification ever overrode them.  So Clang's historical behaviour of following C99 seems correct, because it had no reason not to.

If so, is substituting `char` for a `signed char` in the API consistent with OpenCL C?  They're not the same type.  Unlike when substituting `int` for a `signed int`, for example.  Perhaps we could "correct" OpenCL C, but consideration for already shipped implementations should be made, though this is the only observable difference that I can think of right now.

> Could we perhaps weaken the statement from a "must" to a "should" so shipping implementations aren't broken after this update?

@bashbaug Suggestion may be a good compromise.   I don't have any actual end-user concerns about the substitution.

StuartDBrady:
> Though I 100% realize that these rules for `char` aren't necessary for OpenCL C, nothing in the specification ever overrode them. So Clang's historical behaviour of following C99 seems correct, because it had no reason not to.

Clang's behavior is correct because it has become the de facto standard here, not because of any implicit deference to the C99 specification.

> If so, is substituting `char` for a `signed char` in the API consistent with OpenCL C? They're not the same type. Unlike when substituting `int` for a `signed int`, for example. Perhaps we could "correct" OpenCL C, but consideration for already shipped implementations should be made, though this is the only observable difference that I can think of right now.

It is consistent, as these are the same type within the runtime. This is, after all, a runtime API call.

> @bashbaug Suggestion may be a good compromise. I don't have any actual end-user concerns about the substitution.

I would prefer that we fix the bug.

AnastasiaStulova:
>If so, is substituting char for a signed char in the API consistent with OpenCL C? They're not the same type. Unlike when substituting int for a signed int, for example. Perhaps we could "correct" OpenCL C, but consideration for already shipped implementations should be made, though this is the only observable difference that I can think of right now.

Just to understand - how do you suggest to "correct" OpenCL C? Do you suggest mapping `char` to `signed char`?

I think perhaps we should just add to this description that `char` can be printed as either `char` or `uchar` depending on the target setup?

StuartDBrady:
> > If so, is substituting char for a signed char in the API consistent with OpenCL C? They're not the same type. Unlike when substituting int for a signed int, for example. Perhaps we could "correct" OpenCL C, but consideration for already shipped implementations should be made, though this is the only observable difference that I can think of right now.
> 
> Just to understand - how do you suggest to "correct" OpenCL C? Do you suggest mapping `char` to `signed char`?

I suggest mapping to `char`. In 99.999% of cases where a signed char type is needed for a kernel argument, `char` will be used rather than `signed char`, anyway.

> I think perhaps we should just add to this description that `char` can be printed as either `char` or `uchar` depending on the target setup?

`char` can never be printed as `uchar`.  Without the `signed` type specifier, `char` is always signed in OpenCL C. However, the signed `char` type is not compatible with `signed char` type as per the footnote I quoted previously.

AnastasiaStulova:
> > > If so, is substituting char for a signed char in the API consistent with OpenCL C? They're not the same type. Unlike when substituting int for a signed int, for example. Perhaps we could "correct" OpenCL C, but consideration for already shipped implementations should be made, though this is the only observable difference that I can think of right now.
> > 
> > 
> > Just to understand - how do you suggest to "correct" OpenCL C? Do you suggest mapping `char` to `signed char`?
> 
> I suggest mapping to `char`. In 99.999% of cases where a signed char type is needed for a kernel argument, `char` will be used rather than `signed char`, anyway.
> 
> > I think perhaps we should just add to this description that `char` can be printed as either `char` or `uchar` depending on the target setup?
> 
> `char` can never be printed as `uchar`. Without the `signed` type specifier, `char` is always signed in OpenCL C. 

Ok I see, `char` is indeed signed in OpenCL C looking at Table 6.3.1. Built-in Scalar Data Types.

> However, the signed `char` type is not compatible with `signed char` type as per the footnote I quoted previously.

Well in C99 yes because it is not fixed to be neither signed nor unsigned, but for OpenCL it is fixed and therefore based on the following statement

_"The implementation shall define char to have the same range, representation, and behavior as either signed char or unsigned char."_

we could or probably even should define `char` and `signed char` as the same type in OpenCL C. The problem is that signed counterparts seem to be completely missing in the table I cited above. So not sure what this all means. I now start to think that with current spec wording `signed` types are only accepted as an implementation defined feature and should perhaps not be covered by OpenCL API spec. Or if we do add them in OpenCL API spec we should also add them to OpenCL C spec. Otherwise, we will introduce inconsistency?

StuartDBrady:
> Ok I see, `char` is indeed signed in OpenCL C looking at Table 6.3.1. Built-in Scalar Data Types.

Indeed.

> > However, the signed `char` type is not compatible with `signed char` type as per the footnote I quoted previously.
> 
> Well in C99 yes because it is not fixed to be neither signed nor unsigned, but for OpenCL it is fixed and therefore based on the following statement
> 
> _"The implementation shall define char to have the same range, representation, and behavior as either signed char or unsigned char."_
> 
> we could or probably even should define `char` and `signed char` as the same type in OpenCL C.

`char` _should_ have been defined this way, but was not, so we're left with Clang's behavior and the language in the C99 spec ("Irrespective of the choice made...").

This "incompatibility" between a signed `char` and `signed char` is merely a technicality, though. In practice, as you quote, they have the same range, representation and behavior. They are fundamentally the same type.

> The problem is that signed counterparts seem to be completely missing in the table I cited above. So not sure what this all means. I now start to think that with current spec wording `signed` types are only accepted as an implementation defined feature and should perhaps not be covered by OpenCL API spec. Or if we do add them in OpenCL API spec we should also add them to OpenCL C spec. Otherwise, we will introduce inconsistency?

I'm not sure about `signed int`, but I would say that the OpenCL C spec simply didn't address the issue for `signed char`. The intent of the OpenCL C spec seems to be that there should be _two_ char types, not three, and it would seem a poor choice in my opinion to entrench implementation details that go _against_ this intent into the specification at this point.

AnastasiaStulova:
> I'm not sure about signed int, but I would say that the OpenCL C spec simply didn't address the issue for signed char. The intent of the OpenCL C spec seems to be that there should be two char types, not three, and it would seem a poor choice in my opinion to entrench implementation details that go against this intent into the specification at this point.

Ok, if we don't plan to add signed prefixes into OpenCL C, we should probably leave them out in the API spec too otherwise it will add a conflict i.e. we will end up promoting non-conformant behavior. However, the wording around pointer/array type etc would still be valuable.

StuartDBrady:
> > I'm not sure about signed int, but I would say that the OpenCL C spec simply didn't address the issue for signed char. The intent of the OpenCL C spec seems to be that there should be two char types, not three, and it would seem a poor choice in my opinion to entrench implementation details that go against this intent into the specification at this point.
> 
> Ok, if we don't plan to add signed prefixes into OpenCL C, we should probably leave them out in the API spec too otherwise it will add a conflict i.e. we will end up promoting non-conformant behavior. However, the wording around pointer/array type etc would still be valuable.

I would agree that we can say that it is for implementations to determine how they handle `signed char`. I do not believe this to be an _intentional_ deference to the C99 spec, and Clang does not match the most likely intent (regardless of whether existence of the `signed` type specifier was even known or considered when the OpenCL spec was first drafted). The OpenCL C spec lists supported types, strongly implying that other types are not supported. However, the `signed` type specifier is used in practice in OpenCL C code.

We would, therefore, _not_ test `signed char` as part of the OpenCL CTS, although in my view we should still test `signed int`, `signed long` and `signed short` and improve spec wording to _explicitly_ permit these. As these are _not_ distinct types, the use of the `signed` type specifier for them _is_ supported through deference to the C99 spec, whereas `signed char` would be a distinct type not mentioned in the OpenCL C spec.

AnastasiaStulova:
> I would agree that we can say that it is for implementations to determine how they handle `signed char`. I do not believe this to be an _intentional_ deference to the C99 spec, and Clang does not match the most likely intent (regardless of whether existence of the `signed` type specifier was even known or considered when the OpenCL spec was first drafted). The OpenCL C spec lists supported types, strongly implying that other types are not supported. However, the `signed` type specifier is used in practice in OpenCL C code.
> 
> We would, therefore, _not_ test `signed char` as part of the OpenCL CTS, although in my view we should still test `signed int`, `signed long` and `signed short` and improve spec wording to _explicitly_ permit these. As these are _not_ distinct types, the use of the `signed` type specifier for them _is_ supported through deference to the C99 spec, whereas `signed char` would be a distinct type not mentioned in the OpenCL C spec.

I agree CTS should take care of conformant behavior and implementation-defined behavior should be tested elsewhere. I wouldn't object however if we add a `signed` type in the spec, although we might only be able to do so for the next version. I would suggest to split it into a separate issue.

StuartDBrady:
> I agree CTS should take care of conformant behavior and implementation-defined behavior should be tested elsewhere. I wouldn't object however if we add a `signed` type in the spec, although we might only be able to do so for the next version. I would suggest to split it into a separate issue.

This seems reasonable to me. I'll amend this change to make the wording for `signed` be specific to `int`, `short` and `long`, given that `signed char` is a distinct type that is _not_ supported according to the specification, even though Clang supports it and there is OpenCL C code that uses it, too. (`signed int`, `signed short` and `signed long` are a different matter, as although the OpenCL C specification doesn't mention the `signed` type specifier, it doesn't need to: given that `signed int` and `int` are the _same_ type, and that OpenCL C supports the `int` type, it therefore supports `signed int`. OpenCL C does not rescind the parts of the C specification that describe the `signed` type specifier and its effect. It merely omits `signed char` from its list of supported types.)

A future version of the language should define `signed char` as an alias for `char`, countermanding C99's definition of the type. With this, the API spec would be updated so that `CL_KERNEL_ARG_TYPE_NAME` would be consistent between `signed char` and `signed int`.

(A general note, BTW, to avoid possible confusion for the wider audience that might be reading this discussion: `signed` is not strictly a prefix: it is a type specifier, just as `long` and `short` are, and just as `int` is. Type specifiers can appear in any order, meaning that `int short signed` is perfectly valid both in C and in OpenCL C and is equivalent to `short int`, and `signed` on its own is also perfectly valid in both languages, being equivalent to just `int`. For `int`, `short` and `long`, the `signed` type specifier is usually omitted. For `long` and `short`, the `int` type specifier is usually omitted, also. There is a distinction between `signed char` and `char`, as we have been discussing, although it is quite uncommon now for C implementations to define the plain `char` type as an unsigned type, and it is explicitly defined as signed in OpenCL C, as we have been discussing.)

AnastasiaStulova:
> This seems reasonable to me. I'll amend this change to make the wording for `signed` be specific to `int`, `short` and `long`, given that `signed char` is a distinct type that is _not_ supported according to the specification, even though Clang supports it and there is OpenCL C code that uses it, too.



> This seems reasonable to me. I'll amend this change to make the wording for `signed` be specific to `int`, `short` and `long`, given that `signed char` is a distinct type that is _not_ supported according to the specification, even though Clang supports it and there is OpenCL C code that uses it, too.
> 
> (Some further clarification for anyone who would like it, as I was fairly concise, previously: `signed int`, `signed short` and `signed long` are a different matter, as although the OpenCL C specification doesn't mention the `signed` type specifier, it doesn't need to: given that `signed int` and `int` are the _same_ type, and that OpenCL C supports the `int` type, it therefore supports `signed int`. OpenCL C does not rescind the parts of the C specification that describe the `signed` type specifier and its effect. It merely omits `signed char` from its list of supported types.)

Signed prefixes are not part of OpenCL C so I think it is better to leave those out completely and not selectively. This will help to eliminate further confusion. If we agree to add them in OpenCL C though then we can also add clarification about the arg info spelling. We should probably create an issue covering both specs.

StuartDBrady:
> Signed prefixes are not part of OpenCL C so I think it is better to leave those out completely and not selectively. This will help to eliminate further confusion. If we agree to add them in OpenCL C though then we can also add clarification about the arg info spelling. We should probably create an issue covering both specs.

The OpenCL C spec does not, for instance, mention empty statements, but I would still expect them to work. The `signed` type specifier is already part of OpenCL C. Clang supports it and there is OpenCL C code that relies upon it. It may be worth clarifying this in the OpenCL C spec but it is too late to change the OpenCL C spec to list the `signed` type specifier as unsupported.

The question for CL_KERNEL_ARG_TYPE_NAME is a different one, though. This is a question of whether the spec was mistaken in (unintentionally) requiring that a `signed int` kernel argument be returned as `signedint` (without whitespace) or whether Clang's behavior is correct (or whether there is some third option).

Given the OpenCL C spec does nothing to forbid the use of the `signed` qualifier, the current wording in the API spec makes Clang's long-established behavior of returning `int`, `long` and `short` for these types non-conformant.  It would be good to resolve this.

AnastasiaStulova:
> The question for CL_KERNEL_ARG_TYPE_NAME is a different one, though. This is a question of whether the spec was mistaken in (unintentionally) requiring that a `signed int` kernel argument be returned as `signedint` (without whitespace) or whether Clang's behavior is correct (or whether there is some third option).

AnastasiaStulova:
> The OpenCL C spec does not, for instance, mention empty statements, but I would still expect them to work. The signed type specifier is already part of OpenCL C.

Can you refer to the spec please?

> Clang supports it and there is OpenCL C code that relies upon it. It may be worth clarifying this in the OpenCL C spec but it is too late to change the OpenCL C spec to list the signed type specifier as unsupported.

If you think it is too late to clarify this in OpenCL C why is it not too late for the API? I don't see what good would it make if OpenCL C doesn't include signed prefixes but OpenCL API does. If anything this will indirectly mean that signed are supported then OpenCL C should just say it straight away.

> The question for CL_KERNEL_ARG_TYPE_NAME is a different one, though. This is a question of whether the spec was mistaken in (unintentionally) requiring that a signed int kernel argument be returned as signedint (without whitespace) or whether Clang's behavior is correct (or whether there is some third option).

This doesn't apply to signed because the OpenCL C spec doesn't include this spelling. Right now the position is very clear to me. Clang support it as an implementation-defined feature and it, therefore, defines what spelling it returns for the arg info metadata too. It is not goverened by OpenCL specifications at all.

StuartDBrady:
> > The OpenCL C spec does not, for instance, mention empty statements, but I would still expect them to work. The signed type specifier is already part of OpenCL C.
> 
> Can you refer to the spec please?
>

Sure, from _Chapter 6, The OpenCL C Programming Language_:

> The OpenCL C programming language (also referred to as OpenCL C) is based on the ISO/IEC 9899:1999 Programming languages - C specification (also referred to as the C99 specification, or just C99), with extensions and restrictions to support parallel kernels. [...]
> 
> This document describes the modifications and restrictions to C99 and C11 in OpenCL C. Please refer to the C99 specification for a detailed description of the language grammar.

The OpenCL C spec does not purport to exhaustively list all allowed type specifiers. Likewise it does not restrict the use of `long int` or `short int` as opposed to simply `long` or `short`, which is also relevant to this discussion.
 
> If you think it is too late to clarify this in OpenCL C why is it not too late for the API? I don't see what good would it make if OpenCL C doesn't include signed prefixes but OpenCL API does. If anything this will indirectly mean that signed are supported then OpenCL C should just say it straight away.

I'm not sure if you misread my comment.  I said that it is too late to restrict the use of `signed` in the OpenCL C specification in such a manner that it would not need to be considered in the runtime specification.  The `signed` type specifier is part of OpenCL C (coming via C99) and does need to be considered in the runtime specification.  It is not too late to clarify the OpenCL C specification to state that the `signed` type specifier is supported but it is too late to change OpenCL C so that `signed` is restricted.

> This doesn't apply to signed because the OpenCL C spec doesn't include this spelling. Right now the position is very clear to me. Clang support it as an implementation-defined feature and it, therefore, defines what spelling it returns for the arg info metadata too. It is not goverened by OpenCL specifications at all.

The OpenCL C specification does not need to include this spelling.  The ISO C specification includes it.  The OpenCL C specification describes types that are allowed, not spellings.

AnastasiaStulova:
> > > The OpenCL C spec does not, for instance, mention empty statements, but I would still expect them to work. The signed type specifier is already part of OpenCL C.
> > 
> > 
> > Can you refer to the spec please?
> 
> Sure, from _Chapter 6, The OpenCL C Programming Language_:
> 
> > The OpenCL C programming language (also referred to as OpenCL C) is based on the ISO/IEC 9899:1999 Programming languages - C specification (also referred to as the C99 specification, or just C99), with extensions and restrictions to support parallel kernels. [...]
> > This document describes the modifications and restrictions to C99 and C11 in OpenCL C. Please refer to the C99 specification for a detailed description of the language grammar.
> 
> The OpenCL C spec does not purport to exhaustively list all allowed type specifiers. Likewise it does not restrict the use of `long int` or `short int` as opposed to simply `long` or `short`, which is also relevant to this discussion.
> 
> > If you think it is too late to clarify this in OpenCL C why is it not too late for the API? I don't see what good would it make if OpenCL C doesn't include signed prefixes but OpenCL API does. If anything this will indirectly mean that signed are supported then OpenCL C should just say it straight away.
> 
> I'm not sure if you misread my comment. I said that it is too late to restrict the use of `signed` in the OpenCL C specification in such a manner that it would not need to be considered in the runtime specification. The `signed` type specifier is part of OpenCL C (coming via C99) and does need to be considered in the runtime specification. It is not too late to clarify the OpenCL C specification to state that the `signed` type specifier is supported but it is too late to change OpenCL C so that `signed` is restricted.
> 
> > This doesn't apply to signed because the OpenCL C spec doesn't include this spelling. Right now the position is very clear to me. Clang support it as an implementation-defined feature and it, therefore, defines what spelling it returns for the arg info metadata too. It is not goverened by OpenCL specifications at all.
> 
> The OpenCL C specification does not need to include this spelling. The ISO C specification includes it. The OpenCL C specification describes types that are allowed, not spellings.

Ok, we don't seem to interpret OpenCL C spec in the same way. My interpretation is that supported data types are those listed in the section 6.3. as there is the following wording:
`The following data types are supported.`

So I assume the other spellings are not supported. But perhaps the wording is not really precise. This is why I think the use of signed and other spellings should be further clarified before we can confidently say whether it is supposed/not supposed to be used.

StuartDBrady:
> Ok, we don't seem to interpret OpenCL C spec in the same way. My interpretation is that supported data types are those listed in the section 6.3. as there is the following wording:
> `The following data types are supported.`
>
> So I assume the other spellings are not supported. But perhaps the wording is not really precise. This is why I think the use of signed and other spellings should be further clarified before we can confidently say whether it is supposed/not supposed to be used.

I agree that section 6.3 lists the supported data types.  I agree that the OpenCL specification is not _explicit_ in this matter, and that it should be clarified, although as I have said, if the original intent was that the `signed` type specifier should not be supported, this has not been properly conveyed in the specification which defers to the C99 specification.  I will raise an issue for this and amend my change to exclude any mention of the `signed` type specifier for now.

However, I will keep the part that addresses the use of `long int` and `short int`, assuming that is not also in dispute.

AnastasiaStulova:
> I agree that section 6.3 lists the supported data types. I agree that the OpenCL specification is not explicit in this matter, and that it should be clarified, although as I have said, if the original intent was that the signed type specifier should not be supported, this has not been properly conveyed in the specification which defers to the C99 specification. I will raise an issue for this and amend my change to exclude any mention of the signed type specifier for now.

Makes sense - I would add if we do clarify that `signed` prefix is intended to be supported we should also clarify `char` and `signed char` because I don't see what different type they could have been mapped to?

> However, I will keep the part that addresses the use of long int and short int, assuming that is not also in dispute.

I think those spellings don't appear in OpenCL C either? I guess they should be threat them the same as `signed` prefixes. Potentially we can follow up on those in the same issue?

AnastasiaStulova:
> Potentially we can follow up on those in the same issue?

I mean the same as for `signed`. :)

StuartDBrady:
> I think those spellings don't appear in OpenCL C either? I guess they should be threat them the same as `signed` prefixes. Potentially we can follow up on those in the same issue?

It seems unthinkable to me that `long int` and `short int` are not equivalent to `long` and `short`.

AnastasiaStulova:
> > I think those spellings don't appear in OpenCL C either? I guess they should be threat them the same as `signed` prefixes. Potentially we can follow up on those in the same issue?
> 
> It seems unthinkable to me that `long int` and `short int` are not equivalent to `long` and `short`.

Well, the issue is not what they are mapped to but whether the type spelling is to be accepted by the conformant implementations. If not Clang docs would be better fit for such documentation. And as a matter of fact those are not even present in this PR at all. But I don agree that if we clarify signed spelling we should clarify all others too.

bashbaug:
This PR appears to have gotten stuck a couple of months ago.  What do we need to do to get it un-stuck?

alycm:
Some of the comments above suggest that type specifiers not explicitly in [OpenCL C 6.3.1](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-scalar-data-types), e.g. `short int`, maybe aren't valid OpenCL C.  I don't think anyone is proposing to change anything, but maybe that discussion should be a new issue?  Hopefully the outcome of that can be more explicit language in the OpenCL C spec on which specifiers are allowed.

Otherwise, for the `signed char` case I can't think of a single OpenCL C program where it matters whether `signed char` and signed `char` are the same type vs equivalent types.  The only affect this has is on `CL_KERNEL_ARG_TYPE_NAME`.  Shipping implementations will report `signed char` for arguments of type `signed char` (but `int` for arguments of type `signed int`), indicating that types are normalized to eliminate whitespace, but that `signed char` can't be normalized to a whitespace-less form.

Either we

1. Declare that `CL_KERNEL_ARG_TYPE_NAME ` **will** drop `signed` in all cases, without making a statement on OpenCL C (the original change).
2. Declare that `CL_KERNEL_ARG_TYPE_NAME` **should** drop `signed` (Ben's suggestion), without making a statement on OpenCL C.
3. Drop or refine the "whitespace removed" wording.
4. Declare that `char` and `signed char` are the same type in OpenCL C (overriding C99).

Note relevant to (4) is that in C++ for OpenCL currently Clang follows C++ and treats the three char types as separate types, and thus the difference matters for things like overloading.  In the below example note that the three calls to the three overloads have three different manglings.  Thus a decision that that `signed char` and signed `char` are the same type in OpenCL C could put C++ for OpenCL in a position of having to choose OpenCL C compatibility versus C++ compatibility.  Hard to say if it will ever matter to a real program, so I may be over-stretching the point.

https://godbolt.org/z/T634oYY7T

The link uses Clang 12.0.0 instead of trunk to avoid the example changing if Clang changes, but Clang trunk behaves identically at time of writing.

PR #560, opened 2021-02-09
Move descriptions of identity value for exclusive scan

StuartDBrady:
The identify value is introduced in the description of the inclusive scan operations, despite only being referenced in the description of the exclusive scan operations.  Move its description into that of the exclusive scan operations, where it belongs.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9996,16 +9996,9 @@ The *<op>* in *work_group_reduce_<op>*, *work_group_scan_exclusive_<op>* and
 *work_group_scan_inclusive_<op>* defines the operator and can be *add*,
 *min* or *max*.
 
-The inclusive scan operation takes a binary operator *op* with an identity I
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
I'm OK with this change, but it'd be nice to more closely align the descriptions of the work-group scans and reductions with the subgroup scans and reductions eventually.

StuartDBrady:
> I'm OK with this change, but it'd be nice to more closely align the descriptions of the work-group scans and reductions with the subgroup scans and reductions eventually.

I would agree. In fact, generally I think it would be good to align the descriptions of workgroup and subgroup functions, and also things like half (cl_khr_fp16), float and double (cl_khr_fp64) functions. Of course, I prefer not to do all of this in one big PR as it would be difficult to review, would suffer from merge conflicts as other parts of the spec get changed over time, and wouldn't allow sharing of any of the workload.

bashbaug:
I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/580 so we don't lose track of the discussion above.  Merging!

PR #561, opened 2021-02-09
Add missing '*' in printf prototype

StuartDBrady:


kpet (approved):

bashbaug (approved):
LGTM, just waiting for the Travis build to complete.

alycm (approved):
LGTM, Travis is done, merging.

StuartDBrady:
The second version that I pushed which has now now been merged matches the function prototype originally used in the [cl_arm_printf extension](https://www.khronos.org/registry/OpenCL/extensions/arm/cl_arm_printf.txt).

On reflection, the following prototype may have been preferable:

> int **printf**(const __constant char *restrict _format_, ...)

This would raise questions regarding forwards compatibility of older implementations compiling code written according to the updated specification.  I would be happy to raise a separate issue for this if it is perceived that these questions can be resolved.

alycm:
@StuartDBrady The difference being the `const`?  Open an issue if you like, I've not spent any time thinking on the implications.

StuartDBrady:
> @StuartDBrady The difference being the `const`? Open an issue if you like, I've not spent any time thinking on the implications.

There was a similar issue discussed recently. The problem would be if any single implementation lacks a `const` qualifier currently.

_Using_ a const qualifier will produce a warning/error when the qualifier is implicitly stripped away when calling a function declared without it.

This would mean code conformant to the updated specification would fail to compile under such implementations.

However, if I have understood correctly, string literals are of the type _constant char *_, not _constant const char *_, and so almost all use of printf would be unaffected. The (main) exception would be where a variable is used to hold the format string.

In any case, there already exists a widespread portability hazard in this regard as numerous implementations already use `constant const char *` for their printf declarations.

If there's some chance of us addressing this, I can raise an issue. Is it conceivable that we might resolve this?

alycm:
Right, so code that looks like the following:

```c
const constant char * constant s = "Hello";

kernel void foo()
{
    printf(s);
}
```

*shouldn't* compile with the current spec definition because of the discarded qualifier, but will compile for some (most?) current implementations.  Changing the spec wouldn't break existing conformant programs, and though in theory it would be possible to now write a new program that wouldn't work on an old implementation, that seems to already be the case in practice.

StuartDBrady:
> Right, so code that looks like the following:
[...]
> _shouldn't_ compile with the current spec definition because of the discarded qualifier, but will compile for some (most?) current implementations.

More or less.

(I wouldn't go so far as to say the code _shouldn't_ compile, though - existing implementations are conformant despite their use of the qualifier, as conforming programs will compile correctly regardless of it.)

> Changing the spec wouldn't break existing conformant programs, and though in theory it would be possible to now write a new program that wouldn't work on an old implementation, that seems to already be the case in practice.

Yeah.  Arm's implementation and upstream Clang both use the qualifier.

Without the qualifier in the prototype, printf won't accept a `constant const char *` as the format string, which is (obviously) bad.

PR #562, opened 2021-02-10
Spec source for cl_intel_command_queue_families extension

bashbaug:
This PR contains the spec source and XML changes for the `cl_intel_command_queue_families` extension.

### DIFF
--- a/extensions/cl_intel_command_queue_families.asciidoc
+++ b/extensions/cl_intel_command_queue_families.asciidoc
@@ -1,0 +1,428 @@
+= cl_intel_command_queue_families
+
+// This section needs to be after the document title.
+:doctype: book
+:toc2:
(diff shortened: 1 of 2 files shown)


kpet (approved):
Interesting extension! This is something we might want to see as an EXT/KHR in the future (with some additional ideas/refinements around cross-queue synchronisation). 

Nothing jumped at me after a cursory read, approving since it's a vendor extension.

bashbaug:
> Interesting extension! This is something we might want to see as an EXT/KHR in the future (with some additional ideas/refinements around cross-queue synchronisation).

Thanks!  We're definitely open to standardizing this feature; we mostly went with a vendor extension initially for TTM reasons (and to work out any bugs).

PR #564, opened 2021-02-11
Fix use of parameter name in work_group_broadcast() decl

StuartDBrady:


StuartDBrady:
I've gone for _a_ here, rather than changing the parameter name to _x_, because we already have _local_id_x_, and so using _a_ avoids any possible confusion.

AnastasiaStulova (approved):
LGTM! Thanks

bashbaug (approved):
Works for me!

The `sub_group_broadcast` description uses `x`, but it doesn't have to match this description, and it doesn't have the _local_id_x_ issue (because subgroups are always 1-dimensional).

PR #565, opened 2021-02-11
Fix naming of layer header file.

Kerilk:
Fix naming of layer header to reflect final file name from https://github.com/KhronosGroup/OpenCL-Headers/pull/148.


bashbaug (approved):
LGTM - thanks!

PR #566, opened 2021-02-12
clarify the size of a pipe is implementation-defined

### LINKED ISSUE #563: OpenCL C: is sizeof(pipe) defined?
See related CTS discussion: https://github.com/KhronosGroup/OpenCL-CTS/pull/1056#issuecomment-776763156

The OpenCL C spec doesn't currently say what the size of a pipe is.  Should it?

FWIW, in the current Clang implementation the size of a pipe is the same as the size of a pointer: https://godbolt.org/z/8xdbWz

The safest change is to specify that the size of a pipe implementation-defined, perhaps by adding pipe objects to this paragraph:

> The behavior of applying the sizeof operator to the bool, image2d_t, image3d_t, image2d_array_t, image1d_t, image1d_buffer_t, image1d_array_t, image2d_depth_t, image2d_array_depth_t, sampler_t, queue_t, ndrange_t, clk_event_t, reserve_id_t, and event_t types is implementation-defined.

bashbaug:
Clarifies that the size of a pipe object is implementation-defined.  Fixes #563 .


AnastasiaStulova (approved):
LGTM! Thanks!

alycm (approved):
LGTM.

PR #569, opened 2021-02-16
Replace uses of CL_EXT_{PRE,SUF}FIX* with CL_API_{PRE,SUF}FIX*

kpet:
See https://github.com/KhronosGroup/OpenCL-Headers/issues/153 for the
discussion.


bashbaug (approved):
LGTM, merging to keep in sync with the generated ICD loader files.

PR #572, opened 2021-02-19
Require the cl_mipmap_filter_mode_img type for cl_img_generate_mipmap.

zfigura:
To help tools which parse the cl.xml file and conditionally output types.


bashbaug (approved):
LGTM, but it'd be great to get a +1 from @Jeremy-Kemp before merging.  Thanks!

Jeremy-Kemp (approved):
Looks good, thank you!

PR #574, opened 2021-02-22
clarify table structure for backwards compatible appendix

### LINKED ISSUE #568: Subgroups optionality clarification
We're looking for some clarification around subgroups that has come up during the process of adding subgroups to an older device that does not support IFP.

The Subgroups section in "Appendix H (OpenCL 3.0 Backwards Compatibility)" currently states:

> `clGetDeviceInfo`, passing `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS`
>
>Returns `CL_FALSE` if device does not support subgroups.

and

>`clGetDeviceInfo`, passing `CL_DEVICE_EXTENSIONS`
>
>Will not describe support for the `cl_khr_subgroups extension` if device does not support subgroups.

Our current interpretation of this is somewhat contradictory. If an implementation wished to support core subgroups but not the `cl_khr_subgroups` extension, `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` should be permitted to be `CL_FALSE` whilst still supporting subgroups.

Is our interpretation here correct - that a 3.0 implementation can support core subgroups, without IFP and without `cl_khr_subgroups`?

bashbaug:
Adds an informative note describing that the first row in the backwards compatibility tables usually describes a detection mechanism and subsequent rows usually describe implications if a feature is not supported.

Fixes #568.

### DIFF
--- a/api/appendix_h.asciidoc
+++ b/api/appendix_h.asciidoc
@@ -9,6 +9,8 @@
 OpenCL 3.0 breaks backwards compatibility with earlier versions of OpenCL by making some features that were previously required for FULL_PROFILE or EMBEDDED_PROFILE devices optional.
 This appendix describes the features that were previously required that are now optional, how to detect whether an optional feature is supported, and expected behavior when an optional feature is not supported.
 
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp (approved):
Thanks for doing this @bashbaug 

LGTM!

Issue #169, opened 2019-11-26
Update doc for `work_group_broadcast` 


mantognini:
The doc for `work_group_broadcast` is currently this:
https://github.com/KhronosGroup/OpenCL-Docs/blob/b86f597faf4829234b68fa72c7f33fa85d8c0566/OpenCL_C.txt#L8315-L8324

There's a small typo: it should _Broadcast the value of **`a`**_, not `x`.

I would suggest renaming `local_id` to `local_id_x` for explicitness.

I would also suggest stating that calling {first, second, third} overload from a NDRange with a dimension that is _not_ {1D, 2D, 3D} (respectively) results in undefined behaviour. 

It might also be worth, for completeness, mentioning that the local id should refer to an "existing" and "valid" id (i.e. to ensure that at lest one value can be returned by this builtin). Those are probably not the right words, but it should give the idea.

Together, these two last suggestions ensure **one and only one** value can be returned by this builtin.

I believe this has no impact on existing CTS tests as it only make undefined behaviour explicit and doesn't change existing expectations.

---

The reasoning is the following:
* When calling `work_group_broadcast(a, local_id_x)` from within a 2D or 3D NDRange, multiple work-items could by identified by `local_id_x`. Hence, more than one value should be returned, which is impossible/ambiguous. The same applies when calling `work_group_broadcast(a, local_id_x, local_id_y)` from a 3D NDRange.
* It would technically possible to identify a unique work-item from a 2/3D NDRange when the second/third dimensions are "1", but I don't think it's worth explicitly supporting that.
* When calling `work_group_broadcast(a, local_id_x, local_id_y)` from a 1D NDRange, no work-item will match the request (unless, maybe, if `local_id_y` is one, but again I don't think it's work explicitly supporting that), hence no value can be returned. The same applies when calling `work_group_broadcast(a, local_id_x, local_id_y, local_id_z)` from a 1D or 2D NDRange.

bashbaug:
Sorry for the slow reply, I think this issue came in around the US Thanksgiving holidays and I missed it.

> I would also suggest stating that calling {first, second, third} overload from a NDRange with a dimension that is _not_ {1D, 2D, 3D} (respectively) results in undefined behaviour.

Does it make sense instead to say that any "missing" local ID coordinates default to zero?  Otherwise I think it's very difficult to write a single kernel that works for multiple dimensions.  Remember that local IDs are zero-based, so zero should be "valid" for any dimension.

[...]

PR #577, opened 2021-03-01
Clarify behavior of work_group_broadcast

mantognini:
See #169 for the discussion behind this change.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9946,7 +9946,7 @@ NOTE: The functionality described in this section <<unified-spec, requires>>
 support for OpenCL C 2.0, or OpenCL C 3.0 or newer and the
 {opencl_c_work_group_collective_functions} feature.
 
-This section decribes built-in functions that perform collective options
+This section describes built-in functions that perform collective options
 across a work-group.
 These built-in functions must be encountered by all work-items in a
 work-group executing the kernel.
@@ -9976,8 +9976,9 @@ footnote:[{fn-double-supported}] as the type for the arguments.
     | Broadcast the value of _a_ for work-item identified by _local_id_ to
       all work-items in the work-group.
 
-      _local_id_ must be the same value for all work-items in the
-      work-group.
+      _local_id_ must be the same value for all work-items in the work-group
+      and represent a work-item in the work-group otherwise the behavior is
+      undefined.
 | gentype *work_group_reduce_<op>*(gentype _x_)
     | Return result of reduction operation specified by *<op>* for all
       values of _x_ specified by work-items in a work-group.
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
Sorry for the slow reply.

This is a definite improvement but I wonder if we can be more precise than "represent a work-item in the work-group".  Could we instead say something like: "Behavior is undefined unless _local_id_ is less than the local work-group size in the corresponding dimension"?

I'm looking for something similar to the text for `sub_group_non_uniform_broadcast` in the [extended subroups extension](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_add_a_new_section_6_15_x_subgroup_ballot_built_in_functions):

> Behavior is undefined when the value of index is not equivalent for all active work items in the subgroup.
>
>The return value is undefined if the work item with subgroup local ID equal to index is inactive or if index is greater than or equal to the size of the subgroup.

(Note, the "inactive" part is not applicable to the "uniform" `work_group_broadcast` or `sub_group_broadcast`.)

Can we clarify the description of the uniform `sub_group_broadcast` in the OpenCL C spec also?  It seems to have the same problem.

Thanks!

mantognini:
Good catch, the wording for `sub_group_non_uniform_broadcast` is indeed clearer. I'll update the PR to use a similar wording for both `work_group_broadcast` and `sub_group_broadcast`.

mantognini:
I went for two "positive" sentences (_Behavior is undefined when ..._, not _unless_) so that they can be composed with each other unambiguously. Let me know if there's anything else.

OpenCL_C.txt
```diff
@@ -11031,8 +11034,11 @@ footnote:[{fn-int64-supported}], `ulong`, `half` footnote:[{fn-half-supported}],
   _sub_group_local_id_ (value returned by *get_sub_group_local_id*) to all
   work-items in the subgroup.
 
-  _sub_group_local_id_ must be the same value for all work-items in the
-  subgroup.
+  Behaviour is undefined when the value of _sub_group_local_id_ is not
```
bashbaug, line 11037: Suggest using the spelling "behavior" instead of "behaviour" since the former is more common in the spec (and is used in the description of `work_group_broadcast` above).

bashbaug (approved):
LGTM, just one minor comment.  We can merge this as-is and I can make the spellings consistent in an editorial pass before we publish the next spec, if desired.  Thanks!

OpenCL_C.txt
mantognini: Indeed! It should be more consistent now.

bashbaug (approved):
Thanks!  Merging.

PR #582, opened 2021-03-05
[C++ for OpenCL] Improve description of restricted C++ features.

AnastasiaStulova:
Added the following restrictions that were omitted:
- references to functions;
- pointers to member functions;

Added clarification for:
- thread-safety of static locals initialization.

Added references to C++ spec sections where applicable.

### DIFF
--- a/cxx4opencl/diff2cxx.txt
+++ b/cxx4opencl/diff2cxx.txt
@@ -23,7 +23,7 @@ The following {cpp} language features are not supported:
   * Standard C++ libraries ({cpp}17 `[library]`).
 
 Thread-safe initialization of static local objects is not guaranteed ({cpp}17
-`[stmt.dcl]`) and whether an implementation provides a thread-safety is
+`[stmt.dcl]`) and whether an implementation provides thread-safety is
 indicated by the presence of `__cpp_threadsafe_static_init` predefined feature
 test macro.
 
--- a/cxx4opencl/diff2cxx.txt
+++ b/cxx4opencl/diff2cxx.txt
@@ -22,10 +22,14 @@ The following {cpp} language features are not supported:
     `[expr.new]`/`[expr.delete]`);
   * Standard C++ libraries ({cpp}17 `[library]`).
 
-Thread-safe initialization of static local objects is not guaranteed ({cpp}17
-`[stmt.dcl]`) and whether an implementation provides thread-safety is
-indicated by the presence of the `__cpp_threadsafe_static_init` feature
-test macro.
+:fn-feature-macro: footnote:[The macro belongs to the list of {cpp}20's \
+feature test macros.]
+
+Simultaneous initialization of static local objects performed by
+different work-items is not guaranteed to be free from race-condition.
+Whether an implementation provides such a guarantee is indicated by the
+presence of the `__cpp_threadsafe_static_init` feature test
+macro{fn-feature-macro}.
 
 The list above only contains extra restrictions that are not detailed in OpenCL
 C. As OpenCL restricts a number of C features, the same restrictions are
--- a/cxx4opencl/diff2cxx.txt
+++ b/cxx4opencl/diff2cxx.txt
@@ -28,6 +28,6 @@ indicated by the presence of the `__cpp_threadsafe_static_init` feature
 test macro.
 
 The list above only contains extra restrictions that are not detailed in OpenCL
-C. As OpenCL restricts a number of C features the same restrictions are
+C. As OpenCL restricts a number of C features, the same restrictions are
 inherited by C++ for OpenCL. The detailed list of C feature restrictions
 is provided in `OpenCL C v2.0 section 6.9`.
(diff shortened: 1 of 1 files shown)


cxx4opencl/diff2cxx.txt
```diff
@@ -12,8 +12,22 @@ however, there are some differences that are documented in this section.
...
+  * Pointers to class member functions (in addition to the regular non-member
+    functions that are already restricted in OpenCL C);
+  * Exceptions ({cpp}17 `[except]`);
+  * `dynamic_cast` operator ({cpp}17 `[expr.dynamic.cast]`);
+  * Non-placement `new`/`delete` operators ({cpp}17
+    `[expr.new]`/`[expr.delete]`);
+  * Standard C++ libraries ({cpp}17 `[library]`).
+
+Thread-safe initialization of static local objects is not guaranteed ({cpp}17
+`[stmt.dcl]`) and whether an implementation provides a thread-safety is
```
svenvh, line 26:
```suggestion
`[stmt.dcl]`) and whether an implementation provides thread-safety is
```

cxx4opencl/diff2cxx.txt
```diff
@@ -12,8 +12,22 @@ however, there are some differences that are documented in this section.
...
+indicated by the presence of `__cpp_threadsafe_static_init` predefined feature
```
svenvh, line 27:
```suggestion
indicated by the presence of the `__cpp_threadsafe_static_init` feature
```

cxx4opencl/diff2cxx.txt
```diff
@@ -12,8 +12,22 @@ however, there are some differences that are documented in this section.
...
+test macro.
+
+The list above only contains extra restrictions that are not detailed in OpenCL
+C. As OpenCL restricts a number of C features the same restrictions are
```
svenvh, line 31:
```suggestion
C. As OpenCL restricts a number of C features, the same restrictions are
```

svenvh (changes requested):
Just some minor nitpicks, otherwise LGTM.

svenvh (approved):
LGTM, thanks!

cxx4opencl/diff2cxx.txt
```diff
@@ -12,8 +12,22 @@ however, there are some differences that are documented in this section.
...
+  * Exceptions ({cpp}17 `[except]`);
+  * `dynamic_cast` operator ({cpp}17 `[expr.dynamic.cast]`);
+  * Non-placement `new`/`delete` operators ({cpp}17
+    `[expr.new]`/`[expr.delete]`);
+  * Standard C++ libraries ({cpp}17 `[library]`).
+
+Thread-safe initialization of static local objects is not guaranteed ({cpp}17
+`[stmt.dcl]`) and whether an implementation provides thread-safety is
+indicated by the presence of the `__cpp_threadsafe_static_init` feature
+test macro.
```
alycm, line 28: I think this is the first time that a C++ feature macro has been referenced in C++ for OpenCL, though I don't know if more from C++ spec are inherited automatically.

For OpenCL C feature test macros we have the `CL_DEVICE_OPENCL_C_FEATURES` device query.  Should a new revision of the `cl_ext_cxx_for_opencl` extension add a `CL_DEVICE_CXX_FOR_OPENCL_FEATURES` query?

It is unlikely that this can be achieved in this PR, so if there is an action to take it could just be an issue for now.

Is threadsafe the correct name here too?  Work-items aren't normally referred to as threads.  I do realize that this name comes from C++ though, so it would be easy to argue that keeping it the same is consistent.

AnastasiaStulova: > I think this is the first time that a C++ feature macro has been referenced in C++ for OpenCL, though I don't know if more from C++ spec are inherited automatically.

Yeah, it is assumed that everything that is not covered in this section is automatically inherited from C++. However, the feature test macros are not in C++17 though. So I have added a footnote to make clear that this is a C++20 functionality.

> For OpenCL C feature test macros we have the CL_DEVICE_OPENCL_C_FEATURES device query. Should a new revision of the cl_ext_cxx_for_opencl extension add a CL_DEVICE_CXX_FOR_OPENCL_FEATURES query?
> 
> It is unlikely that this can be achieved in this PR, so if there is an action to take it could just be an issue for now.
> 

We haven't ported C++ for OpenCL to OpenCL 3.0 yet, so there were no feature test macros provided so far. But I think it would indeed be good to record this for the future.

> Is threadsafe the correct name here too? Work-items aren't normally referred to as threads. I do realize that this name comes from C++ though, so it would be easy to argue that keeping it the same is consistent.

Good point. We don't have any equivalent term for work-items so I have rephrased it now to:

```
Simultaneous initialization of static local objects performed by
different work-items is not guaranteed to be free from race-condition.
Whether an implementation provides such a guarantee is indicated by the
presence of the `__cpp_threadsafe_static_init` feature test
macro{fn-feature-macro}.
```

Thanks for your feedback!

svenvh (approved):

PR #583, opened 2021-03-05
[C++ for OpenCL] Add section w/ kernel function restrictions.

AnastasiaStulova:


cxx4opencl/kernel.txt
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+[[kernel_function]]
+=== Kernel functions
+
+Kernel functions have a C linkage ({cpp}17 `[dcl.link]`) implicitly which
```
svenvh, line 8:
```suggestion
Kernel functions have implicit C linkage ({cpp}17 `[dcl.link]`) which
```

cxx4opencl/kernel.txt
```diff
@@ -0,0 +1,23 @@
...
+functions:
+
+  * Can not be class members ({cpp}17 `[class.mfct]`);
+  * Can not be overloaded ({cpp}17 `[over]`);
+  * Can not be function templates ({cpp}17 `[temp.fct]`).
+
+Moreover the types used in parameters of the kernel functions must be:
+
+  * Trivial and standard-layout types {cpp}17 `[basic.types]` (plain old data
+    types) for parameters passes by value;
```
svenvh, line 19:
```suggestion
    types) for parameters passed by value;
```

svenvh (changes requested):

svenvh (approved):

cxx4opencl/kernel.txt
```diff
@@ -0,0 +1,28 @@
...
+  * Can not be function templates ({cpp}17 `[temp.fct]`).
+
+:fn-ker-par-ref: footnote:[Whether C++ features (e.g references) can be used \
+in functions with C linkage is implementation-defined ({cpp}17 [dcl.link\]).]
+
+Moreover the types used in parameters of the kernel functions must be:
+
+  * Trivial and standard-layout types {cpp}17 `[basic.types]` (plain old data
+    types) for parameters passed by value;
+  * Standard layout types for pointer parameters. The same applies to
```
AnastasiaStulova, line 23: Discussed w/ @mantognini that references can only be supported as an implementation defined feature.

Btw the footnotes are really a nightmare in asciidoc. Currently, they always goes to a separate page instead of the same page: https://github.com/asciidoctor/asciidoctor-pdf/issues/1833.

Common work arouns if to add a separate section with all footnotes in the docukent. I don't really like it but we might have to do this though. I will handle it separately if I don't find any better solution.

mantognini (approved):
LGTM

PR #593, opened 2021-04-12
Removed void parameter from clUnloadCompiler in cl.xml

johnplate:
The function clUnloadCompiler has no parameters and should not have a void parameter in the xml. This would be consistent with other Khronos xml specs.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -3521,7 +3521,6 @@ server's OpenCL/api-docs repository.
         </command>
         <command prefix="CL_API_PREFIX__VERSION_1_1_DEPRECATED" suffix="CL_API_SUFFIX__VERSION_1_1_DEPRECATED">
             <proto><type>cl_int</type> <name>clUnloadCompiler</name></proto>
-            <param><type>void</type>                                    </param>
         </command>
         <command prefix="CL_API_PREFIX__VERSION_1_1_DEPRECATED" suffix="CL_API_SUFFIX__VERSION_1_1_DEPRECATED">
             <proto><type>void</type>*  <name>clGetExtensionFunctionAddress</name></proto>


bashbaug:
If we accept this change then we'll need a few tweaks to the ICD loader generation scripts (and possibly a few others also) to  robustly handle the case when there truly are no function parameters.  This is probably a good idea regardless, but it's not done currently.

I do question whether this is the right change though, since the `clUnloadCompiler` function has explicitly accepted `void` as its parameter all the way back to the OpenCL 1.0 spec, rather an empty parameter list.  Other than consistency with other Khronos XML files, is there a reason to make this change?  Which other Khronos APIs accept no parameters, and would it make more sense to add an explicit `void` parameter for them if consistency is important?

Kerilk:
I would be very cautious about this change:
https://stackoverflow.com/a/5929736

If we remove the `void` from the xml, we must not forget to add it back during generation, as the API is in C and not C++.

johnplate:
> Which other Khronos APIs accept no parameters, and would it make more sense to add an explicit `void` parameter for them if consistency is important?
These are a lot Khronos functions that accept no parameters: glEnd, glEndList, glFinish, glFlush, glGetError, glInitNames, glLoadIdentity, glPopAttrib, glPopMatrix, glPopName, glPushMatrix, glPopClientAttrib, glCreateProgram, glEndConditionalRender, glEndTransformFeedback, glPauseTransformFeedback, glResumeTransformFeedback, glReleaseShaderCompiler, glPopDebugGroup, glGetGraphicsResetStatus, glTextureBarrier, eglGetCurrentDisplay, eglGetError, eglWaitGL, eglQueryAPI, eglReleaseThread, eglWaitClient, eglGetCurrentContext, and more ...

None of those functions has a 'param' member in their xml entries in gl.xml and egl.xml. I personally know at least 5 scripts in multiple companies which will break if you add a void parameter, so please don't consider that. I only found the problem because our generator that we use since years for gl.xml and egl.xml didn't work for cl.xml.
I am not aware of any script outside of Khronos that will fail if you remove the void parameter. 'clUnloadCompiler' is a deprecated command anyway, so often it is not used. We have to include it for conformance tests.

> I do question whether this is the right change though, since the `clUnloadCompiler` function has explicitly accepted `void` as its parameter all the way back to the OpenCL 1.0 spec, rather an empty parameter list. Other than consistency with other Khronos XML files, is there a reason to make this change?

The reason to make the change is how you define the content of the xml. An xml file does not have to be parsed by a C compiler, there is no need to stick to C language rules. Many projects use the xml to produce code layers in other languages, for example C++ and Go. The common assumption is that each 'param' entry describes a 'real' named parameter. It is easy to process the absence of parameters in a generator, and a bit more complicated to check if there is only one 'void' parameter, because you have to check the absence of '*' and name. On the other side it is trivial for a C generator to produce a 'void' if no parameter is present.

bashbaug:
FWIW, I explicitly added the void parameter a year or so ago since it was breaking one of my scripts:

https://github.com/KhronosGroup/OpenCL-Docs/pull/95

Regardless, I'm OK making this change, though I agree with @kerilk that we will need to be very clear that the lack of any parameters really truly does mean that there are no parameters, and hence generating a function with `void` is required for a C function.  We'll also need to update a few of the OpenCL scripts to handle this change.

alycm:
> We'll also need to update a few of the OpenCL scripts to handle this change.

I checked the generator scripts for the specification itself and it already treats no parameters as explicitly void: https://github.com/KhronosGroup/OpenCL-Docs/blob/ed25b04694c50064d763c0f7dffbbf02ed3360ba/scripts/generator.py#L900

The formatting changes, I don't think that matters, it would be trivial to adjust if it does.

I haven't checked any of the code generator scripts in other repositories though.

bashbaug:
Quick update: I think we need to update the ICD loader generation script to handle this change and I'd like to update my extension header generation script as well (though this is less critical since it isn't merged yet).  I'm hoping to do this shortly.

Are there any other scripts we need to update before merging this PR?

bashbaug (approved):
I just merged https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/150 so all of the code generation scripts I am aware of will work with this change.

PR #597, opened 2021-04-19
Unsafe math optimizations implies FTZ

### LINKED ISSUE #579: flush-to-zero behavior in case of cl-fast-relaxed-math
It is unclear from the spec if cl-fast-relaxed-math allows flush-to-zero as one of the optimizations, especially if device supports CL_FP_DENORM.

alycm:
Fixes #579

### DIFF
--- a/api/footnotes.asciidoc
+++ b/api/footnotes.asciidoc
@@ -148,6 +148,10 @@ Please refer to the OpenCL glossary for the OpenCL definition of thread-safe. \
 This definition may be different from usage of the term in other contexts. \
 ]
 
+:fn-unsafe-denorms-are-zero: pass:n[ \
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
LGTM, adding the footnote is a nice clarification.

Do we need to make any changes to the CTS tests as a result of this change?

alycm:
> Do we need to make any changes to the CTS tests as a result of this change?

I'm not sure, I've left a comment about how to check on the issue: https://github.com/KhronosGroup/OpenCL-Docs/issues/579#issuecomment-822856376

bashbaug:
Merging... no reason to hold up these changes while we check the CTS.

PR #600, opened 2021-04-20
Within extension template use `_khr` suffix in new type example

alycm:


alycm:
If bumping the version for this seems excessive it is trivial to cut that line.

bashbaug (commented):
LGTM.  Merging immediately since these changes are straightforward.

bashbaug (approved):

PR #605, opened 2021-04-23
cl_khr_extended_bit_ops, cl_khr_suggested_local_work_size

bashbaug:
Please publish the asciidoctor source for the following extensions:

* cl_khr_extended_bit_ops
* cl_khr_suggested_local_work_size


ext/cl_khr_extended_bit_ops.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+This extension is written against the OpenCL 3.0 C Language Specification and the OpenCL SPIR-V Environment Specification Version V3.0.6.
+
+This extension requires OpenCL 1.0.
+
+=== New OpenCL C Functions
+
+[source]
+----
+gentype bitfield_insert( gentype base, gentype insert, uint offset, uint count )
+igentype bitfield_extract_signed( gentype base, uint offset, uint count )
```
kpet, line 40: `i` is typically used as a prefix where signedness doesn't matter. `sgentype` might be clearer.

ext/cl_khr_extended_bit_ops.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+--
+The functions described in the following table can be used with built-in scalar or vector integer types to perform extended bit operations.
+The functions that operate on vector types operate component-wise.
+The description is per-component.
+
+In the table below, the generic type name `gentype` refers to the built-in integer types `char`, `char__n__`, `uchar`, `uchar__n__`, `short`, `short__n__`, `ushort`, `ushort__n__`, `int`, `int__n__`, `uint`, `uint__n__`, `long`, `long__n__`, `ulong`, and `ulong__n__`.
+The generic type name `igentype` refers to the built-in signed integer types `char`, `char__n__`, `short`, `short__n__`, `int`, `int__n__`, `long`, and `long__n__`.
+The generic type name `ugentype` refers to the built-in unsigned integer types `uchar`, `uchar__n__`, `ushort`, `ushort__n__`, `uint`, `uint__n__`, `ulong`, and `ulong__n__`.
+_n_ is 2, 3, 4, 8, or 16.
+
```
kpet, line 60: We may want to mention the interaction with `cles_khr_int64` on embedded profile devices.

kpet (commented):
I see you've included the test specs as well. Doesn't hurt for now but it raises the question of whether/how these should be maintained when the CTS changes.

ext/cl_khr_extended_bit_ops.asciidoc
bashbaug: Hmm, looking at the OpenCL C spec I think it's the other way around:

* `sgentype` is used to represent a scalar type, for `clamp`, `min`, and `max`.
* `igentype` is used to represent a signed type, for a handful of relational functions.

I think this means that this extension spec is consistent with the rest of the OpenCL C spec?

ext/cl_khr_extended_bit_ops.asciidoc
bashbaug: How important is this?

Looking through the rest of the extensions spec we seem to be fairly dilligent about qualifying `double` and `half` support, but we generally don't qualify `long` support.  Options are: fix them all in a subsequent PR or qualify usage of the `long` type here at least, even though it's not qualified elsewhere.

I'd prefer to fix this in a subsequent PR.

ext/cl_khr_extended_bit_ops.asciidoc
kpet: Unfortunately you're right :). Agree consistency is more important so no change it is.

ext/cl_khr_extended_bit_ops.asciidoc
kpet: A subsequent PR works for me.

kpet (approved):

ext/cl_khr_suggested_local_work_size.asciidoc
```diff
@@ -0,0 +1,95 @@
...
+* `CL_INVALID_KERNEL_ARGS` if all argument values for _kernel_ have not been set.
+* `CL_MISALIGNED_SUB_BUFFER_OFFSET` if a sub-buffer object is set as an argument to _kernel_ and the offset specified when the sub-buffer object was created is not aligned to `CL_DEVICE_MEM_BASE_ADDR_ALIGN` for the device associated with _command_queue_.
+* `CL_INVALID_IMAGE_SIZE` if an image object is set as an argument to _kernel_ and the image dimensions are not supported by device associated with _command_queue_.
+* `CL_IMAGE_FORMAT_NOT_SUPPORTED` if an image object is set as an argument to _kernel_ and the image format is not supported by the device associated with _command_queue_.
+* `CL_INVALID_OPERATION` if an SVM pointer is set as an argument to _kernel_ and the device associated with _command_queue_ does not support SVM or the required SVM capabilities for the SVM pointer.
+* `CL_INVALID_WORK_DIMENSION` if _work_dim_ is not a valid value (i.e. a value between 1 and `CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS`).
+* `CL_INVALID_GLOBAL_WORK_SIZE` if _global_work_size_ is NULL or if any of the values specified in _global_work_size_ are 0.
```
alycm, line 89: I can't quite decide if this is sensible for OpenCL 2.1+ where this would be a legal call to `clEnqueueNDRangeKernel`, but returning values of `0` because of it seems dubious.

This is in contradiction with the "These error conditions are consistent with error conditions for clEnqueueNDRangeKernel." info box below, for OpenCL 2.1+.

The extension is explicitly "requires OpenCL 1.0" though, and this is too technical a thought for final editorial pass.  If I can't figure it out I'll file an issue.

alycm: To be clear: This comment does not block merging, I just wrote down my thoughts.

alycm (approved):
LGTM.

ext/cl_khr_suggested_local_work_size.asciidoc
alycm: This was already discussed and resolved, I'm just late to the party.

```
//. What should the behavior be for a global work size equal to zero?
//+
//--
//*RESOLVED*: This is an error, both for a `NULL` global work size and when the global work size is equal to zero.
//The error code is `CL_INVALID_GLOBAL_WORK_SIZE`, consistent with *clEnqueueNDRangeKernel*.
```

PR #613, opened 2021-05-04
Update copyrights to 2021

alycm:
Only update Khronos copyrights, but update all such copyrights whether the file
has been modified in 2021 so far or not.


bashbaug (approved):
LGTM - thank you for taking care of this!

I didn't do an exhaustive check but I did grep for `2020` and didn't see anything obviously missing in this PR.

bashbaug:
I'm going to go ahead and merge this since it's clearly an improvement over what we have.  If we missed anything here, we can tidy it up in a subsequent PR.

PR #615, opened 2021-05-04
Update OpenGL trademark

alycm:
Bring the description of the OpenGL trademarks in line with this text:
https://www.khronos.org/legal/Khronos_Specification_Copyright_License_Header

An alternative source for this being the up to date information is the Vulkan
specification:
https://github.com/KhronosGroup/Vulkan-Docs/blob/88da24862cf91d692801c6ae64665d26d39f06c4/copyright-spec.txt#L80-L81


bashbaug (commented):
This makes sense to me but it'd be good to get a +1 from @oddhack or @neiltrevett .

oddhack (approved):
LGTM.

Issue #474, opened 2020-10-07
additional hyphenation changes for consistency


bashbaug:
PR #470 was a big improvement in hyphenation consistency but there are still a few other terms that are hyphenated or otherwise used inconsistently:

* `subgroup` or `sub-group` (I didn't find any cases of `sub group`)
* `command queue` or `command-queue`
* `NDRange` or `ND-range` (I didn't find any cases of `ND-Range`)

If there are any others we should fix please add them here and I'll fix them all in one PR.

bashbaug:
One more:

* `subdevice` or `sub-device`

bashbaug:
Another:

* `out of order` or `out-of-order`

bashbaug:
Another:
* `implementation defined` or `implementation-defined`

EwanC:
A few more i've seen

- `in order` v `in-order`
- `as if` v `as-if`
- `thread-safe` v `thread safe`
- `floating-point` v `floating point`
- `happens-before` v `happens before`
- `device-side` v `device side`

For `subgroup`/`sub-group` there are also a couple of places where there is an accidental space before the hyphen, i.e `sub -group`

bashbaug:
I found one more: `kernel instance` vs. `kernel-instance`.

PR #617, opened 2021-05-07
Clarify optionality support for the double precision literals.

AnastasiaStulova:
This change adds clarifications regarding double precision
literals when double precision is not supported.

It also improves wording for -cl-single-precision-constant
flag to make it more precise.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -336,8 +336,8 @@ The following table describes the list of built-in scalar data types.
 If double precision floating-point extension {cl_khr_fp64} or feature
 {opencl_c_fp64} is not supported, implementations are allowed to
 implicitly cast double precision floating-point literals to
-single-precision literals. The use of double precision literals without
-double-precision support should result in a diagnostic.
+single precision literals. The use of double precision literals without
+double precision support should result in a diagnostic.
 
 Most built-in scalar data types are also declared as appropriate types in
 the OpenCL API (and header files) that can be used by an application.
@@ -481,7 +481,8 @@ The following table describes the list of built-in vector data types.
 | `double__n__` footnote:[{fn-double-vec}]
     | A vector of _n_ 64-bit floating-point values.
 
-      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer.
+      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer. In
+      OpenCL C 3.0 it requires support of {opencl_c_fp64} feature.
       Also see extension *cl_khr_fp64*.
 |====
 
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -4822,7 +4822,7 @@ all arguments and the return type, unless otherwise specified.
     | Returns a quiet NaN.
       The _nancode_ may be placed in the significand of the resulting NaN.
 | gentype *nextafter*(gentype _x_, gentype _y_)
-    | Computes the next representable single-precision floating-point value
+    | Computes the next representable single precision floating-point value
       following _x_ in the direction of _y_.
       Thus, if _y_ is less than _x_, *nextafter*() returns the largest
       representable floating-point number less than _x_.
@@ -5064,7 +5064,7 @@ single precision floating-point number.
 |====
 | *Constant Name* | *Description*
 | `MAXFLOAT`
-    | Value of maximum non-infinite single-precision floating-point number.
+    | Value of maximum non-infinite single precision floating-point number.
 | `HUGE_VALF`
     | A positive `float` constant expression.
       `HUGE_VALF` evaluates to +infinity.
@@ -8189,7 +8189,7 @@ a following *a*, *A*, *e*, *E*, *f*, *F*, *g*, or *G* conversion specifier
 applies to a `double__n__` argument.
 The *l* modifier is supported by the full profile.
 For the embedded profile, the *l* modifier is supported only if 64-bit
-integers or double-precision floating-point are supported by the device.
+integers or double precision floating-point are supported by the device.
 
 If a vector specifier appears without a length modifier, the behavior is
 undefined.
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -334,10 +334,10 @@ The following table describes the list of built-in scalar data types.
 |====
 
 If double precision floating-point extension {cl_khr_fp64} or feature
-{opencl_c_fp64} is not supported, the implementations are allowed to
+{opencl_c_fp64} is not supported, implementations are allowed to
 implicitly cast double precision floating-point literals to
 single-precision literals. The use of double precision literals without
-the double-precision support should results in a diagnostic.
+double-precision support should result in a diagnostic.
 
 Most built-in scalar data types are also declared as appropriate types in
 the OpenCL API (and header files) that can be used by an application.
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -307,7 +307,8 @@ The following table describes the list of built-in scalar data types.
       The `double` data type must conform to the IEEE 754 double precision
       storage format.
 
-      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer.
+      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer. In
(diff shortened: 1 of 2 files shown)


AnastasiaStulova:
@b-sumner and @jlewis-austin I was not able to request your review... Perhaps you are not in OpenCL group yet?

jlewis-austin:
> @b-sumner and @jlewis-austin I was not able to request your review... Perhaps you are not in OpenCL group yet?

@AnastasiaStulova thanks I'll take a look, and emailed James regarding getting us added.

OpenCL_C.txt
```diff
@@ -307,7 +307,8 @@ The following table describes the list of built-in scalar data types.
       The `double` data type must conform to the IEEE 754 double precision
       storage format.
 
-      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer.
+      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer. In
+      OpenCL C 3.0 it requires support of {opencl_c_fp64} feature.
```
jlewis-austin, line 311:
```suggestion
      OpenCL C 3.0 it requires support of the {opencl_c_fp64} feature.
```
Elsewhere it looks like we refer to the feature macro rather than the feature. I'm not sure if there is any distinction in meaning?

OpenCL_C.txt
```diff
@@ -307,7 +307,8 @@ The following table describes the list of built-in scalar data types.
       The `double` data type must conform to the IEEE 754 double precision
       storage format.
 
-      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer.
+      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer. In
```
jlewis-austin, line 310: Should this change be duplicated for the double vector types?

jlewis-austin (commented):
Thanks, these look good, just a few editorial suggestions.

OpenCL_C.txt
AnastasiaStulova: Ah I think it was done when we didn't have wording saying that features inherit names from the macros but now we do. For extensions we used to refer to the name rather than the macro. This is why we wanted to update the features but it has not been done yet.

I can reword it to alight with the other features and we change all of them later or I can keep it as is since this is what we aim at but it will be inconsistent with the other features but consistent with extensions...

OpenCL_C.txt
AnastasiaStulova: Done! Thanks!

AnastasiaStulova:
FYI I also fixed spelling of `double precision` and `single precision` in various places since it was inconsistent. I can undo it though.

OpenCL_C.txt
jlewis-austin: OK thanks, however you want to stage it is fine with me

jlewis-austin:
> FYI I also fixed spelling of `double precision` and `single precision` in various places since it was inconsistent. I can undo it though.

The rule I've seen most often elsewhere is to hyphenate when used as an adjective, e.g. "single-precision floating-point literal" vs "operator supports single precision floating point". That being said, I'm OK with any choice consistently applied.

OpenCL_C.txt
bashbaug: I do think we should accept this suggestion though and add `the`.

bashbaug:
> Maybe we can use the most common spelling in this PR, but it can be fully renamed separately?

Yes, absolutely, we shouldn't "fix" the hyphenation issues in this PR, but I'm hoping we don't introduce new issues either.

We can add these terms to this issue if you'd like: https://github.com/KhronosGroup/OpenCL-Docs/issues/474

Remember: we can't simply use the hyphenated or non-hyphenated version exclusively since the correct version depends on how it is used.

OpenCL_C.txt
```diff
@@ -332,6 +333,12 @@ The following table describes the list of built-in scalar data types.
       type that cannot be completed.
 |====
 
+If double precision floating-point extension {cl_khr_fp64} or feature
+{opencl_c_fp64} is not supported, implementations are allowed to
+implicitly cast double precision floating-point literals to
+single precision literals. The use of double precision literals without
```
AnastasiaStulova, line 339:
```suggestion
single-precision literals. The use of double-precision literals without
```

OpenCL_C.txt
```diff
@@ -332,6 +333,12 @@ The following table describes the list of built-in scalar data types.
...
+double precision support should result in a diagnostic.
```
AnastasiaStulova, line 340:
```suggestion
double-precision support should result in a diagnostic.
```

OpenCL_C.txt
```diff
@@ -332,6 +333,12 @@ The following table describes the list of built-in scalar data types.
       type that cannot be completed.
 |====
 
+If double precision floating-point extension {cl_khr_fp64} or feature
+{opencl_c_fp64} is not supported, implementations are allowed to
+implicitly cast double precision floating-point literals to
```
AnastasiaStulova, line 338:
```suggestion
implicitly cast double-precision floating-point literals to
```

OpenCL_C.txt
```diff
@@ -474,7 +481,8 @@ The following table describes the list of built-in vector data types.
 | `double__n__` footnote:[{fn-double-vec}]
     | A vector of _n_ 64-bit floating-point values.
 
-      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer.
+      <<unified-spec, Requires>> support for OpenCL C 1.2 or newer. In
+      OpenCL C 3.0 it requires support of {opencl_c_fp64} feature.
```
AnastasiaStulova, line 485:
```suggestion
      OpenCL C 3.0 it requires support of the {opencl_c_fp64} feature.
```

OpenCL_C.txt
```diff
@@ -4814,7 +4822,7 @@ all arguments and the return type, unless otherwise specified.
     | Returns a quiet NaN.
       The _nancode_ may be placed in the significand of the resulting NaN.
 | gentype *nextafter*(gentype _x_, gentype _y_)
-    | Computes the next representable single-precision floating-point value
+    | Computes the next representable single precision floating-point value
```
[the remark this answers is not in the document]
AnastasiaStulova:
```suggestion
    | Computes the next representable single-precision floating-point value
```

OpenCL_C.txt
```diff
@@ -5056,7 +5064,7 @@ single precision floating-point number.
 |====
 | *Constant Name* | *Description*
 | `MAXFLOAT`
-    | Value of maximum non-infinite single-precision floating-point number.
+    | Value of maximum non-infinite single precision floating-point number.
```
AnastasiaStulova, line 5067:
```suggestion
    | Value of maximum non-infinite single-precision floating-point number.
```

OpenCL_C.txt
```diff
@@ -8181,7 +8189,7 @@ a following *a*, *A*, *e*, *E*, *f*, *F*, *g*, or *G* conversion specifier
 applies to a `double__n__` argument.
 The *l* modifier is supported by the full profile.
 For the embedded profile, the *l* modifier is supported only if 64-bit
-integers or double-precision floating-point are supported by the device.
+integers or double precision floating-point are supported by the device.
```
AnastasiaStulova, line 8192:
```suggestion
integers or double-precision floating-point are supported by the device.
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -6211,8 +6211,8 @@ These options control compiler behavior regarding floating-point arithmetic.
 These options trade off between speed and correctness.
 
 `-cl-single-precision-constant` ::
-    Treat double precision floating-point constant as single precision
-    constant.
+    This option forces implicit conversions of double precision floating-point
```
AnastasiaStulova, line 6214:
```suggestion
    This option forces implicit conversions of double-precision floating-point
```

AnastasiaStulova:
> Remember: we can't simply use the hyphenated or non-hyphenated version exclusively since the correct version depends on how it is used.

Thanks for the pointer. It really helped to remove the confusion. I think I have fixed those issues now.

bashbaug (approved):
Thanks LGTM!

PR #622, opened 2021-06-01
cl_intel_sharing_format_query

bashbaug:
This PR adds the extension spec source and XML changes for the `cl_intel_sharing_format_query` extension.

The APIs for this extension are slightly unconventional in that they effectively extend their corrseponding sharing extensions - for example, when `cl_khr_gl_sharing` is supported, this extension adds the `clGetSupportedGLTextureFormatsINTEL` API.  I've represented this in the XML file by defining specific OpenGL, DX9, D3D10, D3D11, and VA_API sections.  I think this works reasonably well for things like extension header generation, but it is an area to pay attention to where we would appreciate feedback.

We would prefer not to define a unique extension string for each interop API to avoid redundancey and to minimize the length of our extension string.

I have header and registry changes for this extension ready to go but I haven't created PRs for them, just in case they need last-minute changes.

### DIFF
--- a/extensions/cl_intel_sharing_format_query.asciidoc
+++ b/extensions/cl_intel_sharing_format_query.asciidoc
@@ -1,0 +1,475 @@
+:data-uri:
+:sectanchors:
+:icons: font
+:source-highlighter: coderay
+
+= cl_intel_sharing_format_query
+
+== Name Strings
+
+`cl_intel_sharing_format_query`
+
+== Contact
+
+Ben Ashbaugh, Intel (ben 'dot' ashbaugh 'at' intel 'dot' com)
+
+== Contributors
+
+// spell-checker: disable
+Ben Ashbaugh, Intel +
+Jacek Danecki, Intel +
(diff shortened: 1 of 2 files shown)


extensions/cl_intel_sharing_format_query.asciidoc
```diff
@@ -0,0 +1,475 @@
...
+--
+
+== Revision History
+
+[cols="5,15,15,70"]
+[grid="rows"]
+[options="header"]
+|========================================
+|Version|Date|Author|Changes
+|1.0.0|2021-05-31|Ben Ashbaugh|*Initial public revision*
```
alycm, line 466:
```suggestion
|1.0.0|2021-05-31|Ben Ashbaugh|*Initial public revision*
|========================================
```

alycm (approved):
LGTM.

bashbaug:
I addressed the code review comments and fixed the conflict in the XML file.  Merging this one too.

Issue #607, opened 2021-04-23
SPIR-V Changes for cl_khr_extended_bit_ops


bashbaug:
The SPIR-V changes for `cl_khr_extended_bit_ops` are currently described in the extension spec.  They should be moved to the SPIR-V environment spec, similar to other KHR extensions.

PR #627, opened 2021-06-17
Add cl_khr_integer_dot_product

kpet:

### DIFF
--- a/ext/cl_khr_integer_dot_product.asciidoc
+++ b/ext/cl_khr_integer_dot_product.asciidoc
@@ -193,17 +193,7 @@ of `a`, `Y` that of the components of `b`. `R` is the return type.
 
 === Modifications to the OpenCL SPIR-V Environment Specification
 
-(Modify Section 5.2, *Full and Embedded Profile Extensions*) ::
-+
---
-If the OpenCL environment supports the extension `cl_khr_integer_dot_product`,
-then the environment accept modules that require `SPV_KHR_integer_dot_product` and
-declare the following SPIR-V capabilities:
-
-* *DotProductKHR*
-* *DotProductInput4x8BitKHR* if `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_KHR` is supported
-* *DotProductInput4x8BitPackedKHR*
---
+See OpenCL SPIR-V Environment Specification.
 
 === Interactions with Other Extensions
 
--- a/ext/cl_khr_integer_dot_product.asciidoc
+++ b/ext/cl_khr_integer_dot_product.asciidoc
@@ -123,6 +123,9 @@ returned for `CL_DEVICE_INTEGER_DOT_PRODUCT_CAPABILITIES_KHR`:
 | Feature bit
 | Feature macro
 
+| `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_PACKED_KHR`
+| `__opencl_c_integer_dot_product_input_4x8bit_packed`
+
 | `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_KHR`
 | `__opencl_c_integer_dot_product_input_4x8bit`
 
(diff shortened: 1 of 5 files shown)


ext/cl_khr_integer_dot_product.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+--
+
+[cols="2,2,4",options="header"]
+|====
+| cl_device_info
+| Return Type
+| Description
+
+| `CL_DEVICE_INTEGER_DOT_PRODUCT_CAPABILITIES_KHR`
+| `cl_device_integer_dot_product_capabilities_khr`
```
bashbaug, line 102: Consider using the auto-generated asciidoctor macros `{CL_DEVICE_INTEGER_DOT_PRODUCT_CAPABILITIES_KHR}` and `{cl_device_integer_dot_product_capabilities_khr_type}` so these table cells wrap and do not widen the table excessively.

ext/cl_khr_integer_dot_product.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+`CL_DEVICE_OPENCL_C_FEATURES` when the corresponding bit is set in the bitfield
+returned for `CL_DEVICE_INTEGER_DOT_PRODUCT_CAPABILITIES_KHR`:
+
+[cols="1,1",options="header"]
+|====
+| Feature bit
+| Feature macro
+
+| `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_KHR`
+| `__opencl_c_integer_dot_product_input_4x8bit`
```
bashbaug, line 127: I think this is fine, but just to confirm: we don't want a feature test macro for the packed variants because they are required and the test for the extension is sufficient?

Would we want to add a feature test macro if this extension ever became a core feature?

We don't want a `_khr` suffix on this feature test macro?

bashbaug (changes requested):
Can you please add an entry to the "Quick Reference" table also?

My other comments are nice-to-haves and not must-fixes.

Thanks!

ext/cl_khr_integer_dot_product.asciidoc
```diff
@@ -0,0 +1,212 @@
...
++
+--
+If the OpenCL environment supports the extension `cl_khr_integer_dot_product`,
+then the environment accept modules that require `SPV_KHR_integer_dot_product` and
+declare the following SPIR-V capabilities:
+
+* *DotProductKHR*
+* *DotProductInput4x8BitKHR* if `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_KHR` is supported
+* *DotProductInput4x8BitPackedKHR*
+--
```
bashbaug, line 206: One more comment (since I'm fixing #607, which is similar): this section should be moved to the SPIR-V environment spec rather than being documented here.

ext/cl_khr_integer_dot_product.asciidoc
kpet: I've tried this and it seems to have no impact on the output for me. Is the behaviour specific to given versions of asciidoctor?

ext/cl_khr_integer_dot_product.asciidoc
kpet: >we don't want a feature test macro for the packed variants because they are required and the test for the extension is sufficient?

That's been the rationale.
>Would we want to add a feature test macro if this extension ever became a core feature?

Probably, if we make it an optional core feature. There could be an argument for adding it now to help with code portability if we make this a core feature in the future.

>We don't want a _khr suffix on this feature test macro?

I'd say no for the same reason we don't add it to khr built-in function names. We want code written for the extension to "just work" if it becomes a core feature.

ext/cl_khr_integer_dot_product.asciidoc
kpet: Yeah, I might as well do that here. We need better cross-linking of the two documents.

ext/cl_khr_integer_dot_product.asciidoc
bashbaug: Are you looking at the HTML output or the PDF output?

The PDF output looks reasonably OK to me without the macros, though the table cells wrap arbitrarily rather than at underscore boundaries.  The HTML output is very wide without the macros, though.

bashbaug (approved):
My important comments are resolved - thanks!

I'll leave this PR open a little longer in case we want to add support for the asciidoctor macros or to add a feature macro for the packed variants.

xml/cl.xml
```diff
@@ -6100,5 +6107,20 @@ server's OpenCL/api-docs repository.
...
+            <require>
+                <type name="CL/cl.h"/>
+            </require>
+            <require>
+                <type name="cl_device_integer_dot_product_capabilities_khr"/>
+            </require>
+            <require comment="cl_device_integer_dot_product_capabilities_khr">
+                <enum name="CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_KHR"/>
+                <enum name="CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_PACKED_KHR"/>
+            </require>
```
bashbaug, line 6120: The current header generation script will generate `#defines` for these enums in the order they are listed in the XML file.  The unpacked `4x8BIT_KHR` enum is bit 1 and the `4x8BIT_PACKED_KHR` enum is bit 0.  If we want bit zero to come before bit one I'll either need to update the script to sort the enum names based on bit number or we should switch the order here in the XML file.

ext/cl_khr_integer_dot_product.asciidoc
kpet: As discussed in the 2021/06/22 teleconference, I've added a feature macro for `CL_DEVICE_INTEGER_DOT_PRODUCT_INPUT_4x8BIT_PACKED_KHR`

ext/cl_khr_integer_dot_product.asciidoc
kpet: I was looking at the PDF only. I can see the impact on the HTML spec as well. I've added a commit to use the macros.

PR #631, opened 2021-06-21
remove unnecessary phrase from subgroup mask function descriptions

### LINKED ISSUE #625: Clarify Description of Extended Subgroup "mask" Builtins
The descriptions of the extended subgroup "mask" built-ins (e.g. `get_sub_group_eq_mask`) currently starts with:

> Generates a bitmask of all work items in the subgroup, where the bit is set in the bitmask if ...

The "of all work items in the subgroup" is extraneous and potentially confusing, especially for partial subgroups where the subgroup size is less than the maximum subgroup size.

Suggest shortening the descriptions of these built-ins to simply be:

> Generates a bitmask, where the bit is set in the bitmask if ...

bashbaug:
Fixes #625.

Removes an extraneous phrase from the descriptions of the subgroup mask built-in functions.

### DIFF
--- a/ext/cl_khr_subgroup_extensions.asciidoc
+++ b/ext/cl_khr_subgroup_extensions.asciidoc
@@ -339,35 +339,35 @@ If no bits representing predicate values from all work items in the subgroup are
 ----
 uint4 get_sub_group_eq_mask()
 ----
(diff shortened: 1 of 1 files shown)


StuartDBrady:
Looks good to me.

("Generates a bitmask where for each bit, the bit is [...]" might be slightly better wording.)

StuartDBrady (approved):

kpet (approved):

PR #635, opened 2021-06-29
Clarify that CL_DEVICE_{GLOBAL_MEM,MAX_MEM_ALLOC}_SIZE return a constant

kpet:
Change-Id: I650a8a41eccaf47689914e8e110227a6253023ef

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -468,7 +468,9 @@ include::{generated}/api/version-notes/CL_DEVICE_MAX_MEM_ALLOC_SIZE.asciidoc[]
       | Max size of memory object allocation in bytes.
         The minimum value is max(min(1024 {times} 1024 {times} 1024, 1/4^th^
         of {CL_DEVICE_GLOBAL_MEM_SIZE}), 32 {times} 1024 {times} 1024) for
-        devices that are not of type {CL_DEVICE_TYPE_CUSTOM}.
+        devices that are not of type {CL_DEVICE_TYPE_CUSTOM}. +
+        The value returned is a property of the device and does not change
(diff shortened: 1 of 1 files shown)


alycm (dismissed):
LGTM.

bashbaug (commented):
It seems that most `clGetDeviceInfo` queries would have this property and be invariant across calls, assuming the call is being made on the same device object.  I took a quick look and these are the only ones I think can vary:

* `CL_DEVICE_AVAILABLE` (maybe?)
* `CL_DEVICE_COMPILER_AVAILABLE`, `CL_DEVICE_LINKER_AVAILABLE` (maybe?)
* `CL_DEVICE_REFERENCE_COUNT` (at least for sub-devices)

Should we add some general text that covers more than just these two queries?

I'm worried that if we only describe the invariance property for these two queries a reasonable interpretation would be that all of the other queries could vary, which would be bad.

FWIW, I think many of the other `clGetXXXInfo` queries are invariant too (but not all!), so if we add general text we may want to extend it to more than just the `clGetDeviceInfo` queries also.

Oblomov:
I don't think we can guarantee that `CL_DEVICE_*MEM*_SIZE` would be constant, in fact. Consider for example a CPU platform where the system has support for memory hotplug. The number of compute units may change too (NUMA node hotplug, unless the platform decides to expose each NUMA node as a separate device).

We might want to be more specific about the “sense” of the constancy (e.g. the fact that it does not depend on the presence of other running programs using the same device for their own purposes).

kpet:
All very good points. For context, we discussed and agreed this change in the 2021/06/29 teleconference but didn't touch on hot-plugable memory or compute resources. I also agree with Ben that we can probably make a more generic change. This needs more thinking.

PR #637, opened 2021-06-30
Increase the number of TOC levels for the extension specification

kpet:
Makes the specificaiton much more navigable. We may want to revisit section
titles for some of the older extensions but overall it feels like a good
change as is to me.

Change-Id: I7ac819888e702075aec02001d47ff3084cab2819

### DIFF
--- a/OpenCL_Ext.txt
+++ b/OpenCL_Ext.txt
@@ -8,7 +8,7 @@ Khronos{R} OpenCL Working Group
 :data-uri:
 :icons: font
 :toc2:
-:toclevels: 1
+:toclevels: 2
 :max-width: 100
 :numbered:
 :imagewidth: 800
--- a/OpenCL_Ext.txt
+++ b/OpenCL_Ext.txt
@@ -8,7 +8,9 @@ Khronos{R} OpenCL Working Group
 :data-uri:
 :icons: font
 :toc2:
-:toclevels: 2
+// Only one TOC level for the HTML output, two TOC levels for everything else.
+ifdef::backend-html5[:toclevels: 1]
+ifndef::backend-html5[:toclevels: 2]
 :max-width: 100
 :numbered:
 :imagewidth: 800


bashbaug:
Maybe I just need to get used to it, but I find the expanded TOC on the left-hand side of the HTML output harder to navigate.  I do like the expanded TOC at the beginning of the PDF document.  Should we consider increasing the number of TOC levels for the PDF output only?

kpet:
Yeah, the TOC is not collapsible in the HTML output. I still find it "okay" but it's not nearly as usable as with the PDF. I'd be fine with doing it only for the PDF if you know the magic runes.

OpenCL_Ext.txt
```diff
@@ -8,7 +8,7 @@ Khronos{R} OpenCL Working Group
 :data-uri:
 :icons: font
 :toc2:
-:toclevels: 1
+:toclevels: 2
```
bashbaug, line 11: If we decide to keep one TOC level for the HTML output but use two TOC levels for the PDF output I think this will do it.

```suggestion
// Only one TOC level for the HTML output, two TOC levels for everything else.
ifdef::backend-html5[:toclevels: 1]
ifndef::backend-html5[:toclevels: 2]
```

kpet: Thanks, I'll commit this.

kpet:
@bashbaug I've applied your proposed change, PTAL.

alycm (approved):
> We may want to revisit section
titles for some of the older extensions but overall it feels like a good
change as is to me.

Yes, the newer extensions have more helpful section titles.

I agree that the HTML version is more manageable with one level.

For the PDF it's still very long (5.5 pages, compared to 3.5 pages for the longer API document).  But this is certainly nicer for large extension chapters, such as extended sub-groups.

Overall it seems an improvement.

bashbaug (approved):
Thanks!

PR #638, opened 2021-07-01
Fix grammar in Environment spec

svenvh:


bashbaug (approved):
LGTM, thanks for fixing these!

PR #639, opened 2021-07-05
Minor program scope variable typo fix

mkinsner:


alycm (approved):
LGTM, merging immediately as trivial.

Issue #501, opened 2020-11-23
Can generic address space become a named address space too?


AnastasiaStulova:
The definition of unnamed/generic address space in OpenCL C doesn't conform to Embedded C section 5.1.3 that defines unnamed address space as default address space i.e. C compiler would default to unnamed everywhere the address space is not specified. OpenCL C compiler will use `private`, `local`, `global`, `constant` or `generic` if not specified depending on the scope and variable type. 

Spec reserved the use of `generic`/`__generic` as identifiers and they were used by Clang to simplify the implementation and make it explicit that the generic address space in OpenCL C and generic address space in Embedded C are not the same.

Therefore, it is suggested to make generic address space a named address space too. The main benefit to do this is to align better with Embedded C terminology and tooling too.

AnastasiaStulova:
Summary from the teleconference on Feb 11, 2021
- OpenCL C derives the semantic of address spaces from Embedded C, where generic is used for flat private memory space. That adds unnecessary confusion.
- In contrast to Embedded C, named address spaces in Embedded C are specified with explicit language construct. Therefore OpenCL C generic AS is not entirely a match to named address spaces either.
- If we allow mapping reserved `generic` keyword to an address space then we will be able to classify generic as named address space in OpenCL. This can be useful for some kernel sources that prefer to state address space explicitly to avoid implicit conversions etc. This will also match Clang implementation of the feature.

**Possible direction:** Make reserved `generic`/`__generic` as explicit address space qualifiers in the next versions.

bashbaug:
From #643:

@svenvh [said](https://github.com/KhronosGroup/OpenCL-Docs/pull/643#issuecomment-879702552):

> Apologies, I wasn't aware of #501 and thought this was simply an oversight, driven by the observation that Clang has been supporting the `generic` qualifiers since 2014(!) There are no CTS tests that use `generic` as an address space qualifier yet.

@AnastasiaStulova [said](https://github.com/KhronosGroup/OpenCL-Docs/pull/643#issuecomment-879715283):

> Just one small clarification here:
> 
> > I'm aware that Clang already supports these keywords, and treats them the same as the unnamed generic address space.
> 
> Generic address space is implemented and specified as a named address space. However, spec incorrectly _pretends_ that it is an unnamed address space adding lots of confusion.
> 
> So it would be good if we could correct this issue at some point.

### Continuing the discussion here:

[...]

PR #643, opened 2021-07-13
Drop "future" status of generic addrspace qualifier

svenvh:


kpet (approved):
LGTM

alycm (commented):
By dropping the "future" status this is effectively adding a new pair of keywords.  This was discussed before, the outcome was summarised in https://github.com/KhronosGroup/OpenCL-Docs/issues/501.  A possible outcome there was to add this keyword in a future version.

Maybe I'm not understanding the intent of this change though, as this PR doesn't actually define a meaning for the keywords.

I'm aware that Clang already supports these keywords, and treats then the same as the unnamed generic address space.

bashbaug (changes requested):
This seems a little premature given the discussion in #501.

Additionally, do we have any CTS tests that explicitly use `generic` as an address space qualifier?

svenvh:
Apologies, I wasn't aware of #501 and thought this was simply an oversight, driven by the observation that Clang has been supporting the `generic` qualifiers since 2014(!)  There are no CTS tests that use `generic` as an address space qualifier yet.

AnastasiaStulova:
Just one small clarification here:

> I'm aware that Clang already supports these keywords, and treats them the same as the unnamed generic address space.
Generic address space is implemented and specified as a named address space. However, spec incorrectly _pretends_ that it is an unnamed address space adding lots of confusion.

So it would be good if we could correct this issue at some point.

PR #645, opened 2021-07-21
Drop claim about not requiring SPIR-V extensions

svenvh:
The Environment Specification later contradicts this claim:

    If the OpenCL environment supports the extension
    cl_khr_extended_bit_ops, then the environment must accept modules
    that declare use of the extension SPV_KHR_bit_instructions


bashbaug (approved):
Thanks, LGTM.  Merging immediately since this fixes an obvious mistake.

PR #616, opened 2021-05-05
C++ for OpenCL 2021 (i.e OpenCL 3.0-compatible)

AnastasiaStulova:
A draft for C++ for OpenCL that is aligned with OpenCL 3.0 making features optional.

Compiled PDF is attached to https://github.com/AnastasiaStulova/OpenCL-Docs/releases/tag/cxxforopencl-v2021-draft

PR #647, opened 2021-07-26
Documentation for __remove_address_space in C++ for OpenCL

Topotuna:
Documentations provided along with `__remove_address_space` functionality in C++ for OpenCL.

Fixes: llvm.org/PR45326

Differential Revision: reviews.llvm.org/D106785


cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
 gengenptr = reinterpret_cast<int**>(locgenptr); // legal.
 gengenptr = reinterpret_cast<int**>(cnstgenptr); // legal.
 ----------
+
+[[remove-addrspace]]
+==== Address space removal
+
+Utility `+__remove address_space+` allows to derive types that have address
```
AnastasiaStulova, line 427: Since we haven't supported this in C++ for OpenCL 1.0 initially, I suggest we add this only in 2021 version? so maybe we can add some text to clarify that and then also dependency on https://github.com/KhronosGroup/OpenCL-Docs/pull/616?

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+space qualifiers removed. The tool only throws away address space qualifiers,
```
AnastasiaStulova, line 428:
```suggestion
space qualifiers removed. The utility only removes an address space qualifier from a given type,
```

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+therefore, other type qualifiers such as const or volatile remain unchanged.
+
+[source,cpp]
```
AnastasiaStulova, line 431: I feel that the example alone is not enough, we should probably add at least a template class declaration:

`template<class T> struct remove_address_space;`

either following C++ spec style or reference pages:
https://www.cplusplus.com/reference/type_traits/remove_const/

Perhaps we should also point out that it is expected to have similar behavior to `remove_const` and others similar traits from C++17? Then it will be easier to explain and to understand the idea behind this feature.

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+void foo(T *par){
+  T var1; // error: function scope variable cannot be declared in global
+          // address space
+  __private T var2; // error: conflicting address space qualifiers are provided
+                    // between types '__private T' and '__global int'
+  __private __remove_address_space<T>::type var3; // type of var3 is __private int
+}
+
+void bar(){
+    __global int* ptr;
```
svenvh, line 443: Indentation is inconsistent with template above (and in fact the rest of the document which seems to be using 2 spaces).

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+==== Address space removal
+
+Utility `+__remove address_space+` allows to derive types that have address
+space qualifiers removed. The utility only removes an address space qualifier from a given type,
+therefore, other type qualifiers such as const or volatile remain unchanged.
+
+[source,cpp]
+----------
+template<typename T>
+void foo(T *par){
```
svenvh, line 434:
```suggestion
void foo(T *par) {
```

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+  T var1; // error: function scope variable cannot be declared in global
+          // address space
+  __private T var2; // error: conflicting address space qualifiers are provided
+                    // between types '__private T' and '__global int'
+  __private __remove_address_space<T>::type var3; // type of var3 is __private int
+}
+
+void bar(){
```
svenvh, line 442:
```suggestion
void bar() {
```

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
+    __global int* ptr;
+    foo(ptr);
```
Topotuna, line 444:
```suggestion
  __global int* ptr;
  foo(ptr);
```

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,27 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
 gengenptr = reinterpret_cast<int**>(locgenptr); // legal.
 gengenptr = reinterpret_cast<int**>(cnstgenptr); // legal.
 ----------
+
+[[remove-addrspace]]
+==== Address space removal
+
+Utility `+__remove address_space+` allows to derive types that have address
+space qualifiers removed. The utility only removes an address space qualifier from a given type,
+therefore, other type qualifiers such as const or volatile remain unchanged.
```
Topotuna, line 429:
```suggestion
C++ for OpenCL 2021 supports utility `+__remove address_space+` which allows to
derive types that have address space qualifiers removed. The utility only removes
an address space qualifier from a given type, therefore, other type qualifiers
such as `+const+` or `+volatile+` remain unchanged.
```

cxx4opencl/address_spaces.txt
```diff
@@ -420,3 +420,28 @@ gengenptr = addrspace_cast<int**>(cnstgenptr); // illegal.
...
 gengenptr = reinterpret_cast<int**>(cnstgenptr); // legal.
 ----------
+
+[[remove-addrspace]]
+==== Address space removal
+
+C++ for OpenCL 2021 supports utility `+__remove address_space+` which allows to
+derive types that have address space qualifiers removed. The utility only removes
+an address space qualifier from a given type, therefore, other type qualifiers
+such as `+const+` or `+volatile+` remain unchanged.
```
Topotuna, line 430:
```suggestion
{cpp} for OpenCL 2021 supports utility `+__remove_address_space+` which allows to
derive types that have address space qualifiers removed. Its effect is analogous
to `+__remove_const+` and other similar traits in {cpp}17 `[meta.trans]`. The
utility only removes an address space qualifier from a given type, therefore, other
type qualifiers such as `+const+` or `+volatile+` remain unchanged.
```

AnastasiaStulova (approved):
LGTM! Thanks

This needs to be merged after https://github.com/KhronosGroup/OpenCL-Docs/pull/616

PR #649, opened 2021-08-03
Added new return code for clGetKernelSuggestedLocalWorkSizeKHR.

### LINKED ISSUE #641: Add extra error condition to `clGetKernelSuggestedLocalWorkSizeKHR`
Currently, there is no wording in the spec to describe an appropriate error code if `NULL` is given as the `suggested_local_work_size` argument to `clGetKernelSuggestedLocalWorkSizeKHR`.

We suggest the following wording:

> CL_INVALID_VALUE if suggested_local_work_size is NULL.

Alternatively a new error code could be added to cover this exact case.

Jeremy-Kemp:
Return CL_INVALID_VALUE if suggested_local_work_size is NULL.

Fixes #641


bashbaug (commented):
Thanks, this mostly looks good.

My one nit-picky suggestion for improvement is that we currently have an (admittedly undocumented) convention of listing error codes roughly in the order they apply to arguments, with `CL_OUT_OF_RESOURCES` and `CL_OUT_OF_HOST_RESOURCES` at the end.  Since this error code apples to the last argument it mostly meets this convention, but would you mind moving this addition up a few lines, before `CL_OUT_OF_RESOURCES`?  I can take care of this when merging if it would be easier.

Thanks again!

bashbaug (approved):
Thanks!

I don't think this should be contentious so I'll go ahead and merge this.

Issue #652, opened 2021-08-12
Use generated command prototypes and definitions in extension specification


kpet:
In the core specification, structures and function prototypes are generated from cl.xml and included as follows:

```
include::{generated}/api/protos/clCreateCommandQueueWithProperties.txt[]
```

In the extension specification, these are all spelled out manually. They should also be generated and included instead.

PR #653, opened 2021-08-12
Generate asciidoc api snippets for all supported extensions

kpet:
Prototypes, structures and definitions are now generated by default
for all extensions that are declared with 'supported="opencl"' in cl.xml.

Also define the missing "unsigned int" type in cl.xml, the generation
tool flagged its absence.

Contributes to #652

Change-Id: I6a56f2c33e59db5bc549780bbcbb6095c2dc8d98

### DIFF
--- a/scripts/gencl.py
+++ b/scripts/gencl.py
@@ -140,7 +140,7 @@ def makeGenOpts(args):
             profile           = None,
             versions          = featuresPat,
             emitversions      = featuresPat,
-            defaultExtensions = None,
+            defaultExtensions = defaultExtensions,
             addExtensions     = addExtensionsPat,
             removeExtensions  = removeExtensionsPat,
             emitExtensions    = emitExtensionsPat,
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
Nice, this will be useful.

I did get a few new warnings building the specs with this PR.  It'd be nice to fix these if it's not too complicated, but it doesn't seem to be hurting anything if we want to fix them later.

```
WARNING: type: D3DFORMAT IS NOT DEFINED
WARNING: type: DXGI_FORMAT IS NOT DEFINED
WARNING: type: DXGI_FORMAT IS NOT DEFINED
WARNING: type: VAImageFormat IS NOT DEFINED
```

kpet:
I guess these should probably be defined as opaque in the XML. I was surprised to find that none of the API calls using these types are declared in the headers. I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/655 to resolve the discrepancy (assigned to you since they're all Intel extensions). I'll merge this PR as is then. Thanks for the quick review!

PR #657, opened 2021-08-14
spec source and xml updates for cl_intel_device_attribute_query

bashbaug:
This PR adds the spec source and XML updates for the `cl_intel_device_attribute_query` extension.

I will create an OpenCL-Headers PR with updates for this extension shortly.

I will create an OpenCL-Registry PR after the spec source is merged.

### DIFF
--- a/extensions/cl_intel_device_attribute_query.asciidoc
+++ b/extensions/cl_intel_device_attribute_query.asciidoc
@@ -1,0 +1,176 @@
+:data-uri:
+:sectanchors:
+:icons: font
+:source-highlighter: coderay
+
+= cl_intel_device_attribute_query
+
+== Name Strings
+
+`cl_intel_device_attribute_query`
+
+== Contact
+
+Ben Ashbaugh, Intel (ben 'dot' ashbaugh 'at' intel 'dot' com)
+
+== Contributors
+
+// spell-checker: disable
+Ben Ashbaugh, Intel +
+Maciej Dziuban, Intel +
+Rafik Saliev, Intel
+// spell-checker: enable
+
+== Notice
+
+Copyright (c) 2021 Intel Corporation. All rights reserved.
+
+== Status
+
+Shipping
+
+== Version
+
+Built On: {docdate} +
+Version: 1.0.0
+
+== Dependencies
+
+This feature is written against the OpenCL API Specification Version V3.0.6.
+
+This extension requires OpenCL 1.0.
+
+== Overview
+
+This extension can be used to query additional information about Intel OpenCL devices.
+The additional information may be useful to tailor a specific workload to the properties of the device.
+
+== New API Functions
+
+None.
+
+== New API Enums
+
+Possible values accepted as the _param_name_ parameter of *clGetDeviceInfo*, depending on the device type and the extension version.
+Additional queries may be added in subsequent versions of the extension:
+
+[source]
+----
+/* For GPU devices, version 1.0.0: */
+
+#define CL_DEVICE_IP_VERSION_INTEL                0x4250
+#define CL_DEVICE_ID_INTEL                        0x4251
+#define CL_DEVICE_NUM_SLICES_INTEL                0x4252
+#define CL_DEVICE_NUM_SUB_SLICES_PER_SLICE_INTEL  0x4253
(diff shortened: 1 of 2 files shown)


extensions/cl_intel_device_attribute_query.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+        If the implementation is driven primarily by a PCI device with a PCI device ID, the low 16 bits of the device identifier must contain that PCI device ID, and the remaining bits must be set to zero.
+        Otherwise, the choice of what to return may be dictated by operating system or platform policies - but should uniquely identify both the device version and any major configuration options (for example, core count in the case of multi-core devices).
+| {CL_DEVICE_NUM_SLICES_INTEL}
+  | `cl_uint`
+      | The number of slices in the device.
+        A slice is a collection of sub-slices.
+| {CL_DEVICE_NUM_SUB_SLICES_PER_SLICE_INTEL}
+  | `cl_uint`
+      | The maximum number of sub-slices in one slice.
+        A slice is a collection of execution units (EUs), fixed-function units, and caches.
```
jlewis-austin, line 144: Slices are defined two different ways, while sub-slices aren't explicitly defined here. Maybe move this definition up to `CL_DEVICE_NUM_SLICE_INTEL` and here talk about what a sub-slice is? I'm not sure if it's relevant or not, but I was curious if the sub-slices are assumed or required to be the equivalent (ie, all contain the same numbers and types of units and caches).

extensions/cl_intel_device_attribute_query.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+      | The maximum number of threads that may be simultaneously resident on one execution unit (EU).
+| {CL_DEVICE_FEATURE_CAPABILITIES_INTEL}
+  | {cl_device_feature_capabilities_intel_TYPE}
+      | Feature flags describing capabilities supported by the device.
+|====
+--
+
+== Issues
+
+. How to express cache sizes?  Is for the entire device?
```
jlewis-austin, line 160: Second sentence could use an edit. Maybe refer to the scope of the cache(s) being queried?

jlewis-austin (approved):
LGTM, just a couple editorial ideas

extensions/cl_intel_device_attribute_query.asciidoc
alycm:
```suggestion
        A sub-slice is a collection of execution units (EUs), fixed-function units, and caches.
```
I had a similar thought as @jlewis-austin when reading this.  No idea if my suggestion is correct, but it seems to fit the context.

extensions/cl_intel_device_attribute_query.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+Additional feature flags may be added in subsequent versions of the extension.
+
+[source]
+----
+typedef cl_bitfield         cl_device_feature_capabilities_intel;
+
+/* For GPU devices, version 1.0.0: */
+
+#define CL_DEVICE_FEATURE_FLAG_DP4A_INTEL         (1 << 0)
+#define CL_DEVICE_FEATURE_FLAG_DPAS_INTEL         (1 << 1)
```
alycm, line 101: It is not required for merging, but pointing out that there is no description of these capabilities in this document.

alycm (approved):
LGTM, neither of my comments are blockers.

extensions/cl_intel_device_attribute_query.asciidoc
bashbaug: Yes, it should be:

```suggestion
. How to express cache sizes?  Is it for the entire device?
```

Thanks!

extensions/cl_intel_device_attribute_query.asciidoc
bashbaug: Thank you for catching this, that's a little embarassing!

Alastair's suggested fix is correct.  I've applied it.

extensions/cl_intel_device_attribute_query.asciidoc
bashbaug: Yes - we will either document these in an updated version of the spec or in a related implementation doc.  Thanks!

PR #662, opened 2021-08-25
move asciidoctor attributes to a separate file for ref pages

bashbaug:
A few of the OpenCL specifications were defining their own asciidoctor attributes.  This PR:

* Moves all asciidoctor attributes to a separate file for easier re-use and includes it instead.
* Updates the reference page tooling to also include the new file.

This PR has no affect on the generated specifications but it does fix a few places where the attributes were not being correclty substituted in the reference pages.

We could put these attributes into the existing "attribs.txt", but this looked like a shared file, so I didn't want to modify it.


kpet (approved):

bashbaug:
Merging - this may benefit 3.0 refpage work (tagging @oddhack).

PR #667, opened 2021-09-06
Fixed documentation about __constant initialization

OleStrohm:
This fixes the documentation about constructing objects in the __constant address space with user defined constructors. It also updates the example to make it compile in the latest versions.

### DIFF
--- a/cxx4opencl/address_spaces.txt
+++ b/cxx4opencl/address_spaces.txt
@@ -368,15 +368,16 @@ kernel void foo() {
 }
 ------------
 
-User defined constructors are not allowed to construct objects in `__constant`
-address space. Such objects can be initialized using literals and
-initialization lists if they do not require any user defined conversions.
+User defined constructors in `__constant` address space must be `constexpr`.
+Such objects can be initialized using literals and initialization lists if they
+do not require any user defined conversions.
 
 Objects in `__constant` address space can be initialized using:
 
  * Literal expressions;
  * Uniform initialization syntax `{}`;
  * Using implicit constructors.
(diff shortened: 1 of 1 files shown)


cxx4opencl/address_spaces.txt
```diff
@@ -386,14 +387,12 @@ struct C1 {
...
 };
 
-kernel void k() {
-  __constant C1 cobj1 = {1};
-  __constant C1 cobj2 = C1();
-  __constant C2 cobj3 = {1}; // error: user defined constructor can't be used
-}
+__constant C1 cobj1 = {1};
+__constant C1 cobj2 = C1();
+__constant C2 cobj3(1);
```
AnastasiaStulova, line 395: Would it make sense to add one negative example where the ctor is invalid which I think was the intent of the original example?

AnastasiaStulova: The original example doesn't seem to compile due to the clang bug but I think it should also be here to illustrate the restriction?
https://godbolt.org/z/Y53joTYen

AnastasiaStulova: I have filed a bug llvm.org/PR52437 for this. Once it's fixed we should finalise this PR.

AnastasiaStulova: This issue seems to be specific to the implementation of OpenCL in Compiler Explorer and not clang. If C++ mode is used the example works correctly:
https://godbolt.org/z/czvYWvK1K

AnastasiaStulova (approved):
LGTM! Thanks

PR #671, opened 2021-09-12
OpenCL 3.0 Refpages

oddhack:
Import relevant changes from update-refpages-2.2-from-0a92eee branch

Update copyrights on relevant files to use SPDX license identifiers
(should do this for all files and make repository REUSE-compliant, but
that's a different issue).

Update refpage generation scripts to use 3.0 specs / version numbers
instead of 2.2.


man/rewritehead
```diff
@@ -70,3 +72,24 @@ RewriteRule ^cl_sampler_properties.html$ enums.html
...
+RewriteRule ^cl_platform_id.html$ abstractDataTypes.html
+RewriteRule ^cl_program.html$ abstractDataTypes.html
+RewriteRule ^cl_sampler.html$ abstractDataTypes.html
+
+# Data types used in APIs, but not otherwise described in the specification
+RewriteRule ^cl_pipe_properties.html$ clCreatePipe.html
+RewriteRule ^cl_queue_properties.html$ clCreateCommandQueueWithProperties.html
+
+# Pages with no documentation at all, yet
+RewriteRule ^clCreateImageWithProperties.html$ deadLinks.html
```
bashbaug, line 93: This should be able to go to the `clCreateImage` page, which also documents `clCreateImageWithProperties`.

https://www.khronos.org/registry/OpenCL/sdk/3.0/docs/man/html/clCreateImage.html

I'm surprised we don't have a similar issue with `clCreateBufferWithProperties`?

bashbaug (approved):
Thanks, this looks really good!

I found one minor issue that we could fix now or could fix later.  I'm inclined to merge what we have now as-is and fix `clCreateImageWithProperties` when we tidy up any of the other 2.2 -> 3.0 changes.

bashbaug:
@oddhack are you OK merging these changes as-is?  Mostly confirming because it's still marked WIP.  Thanks!

oddhack:
> @oddhack are you OK merging these changes as-is? Mostly confirming because it's still marked WIP. Thanks!

Sure, we could merge now and go for incremental improvements. I just don't have the deep CL knowledge to fill in such missing content as might exist.

man/rewritehead
oddhack: I think there's a slight problem in the build toolchain - the refpage is tagged as an alias of clCreateImage but doesn't seem to be getting put into .htaccess as a redirect as it should. I'll take a look at that shortly.

oddhack: Fixed.

oddhack:
@bashbaug please go ahead and merge when you're ready, then let me know and I'll generate 3.0 refpage HTML for the sdk area. I removed the 3.0 page drafts I had put on the webserver (not checked in) for review, so the auto-pull of the new sdk/3.0 directory will work.

alycm (approved):
LGTM, this is a second review so I'll merge immediately.

I did spot an issue with some of the changed URLs, but the issue pre-exists this PR so I've just filed an issue.  https://github.com/KhronosGroup/OpenCL-Docs/issues/681

alycm:
> @bashbaug please go ahead and merge when you're ready, then let me know and I'll generate 3.0 refpage HTML for the sdk area. I removed the 3.0 page drafts I had put on the webserver (not checked in) for review, so the auto-pull of the new sdk/3.0 directory will work.

@oddhack This is now merged.

PR #679, opened 2021-09-24
Use correct names for clustered reduce functions in SPIR-V mapping table

StuartDBrady:


bashbaug (approved):
That's embarassing - thank you for finding and fixing this!

I think we renamed these functions at the last minute and my search-and-replace didn't find these names.  It'd be really great to have a better solution to the awkward wrapping...

PR #684, opened 2021-09-28
Relax memory object acquire error checking requirements in EGL and D3D extensions

### LINKED ISSUE #669: Memory object acquire error checking in cl_khr_egl_image
The cl_khr_egl_image extension specification currently states:

>OpenCL memory objects created from EGL resources must be acquired before they can be used by
any OpenCL commands queued to a command-queue. If an OpenCL memory object created from a
EGL resource is used while it is not currently acquired by OpenCL, the call attempting to use that
OpenCL memory object will return CL_EGL_RESOURCE_NOT_ACQUIRED_KHR.

This text could be clarified and I'm assuming that what is meant by "the call [...] CL_EGL_RESOURCE_NOT_ACQUIRED_KHR" is that

> commands attempting to use that OpenCL memory object will fail execution and their associated event's execution status will be set to `CL_EGL_RESOURCE_NOT_ACQUIRED_KHR`.

Since cheking that memory objects have been acquired comes at a non-negligible complexity and performance cost in some scenarios and this behaviour is not currently tested (https://github.com/KhronosGroup/OpenCL-CTS/issues/1321), I'd like to propose we relax the specification as follows:

>OpenCL memory objects created from EGL resources must be acquired before they can be used by
any OpenCL commands queued to a command-queue. If an OpenCL memory object created from a
EGL resource is used while it is not currently acquired by OpenCL, the behavior is undefined. Implementations may fail the execution of commands attempting to use that OpenCL memory object and set their associated event's execution status to `CL_EGL_RESOURCE_NOT_ACQUIRED_KHR`.

kpet:
Fixes #669

Change-Id: Idbc49280d3eb49ae4af27838eb60c7022b0e7981

### DIFF
--- a/ext/cl_khr_d3d10_sharing.asciidoc
+++ b/ext/cl_khr_d3d10_sharing.asciidoc
@@ -678,8 +678,10 @@ with the OpenCL context.
 OpenCL memory objects created from Direct3D 10 resources must be acquired
 before they can be used by any OpenCL commands queued to a command-queue.
 If an OpenCL memory object created from a Direct3D 10 resource is used while
-it is not currently acquired by OpenCL, the call attempting to use that
-OpenCL memory object will return CL_D3D10_RESOURCE_NOT_ACQUIRED_KHR.
+it is not currently acquired by OpenCL, the behavior is undefined.
+Implementations may fail the execution of commands attempting to use that
+OpenCL memory object and set their associated event's execution status to
+{CL_D3D10_RESOURCE_NOT_ACQUIRED_KHR}.
 
 If CL_CONTEXT_INTEROP_USER_SYNC is not specified as CL_TRUE during context
 creation, *clEnqueueAcquireD3D10ObjectsKHR* provides the synchronization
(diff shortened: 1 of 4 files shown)


bashbaug (approved):
LGTM.

If you feel up to it I think the text for `cl_khr_gl_sharing` could be clarified similarly, something like:

> These objects need to be acquired before they can be used by any OpenCL
commands queued to a command-queue **or the behavior is undefined**.

alycm (approved):
Text LGTM, I have no opinion on the technicality of the change.

Minutes for 2021-09-28 call say:

> maybe wait until we have tooling before we merge

Do we want to wait?  I'm assuming no one is currently writing that.

kpet:
@bashbaug I've made the proposed change to the `cl_khr_gl_sharing` specification

@alycm I don't think we want to wait. My suggestion would be to create an issue on the layers repo (once we've transfered it, I'll start a Slack discussion about this)

Kerilk:
@kpet given that the first strategy didn't yield anything concrete, I reverted to used a more crude approach. This must be pretty close semantically from what you would get if you had a flag on the cl_mem object that you would set to remember if it was acquired. This is still a WIP, and I am not sure I have time to build a test case for this one, but I intend to complete it.
 - I need to add the infrastructure to store kernel arguments, but this is not different semantically from the clEnqueueCopyImage, and not very complicated, as I do it in other layers.
 - I need to add support for the other types of acquired image, but there shouldn't be much difference, just a type to carry around
 - I need to store more meta data to give more context so the user can more easily identify where the problem was
If you have some use cases ready and want to give it a spin (the code is untested yet, so no promise), feel free.

https://github.com/Kerilk/OpenCL-Layers/tree/object-acquire

PR #685, opened 2021-09-30
Reserve 4 bits in cl_mem_flags for Qualcomm

lakshmih:


bashbaug (approved):
LGTM.

We still have plenty of  breathing room but we've almost exhausted half of the 64 mem flags bits.  We should start thinking about what to do if we eventually run out.  May be worth a quick discussion in the next teleconference.

alycm (approved):
LGTM.

alycm:
> We still have plenty of breathing room but we've almost exhausted half of the 64 mem flags bits. We should start thinking about what to do if we eventually run out. May be worth a quick discussion in the next teleconference.

Yes, I can see that will be a problem at some point.  I don't think there is any need to delay the current two PRs requesting bits, but it would be good to add any guidance soon so that the next extension writers can design it in from the start.

PR #687, opened 2021-10-04
reserve cl_mem_flags bits for an upcoming Intel extension

bashbaug:
Please reserve two cl_mem_flags bits for an upcoming Intel extension.

Note, we have started to move functionality that we would have traditionally expressed via cl_mem_flags to properties instead in order to conserve cl_mem_flags bits, but for this case a significant amount of internal code has already been written to use cl_mem_flags, so we would prefer to continue using cl_mem_flags.

If both this PR and #685 is merged we will have allocated 32 of the possible 64 cl_mem_flags bits.


alycm (approved):
LGTM, comment on remain namespace here: https://github.com/KhronosGroup/OpenCL-Docs/pull/685#issuecomment-933869368

PR #688, opened 2021-10-06
add recommendation for bits in bitfields

bashbaug:
As discussed in the October 4th teleconference, adds a comment recommending usage of properties instead of bits int a bitfield to the XML file and the extension template, which may help to conserve bitfield bits.


alycm (approved):
LGTM.

Issue #626, opened 2021-06-16
Subgroup Ballots and Remainder Subgroups


bashbaug:
The extended subgroup ballot functions `sub_group_ballot_bit_count`, `sub_group_ballot_find_lsb`, and `sub_group_ballot_find_msb` are currently written to only consider bits in the ballot corresponding to work-items in the subgroup.  Example for `sub_group_ballot_bit_count`:

> Returns the number of bits that are set in the bitfield value, only considering the bits in value that represent predicate values from all work items in the subgroup.

The size of the subgroup isn't known at compile-time though, so an implementation may need special code to consider the correct number of bits in the bitfield at runtime.  For some implementations, it would be more efficient to consider the **maximum** subgroup size instead, which is known at compile-time.

**Are these functions intended to work on the maximum subgroup size instead (possibly improving efficiency), or is it important that they only work on the work-items in the subgroup?**

Note that for most typical usages that operate on ballots obtained via subgroup built-ins there is no difference in functionality between the two.  This is because a subgroup ballot bit is set if and only if a) the work-item is active, and b) the predicate is true.  Therefore, the ballot bits corresponding to work-items in "full" subgroups but not in the "remainder" subgroup are unset and hence do not modify the bit count nor affect the least or most significant bit set:

```c
// This will return the same count regardless whether the mask bits
// corresponding to work-items in a "partial" subgroup are considered or
// whether mask bits for "full" subgroups are considred unconditionally.
uint4 my_ballot = sub_group_ballot(my_condition);
uint my_count = sub_group_ballot_bit_count(my_ballot);
```

If the ballot is manipulated manually, the results can be different.  Consider, for example:

```c
uint4 my_manipulated_ballot = ~0; // all bits set

// Is this the subgroup size or the max subgroup size?
uint my_count = sub_group_ballot_bit_count(my_ballot);
```

This doesn't seem to be a common case though.

A few other brief notes worth mentioning:

* I initially thought that the ballot scans could be affected also, but since they only consider bits up to this work-item's subgroup local ID I don't think they are affected.
* Vulkan has a similar issue in theory, although subgroups in Vulkan are a little different, and a remainder subgroup in Vulkan has the same subgroup size as a full subgroup it's just that some of the work-items in the subgroup are inactive by default ([link](https://www.khronos.org/blog/vulkan-subgroup-tutorial)).
[...]

PR #689, opened 2021-10-08
Clarify ballot ops for remainder subgroups

StuartDBrady:
As discussed in issue #626, the wording used to describe the `sub_group_ballot_bit_count()` and `sub_group_ballot_find_msb()` operations included the phrase "only considering the bits in _value_ that represent predicate values from all work items in the subgroup", but "remainder" subgroups were not considered when arriving at this wording.

In Vulkan, the subgroup size is defined in a way that is equivalent to the definition of the maximum subgroup size in OpenCL.  Therefore, in Vulkan, "remainder" subgroups start with work-items that are initially inactive, but are still considered work-items within the subgroup.

The ballot operations in OpenCL were intended to be equivalent to those in Vulkan.  Therefore, for a "remainder" subgroup, predicate bits that would correspond to non-existent work-items that would be considered initially inactive in Vulkan should still contribute to ballot operations that concern predicate bits and are always required to be well-defined.

The `sub_group_ballot_bit_count()` and `sub_group_ballot_find_msb()` operations are affected by this.

Due to the manner of their operation, the `sub_group_ballot_find_lsb()`, `sub_group_ballot_inclusive_scan()` and `sub_group_ballot_exclusive_scan()` operations are not affected.  However, as `sub_group_ballot_find_lsb()` is described using the same wording, and altering the wording does not affect its semantics, it has been updated for consistency.


bashbaug (approved):
I think this looks OK and we can merge this anytime if needed.

A few possible improvements to consider:

* I think the description is correct but it took me a bit to convince myself there wasn't an off-by-one error.  My confusion was around "subgroup local IDs less than the maximum subgroup size", where it could be "less than" or "less than or equal to".  I wonder if it makes more sense to describe the number of predicate bits that are valid vs. the predicate values corresponding to subgroup local IDs?

* Since some devices may support different maximum subgroup sizes it may be worth disambiguating "the maximum subgroup size" by adding text like "the maximum subgroup size within the dispatch", or stating explicitly that the maximum subgroup size is the value returned by `get_max_sub_group_size`.

Thanks!

StuartDBrady:
> I think this looks OK and we can merge this anytime if needed.
> 
> A few possible improvements to consider:
> 
> * I think the description is correct but it took me a bit to convince myself there wasn't an off-by-one error.  My confusion was around "subgroup local IDs less than the maximum subgroup size", where it could be "less than" or "less than or equal to".  I wonder if it makes more sense to describe the number of predicate bits that are valid vs. the predicate values corresponding to subgroup local IDs?

I agree, but I cannot find a wording that achieves this.  If you can think of a wording, please feel free to suggest this.  Otherwise, would it be okay to accept the change as worded, and reopen issue #626 or raise a new one to improve this wording?

> * Since some devices may support different maximum subgroup sizes it may be worth disambiguating "the maximum subgroup size" by adding text like "the maximum subgroup size within the dispatch", or stating explicitly that the maximum subgroup size is the value returned by `get_max_sub_group_size`.

I'm happy to say "maximum subgroup size within the dispatch (as returned by `get_max_sub_group_size`)" if that would help?  I was a bit concerned about being overly verbose but I agree it's good to be unambiguous here.

bashbaug:
> I'm happy to say "maximum subgroup size within the dispatch (as returned by `get_max_sub_group_size`)" if that would help?

Yes, would you mind making this change?

I'm also fine merging as-is if this is time critical - let me know.  Thanks!

(I've added the needs-cts-coverage tag also, because I think the CTS will need to be modified to test this?)

StuartDBrady:
@bashbaug, I notice my force-push hasn't cleared your existing review.

Leaving this for you to merge, assuming you're happy to do so.

PR #691, opened 2021-10-12
cosmetic updates for the OpenCL specifications

bashbaug:
This PR contains cosmetic updates to the OpenCL specifications.  There should be no content changes.

Specific changes are:

* Switched to the new Khronos stylesheets to fix accessibility issues.
* Switched to the "rouge" source highligher instead of the previous "coderay" source highlighter.  Please note that only the OpenCL API, OpenCL C, OpenCL Extensions, and OpenCL SPIR-V Environment specs were updated.  I have not updated the OpenCL C++ kernel language, the C++ for OpenCL, or any vendor extension specs at this time.
* Added customized rouge source highlighting for OpenCL API types and OpenCL C keywords.  This is easy to further customize, if desired, either by modifying the colors or styles or by adding additional keywords or types.

To more easily review these changes, please view the specs on my GitHub pages, which were built from this branch.  For example:

* https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html
* https://bashbaug.github.io/OpenCL-Docs/pdf/OpenCL_C.pdf

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
   - gem install rouge -v 3.19.0
   - gem install ttfunk -v 1.5.1
   - gem install asciidoctor-pdf -v 1.5.0
-  - MATHEMATICAL_SKIP_STRDUP=1 gem install asciidoctor-mathematical -v 0.2.2
+  - gem install asciidoctor-mathematical -v 0.3.5
 
 script:
   - git describe --tags --dirty
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,9 @@ rvm:
 
 before_install:
   - sudo apt-get install -y libpango1.0-dev ghostscript fonts-lyx
-  - gem install asciidoctor -v 1.5.8
+  - gem install asciidoctor -v 2.0.16
   - gem install coderay -v 1.1.1
+  - gem install rouge -v 3.19.0
   - gem install ttfunk -v 1.5.1
   - gem install asciidoctor-pdf -v 1.5.0
   - MATHEMATICAL_SKIP_STRDUP=1 gem install asciidoctor-mathematical -v 0.2.2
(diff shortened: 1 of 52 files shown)


alycm (approved):
LGTM, I did random spot checks across the specs, I didn't see anything out of place.

I particularly like the highlighting on entry points in the API spec.

The recently added extension chapters could also use the new syntax highlighting, that doesn't need to happen in this PR though.

keryell:
Why is there both `opencl.rb` and `opencl_c.rb`?
At least it should be explained in the comments what they are really doing.

bashbaug:
The "opencl" source type is for OpenCL API syntax highlighting, so it primarily includes the OpenCL API types for things like `cl_platform_id`, `cl_context`, `cl_command_queue`, etc.

The "opencl_c" source type is for the OpenCL C programming language, so it includes things like the `__kernel` and `__global` keywords, types like `char2` and `image2d_t`, and built-in functions like `get_global_id`.

I'm totally open to clarifying this further.  Where would you like to see more comments?

keryell:
The difference make sense.
But reading the `.rb` should make it clear:
```ruby
    class OpenCL < Cpp
      title "OpenCL"
      desc "The OpenCL standard for heterogeneous computing from the Khronos Group"
```
perhaps with `OpenCL host API` instead of just `OpenCL`?
Also clarify the tags?
The comments are always useful at the top of the `.rb` to explain what is happening in the file but probably that if the `title`, `desc` and `tag` are clearer one can understand.
In the host API, it is probably the C or the C++ host API, then it would be nice to have also the API calls highlighted?
At the end I realize that having a single-source language standard like SYCL brings some simplification here too by not requiring 2 source highlighters for the host part and the device part. ;-)

bashbaug:
I created https://github.com/KhronosGroup/OpenCL-Docs/issues/715 to better track this request.  Thanks!

PR #692, opened 2021-10-13
Elaborate on C++ for OpenCL preprocessor macros

svenvh:


extensions/cl_ext_cxx_for_opencl.asciidoc
```diff
@@ -66,6 +66,15 @@ Accepted value for the _param_name_ parameter to *clGetDeviceInfo*:
...
 ----
 
+== Preprocessor Macros
+
+This extension defines a new language, instead of extending an existing
+language. As such, there will be no preprocessor `#define` matching the
+extension name string. Instead, dedicated preprocessor macros conveying
+language version information are available as described in the C++ for
+OpenCL Programming Language Documentation, section 2.2.2.2 "Predefined
+macros".
```
AnastasiaStulova, line 76: do you think it would make sense to explicitly list the relevant macros `__cplusplus` and `__OPENCL_CPP_VERSION__`?

AnastasiaStulova: Also do you think we could refer to C++ predefined macros section i.e. **C++17 [cpp.predefined]**?

AnastasiaStulova (approved):

PR #693, opened 2021-10-14
fix the OpenCL logo for the online reference pages

bashbaug:
This fixes a broken image on the front page of the OpenCL 3.0 reference pages and updates it to the new OpenCL logo.

I'm open to suggestions if there is a less intrusive fix for this issue but I couldn't figure out how to do it without an asciidoctor attribute to specify the image path.


Makefile
```diff
@@ -453,7 +454,7 @@ $(REFPATH)/apispec.txt: $(SPECFILES) $(GENREF) $(SCRIPTS)/reflib.py $(GENERATED)
 # running the recursive make, so it doesn't trigger twice
 manhtmlpages: $(REFPATH)/apispec.txt $(GENDEPENDS)
 	$(MAKE) -e buildmanpages
-	$(CP) $(MANDIR)/*.html $(MANDIR)/*.css $(MANDIR)/*.gif $(MANDIR)/*.jpg $(MANHTMLDIR)
+	$(CP) $(MANDIR)/*.html $(MANDIR)/*.css $(MANDIR)/*.gif $(MANHTMLDIR)
```
oddhack, line 457: I'm unsure why this was removed. Is the .jpg getting inlined, or sourced from somewhere else?

When we build the Vulkan HTML we inline CSS and images, using some attributes I'd have to look up, to reduce external dependencies as much as possible (unfortunately, KaTeX is still an external dependency, which reminds me that we never finished setting up the redirects on the registry repo - will look at James' last comments again tonight).

oddhack (approved):
LGTM.

Makefile
bashbaug: > Is the .jpg getting inlined, or sourced from somewhere else?

Yes, it seems to be inlined.

Since I removed the old .jpg from the directory there are no .jpgs left and the makefile was barfing.

I thought about putting the new .jpg in the directory but since it seemed to be unnecessary I chose to update the makefile instead.  This does mean we'll need to add it back if we start using .jpgs that aren't inlined, but I suppose that's OK?

oddhack: Should be fine. AFAIK either all the images get inlined, or none of them.

PR #705, opened 2021-10-29
Specify what state kernel objects are in after a failed call to clSetKernelArg

### LINKED ISSUE #548: What state does a failed clSetKernelArg call leave kernel arguments in?
The specification apparently says nothing about the state a failed call to `clSetKernelArg` leaves kernel arguments in? Are they (a few ideas):
1. Set to their previous value (or left unset)? This would put a high burden on implementations.
1. Forced unset? Probably the easiest for implementations and applications.
1. In an implementation-defined/undefined state? Leaves a lot of questions open. Is it valid to enqueue the kernel? Can applications ever be confident that a previously set argument still is? Is `clEnqueueNDRangeKernel` guaranteed to fail cleanly if arguments are invalid?

kpet:
Fixes #548

Change-Id: Ibcc63c7f6e19bff4169c22e16f3ff4dacf2b6e25

### DIFF
--- a/api/footnotes.asciidoc
+++ b/api/footnotes.asciidoc
@@ -162,3 +162,7 @@ Only once the ID has been allotted may it be exposed to OpenCL by proposing a me
 The merge must define a new enumerant by adding an `<enum>` tag to the {cl_khronos_vendor_id_TYPE} `<enums>` tag, with the `<value>` attribute set as the acquired Khronos vendor ID. \
 The `<name>` attribute must identify the vendor/adopter, and be of the form `CL_KHRONOS_VENDOR_ID_<vendor>`. \
 ]
+
+:fn-setkernelarg-prefer-unset-on-error: pass:n[ \
+Implementations are encouraged to favor this option as it makes it more likely that errors will be managed by applications. \
+]
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -7228,8 +7228,7 @@ of _kernel_ is modified, it is implementation-defined whether:
   * The argument value that was previously set is kept so that it can be used in
     further kernel enqueues.
   * The argument value is unset such that a subsequent kernel enqueue fails with
-    {CL_INVALID_KERNEL_ARGS}. Implementations are encouraged to favour this option
-    as it makes it more likely that errors will be managed by applications.
+    {CL_INVALID_KERNEL_ARGS}. footnote:[{fn-setkernelarg-prefer-unset-on-error}]
 --
 
 [open,refpage='clSetKernelArgSVMPointer',desc='Set a SVM pointer as the argument value for a specific argument of a kernel.',type='protos']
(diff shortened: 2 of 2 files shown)


jrprice (approved):

jrprice:
If we don't already, we should make sure we have a CTS issue to cover the error cases where the argument is guaranteed to be unmodified.

bcalidas:
Overall the update looks good. I do have some minor concerns about this text  "Implementations are encouraged to favour this option as it makes it more likely that errors will be managed by applications".

Is this text necessary.

api/opencl_runtime_layer.asciidoc
```diff
@@ -7219,6 +7219,17 @@ Otherwise, it returns one of the following errors:
...
+
+When {clSetKernelArg} returns an error code different from {CL_SUCCESS}, the
+internal state of _kernel_ may only be modified when that error code is
+{CL_OUT_OF_RESOURCES} or {CL_OUT_OF_HOST_MEMORY}. When the internal state
+of _kernel_ is modified, it is implementation-defined whether:
+
+  * The argument value that was previously set is kept so that it can be used in
+    further kernel enqueues.
+  * The argument value is unset such that a subsequent kernel enqueue fails with
+    {CL_INVALID_KERNEL_ARGS}. Implementations are encouraged to favour this option
```
bashbaug, line 7231:
```suggestion
    {CL_INVALID_KERNEL_ARGS}. Implementations are encouraged to favor this option
```

Minor: If we decide to keep this sentence then I think we should use the American English spelling.  If we decide to keep it should it be a note or a footnote instead?

bashbaug (approved):

alycm (approved):
LGTM.

CTS issue filed: https://github.com/KhronosGroup/OpenCL-CTS/issues/1356

kpet:
@alycm Thanks for creating the CTS issue.

I've changed the recommendation into a footnote and used American spelling. PTAL.

bashbaug (approved):

api/footnotes.asciidoc
```diff
@@ -162,3 +162,7 @@ Only once the ID has been allotted may it be exposed to OpenCL by proposing a me
 The merge must define a new enumerant by adding an `<enum>` tag to the {cl_khronos_vendor_id_TYPE} `<enums>` tag, with the `<value>` attribute set as the acquired Khronos vendor ID. \
 The `<name>` attribute must identify the vendor/adopter, and be of the form `CL_KHRONOS_VENDOR_ID_<vendor>`. \
 ]
+
+:fn-setkernelarg-prefer-unset-on-error: pass:n[ \
```
alycm, line 166: The top of this file says

```c
// Please keep footnotes in alphabetical order!
```

But this can be moved afterwards, I don't want to block merging.

bashbaug: I'll fix this.  If there are no other issues I think we should merge this now.

kpet: Sorry I missed this.

PR #706, opened 2021-10-29
Explain the location of language doc in cl_ext_cxx_for_opencl.

AnastasiaStulova:
C++ for OpenCL documentation is not published in the Registry
like all other documents and therefore there is a need to explain
where the documentation can be found.


svenvh (approved):
LGTM

PR #709, opened 2021-11-08
add the spec source for cl_intel_unified_shared_memory

bashbaug:
This PR adds the spec source and a few minor XML updates for the latest version of the `cl_intel_unified_shared_memory` extension.

I will create an OpenCL-Headers PR with the minor updates for this extension shortly.

I will create an OpenCL-Registry PR after the spec source is merged.

### DIFF
--- a/extensions/cl_intel_unified_shared_memory.asciidoc
+++ b/extensions/cl_intel_unified_shared_memory.asciidoc
@@ -465,6 +465,7 @@ Otherwise, `NULL` will be returned, and _errcode_ret_ will be set to one of the 
 * `CL_INVALID_VALUE` if _alignment_ is not zero or a power of two.
 * `CL_INVALID_VALUE` if _alignment_ is greater than the size of the largest data type supported by any OpenCL device in _context_ that supports host Unified Shared Memory allocations.
 * `CL_INVALID_PROPERTY` if a memory property name in _properties_ is not a supported property name, if the value specified for a supported property name is not valid, or if the same property name is specified more than once.
+* `CL_INVALID_PROPERTY` if either the `CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL` or `CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL` flags are specified.
 * `CL_INVALID_BUFFER_SIZE` if _size_ is zero or greater than `CL_DEVICE_MAX_MEM_ALLOC_SIZE` for any OpenCL device in _context_ that supports host Unified Shared Memory allocations.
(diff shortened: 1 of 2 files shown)


extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -0,0 +1,1264 @@
...
+    | Flags specifying allocation and usage information.
+      This is a bitfield type that may be set to a combination of the following values:
+
+      `CL_MEM_ALLOC_WRITE_COMBINED_INTEL`:
+      Request write combined (WC) memory.
+      Write combined memory may improve performance in some cases, however write combined memory must be used with care since it may hurt performance in other cases or use different coherency protocols than non-write combined memory.
+
+      `CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL`:
+      Request the implementation to optimize for first access being done by the device.
+      This flag is valid only for *clSharedMemAllocINTEL*.
```
EwanC, line 633: This states that the flag is valid only for `clSharedMemAllocINTEL`, but I only see error wording for it's usage under `clHostMemAllocINTEL`. Should this wording be under `clSharedMemAllocINTEL` ?

> CL_INVALID_PROPERTY if both CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL and CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL flags are specified.

bashbaug: Good catch, yes, this error condition should definitely be under `clSharedMemAllocINTEL` and not `clHostMemAllocINTEL`.  I've moved it.

I'm also following up to see if any error conditions also apply to `clHostMemAllocINTEL` and `clDeviceMemAllocINTEL` or if the "valid only for `clSharedMemAllocINTEL`" phrase needs to be weakened.

Thanks for the careful review!

bashbaug: I've also added similar error conditions for `clHostMemAllocINTEL` and `clDeviceMemAllocINTEL` if either flag is specified:

> * `CL_INVALID_PROPERTY` if either the `CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL` or `CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL` flags are specified.

I don't think our implementation is checking for these conditions right now, but I think an error makes the most sense since the initial placement flags do not apply to host or device allocations.

Thanks again!

EwanC: Makes sense, thanks for following up about this

EwanC (approved):

PR #710, opened 2021-11-10
WIP cl_ext_image_tiling_control

kpet:
Change-Id: I6c391b5ecaa203f7db566db68a7e3d124d6038a2

### DIFF
--- a/extensions/cl_ext_image_tiling_control.asciidoc
+++ b/extensions/cl_ext_image_tiling_control.asciidoc
@@ -1,0 +1,211 @@
+// Copyright 2018-2021 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+:data-uri:
+:icons: font
+include::../config/attribs.txt[]
+include::{generated}/api/api-dictionary-no-links.asciidoc[]
+:source-highlighter: coderay
+
+= cl_ext_image_tiling_control
+
+== Name Strings
+
+`cl_ext_image_tiling_control`
+
+== Contact
+
+Kevin Petit (kevin.petit 'at' arm.com)
+
+== Contributors
+
+Kevin Petit, Arm Ltd. +
+
+== Notice
+
+Copyright (c) 2021 Arm Ltd.
+
+== Status
+
+Draft
+
+== Version
+
+Built On: {docdate} +
+Version: 1.0.0
+
+== Dependencies
+
+This extension is written against the OpenCL Specification version 3.0.9.
+
+This extension requires OpenCL 3.0.
+
+== Overview
+
+This extension gives applications explicit control over how image data is laid out in memory.
+
+== New API Types
+
+[source,c]
+----
+typedef cl_uint cl_image_tiling_ext;
+
+#define CL_IMAGE_TILING_LINEAR_EXT   1
+#define CL_IMAGE_TILING_OPTIMAL_EXT  2
+
+typedef cl_bitfield cl_device_image_tiling_capabilities_ext;
+
+#define CL_DEVICE_IMAGE_TILING_HOST_ACCESS_EXT (1 << 0)
+----
+
+== New API Enums
+
+Accepted value for the _param_name_ parameter to *clGetDeviceInfo*:
+
+[source,c]
+----
+CL_DEVICE_IMAGE_TILING_CAPABILITIES_EXT 0x4231
+----
+
+Accepted value for the _properties_ parameter to *clCreateImageWithProperties*:
+
+[source,c]
+----
+CL_MEM_IMAGE_TILING_EXT       0x4232
+----
+
+Accepted value for the _param_name_ parameter to *clGetImageInfo*:
+
+[source,c]
+----
+CL_IMAGE_TILING_EXT 0x4233
+----
+
+== Modifications to the OpenCL API Specification
+
+(Modify Section 4.2, *Querying Devices*) ::
++
+--
+
+(Add the following to Table 5, _List of supported param_names by *clGetDeviceInfo*_) ::
++
+--
+
+[cols="1,1,4",options="header"]
+|====
+| cl_device_info
+| Return Type
+| Description
+
+| {CL_DEVICE_IMAGE_TILING_CAPABILITIES_EXT}
+| {cl_device_image_tiling_capabilities_ext_TYPE}
+| Returns the image tiling capabilities supported by the device.
+ +
+- {CL_DEVICE_IMAGE_TILING_HOST_ACCESS_EXT} is set when it is allowed to pass
+    {CL_MEM_IMAGE_TILING_EXT} with a value different from {CL_IMAGE_TILING_LINEAR_EXT}
+    in the _properties_ given to *clCreateImageWithProperties* when creating host-accessible images.
+    Host-accessible images are all images created without {CL_MEM_HOST_NO_ACCESS}.
+|====
+
+--
+--
+
+(Modify section 5.3.1, *Creating Image Objects*) ::
++
+--
+{CL_MEM_IMAGE_TILING_EXT} can be passed as part of the _properties_ parameter
+to *clCreateImageWithProperties* to control the tiling used for the image being
+created. The following values are accepted:
+
+  * {CL_IMAGE_TILING_LINEAR_EXT}, which is the default value used if
+    {CL_MEM_IMAGE_TILING_EXT} is not specified in the _properties_. Image data is
+    laid out in so-called raster order, one row after the other, one slice or array
+    layer after the other in memory according to the pitches provided by the application
+    or calculated.
+
+  * {CL_IMAGE_TILING_OPTIMAL_EXT} requires image data be laid out in an implementation-defined
+    format which can vary depending on the type of image, its format and/or dimensions.
+    The will attempt to select the best layout for the image.
+
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT}, {CL_MEM_USE_HOST_PTR} must not be set in _mem_flags_.
+
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT} and {CL_MEM_COPY_HOST_PTR} is set in _mem_flags_, the data
+pointed to by _host_ptr_ is laid out using linear tiling and its layout is described
+in the same way as for case whe {CL_MEM_IMAGE_TILING_EXT} is not set or set to
+{CL_IMAGE_TILING_LINEAR_EXT}.
+
+When creating an image from a buffer, _image_row_pitch_ and _image_slice_pitch_
+must both be `0` if {CL_MEM_IMAGE_TILING_EXT} is passed and set to
+{CL_IMAGE_TILING_OPTIMAL_EXT}. The data in the buffer and its underlying memory
+are reused as-is and the exact behaviour is implementation-defined.
+--
+
+(Modify section 5.3.6, *Mapping Image Objects*) ::
++
+--
+If the image object is created with {CL_MEM_IMAGE_TILING_EXT} set to a value
+different from {CL_IMAGE_TILING_LINEAR_EXT}, the implementation may need to
+allocate memory and perform a copy as part of the map operation. The pointer
+returned to the application will always point to data in linear tiling
+laid out according to the returned _image_row_pitch_ and _image_slice_pitch_.
+--
+
+(Modify section 5.3.7, *Image Object Queries*) ::
++
+--
+
+If the image object is created with {CL_MEM_IMAGE_TILING_EXT} set to a value
+different from {CL_IMAGE_TILING_LINEAR_EXT}, the value returned for
+
+* {CL_IMAGE_ROW_PITCH} is the value that would be returned as _image_row_pitch_
+  by *clEnqueueMapImage* when mapping the entire image.
+* {CL_IMAGE_SLICE_PITCH} is the value that would be returned as _image_slice_pitch_
+  by *clEnqueueMapImage* when mapping the entire image.
+
+The following is added to _Table 22: List of supported param_names by *clGetImageInfo*_:
+--
+
+[cols="1,1,4",options="header"]
+|====
+| Image info
+| Return Type
+| Description
+
+| {CL_IMAGE_TILING_EXT}
+| {cl_image_tiling_ext_TYPE}
+| Return the tiling passed using the {CL_MEM_IMAGE_TILING_EXT} property or
+  {CL_IMAGE_TILING_LINEAR_EXT} if none was passed at image creation time.
+|====
+
+== Interactions with Other Extensions
+
+None.
+
+TODO unpublished KHR extensions
+
+== Conformance tests
+
+None.
+
+== Issues
+
+. How can an application get a view of tiled image data?
++
+--
+*RESOLVED*: An application wishing to get visibility of tiled data can do so by
+creating images from a buffer and mapping the buffer directly.
+--
+
+== Version History
+
+[cols="5,15,15,70"]
+[grid="rows"]
+[options="header"]
+|====
+| Version | Date       | Author       | Changes
+| 1.0.0   | 2021-11-01 | Kevin Petit  | *Initial revision*
+|====
+
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -234,6 +234,8 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_ulong</type>         <name>cl_semaphore_payload_khr</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_external_semaphore_handle_type_khr</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_external_memory_handle_type_khr</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_image_tiling_ext</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_image_tiling_capabilities_ext</name>;</type>
 
             <comment>Structure types</comment>
         <type category="struct" name="cl_dx9_surface_info_khr">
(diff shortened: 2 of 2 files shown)


extensions/cl_ext_image_tiling_control.asciidoc
```diff
@@ -0,0 +1,214 @@
...
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT}, {CL_MEM_USE_HOST_PTR} must not be set in _mem_flags_.
+
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT} and {CL_MEM_COPY_HOST_PTR} is set in _mem_flags_, the data
+pointed to by _host_ptr_ is laid out using linear tiling and its layout is described
+in the same way as for case whe {CL_MEM_IMAGE_TILING_EXT} is not set or set to
+{CL_IMAGE_TILING_LINEAR_EXT}.
+
+When creating an image from a buffer, _image_row_pitch_ and _image_slice_pitch_
```
Jeremy-Kemp, line 145: Are we saying here that the implementation *may* re-order the data in-place to the optimal tiled format, rendering the original buffer effectively unusable by an application?

kpet: No, the next sentence says that the data in the buffer is reused as-is. It means whatever produces the buffer data needs to know the tiling scheme used by the implementation to initialise the buffer. Of course, you can create an image from a buffer without initialising it and use buffer access to get a view of the tiled data if you want.

bashbaug:
Discussed in the October 4th teleconference:

- Do we need to define and document any interactions with YUV extensions (#722)?
- Are there any interactions to consider when creating an image from a buffer, or from external memory?
- It would be helpful to clearly describe the parts of this extension that provide additional functional guarantees beyond "performance hints".

bcalidas:
Some initial thoughts

1) It should be possible for (non-linear) tiled images to be created with CL_MEM_USE_HOST_PTR. I believe the spec already allows for this implicitly.
2) It would be helpful to allow tiled  images to be created with a flag that permits  host access but does not detile the image for host access. Therefore EnqueueMapImage would return a host pointer to the tiled image data. This could be useful in cases where the CPU has to transfer image data between the GPU and another IP block or the GPU and storage.
3) As mentioned earlier we should clarify the interaction of this extension (specifically tiled images) with ext_image_from_buffer and external_memory_import with attention given to the role of row_pitch and slice_pitch.

Thanks,
Balaji Calidas
Qualcomm

bashbaug (commented):
One high-level thought: Especially because this extension leaves "optimal tiling" implementation-defined (this is a good thing), should the controls be a bit more descriptive and a bit less prescriptive?

Said another way, I think some implementations could benefit from this extension even if they don't support "tiling", by assuming that a "linear" image should be optimized for host access and an "optimal" image should be optimized for device access.

One change we could consider is to switch from `CL_IMAGE_TILING_{LINEAR|OPTIMAL}` to `CL_IMAGE_LAYOUT_{LINEAR|OPTIMAL}` (so tiling -> layout) to be a little more general.

Or, if we wanted to go a bit further, we could switch to something like `CL_IMAGE_LAYOUT_{DYNAMIC|STATIC}`, or even `CL_IMAGE_LAYOUT_{HOST|DEVICE}_OPTIMAL`.

What do you think?

bcalidas:
I wanted to leave some high-level thoughts about how this extension could work. 

The major impact of linear vs optimal tiling modes would be when the image is created from a buffer or from external memory. In these cases, if the image has optimal tiling mode, then the application should be required to set row_pitch and/or slice_pitch to zero.

Applications should be able to choose whether an optimally tiled image should present a linear view when mapped for host CPU access. Currently, implementations are implicitly required to present a linear view of image data when mapped for host access. This extension could possibly give implementations the ability to opt out of providing that linear view. How the image data is finally presented when mapped for host access would then depend on both implementation capabilities and application selection.

Ben has suggested that applications make the linear view vs. raw view selection as part of the EnqueueMap command. This could also be specified when creating the image. We'll need to discussion these options further.

Thanks,
Balaji

bashbaug:
> I wanted to leave some high-level thoughts about how this extension could work....

I think the comment above describes one way this extension could work, but not necessarily the only way.

IMHO the key value this extension provides is a hint (or assertion?) to the driver that the image should (or must?) be stored internally in a specific format - either "linear" (well-specified) or "tiled" (implementation-defined, in this extension, at least).  This will primarily change the performance characteristics of an image, especially when it is mapped, though likely also when it is accessed within a kernel.

Once we have the ability to control or otherwise influence the internal layout it gives us the ability to add additional functionality, either in this extension or in other layered extensions.  For example, we could:

1. Influence or mandate an internal layout when creating an image from a buffer or from an external memory handle.
2. Map the image but provide direct access to image data in the internal layout vs. converting to a linear layout.

One of the key questions I think we'll need to answer is whether this extension is purely a performance hint or whether it is something stronger.  For the base-level functionality I think we could go either way, but for the additional functionality (external memory and direct access mapping) I think we'd want stronger guarantees (at least in some cases).

extensions/cl_ext_image_tiling_control.asciidoc
```diff
@@ -0,0 +1,214 @@
...
+--
+
+(Modify section 5.3.1, *Creating Image Objects*) ::
++
+--
+{CL_MEM_IMAGE_TILING_EXT} can be passed as part of the _properties_ parameter
+to *clCreateImageWithProperties* to control the tiling used for the image being
+created. The following values are accepted:
+
+  * {CL_IMAGE_TILING_LINEAR_EXT}, which is the default value used if
```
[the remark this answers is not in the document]
bashbaug: Yes, I agree, and I had a similar comment above.

One possible way to solve this problem is to have three "tiling" states:

1. An explicitly linear image.
2. An explicitly tiled image.
3. A "default" or "unspecified" layout where the implementation can choose whether the image is linear or tiled or something else.

(1) and (2) are already defined by this extension.  (3) would need to be added.  If we added (3) and made it the default value I think it would retain backwards compatibility and avoid any performance penalties?

bcalidas:
I posted some comments on https://github.com/KhronosGroup/OpenCL-Docs/issues/861 which include updates to cl_ext_image_tiling_control. These comments also discuss how cl_ext_image_tiling_control could bring clarity to external memory and image_from_buffer usage.

Thanks,
Balaji Calidas
Qualcomm

bcalidas:
Following up on the discussion around #861 from Jan 10, 2023 We'd like to propose the following outline for his extension.

First we recommend that this extension be made khr since it is foundational to other extensions including external memory.
Image_tiling_control defines 2 new  properties for image tiling.

 CL_IMAGE_TILING_LINEAR_KHR and CL_IMAGE_TILING_OPTIMAL_KHR

These properties can be used when calling clCreateImageWithProperties and bring clarity to the image layout when used.

Optimal tiling can be thought of as a superset of linear tiling. What it does it to let implementations decide how the image should be tiled.   An implementation can continue to use linear tiling even if the application selects CL_IMAGE_TILING_OPTIMAL_KHR With that in mind,  we believe  all implementations should be able to support both linear and optimal tiling.

We also recommend that all implementations be required to support device access to optimally tiled and linear images. It seems that device access is a minimum for the image to be useful. Regarding host access, we propose that the default behavior be to present a linear view of the image data  when clEnqueueMapBuffer is called. This means that implementations will need to detile optimally tiled images ( if needed ) when clEnqueueMapBuffer is called. If this is problematic for some vendors, we can add a cl_device_info param that lets the application know if the implementation can detile optimally tiled images when clEnqueueMapBuffer is called.

For many use cases having direct access to tiled image data is useful. For this reason, we are proposing a  new flag CL_MAP_IMAGE_NO_DETILE_KHR. When used in clEnqueueMapImage, the implementation will not detile the data for host access.

Thanks,
Balaji Calidas
Qualcomm

kpet:
I've rebased the specification draft and captured what I believe to be an exhaustive list of all the open topics in the PR description. We seem to be mostly aligned on where this extension should go.

extensions/cl_ext_image_tiling_control.asciidoc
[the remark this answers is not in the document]
nikhiljnv: I think both having a separate default or optimal covering non-linear as well as linear tiling modes should work.
Default may be unnecessary if optimal allows both.
In case of default or optimal tiling mode, if implementation is free to choose optimal or linear mode, what should clGetImageInfo CL_IMAGE_TILING_EXT query return? Should it return default/optimal tiling/layout or should it return optimal / linear as chosen by implementation?

bcalidas:
We uploaded an update draft that has the following major modifications.

1) Implementations supporting this extension will support both  linear  and optimal  image layouts. Optimal layouts can be mapped to linear by an implementation.
2) Allows for per image query of possible tiling modes, actual tiling mode and behavior of EnqueueMapImage.
3) By default EnqueueMapImage will present a linear view of the image data. However, for some images ( likely extension images as opposed to core image formats )  a linear view will not be guaranteed when mapping the image. Instead the raw image data will be presented.
4) Applying this extension to images that are created without explicitly specifying the tiling - 
   a) Most images will be considered to have optimal tiling. This is the default. When these images are mapped, a linear view of the image data will be presented.
  b) When an image is created from a buffer, the image is considered to have linear tiling if a row pitch and/or slice pitch are specified.

Thanks,
Balaji

kpet:
As agreed in recent teleconferences, I have now:
- Reverted changes made by Qualcomm to the extension draft
- Rebased the extension draft and integrated it into the unified specification

I have also fixed minor typos and issues. We should now be in a good position to resume discussions and enact outcomes.

Doing this work made it pretty clear that we will need to integrate cl_ext_image_requirements_info and cl_ext_image_from_buffer to the unified specification before finalising cl_ext_image_tiling_control if we are to cleanly document interactions. I'll work on doing this in parallel to the discussions on cl_ext_image_tiling_control.

xml/cl.xml
```diff
@@ -1383,6 +1385,17 @@ server's OpenCL/api-docs repository.
         <enum bitpos="2"            name="CL_DEVICE_KERNEL_CLOCK_SCOPE_SUB_GROUP_KHR"/>
             <unused start="3" end="63"/>
     </enums>
+    <enums name="cl_image_tiling_ext" vendor="Multiple">
+            <unused start="0" end="0"/>
+        <enum value="1"             name="CL_IMAGE_TILING_LINEAR_EXT"/>
+        <enum value="2"             name="CL_IMAGE_TILING_OPTIMAL_EXT"/>
+            <unused start="3" end="0xFFFFFFFF"/>
```
SunSerega, line 1392: Hmm, is this intentionally a separate value range, parallel to everything else?

api/opencl_platform_layer.asciidoc
```diff
@@ -2097,6 +2097,24 @@ include::{generated}/api/version-notes/CL_DEVICE_CXX_FOR_OPENCL_NUMERIC_VERSION_
...
+    | Returns the image tiling capabilities supported by the device. +
+
+      {CL_DEVICE_IMAGE_TILING_DEVICE_ACCESS_EXT_anchor} is always set indicating
+      that all implementations that support {cl_ext_image_tiling_control_EXT}
+      must support device access to tiled images.
+
+      {CL_DEVICE_IMAGE_TILING_HOST_ACCESS_EXT_anchor} is set when it is allowed
+      to pass {CL_MEM_IMAGE_TILING_EXT} with a value different from
+      {CL_IMAGE_TILING_LINEAR_EXT} in the _properties_ given to
+      {clCreateImageWithProperties} when creating host-accessible images.
```
joshqti, line 2115: Does this work?  If you want raw host access to tiled data, you would create a buffer from the image.  Wouldn't the buffer inherit the no host access flag?  Couldn't the image be host-inaccessible for reasons other than tiling?

api/opencl_runtime_layer.asciidoc
```diff
@@ -1996,6 +1996,39 @@ The following restrictions apply when mipmapped images are created with
...
 
+ifdef::cl_ext_image_tiling_control[]
+*Tiling Control*
+
+{CL_MEM_IMAGE_TILING_EXT_anchor} can be passed as part of the _properties_ parameter
+to {clCreateImageWithProperties} to control the tiling used for the image being
+created. The following values are accepted:
+
+  * {CL_IMAGE_TILING_LINEAR_EXT_anchor}, which is the default value used if
+    {CL_MEM_IMAGE_TILING_EXT} is not specified in the _properties_. Image data
```
joshqti, line 2007: Should me disallow the use if this propery when importing external memory?  This may contradict an external import.

api/opencl_runtime_layer.asciidoc
```diff
@@ -4344,8 +4385,27 @@ include::{generated}/api/version-notes/CL_IMAGE_D3D11_SUBRESOURCE_KHR.asciidoc[]
...
+    | Return the tiling passed using the {CL_MEM_IMAGE_TILING_EXT} property or
+      {CL_IMAGE_TILING_LINEAR_EXT} if none was passed at image creation time.
+endif::cl_ext_image_tiling_control[]
+
 |====
 
+ifdef::cl_ext_image_tiling_control[]
+If the image object is created with {CL_MEM_IMAGE_TILING_EXT} set to a value
+different from {CL_IMAGE_TILING_LINEAR_EXT}, the value returned for
+
```
joshqti, line 4402: If an externally imported image is tiled, should we return optimal?

api/opencl_runtime_layer.asciidoc
```diff
@@ -4344,8 +4385,27 @@ include::{generated}/api/version-notes/CL_IMAGE_D3D11_SUBRESOURCE_KHR.asciidoc[]
       specified when _image_ was created.
 endif::cl_khr_d3d11_sharing[]
 
+ifdef::cl_ext_image_tiling_control[]
+| {CL_IMAGE_TILING_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_IMAGE_TILING_EXT.asciidoc[]
+  | {cl_image_tiling_ext_TYPE}
+    | Return the tiling passed using the {CL_MEM_IMAGE_TILING_EXT} property or
+      {CL_IMAGE_TILING_LINEAR_EXT} if none was passed at image creation time.
```
kpet, line 4394: Discussed in the 2025/01/28 memory TSG: agreement to return the tiling that was either specified by the application at image creation time or selected by the implementation in the absence of tiling specified by the application.

api/opencl_runtime_layer.asciidoc
kpet, line 2007: Discussed in the 2025/01/28 memory TSG: agreement to not require linear tiling by default but let the implementation pick the tiling.

kpet (commented):
Recording a few changes we agreed to make in the 2025/01/28 memory TSG.

api/opencl_runtime_layer.asciidoc
bcalidas: These look good

api/opencl_runtime_layer.asciidoc
kpet: I have now reflected the agreed change in the draft specification.

api/opencl_runtime_layer.asciidoc
kpet: I have now reflected the agreed change in the draft specification.

xml/cl.xml
kpet: Yes. There are other examples for this pattern (see `cl_command_buffer_state_khr`, `cl_command_termination_reason_arm`, etc). Does it create a concrete problem?

api/opencl_runtime_layer.asciidoc
kpet: I'm not sure I understand what you meant.

api/opencl_runtime_layer.asciidoc
kpet: I don't think we want to disallow providing the `CL_MEM_IMAGE_TILING_EXT` property (or specifically the ` CL_IMAGE_TILING_LINEAR_EXT` value) when importing external memory. It is particularly useful when importing external memory as it enables applications to clearly communicate the tiling of the data. If in the future we add mechanisms that lead to a contradiction, then the later of the two extensions (the one adding the mechanism or this extension) should describe the interaction and forbid the problematic case.

api/opencl_runtime_layer.asciidoc
kpet: Discussed in the 2025/05/06 memory subgroup teleconference. The question was: can an implementation select the tiling (and return it when an application queries the tiling of an image) when importing image data from external memory? Answer: yes, nothing in this proposed specification forbids this.

api/opencl_runtime_layer.asciidoc
```diff
@@ -2102,6 +2102,42 @@ The following restrictions apply when mipmapped images are created with
...
+  * {CL_IMAGE_TILING_OPTIMAL_EXT_anchor} requires image data be laid out in an
+    implementation-defined format which can vary depending on the type of image,
+    its format and/or dimensions. The implementation will attempt to select the
+    best layout for the image.
+
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT}, {CL_MEM_USE_HOST_PTR} must not be set in _mem_flags_.
+
+If {CL_MEM_IMAGE_TILING_EXT} is passed and set to a value different from
+{CL_IMAGE_TILING_LINEAR_EXT} and {CL_MEM_COPY_HOST_PTR} is set in _mem_flags_, the data
```
xinjin01, line 2126: Could you clarify the expected behavior when creating an image from a buffer that was created with the CL_MEM_COPY_HOST_PTR flag? According to the OpenCL specification, the CL_MEM_COPY_HOST_PTR, CL_MEM_USE_HOST_PTR, and CL_MEM_ALLOC_HOST_PTR flags are inherited as memory access qualifiers when creating an image from a buffer, but “if CL_MEM_COPY_HOST_PTR is specified in the memory access qualifier values associated with mem_object it does not imply any additional copies when the image is created from mem_object.”

So, am I correct in understanding that when creating an image from such a buffer, no additional copy is performed—the only copy happens at buffer creation, and the image simply views the buffer’s storage? This is different from passing CL_MEM_COPY_HOST_PTR directly to clCreateImage, which does cause a copy from host memory.

It might be helpful to make this distinction explicit in the documentation to avoid confusion. Thanks!

xinjin01: and the same for CL_MEM_USE_HOST_PTR as well

PR #711, opened 2021-11-11
spec source for new extensions

bashbaug:
This PR contains the spec source for the following new extensions:

* `cl_khr_async_copy_fence` (was provisional, now final)
* `cl_khr_extended_async_copies` (was provisional, now final)
* `cl_khr_expect_assume`
* `cl_khr_command_buffer` (provisional)

### DIFF
--- a/ext/cl_khr_command_buffer.asciidoc
+++ b/ext/cl_khr_command_buffer.asciidoc
@@ -161,12 +161,14 @@ typedef cl_uint cl_sync_point_khr;
 // Handle returned on command recording
 typedef struct _cl_mutable_command_khr* cl_mutable_command_khr;
 
-// Bitfield representing mutable properties of a clCommandNDRangeKernelKHR
-// command
-typedef cl_bitfield cl_ndrange_kernel_command_properties_khr;
+// Mutable properties of a clCommandNDRangeKernelKHR command
+typedef cl_properties cl_ndrange_kernel_command_properties_khr;
 
-// Bitfield representing properties for command-buffer creation.
-typedef cl_bitfield cl_command_buffer_properties_khr;
+// Properties for command-buffer creation
+typedef cl_properties cl_command_buffer_properties_khr;
+
+// Bitfield representing flags for command-buffers
+typedef cl_bitfield cl_command_buffer_flags_khr;
 
 // Enumerated type for use in clGetCommandBufferInfoKHR()
 typedef cl_uint cl_command_buffer_info_khr;
@@ -359,9 +361,9 @@ CL_INVALID_SYNC_POINT_WAIT_LIST_KHR                -1139
 CL_INCOMPATIBLE_COMMAND_QUEUE_KHR                  -1140
 
 // Bitfield to clCreateCommandBufferKHR
-CL_COMMAND_BUFFER_PROPERTIES_KHR                   0x1293
+CL_COMMAND_BUFFER_FLAGS_KHR                        0x1293
 
-// Bits for cl_command_buffer_properties_khr bitfield
+// Bits for cl_command_buffer_flags_khr bitfield
 CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR             (0x1 << 0)
 
 // cl_command_buffer_info_khr queries to clGetCommandBufferInfoKHR
@@ -514,8 +516,8 @@ default values for supported command-buffer properties will be used.
 | *Property Value*
 | *Description*
 
-| *CL_COMMAND_BUFFER_PROPERTIES_KHR*
-| `cl_command_buffer_properties_khr`
+| *CL_COMMAND_BUFFER_FLAGS_KHR*
+| `cl_command_buffer_flags_khr`
 | This is a bitfield and can be set to a combination of the following values:
 
   `CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR` - Allow multiple instances of the
--- a/ext/cl_khr_command_buffer.asciidoc
+++ b/ext/cl_khr_command_buffer.asciidoc
@@ -1,0 +1,1875 @@
+// Copyright 2018-2021 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+[[cl_khr_command_buffer]]
+== Command Buffers (Provisional)
+
+This extension adds the ability to record and replay buffers of OpenCL commands.
+
+=== General Information
+
+==== Name Strings
+
+`cl_khr_command_buffer`
+
+==== Version History
+
+[cols="1,1,3",options="header",]
+|====
+| *Date*     | *Version* | *Description*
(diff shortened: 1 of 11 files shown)


ext/cl_khr_command_buffer.asciidoc
```diff
@@ -0,0 +1,1875 @@
...
+CL_INCOMPATIBLE_COMMAND_QUEUE_KHR                  -1140
+
+// Bitfield to clCreateCommandBufferKHR
+CL_COMMAND_BUFFER_PROPERTIES_KHR                   0x1293
+
+// Bits for cl_command_buffer_properties_khr bitfield
+CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR             (0x1 << 0)
+
+// cl_command_buffer_info_khr queries to clGetCommandBufferInfoKHR
+CL_COMMAND_BUFFER_INFO_QUEUES_KHR                  0x1294
```
bashbaug, line 368: Minor/consistency: Most of the other query enums do not include "INFO" in their names.  For example, we have a query enum `CL_CONTEXT_REFERENCE_COUNT` and not `CL_CONTEXT_INFO_REFERENCE_COUNT`.  Should we drop "INFO" in the names of these enums?

Note that the enum type does typically include "info" in the name, so we should keep `cl_command_buffer_info_khr`.

bashbaug: Tagging @EwanC for awareness.

alycm (commented):
Not reviewed `cl_khr_command_buffer` yet, but reviewed the others.

Hopetech:
Thanks you Ben,
`cl_khr_extended_async_copies` looks good.

ext/cl_khr_command_buffer.asciidoc
EwanC: Ah yes, good catch Ben. Happy for you to remove the `_INFO_` component of these enums.

ext/cl_khr_expect_assume.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+kernel void test(global int* dst, global int* src)
+{
+    int value = src[get_global_id(0)];
+
+    // Tell the compiler that the most likely source value is zero.
+#if __has_builtin(__builtin_expect)
+    value = __builtin_expect(value, 0);
+#endif
+
+    // Tell the compiler that the source value is non-negative.
```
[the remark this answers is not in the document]
bashbaug: Good suggestion, it couldn't hurt, especially because this behavior isn't described specifically here.

ext/cl_khr_command_buffer.asciidoc
bashbaug: I removed "INFO" in commit https://github.com/KhronosGroup/OpenCL-Docs/pull/711/commits/929586586711de05bd45c05840c3b3d147d4e949 - please take a look.

EwanC: Looks good, updated my headers PR in https://github.com/KhronosGroup/OpenCL-Headers/pull/182/commits/eba2023490a4071718bad97ffbef6f877f41bb2e to match

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -0,0 +1,1875 @@
...
+| Return the list of command-queues specified when the _command_buffer_ was
+  created.
+
+| *CL_COMMAND_BUFFER_REFERENCE_COUNT_KHR*
+| cl_uint
+| Return the _command_buffer_ reference count.
+
+  The reference count returned with `CL_COMMAND_BUFFER_REFERENCE_COUNT_KHR`
+  should be considered immediately stale. It is provided for identifying
+  memory leaks.
```
alycm, line 1641:
```suggestion
| *CL_COMMAND_BUFFER_REFERENCE_COUNT_KHR* footnote:[{fn-reference-count-usage}]
| cl_uint
| Return the _command_buffer_ reference count.
```

Will only work if https://github.com/KhronosGroup/OpenCL-Docs/pull/712 is merged first.

alycm (commented):
LGTM, but one area that I think we'll have to take to an issue.

ext/cl_khr_command_buffer.asciidoc
bashbaug: I think this makes sense, but I'll hold off committing this suggestion until #712 is merged.

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -0,0 +1,1875 @@
...
+_queues_  Is a pointer to a command-queue that the command-buffer commands will
+be recorded to. _queues_ must be a non-`NULL` value.
+
+_properties_  Specifies a list of properties for the command-buffer and their
+corresponding values. Each property name is immediately followed by the
+corresponding desired value. The list is terminated with 0.
+The list of supported properties is described in the table below. If a
+supported property and its value is not specified in properties, its
+default value will be used. _properties_ can be `NULL` in which case the
+default values for supported command-buffer properties will be used.
```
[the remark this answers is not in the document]
bashbaug: I don't think this is wrong per se, though it is a little odd and I agree it is semi-recursive.

FWIW, command queue properties has a similar issue, and it's "solved" by having `cl_command_queue_properties` be the bitfield type specifying the actual property bits and `cl_queue_properties` be the pointer to the property-value pairs passed to `clCreateCommandQueueWithProperties`.  This always seemed a little icky and I can never remember which type gets used for which purpose.  I wonder if we would have been better off with just `cl_command_queue_properties` and a similar semi-recursive usage.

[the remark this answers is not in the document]
EwanC: When I based this off of `clCreateCommandQueueWithProperties`, I completely missed that `cl_queue_properties` != `cl_command_queue_properties` 🤦 So the semi-recursive relationship wasn't intentional. I think my preference would be for two separate types but not sure if that's too significant a change at this point.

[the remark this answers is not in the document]
alycm: > I don't think this is wrong per se

> I think my preference would be for two separate types but not sure if that's too significant a change at this point.

Stepping back for a moment, `cl_bitfield` and `cl_properties` are all just a `cl_ulong`, so this may be more trivial that I was thinking.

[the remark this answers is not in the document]
bashbaug: I think I moved the changes we discussed over correctly, please check https://github.com/KhronosGroup/OpenCL-Docs/pull/711/commits/5dda69bfb85d13f7a86d4f847b6036b70a6b7a96.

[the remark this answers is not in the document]
EwanC: change LGTM, updated headers PR to reflect it in https://github.com/KhronosGroup/OpenCL-Headers/pull/182/commits/eafab6a1c77007d30f334db475949cf983009307

alycm (approved):

PR #714, opened 2021-11-12
fixes whitespace error for union generation

### LINKED ISSUE #682: Whitespace formatting of `union` in `cl_image_desc`
PR https://github.com/KhronosGroup/OpenCL-Docs/pull/671 synchronized scripts with Vulkan-Docs.

This has resulted in an odd whitespace change in the generated `cl_image_desc` struct in section 5.3.1.2.

Old:

image

New:

image

HTML diff:

```diff
@@ -11500,10 +11500,10 @@
     size_t                image_slice_pitch;
     cl_uint               num_mip_levels;
     cl_uint               num_samples;
-    <span class="keyword">union</span> {
+<span class="keyword">union</span> {
         cl_mem buffer;
         cl_mem mem_object;
-    };
+    }    ;
 } cl_image_desc;</code></pre>
 </div>
 </div>
```

I suspect that fixing this will be trivial, I just didn't have time to look right now.

bashbaug:
fixes #682 

I think the issue here is that the whitespace gets included on the RHS of the generated string instead of on the LHS, so swap the order of the "prefix" and the "paramdecl".

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -942,7 +942,7 @@ class OutputGenerator:
 
         # If prefix was originally non-empty and the param has no elements
         # (e.g. is nothing but text), preserve it.
-        paramdecl = prefix + paramdecl
+        paramdecl = paramdecl + prefix
 
         if aligncol == 0:
             # Squeeze out multiple spaces other than the indentation


alycm (approved):
LGTM, confirmed that this fixes the rendering issue.

oddhack (approved):
LGTM

PR #718, opened 2021-11-16
add a specification change log

### LINKED ISSUE #707: Spec Revision Change Log
It would be nice to provide a brief "change log" for each spec revision so it is easier to know what is included in each revision.  We should start to do this at least for each new spec release, even if we don't go back and document past spec releases.

One way to do this is to add a sub-section of Appendix E, which currently documents changes for each minor OpenCL revision.

We should add a similar change log appendix to the OpenCL C specification as well (see #159).

bashbaug:
fixes #707 

This PR adds a specification "change log" for published revisions of the OpenCL 3.0 specifications.  I did this by adding an appendix to the OpenCL C and OpenCL SPIR-V Environment specifications and by adding to the existing appendix in the OpenCL API specification.

I've documented most of the extension changes in the OpenCL API specification appendix for now, but I'm not entirely sure if that makes the most sense.  It's easy to move these to an appendix in the extension spec if desired, though I get nervous spreading information around too many places.  By this same logic I suppose the OpenCL SPIR-V Environment changes could be merged into the API spec also, if desired.

### DIFF
--- a/c/appendix_a.asciidoc
+++ b/c/appendix_a.asciidoc
@@ -13,7 +13,7 @@ summarized below.
 
 The first non-provisional version of the OpenCL 3.0 specifications was *v3.0.5*.
 
-Changes in *v3.0.6*:
+Changes from *v3.0.5*:
 
   * Clarified that `memory_scope_all_devices` is supported only for OpenCL C 3.0 or newer.
   * Defined ULP overflow leniency.
@@ -22,7 +22,7 @@ Changes in *v3.0.6*:
   * Clarified relationship between optional core features and extensions.
   * Deprecated the `+__OPENCL_C_VERSION__+` predefined macro and clarified possible values of the macro for different versions of OpenCL.
 
-Changes in *v3.0.7*:
+Changes from *v3.0.6*:
 
   * Clarified the argument to *vec_step* is not evaluated.
   * Improved description for pipe specifier.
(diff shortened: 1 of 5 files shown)


c/appendix_a.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+  * Removed a confusing phrase about kernel argument pointer types.
+  * Clarified usage of feature test macros pre-OpenCL C 3.0.
+  * Clarified relationship between optional core features and extensions.
+  * Deprecated the `+__OPENCL_C_VERSION__+` predefined macro and clarified possible values of the macro for different versions of OpenCL.
+
+Changes from *v3.0.6*:
+
+  * Clarified the argument to *vec_step* is not evaluated.
+  * Improved description for pipe specifier.
+  * Fixed spelling of `-cl-no-signed-zeros`.
```
alycm, line 29: Technically an API change rather than an OpenCL C change.

alycm (approved):
LGTM.

alycm:
> I've documented most of the extension changes in the OpenCL API specification appendix for now, but I'm not entirely sure if that makes the most sense. It's easy to move these to an appendix in the extension spec if desired, though I get nervous spreading information around too many places. By this same logic I suppose the OpenCL SPIR-V Environment changes could be merged into the API spec also, if desired.

Based on the changes here an extension changelog would be quite sparse.

I don't have any problem with the split as it is in this PR.

c/appendix_a.asciidoc
bashbaug: Whoops!  Yes, you're right.  I've moved this line item to the API spec.

PR #725, opened 2021-11-23
Include generated version notes in API reference pages

### LINKED ISSUE #724: API entry point reference pages do not have first "missing before" style notes
The reference pages for API entry points do not having "missing before" or "deprecated by" version notes for the API entry points themselves.  Parameters or error codes within a reference page do have these notes.

E.g. the page for [clGetKernelSubGroupInfo](https://www.khronos.org/registry/OpenCL/sdk/3.0/docs/man/html/clGetKernelSubGroupInfo.html), at the time of writing, has notes about "missing before" within the parameter descriptions but nothing on the page says the entry point itself is missing before version 2.1.

This is because `genRef.py` and `reflib.py` are looking for specific patterns and copy ranges of the specification into the reference pages.  The version notes fall outside of these ranges.

alycm:
Change the range of text copied into the API entry point reference pages so as
to include the generated version notes.

Before: 
image

After:
image

Closes #724

### DIFF
--- a/scripts/genRef.py
+++ b/scripts/genRef.py
@@ -393,8 +393,10 @@ def emitPage(baseDir, specDir, pi, file):
             logWarn('emitPage:', pageName, 'INCLUDE is None, no page generated')
             return
 
-        # Specification text
-        lines = remapIncludes(file[pi.begin:pi.include + 1], baseDir, specDir)
+        # Specification text from beginning to just before the parameter
+        # section. This covers the description, the prototype, the version
+        # note, and any additional version note text.
+        lines = remapIncludes(file[pi.begin:pi.param], baseDir, specDir)
         specText = ''.join(lines)
 
         if pi.param is not None:
--- a/scripts/genRef.py
+++ b/scripts/genRef.py
@@ -395,8 +395,10 @@ def emitPage(baseDir, specDir, pi, file):
 
         # Specification text from beginning to just before the parameter
         # section. This covers the description, the prototype, the version
-        # note, and any additional version note text.
-        lines = remapIncludes(file[pi.begin:pi.param], baseDir, specDir)
+        # note, and any additional version note text. If a parameter section
+        # is absent then go a line beyond the include.
+        remap_end = pi.include + 1 if pi.param is None else pi.param
+        lines = remapIncludes(file[pi.begin:remap_end], baseDir, specDir)
         specText = ''.join(lines)
 
         if pi.param is not None:


alycm:
One downside of this is notices such as:

> clCreateCommandQueue is deprecated by version 2.0.

without any hints of what to use instead.  In the specification text this isn't much of a problem because this deprecated entry point is immediately after it's replacement.  That doesn't apply to the independent reference pages.

bashbaug (approved):
LGTM.  I diff'd the generated reference pages after this change and didn't observe anything unexpected.

scripts/genRef.py
```diff
@@ -393,8 +393,10 @@ def emitPage(baseDir, specDir, pi, file):
             logWarn('emitPage:', pageName, 'INCLUDE is None, no page generated')
             return
 
-        # Specification text
-        lines = remapIncludes(file[pi.begin:pi.include + 1], baseDir, specDir)
+        # Specification text from beginning to just before the parameter
+        # section. This covers the description, the prototype, the version
+        # note, and any additional version note text.
+        lines = remapIncludes(file[pi.begin:pi.param], baseDir, specDir)
```
oddhack, line 399: Under some conditions, pi.param can be None (thus the following logic). The reflib.py logic tries to find the 'parameter section' (which is really the member / parameter / enum description / etc. section following the include), but there may not be one - though there should be. So while this may work with the current spec text, and its failure may be a good signal that something needs to be changed, I wouldn't recommend this change just as it stands. Playing with this in the Vulkan spec repo and constructing various problematic test cases to try things out with.

bashbaug: This has been open a while and I'd like to merge it - as observed above it seems to work fine with the current spec text.

@oddhack, were you able to find any problematic test cases?

bashbaug:
This has been open for a while now.  What do we want to do with it?

At least for the OpenCL spec it seems like a nice upgrade so I'd like to merge it (or something like it).

scripts/genRef.py
alycm: I suppose the main argument against this change is that divergence in these scripts between specifications isn't great, and this may not work elsewhere.

```suggestion
        remap_end = pi.include + 1 if pi.param is None else pi.param
        lines = remapIncludes(file[pi.begin:remap_end], baseDir, specDir)
```

The above would fall back to the existing search range if `pi.param` is `None`.  Perhaps that's a pragmatic middle ground?  Though for the OpenCL specification I think having the generation scripts crash on a missing section would be an acceptable outcome.  I haven't actually tested the above change, but for OpenCL it should always take the `else` path.

oddhack: Sorry, missed responding earlier. With @alycm's change this is a no-op on the Vulkan spec repository. If you are happy with it for the OpenCL refpages, please proceed.

oddhack (changes requested):
Would not want this without @alycm's change, but with it, LGTM.

scripts/genRef.py
alycm: I've made this change, updated the comment, and checked that the output is unchanged for the OpenCL reference pages.

bashbaug:
@oddhack can you please approve this PR if your requested changes have been made (I think they have)?  Thanks!

oddhack (approved):

SunSerega:
[The page linked in the related issue](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetKernelSubGroupInfo.html) still doesn't have a "Missing before" section for the function. Is that supposed to be so?

bashbaug:
Yes, this is expected, for now.  Building and publishing the reference pages is still a manual process, and we haven't published updated reference pages - yet.  Should be soon, stay tuned.  Good observation, thanks for checking!

PR #728, opened 2021-11-24
[WIP] Add usm buffer location API enum CL_MEM_ALLOC_BUFFER_LOCATION_INTEL

sherry-yuan:
CL_MEM_ALLOC_BUFFER_LOCATION_INTEL usm extension enum specifies ID of the global memory type to allocate device/usm/host memory to.

Where ID for device memory is the index of the global_mem as it appears in board_spec.xml

To use the flag:
```
cl_mem_properties_intel good_property[3] = {
CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, 0,
0};
```

Then pass this into clDeviceMemAllocINTEL:
`void *test_ptr = clDeviceMemAllocINTEL(m_context, m_device[0], &(good_property[0]), 8, alignment, &status);`


extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -231,12 +231,13 @@ Optional allocation properties may also be queried using *clGetMemAllocInfoINTEL
...
 
-#define CL_MEM_ALLOC_TYPE_INTEL         0x419A
-#define CL_MEM_ALLOC_BASE_PTR_INTEL     0x419B
-#define CL_MEM_ALLOC_SIZE_INTEL         0x419C
-#define CL_MEM_ALLOC_DEVICE_INTEL       0x419D
+#define CL_MEM_ALLOC_TYPE_INTEL               0x419A
+#define CL_MEM_ALLOC_BASE_PTR_INTEL           0x419B
+#define CL_MEM_ALLOC_SIZE_INTEL               0x419C
+#define CL_MEM_ALLOC_DEVICE_INTEL             0x419D
+#define CL_MEM_ALLOC_BUFFER_LOCATION_INTEL    0x419E
```
EwanC, line 238: Could you add some spec wording about the usage of `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL`  please

Additionally, your example
```c
cl_mem_properties_intel good_property[3] = {
CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, 0,
0};
```

Is using `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL` as a property but the above enums are for `clGetMemAllocInfoINTEL` so I think this new enum should be split off rather than grouped with them

sherry-yuan: Thank you for reviewing and providing suggestions!

I have added the spec in `cl_intel_mem_alloc_buffer_location.asciidoc`

Regarding CL_MEM_ALLOC_BUFFER_LOCATION_INTEL shouldn't be `CL_MEM_ALLOC...` group: This enum will be used in the clGetMemAllocInfoINTEL for user to check whether the place they are allocating is correct. Sorry I forgot to mention this in the commit message, I now have added it to the spec.

EwanC: Thanks for the response,

> I have added the spec in cl_intel_mem_alloc_buffer_location.asciidoc

Creating a new document `cl_intel_mem_alloc_buffer_location.asciidoc` implies that this is a new extension that depends on `cl_intel_unified_shared_memory`, is this the intention?

sherry-yuan: Sorry I did not know that is the implication of a new doc. That is indeed no the intention. I have moved the info from the other doc to current doc. Thanks for letting me know!

PR #729, opened 2021-11-25
Generate enum dictionaries for all extensions

kpet:
So that we can use them in vendor extensions. This also guarantees that
generation is always in order for vendor extensions.

Change-Id: I5708ad6b1e090678fccb6bb335e633081d6f58fb


bashbaug (approved):
LGTM.  It's nice to get rid of the "include anyway" messiness!

alycm (approved):
LGTM, I'll merge this now.

PR #735, opened 2021-12-06
C++ for OpenCL Documentation for Ternary Operator (?:)

Topotuna:
Since C++ and OpenCL C use diferent rules for ternary operator,
C++ for OpenCL has to adapt to those differences. The documentation
section is written that explains usage of ternary operator in C++
for OpenCL.

### DIFF
--- a/cxx4opencl/diff2openclc.txt
+++ b/cxx4opencl/diff2openclc.txt
@@ -129,6 +129,18 @@ label:  // invalid: jumping forward over declaration of n
    // ...
 ----------
 
+===== Ternary Selection Operator
+
+The ternary selection operator (`?:`) inherits its behaviour from both
+{cpp} and OpenCL C. It operates on three expressions
+`(exp1 ? exp2 : exp3)`. If all three expressions are scalar values,
+the {cpp}17 rules for ternary operator are followed. If the result is
+a vector value, then this is equivalent to calling
+`select(exp3, exp2, exp1)` as described in `OpenCL C v2.0 s6.13.6`.
+The rules from OpenCL C impose limitation that `exp1` cannot be a
+vector of float values. However, `exp1` can be evaluated to a scalar
+float as it is contextually convertible to bool in {cpp}.
+
 ==== OpenCL specific difference
 
 This section describes where {cpp} for OpenCL differs from OpenCL C


cxx4opencl/diff2openclc.txt
```diff
@@ -129,6 +129,18 @@ label:  // invalid: jumping forward over declaration of n
    // ...
 ----------
 
+===== Ternary Selection Operator
+
+The ternary selection operator (`?:`) inherits its behaviour from both
+{cpp} and OpenCL C. It operates on three expressions
+`(exp1 ? exp2 : exp3)`. If all three expressions are scalar values,
+the {cpp}17 rules for ternary operator are followed. If the result is
```
AnastasiaStulova, line 137: I am not sure if `the result` is clearly defined in this context. Maybe it can be explained before this sentence?

AnastasiaStulova: Although I appreciate that OpenCL C wording is the same...

cxx4opencl/diff2openclc.txt
```diff
@@ -129,6 +129,18 @@ label:  // invalid: jumping forward over declaration of n
...
+a vector value, then this is equivalent to calling
+`select(exp3, exp2, exp1)` as described in `OpenCL C v2.0 s6.13.6`.
+The rules from OpenCL C impose limitation that `exp1` cannot be a
+vector of float values. However, `exp1` can be evaluated to a scalar
```
AnastasiaStulova, line 141: Btw is a vector of float allowed? Is this something we should mention explicitly?

Topotuna: No, vector of floats is not allowed because we reuse logic from OpenCL C. I can reword this sentence to be more explicit about C++ for OpenCL:
```suggestion
vector of float values so it is also disallowed in C++ for OpenCL. However, `exp1` can be evaluated to a scalar
```

cxx4opencl/diff2openclc.txt
Topotuna: If I remember correctly, using `the result` is the most precise about the actual implementation inside clang.

cxx4opencl/diff2openclc.txt
AnastasiaStulova: Maybe we  can leave out what is identical to OpenCL and only document the differences since this section is dedicated to differences in behavior?

cxx4opencl/diff2openclc.txt
AnastasiaStulova: ok, but the language documentation should be agnostic to the exact compiler. We can document clang specific aspects in the clang internal documentation... However,  here I don't think the result is clear here. I would at least change this to something like the expression result or the operator result?

cxx4opencl/diff2openclc.txt
AnastasiaStulova: While this discusses some improvements, there is no incorrect wording in this PR, and since we are close to the release date I would suggest to merge it as it and any improvements can be made separately.

AnastasiaStulova (approved):
LGTM, improvements in unresolved comments can be made through separate PRs.

cxx4opencl/diff2openclc.txt
Topotuna: I think it is good to be explicit here about what is identical and what differs between languages because ternary operator inherits its behavior from both C++ and OpenCL C and differs from both of them. I feel that direct comparison with the two component languages makes it easier to explain how operator acts.

PR #738, opened 2021-12-09
Couple device enqueue and program scope global variables features.

azabazno:
Some ABIs (for example, clang blocks ABI) may use program scope variables as part of optimization when generating block constructions for blocks without captures/global blocks. Couple device enqueue and program scope global variables to allow such behaviour.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -161,8 +161,9 @@ share SVM memory with each other and the host process.
 | The OpenCL C compiler supports built-in functions to enqueue additional work
 from the device.
 
-OpenCL C compilers that define the feature macro {opencl_c_device_enqueue}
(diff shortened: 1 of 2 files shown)


bashbaug (commented):
LGTM though should have an approval from a non-Intel reviewer.

It'd be nice to have a CTS test that checked for this required coupling but it doesn't look like we have a test for any of the other couplings either.  I'd recommend filing a CTS issue to track this - it should be a straightforward addition to the compiler test.

kpet (approved):
LGTM. +1 on making sure we have consistency checks in the CTS.

bashbaug:
I created https://github.com/KhronosGroup/OpenCL-CTS/issues/1366 to track the CTS consistency check.  Merging!

azabazno:
Thanks!

PR #739, opened 2021-12-10
Add doc revision to C++ for OpenCL in Makefile

AnastasiaStulova:
With the addition of C++ for OpenCL 2021, we will now end up with 2 language versions defined in the same document.

Hence we need some way to reference the document independent from the language versions.

This change adds the versioning of the document taken from its publishing date. For example, document published in Dec this year will have a version:

DOCREV2021.12

Listed in the front page.

Then  if we need to reference it to describe the semantics of C++ for OpenCL 1.0 and C++ for OpenCL 2021, we can do something like:

> C++ for OpenCL DOCREV2021.12: s2.3.3 by default implicit object parameter is in generic address space for v1.0 and in private address space for v2021.

While this is a bit longer than what we used to do with OpenCL C or C and C++, it can allow uniquely identify the document that explains semantics for multiple language versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,12 +68,12 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
-# C++ for OpenCL document revision scheme in aligned with its release date.
+# The C++ for OpenCL document revision scheme is aligned with its release date.
 # Revision naming scheme is as follows:
 # DocRevYYYY.MM,
 # where YYYY corresponds to its release year,
 # MM corresponds to its release month.
-# Example for the release in Dec 2021 the revision is DOCREV2021.12.
+# Example for the release in Dec 2021 the revision is DocRev2021.12.
 # Leave as 'DocRevYYYY.MM-Next' if the doc content does not correspond to any official revision.
 # where DocRevYYYY.MM is the last released revision.
 CXX4OPENCL_DOCREVISION = DocRev2021.12
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ endif
 # C++ for OpenCL doc revision scheme in aligned with its release date.
 # Revision naming scheme is as follows:
 # DOCREVYYYY.MM,
-# where YYYY correspomds to its release year,
+# where YYYY corresponds to its release year,
 # MM corresponds to its release month.
 # Example for the release in Dec 2021 the revision is DOCREV2021.12.
 # Leave blank if the doc content does not correspond to any official revision.
--- a/Makefile
+++ b/Makefile
@@ -70,12 +70,13 @@ SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 endif
 # C++ for OpenCL document revision scheme in aligned with its release date.
 # Revision naming scheme is as follows:
-# DOCREVYYYY.MM,
+# DocRevYYYY.MM,
 # where YYYY corresponds to its release year,
 # MM corresponds to its release month.
 # Example for the release in Dec 2021 the revision is DOCREV2021.12.
-# Leave blank if the doc content does not correspond to any official revision.
-CXX4OPENCL_DOCREVISION = DOCREV2021.12
+# Leave as 'DocRevYYYY.MM-Next' if the doc content does not correspond to any official revision.
+# where DocRevYYYY.MM is the last released revision.
+CXX4OPENCL_DOCREVISION = DocRev2021.12
 CXX4OPENCL_DOCREMARK = $(SPECREMARK) \
 			tag: $(SPECREVISION)
 
--- a/Makefile
+++ b/Makefile
@@ -68,20 +68,30 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
+# C++ for OpenCL doc revision scheme in aligned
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -68,20 +68,30 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
+# C++ for OpenCL doc revision scheme in aligned
+# with its release date.
+CXX4OPENCL_DOCREVISION = DOCREV$(shell echo `date +"%Y.%m"`)
```
svenvh, line 73: This would make the date the spec is built to be the identifier of a particular spec version.  I think this opens up too many opportunities for confusion: someone building an older revision of the repo will produce a document that "looks like" the most recent version while it is older.  I would strongly advise against using the build date alone as a document revision identifier.  Instead, I would suggest to use a variable that is manually updated with every release.

AnastasiaStulova: Ok, thanks. This makes sense... other approaches could be to extract the revision from the tag... this is what we do for other specs... but however we add a `cxxforopencl` prefix to the tag of C++ for OpenCL document... maybe we don't want to add too much complexity in Makefile for extracting the subsrtings...

Makefile
```diff
@@ -68,20 +68,35 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
-
-ATTRIBOPTS_NO_VERSION	= -a revdate="$(SPECDATE)" \
-			  -a revremark="$(SPECREMARK)" \
+# C++ for OpenCL doc revision scheme in aligned with its release date.
+# Revision naming scheme is as follows:
+# DOCREVYYYY.MM,
+# where YYYY correspomds to its release year,
```
AnastasiaStulova, line 74:
```suggestion
# where YYYY corresponds to its release year,
```

Makefile
AnastasiaStulova: I have updated this PR now to set the revision manually and updated the comments explaining the revision scheme.

For WIP doc builds we should just leave the revision blank.

Makefile
```diff
@@ -68,20 +68,35 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
-
-ATTRIBOPTS_NO_VERSION	= -a revdate="$(SPECDATE)" \
-			  -a revremark="$(SPECREMARK)" \
+# C++ for OpenCL doc revision scheme in aligned with its release date.
+# Revision naming scheme is as follows:
+# DOCREVYYYY.MM,
```
svenvh, line 73: Any reason for DOCREV to be in all capitals?  It feels a bit shouty.  "docrev" or "DocRev" would seem a bit friendlier to me.

Makefile
```diff
@@ -68,20 +68,35 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD`) \
 	     commit: $(shell echo `git log -1 --format="%H"`)
 endif
-
-ATTRIBOPTS_NO_VERSION	= -a revdate="$(SPECDATE)" \
-			  -a revremark="$(SPECREMARK)" \
+# C++ for OpenCL doc revision scheme in aligned with its release date.
```
svenvh, line 71:
```suggestion
# C++ for OpenCL document revision scheme in aligned with its release date.
```

Makefile
```diff
@@ -68,20 +68,35 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
...
+# Revision naming scheme is as follows:
+# DOCREVYYYY.MM,
+# where YYYY corresponds to its release year,
+# MM corresponds to its release month.
+# Example for the release in Dec 2021 the revision is DOCREV2021.12.
+# Leave blank if the doc content does not correspond to any official revision.
+CXX4OPENCL_DOCREVISION = DOCREV2021.12
```
svenvh, line 78: As an alternative to leaving it blank, you could write `-unreleased` or `2021.12-next` perhaps?

Makefile
AnastasiaStulova: I would like to start from the upper case because it gets capitalized in HTML by default for some reasons but not in PDF.

DocRev would also be ok though...

 ```suggestion
# DocRevYYYY.MM,
```

File: Makefile (hunk 5 above)
Comment (AnastasiaStulova) [reply Makefile:78]: Yes, I would quite like to indicate that it is derived/on top of 2021.12, so adding `next` might be a good option
```suggestion
# Leave as 'DocRevYYYY.MM-Next' if the doc content does not correspond to any official revision.
# where DocRevYYYY.MM is the last released revision.
CXX4OPENCL_DOCREVISION = DocRev2021.12
```

Makefile
```diff
@@ -68,20 +68,36 @@ SPECREVISION = $(shell echo `git describe --tags --dirty`)
...
 endif
-
-ATTRIBOPTS_NO_VERSION	= -a revdate="$(SPECDATE)" \
-			  -a revremark="$(SPECREMARK)" \
+# C++ for OpenCL document revision scheme in aligned with its release date.
+# Revision naming scheme is as follows:
+# DocRevYYYY.MM,
+# where YYYY corresponds to its release year,
+# MM corresponds to its release month.
+# Example for the release in Dec 2021 the revision is DOCREV2021.12.
```
svenvh, line 76:
```suggestion
# Example for the release in Dec 2021 the revision is DocRev2021.12.
```

svenvh (approved):

PR #741, opened 2021-12-10
[NFC] Collocate atomic instructions requirements

bader:
This change collocates storage class requirements for atomic instructions pointer argument to make it easier to find/read. No functional changes intended.


bashbaug (approved):
Thanks, this is a nice improvement.

No reason to wait to merge this.

PR #743, opened 2021-12-16
Clarify clGetKernelSubGroupInfo error codes

FranklandJack:
Return `CL_INVALID_VALUE` for any query to `clGetKernelSubGroupInfo`
that takes an input value that is `NULL` or has an invalid size.


alycm (approved):
LGTM.  Technically this is a functional change, but this isn't much else an implementation could do (except not check the value at all).

Needs a non-Codeplay reviewer.

PR #746, opened 2022-01-07
Add CL_MEM_ALLOC_BUFFER_LOCATION_INTEL enum

sherry-yuan:
This allows the user to specify which device memory to allocate to.

CC: @mkinsner @bashbaug @GarveyJoe

### DIFF
--- a/extensions/cl_intel_unified_shared_memory.asciidoc
+++ b/extensions/cl_intel_unified_shared_memory.asciidoc
@@ -236,7 +236,6 @@ typedef cl_uint cl_mem_info_intel;
 #define CL_MEM_ALLOC_SIZE_INTEL         0x419C
 #define CL_MEM_ALLOC_DEVICE_INTEL       0x419D
 /* CL_MEM_ALLOC_FLAGS_INTEL - defined above */
-/* Enum values 0x419E-0x419F are reserved for future queries. */
 ----
 
 Enumeration type and values describing the type of Unified Shared Memory allocation.
@@ -721,7 +720,6 @@ Otherwise, it will return one of the following error values:
       | Returns allocation flags for the Unified Shared Memory allocation.
 
         Returns `0` if no allocation flags were specified for the Unified Shared Memory allocation and for `CL_MEM_TYPE_UNKNOWN_INTEL` allocations.
-
 |====
 
 ==== Using Unified Shared Memory with Kernels
@@ -1013,6 +1011,11 @@ Otherwise, it will return one of the following errors:
 * `CL_INVALID_EVENT_WAIT_LIST` if _event_wait_list_ is `NULL` and _num_events_in_wait_list_ is greater than zero, or if _event_wait_list_ is not `NULL` and _num_events_in_wait_list_ is zero, or if event objects in _event_wait_list_ are not valid events.
 * `CL_OUT_OF_RESOURCES` if there is a failure to allocate resources required by the OpenCL implementation on the device.
 * `CL_OUT_OF_HOST_MEMORY` if there is a failure to allocate resources required by the OpenCL implementation on the host.
+
+== Interactions with Other Extensions
+
+If `cl_intel_mem_alloc_buffer_location` is supported then *clDeviceMemAllocINTEL*, *clSharedMemAllocINTEL*, *clHostMemAllocINTEL*, *clGetMemAllocInfoINTEL* also
+accepts `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL` for _cl_mem_properties_intel_.
 
 == Issues
 
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -2056,7 +2056,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x419B"        name="CL_MEM_ALLOC_BASE_PTR_INTEL"/>
         <enum value="0x419C"        name="CL_MEM_ALLOC_SIZE_INTEL"/>
         <enum value="0x419D"        name="CL_MEM_ALLOC_DEVICE_INTEL"/>
-            <unused start="0x419E" end="0x419F"/>
+        <enum value="0x419E"        name="CL_MEM_ALLOC_BUFFER_LOCATION_INTEL"/>
+            <unused start="0x419F" end="0x419F"/>
     </enums>
 
     <enums start="0x41A0" end="0x41DF" name="enums.41A0" vendor="Qualcomm">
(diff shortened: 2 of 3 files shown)


mkinsner:
LGTM

extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -721,7 +722,11 @@ Otherwise, it will return one of the following error values:
       | Returns allocation flags for the Unified Shared Memory allocation.
 
         Returns `0` if no allocation flags were specified for the Unified Shared Memory allocation and for `CL_MEM_TYPE_UNKNOWN_INTEL` allocations.
+| `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL`
+  | cl_intel_mem_alloc_buffer_location
```
EwanC, line 726: Does the `cl_intel_mem_alloc_buffer_location` type need defined, or is it defined somewhere already? Think `intel` part should also be moved so it's a suffix of the type.

sherry-yuan: `cl_intel_mem_alloc_buffer_location` is defined in `cl_intel_mem_alloc_buffer_location.asciidoc`, I kept intel to be at the beginning because all other extensions eg. `cl_intel_mem_channel_property.asciidoc` uses `intel` before type (specifically it uses `cl_intel_mem_channel_property`).

EwanC: The naming convention for the extension name itself is all good, but the convention for types is different and is to have the suffix at the end. E.g `cl_mem_properties_intel`, `cl_mem_alloc_flags_intel`.  So i'd expect to see this similarly defined somewhere like `typedef foo cl_mem_alloc_buffer_location _intel`

sherry-yuan: Thank you, I have changed it to align with the types in `cl_intel_mem_alloc_buffer_location.asciidoc`

xml/cl.xml
```diff
@@ -6326,6 +6327,7 @@ server's OpenCL/api-docs repository.
             </require>
             <require comment="cl_mem_properties_intel">
                 <enum name="CL_MEM_ALLOC_FLAGS_INTEL"/>
+                <enum name="CL_MEM_ALLOC_BUFFER_LOCATION_INTEL"/>
```
bashbaug, line 6330: This enum needs to be defined in its own block specifically for this new extension vs. adding it ot the USM block.  It's very possible that an implementation will support USM but not this enum.

The XML block specifically for this new extension could look something like:

```xml
        <extension name="cl_mem_alloc_buffer_location_intel" supported="opencl">
            <require>
                <type name="CL/cl.h"/>
            </require>
            <require comment="cl_mem_properties_intel">
                <enum name="CL_MEM_ALLOC_BUFFER_LOCATION_INTEL"/>
            </require>
            <require comment="cl_mem_alloc_info_intel">
                <enum name="CL_MEM_ALLOC_BUFFER_LOCATION_INTEL"/>
            </require>
        </extension>
```

Debatably there should also be sections for `cl_mem_properties` and `cl_mem_info` but this won't matter in practice.

extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -231,12 +231,13 @@ Optional allocation properties may also be queried using *clGetMemAllocInfoINTEL
...
 
-#define CL_MEM_ALLOC_TYPE_INTEL         0x419A
-#define CL_MEM_ALLOC_BASE_PTR_INTEL     0x419B
-#define CL_MEM_ALLOC_SIZE_INTEL         0x419C
-#define CL_MEM_ALLOC_DEVICE_INTEL       0x419D
+#define CL_MEM_ALLOC_TYPE_INTEL               0x419A
+#define CL_MEM_ALLOC_BASE_PTR_INTEL           0x419B
+#define CL_MEM_ALLOC_SIZE_INTEL               0x419C
+#define CL_MEM_ALLOC_DEVICE_INTEL             0x419D
+#define CL_MEM_ALLOC_BUFFER_LOCATION_INTEL    0x419E
```
bashbaug, line 238: I'm fine saying that the USM extension interacts with the new `cl_intel_mem_alloc_buffer_location` extension (see the [section](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/extensions/cl_extension_template.asciidoc#interactions-with-other-extensions) of the extension template that describes how to do this), but I don't think that this enum should be listed directly in the USM spec.  Many USM implementations will not support the buffer location feature.

extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -721,7 +722,11 @@ Otherwise, it will return one of the following error values:
       | Returns allocation flags for the Unified Shared Memory allocation.
 
         Returns `0` if no allocation flags were specified for the Unified Shared Memory allocation and for `CL_MEM_TYPE_UNKNOWN_INTEL` allocations.
+| `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL`
+  | cl_uint
+      | Returns buffer location for the Unified Shared Memory allocation.
 
+        Return the ID of default global memory if no buffer location property were specified.
```
bashbaug, line 729: This description should also move out of the USM spec and into the `cl_intel_mem_alloc_buffer_location` extension spec.

bashbaug (changes requested):
The content of this extension looks good but I think we need to move a few things around to keep this extension distinct from the USM extension.  Thanks!

xml/cl.xml
sherry-yuan: Thank you! I have moved buffer location reference in USM to extension's own definition (both in xml and asciidoc). Specifically:

Moved the definition in xml from `cl_mem_properties_intel` and `cl_mem_alloc_info_intel` to self contained `<extension name="cl_mem_alloc_buffer_location_intel"...`

Also removed extention's reference in `cl_intel_unified_shared_memory.asciidoc` and moved table buffer location row of `cl_mem_info_intel` to `extensions/cl_intel_mem_alloc_buffer_location.asciidoc`

bashbaug: Thanks, the XML updates look good now.

extensions/cl_intel_unified_shared_memory.asciidoc
bashbaug: This change still needs to be reverted, and an "Interactions with Other Extensions" could be added in its place.

extensions/cl_intel_unified_shared_memory.asciidoc
bashbaug: I see that this section has been moved to the `cl_intel_mem_alloc_buffer_location` spec but it hasn't been removed from the USM spec.

extensions/cl_intel_unified_shared_memory.asciidoc
sherry-yuan: I have removed all reference of buffer_location, and moved some info under interaction section. Sorry I originally misunderstood it to be adding "interaction" to buffer_location extension doc (the interaction section is still present in buffer_location doc).

bashbaug (approved):
Thanks, all of my issues have been addressed, LGTM.

@mkinsner @garveyjoe @EwanC would you like to take one last look before merging?

EwanC (approved):
LGTM, thanks for the changes to help keep this distinct from USM

xml/cl.xml
```diff
@@ -6401,6 +6402,17 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_MEM_CHANNEL_INTEL"/>
             </require>
         </extension>
+        <extension name="cl_mem_alloc_buffer_location_intel" supported="opencl">
```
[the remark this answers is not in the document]
sherry-yuan: Thank you, I have changed its name and squashed the commits :)

mkinsner (approved):
LGTM

PR #748, opened 2022-01-14
Fix parameter description of command-buffer API

EwanC:
The `clCommandCopyBufferRectKHR` entry-point defined by `cl_khr_command_buffer` has an incorrect description for parameter `mutable_handle`.

It should be identical to the `mutable_handle` descriptions for all the other command recording entry-points in the extension.


alycm (approved):
LGTM.  Not a functional change in the sense that mutability is not possible with this extension alone.

Needs a non-Codeplay reviewer.

bashbaug (approved):
LGTM, aligns the description with the descriptions for the other APIs.

PR #749, opened 2022-01-14
Reserve additional MR 193 enum

EwanC:
After revision 8 of internal MR 193 we need to reserve another enum value, `0x12B1`. Apologies for following quite quickly after a previous reservation for the same internal MR in https://github.com/KhronosGroup/OpenCL-Docs/pull/747


bashbaug (approved):
> Apologies for following quite quickly after a previous reservation

Not a problem, these are easy to review and it beats incorrectly assigning an enum to two features :).

PR #756, opened 2022-02-09
Fix clGetLayerInfo signature.

Kerilk:
The first two parameters were swapped. See: https://github.com/KhronosGroup/OpenCL-Docs/blob/master/ext/cl_loader_layers.asciidoc
I discovered this while trying to build the layers from the generated extension headers.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -4001,8 +4001,8 @@ server's OpenCL/api-docs repository.
         </command>
         <command>
             <proto><type>cl_int</type>                                  <name>clGetLayerInfo</name></proto>
-            <param><type>size_t</type>                                  <name>param_value_size</name></param>
             <param><type>cl_layer_info</type>                           <name>param_name</name></param>
+            <param><type>size_t</type>                                  <name>param_value_size</name></param>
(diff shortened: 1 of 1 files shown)


Kerilk:
I was lucky to find this one, and I was responsible for it, but I overlooked it while reviewing https://github.com/KhronosGroup/OpenCL-Headers/pull/161 because the generated file layout is quite different.
It could be useful to use clang format or another formatting tool (see https://github.com/KhronosGroup/OpenCL-Headers/pull/161#issuecomment-830451456) on the repository before merging the code generation, and after generating the files. This way the comparison would be easier and mistake like these would have a better chance of being caught. If that sounds reasonable I could work on a PR for this.

alycm (approved):
LGTM, the XML is clearly the file in the wrong as `param_value_size` comes directly before `param_value` in all other API calls; and because the spec, headers and ICD loader all agree with the suggested change.

bashbaug (approved):
Nice catch!

I did a pretty thorough review of the generated files but I think I missed this one since we just started generating the header for the layers recently.

PR #757, opened 2022-02-13
Update Khronos Group copyrights to 2022

alycm:
This PR does not change non-Khronos Group copyrights.

For reference in 2023: I updated the copyrights using the below command and
then manually checked the result -- `cl_extension_template.asciidoc` is the
only file with an oddity in that regard.

```sh
find . \
     -type f \
     ! -path './.git/*' \
     ! -path './generated/*' \
     ! -path './out/*' \
     -exec sed -i {} -e 's/\(Copyright.*20[0-9][0-9]\)-2021/\1-2022/g' \
                     -e 's/\(Copyright.*2021\)/\1-2022/g' \;
```

### DIFF
--- a/CXX_for_OpenCL.txt
+++ b/CXX_for_OpenCL.txt
@@ -1,4 +1,4 @@
-// Copyright 2019-2021 The Khronos Group. This work is licensed under a
+// Copyright 2019-2022 The Khronos Group. This work is licensed under a
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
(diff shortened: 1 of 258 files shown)


alycm:
> Would it make sense to add your script to the "scripts" directory to make it easier to find next year?

Yes, that would make it easier to find.  I'll add a few basic sanity checks.

bashbaug:
I'm going to go ahead and merge this.  It's almost May, seems like it's been around long enough!

Hopefully this doesn't introduce many additional merge conficts...

PR #758, opened 2022-02-14
Update Arm copyrights to 2022

alycm:
One file listed both a Khronos Group and Arm copyright, as this is a vendor
extension I assumed that Arm was the correct choice.  No other vendor extension
has a copyright comment, just the copyright in the text, so I just deleted the
Khronos Group lines.

### DIFF
--- a/extensions/cl_arm_controlled_kernel_termination.asciidoc
+++ b/extensions/cl_arm_controlled_kernel_termination.asciidoc
@@ -21,7 +21,7 @@ Anastasia Stulova, Arm Ltd. +
 
 == Notice
 
-Copyright (c) 2021 Arm Ltd.
+Copyright (c) 2021-2022 Arm Ltd.
 
 == Status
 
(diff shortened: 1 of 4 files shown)


alycm:
@kpet As far as I know it is standard practice in Khronos repo's to just mass-update the copyright dates, e.g. as recently done by [Vulkan](https://github.com/KhronosGroup/Vulkan-Docs/commit/15d807ce4839d8feb523ca5c133a42a2aa448ade) and [OpenXR](https://github.com/KhronosGroup/OpenXR-Docs/commit/ea5b74076bb92cf84eddb4ed312d3534a08dccb1).

I'm not really sure if it serves much purpose, or has advantages over excluding the date, but to be honest just doing the formulaic update is less work than thinking about that!

kpet:
Right, that was my guess. Agree it's not worth spending too much time discussing the relative merits of all options :). Thanks for doing this! I'll merge this one since it's Arm-only and very straightforward.

PR #762, opened 2022-02-18
cl_intel_bfloat16_conversions

bashbaug:
This PR adds the cl_intel_bfloat16_conversions specification that adds built-in functions to convert between 32-bit floating point values and bfloat16 values.

The related SPIR-V extension draft for bfloat16 conversions can be found here:

https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_bfloat16_conversion.asciidoc

I will keep this as a draft PR and extension version 0.9.0 initially to allow for last minute changes before publication, but I believe the writeup is complete and I would appreciate any feedback.  Thanks!

### DIFF
--- a/extensions/cl_intel_bfloat16_conversions.asciidoc
+++ b/extensions/cl_intel_bfloat16_conversions.asciidoc
@@ -27,12 +27,12 @@ Copyright (c) 2022 Intel Corporation. All rights reserved.
 
 == Status
 
-Working Draft
+Shipping
 
 == Version
 
 Built On: {docdate} +
-Version: 0.9.0
+Revision: 1.0.0
 
 == Dependencies
 
@@ -217,6 +217,7 @@ If we decide to add a true `bfloat16` type we should consider other names that d
 |0.9.0|2021-09-03|Ben Ashbaugh|*Initial revision*
 |0.9.0|2021-10-01|Ben Ashbaugh|Reduced scope, resolved all open issues.
 |0.9.0|2021-10-19|Ben Ashbaugh|Fixed the names of the vector conversion functions.
+|1.0.0|2022-08-26|Ben Ashbaugh|Updated version.
 |========================================
 
 
--- a/extensions/cl_intel_bfloat16_conversions.asciidoc
+++ b/extensions/cl_intel_bfloat16_conversions.asciidoc
@@ -1,0 +1,230 @@
+:data-uri:
+:sectanchors:
+:icons: font
+:source-highlighter: coderay
+
+= cl_intel_bfloat16_conversions
+
+== Name Strings
+
+`cl_intel_bfloat16_conversions`
+
+== Contact
+
+Ben Ashbaugh, Intel (ben 'dot' ashbaugh 'at' intel 'dot' com)
+
+== Contributors
+
+// spell-checker: disable
+Ben Ashbaugh, Intel +
+Alexey Sotkin, Intel +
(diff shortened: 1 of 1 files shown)


extensions/cl_intel_bfloat16_conversions.asciidoc
```diff
@@ -0,0 +1,231 @@
...
+The only supported rounding mode is implicitly round-to-nearest-even.
+No explicit rounding modes are supported.
+
+Supported scalar and vector data types:
+
+[width="100%",options="header"]
+|====
+| destType | sourceType
+
+| `bloat16` (as `ushort`)
```
alycm, line 120:
```suggestion
| `bfloat16` (as `ushort`)
```

extensions/cl_intel_bfloat16_conversions.asciidoc
```diff
@@ -0,0 +1,231 @@
...
+  | `float`
+
+| `bloat162`, `bfloat163`, `bfloat164`, `bfloat168`, `bfloat1616` +
```
alycm, line 123:
```suggestion
| `bfloat162`, `bfloat163`, `bfloat164`, `bfloat168`, `bfloat1616` +
```

extensions/cl_intel_bfloat16_conversions.asciidoc
```diff
@@ -0,0 +1,231 @@
...
+  (as `ushort2`, `ushort3`, `ushort4`, `ushort8`, `ushort16`)
+  | `float2`, `float3`, `float4`, `float8`, `float16`
+
+| `float`
+  | `bfloat16` (as `ushort`)
+
+| `float2`, `float3`, `float4`, `float8`, `float16`
+  | `bloat162`, `bfloat163`, `bfloat164`, `bfloat168`, `bfloat1616` +
```
alycm, line 131:
```suggestion
  | `bfloat162`, `bfloat163`, `bfloat164`, `bfloat168`, `bfloat1616` +
```

alycm (approved):
LGTM, just a few typos.

bashbaug:
Thanks, that's embarassing.  I'm not sure if my spell checker was too aggressive or not aggressive enough!  I've accepted all three of the spelling fixes.

alycm:
Some comments that are not part of my review because they are technical comments on a shipping vendor extension, but leaving them here for posterity should this ever be considered more widely.

> Please note that this extension currently does not introduce a `bfloat16` type to OpenCL C and instead the built-in functions convert to or from a `ushort` 16-bit unsigned integer type with a bit pattern that represents a `bfloat16` value.

Not adding a new type is interesting as there is precedence for a storage-only floating-point type in OpenCL (`half` when `cl_khr_fp16` is not supported).  The only advantages that I can think of doing so is reduced implementation work, and opening the possibility for using integer arithmetic operations on the type, i.e. bit-level manipulation of the bfloat16.

> Because `bfloat16` ends with a number this does lead to awkward function names like `intel_convert_bfloat1616_as_ushort16`, but the awkward-ness is preferable to the ambiguity without the vector size suffix.
>
> If we decide to add a true `bfloat16` type we should consider other names that do not end in a number (`bfloat16_t`?).

Yes, the paired numbers aren't great for reading, it requires both knowing the `bfloat16` is a type, and that `type**n**` is the vector convention to be able to read something like `bloat162`.

PR #772, opened 2022-03-11
Add additional enum names within already allocated block

mkinsner:


alycm (approved):
LGTM, merging immediately.

bashbaug:
It's fine to have merged this as-is, but I do think these enums are now going to be considered "orphaned": https://github.com/KhronosGroup/OpenCL-Docs/pull/730

mkinsner:
> It's fine to have merged this as-is, but I do think these enums are now going to be considered "orphaned": #730

Agreed.  I'll add an extension block with another PR shortly to fix that.

alycm:
> It's fine to have merged this as-is, but I do think these enums are now going to be considered "orphaned": #730

That's a good point!  Hopefully we can get that script into the pre-merge checks in the coming months.

PR #775, opened 2022-03-23
Add missing __opencl_c_atomic_scope_device guards

svenvh:
The OpenCL 3.0 specification states that:

>   functions that do not have memory_scope argument have the same
>   semantics as the corresponding functions with the memory_scope
>   argument set to memory_scope_device.

and that memory_scope_device requires:

>   [...] OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device
>   feature.

The fix for issue #357 added the `__opencl_c_atomic_scope_device`
feature macro to most atomic builtin functions that do not take a
scope argument, but not for the scope-less overloads of
atomic_compare_exchange_strong/weak_explicit.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -7079,7 +7079,8 @@ bool atomic_compare_exchange_strong(
     volatile A *object,
     C *expected, C desired)
 
-// Requires OpenCL C 3.0 or newer.
+// Requires OpenCL C 3.0 or newer and the __opencl_c_atomic_scope_device
+// feature.
 bool atomic_compare_exchange_strong_explicit(
     volatile __global A *object,
     __global C *expected,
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
Mostly LGTM, nice catch.

Is there one more overload that needs to be updated similarly, around line 7253 (the generic adddress space overload that doesn't have the memory scope argument)?

svenvh:
> Is there one more overload that needs to be updated similarly, around line 7253 (the generic adddress space overload that doesn't have the memory scope argument)?

Nice catch from your side!  And there's a similar generic address space overload for the "_strong" builtin; I'll update both.

bashbaug:
Actually, looks like there are two overloads that should be updated similarly:

1. The generic address space overload for `atomic_compare_exchange_weak_explicit` mentioned above around line 7253.
2. The generic address space overload for `atomic_compare_exchange_strong_explicit` around line 7123.

Thanks!

bashbaug (approved):

PR #776, opened 2022-03-25
Fix font style for some atomic argument references

svenvh:


bashbaug (approved):
Thanks for fixing these!

PR #777, opened 2022-04-05
Tighten the definition of CL_IMAGE_{ROW,SLICE}_PITCH queries

### LINKED ISSUE #763: Value returned for CL_IMAGE_ROW_PITCH and CL_IMAGE_SLICE_PITCH queries
Both queries are currently defined rather loosely and the value returned by implementations is not validated by the CTS.

Taking the example of `CL_IMAGE_ROW_PITCH`, it is currently defined as:

> Return calculated row pitch in bytes of a row of elements of the image object given by image.

What "calculated" means isn't made explicit in the description for the query itself but several other sections of the specification describe that pitch "calculation" as "element size * width" which is indeed what most implementations seem to return for the query.

There are cases where a user can observe a linearly tiled representation of the image with a pitch that's different from this "calculated" pitch:
- When an image is created with a host pointer and `image_row_pitch` is set by the application to a non-zero value that is different from the calculated pitch
- When an image is created from a buffer and `image_row_pitch` is set by the application to a non-zero value that is different from the calculated pitch
- When an image that was not created with a host pointer is mapped for host access (the pitch is chosen and returned by the implementation).

This means an application that queries `CL_IMAGE_ROW_PITCH` may get a value back that has no relationship to any observable layout of the image data which makes that query essentially useless. The lack of CTS testing means that returning any value has been considered "conformant" so far.

I suggest we redefine those queries such that they return the pitches used for all application visible linearly tiled representations of the full image's data and update the CTS to require implementations adhere to the new behaviour:
- The data pointed to by an image's host pointer
- The data in a buffer the image was created from
- The mapped view of the image (when the entire image is mapped)

kpet:
Fixes #763

Change-Id: I50875d6aa1ffdf6152832c68ac6cf6bdcb434304

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -3430,7 +3430,7 @@ include::{generated}/api/version-notes/CL_IMAGE_ELEMENT_SIZE.asciidoc[]
 
 include::{generated}/api/version-notes/CL_IMAGE_ROW_PITCH.asciidoc[]
   | {size_t_TYPE}
-      | Returns the row pictch in bytes of a row of elements of the
+      | Returns the row pitch in bytes of a row of elements of the
         image object given by _image_ when image data is linearly tiled.
         If _image_ was created with a non-zero value for `image_row_pitch`,
         then the value provided for `image_row_pitch` by the application is
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -3431,7 +3431,7 @@ include::{generated}/api/version-notes/CL_IMAGE_ELEMENT_SIZE.asciidoc[]
 include::{generated}/api/version-notes/CL_IMAGE_ROW_PITCH.asciidoc[]
   | {size_t_TYPE}
       | Returns the row pitch in bytes of a row of elements of the
-        image object given by _image_ when image data is linearly tiled.
+        image object given by _image_. +
         If _image_ was created with a non-zero value for `image_row_pitch`,
         then the value provided for `image_row_pitch` by the application is
         returned, otherwise the returned value is calculated as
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -3430,15 +3430,25 @@ include::{generated}/api/version-notes/CL_IMAGE_ELEMENT_SIZE.asciidoc[]
 
 include::{generated}/api/version-notes/CL_IMAGE_ROW_PITCH.asciidoc[]
   | {size_t_TYPE}
-      | Return calculated row pitch in bytes of a row of elements of the
-        image object given by _image_.
+      | Returns the row pictch in bytes of a row of elements of the
```
bashbaug, line 3433:
```suggestion
      | Returns the row pitch in bytes of a row of elements of the
```

bashbaug (approved):

api/opencl_runtime_layer.asciidoc
kpet: Thanks!

bcalidas:
I feel that it may be premature to make this edit since the concept of linear tiling has not been explained in the spec. Also, since  row_pitch = image_width × size of element applies only when the image is created with host_ptr != NULL, I feel it does not make sense to return this value back.  My suggestion would be -  if the image has linear tiling then image_row_pitch should reflect the true image row pitch as laid out in device memory. In all other cases the queried  image-row_pitch would not have meaning.

kpet:
@bcalidas As discussed in the teleconference I've removed the mention of linear tiling. I agree we haven't introduced the concept to the specification yet and it is not necessary to mention as part of this change. PTAL.

bashbaug:
Discussed on the call May 3rd.  We're not sure if it is possible to create an image from a buffer and to also pass a non-NULL host_ptr.  Need to create an issue for tracking.  Otherwise, ok to merge.

bcalidas:
The current text looks ok.

alycm (approved):
LGTM.

bashbaug:
I filed https://github.com/KhronosGroup/OpenCL-Docs/issues/789 to track the possible issue we discussed on the call.  Merging!

PR #779, opened 2022-04-14
Consistently mark enumerant names

Alexx999:
Tokens from the main block are marked rather inconsistently - some are separated in their own "enums" elements, but many are not.
This adds some of the missing enums.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1438,16 +1438,28 @@ server's OpenCL/api-docs repository.
         <enum value="0x1074"        name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_8BIT_KHR"/>
         <enum value="0x1075"        name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_4x8BIT_PACKED_KHR"/>
             <unused start="0x1076" end="0x107F" comment="Reserved for cl_device_info"/>
+    </enums>
+
+    <enums start="0x1080" end="0x1083" name="cl_context_info" vendor="Khronos">
         <enum value="0x1080"        name="CL_CONTEXT_REFERENCE_COUNT"/>
         <enum value="0x1081"        name="CL_CONTEXT_DEVICES"/>
         <enum value="0x1082"        name="CL_CONTEXT_PROPERTIES"/>
         <enum value="0x1083"        name="CL_CONTEXT_NUM_DEVICES"/>
+    </enums>
+
+    <enums start="0x1084" end="0x1085" name="cl_context_properties" vendor="Khronos">
         <enum value="0x1084"        name="CL_CONTEXT_PLATFORM"/>
         <enum value="0x1085"        name="CL_CONTEXT_INTEROP_USER_SYNC"/>
+    </enums>
+
+    <enums start="0x1086" end="0x108F" name="cl_device_partition_property" vendor="Khronos">
         <enum value="0x1086"        name="CL_DEVICE_PARTITION_EQUALLY"/>
         <enum value="0x1087"        name="CL_DEVICE_PARTITION_BY_COUNTS"/>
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -1648,33 +1723,57 @@ server's OpenCL/api-docs repository.
...
         <enum value="0x1297"             name="CL_COMMAND_BUFFER_STATE_KHR"/>
         <enum value="0x1298"             name="CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR"/>
             <unused start="0x1299" end="0x12A7" comment="Used by future command-buffer extensions"/>
+    </enums>
+
+    <enums start="0x12A8" end="0x12A8" name="cl_command_type" vendor="Khronos">
         <enum value="0x12A8"             name="CL_COMMAND_COMMAND_BUFFER_KHR"/>
+    </enums>
+
+    <enums start="0x12A9" end="0x12B1" name="cl_device_info" vendor="Khronos">
```
bashbaug, line 1765: This change cause one of the scripts ("gencl") executed by the spec Makefile to start to complain:

```sh
scripts/gencl.py    -registry xml/cl.xml -o /home/bashbaug/git/OpenCL-Docs/generated/api apiinc
WARNING: Attempt to redefine cl_command_type (this should not happen)
WARNING: Attempt to redefine cl_device_info (this should not happen)
WARNING: Attempt to redefine cl_kernel_sub_group_info (this should not happen)
```

I'm not sure if the best way to fix this is in the script or in the XML file, but I bet @oddhack knows...

xml/cl.xml
```diff
@@ -1714,10 +1813,16 @@ server's OpenCL/api-docs repository.
...
+    </enums>
+
+    <enums start="0x2033" end="0x2034" name="cl_kernel_sub_group_info" vendor="Khronos">
         <enum value="0x2033"        name="CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE"/>
         <enum value="0x2033"        name="CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE_KHR"/>
         <enum value="0x2034"        name="CL_KERNEL_SUB_GROUP_COUNT_FOR_NDRANGE"/>
         <enum value="0x2034"        name="CL_KERNEL_SUB_GROUP_COUNT_FOR_NDRANGE_KHR"/>
+    </enums>
+
+    <enums start="0x2035" end="0x201F" name="enums.2035" vendor="Khronos" comment="Reserved for interop with other APIs">
```
bashbaug, line 1825: This is no worse than it was previously, but not all of these enums are used for "interop with other APIs".

I'm fine addressing this with a subsequent PR if you'd like, but please file an issue referencing this comment so we don't forget about it.  Thanks!

bashbaug (changes requested):
Mostly LGTM, though I'd like to see the warning in the spec Makefile cleaned up before merging.

We should also be ready just in case these changes break somebody's workflow - the XML file tends to be used in surprising ways.

Thanks for fixing all of these!  This is a nice improvement.

xml/cl.xml
oddhack: Probably best to suppress the warning in the script *if* there is no 'type' attribute on the \<enums> tag. If there's a 'type' attribute, then the \<enums> are defining a derived type with that name, rather than the name simply being a label on a bunch of constants. If you want I'll propose an MR to that effect.

xml/cl.xml
Alexx999: the comment is obviously wrong changing _everything_ at once is usually not the best way to get PRs accepted :)

oddhack: This is just historical hangover - when the block was initially reserved that was its purpose. Sometimes it's useful to actually keep related enums contiguous and in those cases actually reserving a block specific to that purpose makes sense, but it doesn't here and now.

PR #780, opened 2022-04-21
Fix minor command-buffer extension typo

EwanC:
The `cl_command_buffer_khr` return type of `clCreateCommandBufferKHR` is missing the `_khr` suffix in its definition.

### DIFF
--- a/ext/cl_khr_command_buffer.asciidoc
+++ b/ext/cl_khr_command_buffer.asciidoc
@@ -481,7 +481,7 @@ The function
 indexterm:[clCreateCommandBufferKHR]
 [source]
 ----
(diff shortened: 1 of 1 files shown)


kpet (approved):
LGTM

bashbaug (approved):
Nice catch, thank you for fixing this.

Seems like this is another reason we should move to using generated snippets from the XML file in these specifications!

kpet:
Not sure people are aware of this but since I landed 1606dc8, it's actually very easy to do and snippets are already generated by default and you could have fixed the prototype by replacing the code with:
```
include::{generated}/api/protos/clCreateCommandBufferKHR.txt[]
```

PR #781, opened 2022-04-22
Add specification for cl_khr_subgroup_rotate

kpet:
Change-Id: I3acf34660b83c65533d244c7ceb17ec5219d1949


kpet:
This version still uses plain English to describe the operations. We may want to use pseudo-code instead as was done in the corresponding SPIR-V specification.

bashbaug (approved):
LGTM!

I have a few minor cosmetic fixes but I'll create a separate PR with them.

PR #786, opened 2022-04-30
publish cl_khr_work_group_uniform_arithmetic

bashbaug:
Publishes the `cl_khr_work_group_uniform_arithmetic` extension.

### DIFF
--- a/env/extensions.asciidoc
+++ b/env/extensions.asciidoc
@@ -367,7 +367,7 @@ For instructions requiring these capabilities, _Scope_ for _Execution_ may be:
 
 * *Workgroup*
 
-For the instructions *OpGroupUniformLogicalAndKHR*, *OpGroupUniformLogicalOrKHR*, and *OpGroupUniformLogicalXorKHR*, the valid type for _X_ is *OpTypeBool*.
+For the instructions *OpGroupLogicalAndKHR*, *OpGroupLogicalOrKHR*, and *OpGroupLogicalXorKHR*, the valid type for _X_ is *OpTypeBool*.
 
 Otherwise, for the *GroupUniformArithmeticKHR* scan and reduction instructions, valid types for _X_ are:
 
--- a/env/extensions.asciidoc
+++ b/env/extensions.asciidoc
@@ -355,6 +355,26 @@ declare the following SPIR-V capabilities:
 
 * *GroupNonUniformRotateKHR*
 
+==== `cl_khr_work_group_uniform_arithmetic`
+
+If the OpenCL environment supports the extension `cl_khr_work_group_uniform_arithmetic`, then the environment must accept modules that declare use of the extension `SPV_KHR_uniform_group_instructions` via *OpExtension*.
+
+If the OpenCL environment supports the extension `cl_khr_work_group_uniform_arithmetic` and use of the SPIR-V extension `SPV_KHR_uniform_group_instructions` is declared in the module via *OpExtension*, then the environment must accept modules that declare the following SPIR-V capabilities:
+
+* *GroupUniformArithmeticKHR*
+
+For instructions requiring these capabilities, _Scope_ for _Execution_ may be:
+
+* *Workgroup*
+
+For the instructions *OpGroupUniformLogicalAndKHR*, *OpGroupUniformLogicalOrKHR*, and *OpGroupUniformLogicalXorKHR*, the valid type for _X_ is *OpTypeBool*.
+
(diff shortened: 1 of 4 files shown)


alycm (approved):
LGTM, only trivial comments.

env/extensions.asciidoc
```diff
@@ -355,6 +355,26 @@ declare the following SPIR-V capabilities:
...
+
+If the OpenCL environment supports the extension `cl_khr_work_group_uniform_arithmetic` and use of the SPIR-V extension `SPV_KHR_uniform_group_instructions` is declared in the module via *OpExtension*, then the environment must accept modules that declare the following SPIR-V capabilities:
+
+* *GroupUniformArithmeticKHR*
+
+For instructions requiring these capabilities, _Scope_ for _Execution_ may be:
+
+* *Workgroup*
+
+For the instructions *OpGroupUniformLogicalAndKHR*, *OpGroupUniformLogicalOrKHR*, and *OpGroupUniformLogicalXorKHR*, the valid type for _X_ is *OpTypeBool*.
```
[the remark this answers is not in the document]
bashbaug: Thank you - not sure how I messed this up!

ext/cl_khr_work_group_uniform_arithmetic.asciidoc
```diff
@@ -0,0 +1,239 @@
...
+int work_group_scan_inclusive_logical_and(int predicate);
+int work_group_scan_inclusive_logical_or(int predicate);
+int work_group_scan_inclusive_logical_xor(int predicate);
+
+int work_group_scan_exclusive_logical_and(int predicate);
+int work_group_scan_exclusive_logical_or(int predicate);
+int work_group_scan_exclusive_logical_xor(int predicate);
+----
+
+For the following functions, the generic type name `gentype` may be one of the supported built-in scalar data types `int`, `uint`, `long`, or `ulong`.
```
StuartDBrady, line 61: I might be a bit too late in noticing this, and this might be part of the SPIR-V extension and too late to fix there too, but I note that `char`, `uchar`, `short` and `ushort` are accepted for the reduce functions in _cl_khr_subgroup_non_uniform_arithmetic_, and for its scan functions too (also absent above), and also for the _cl_khr_subgroup_clustered_reduce_ reductions.

StuartDBrady: Ah, I see that this is listed as resolved further below, but the rationale of the resolution isn't especially clear. I can't imagine bitwise operations for `short`/`ushort` to be _especially_ useful but it might have been somewhat useful to have allowed `char`/`uchar`. Either way, I presume it's now too late.

bashbaug: I do think it's too late for this version of this extension, but it could always be added in a subsequent extension, or perhaps even as a newer version of this extension if this functionality is useful.

I suspect it may also be possible to pattern-match most cases that upconvert to int -> scan -> downconvert?  Or are there other reasons to prefer "native" versions that accept types smaller than 32-bit types?

StuartDBrady: Narrowing conversions are a potential source of bugs (hence `-Wnarrowing` etc.).

Widening, performing the scan, and narrowing might well be valid, but any code that _contains_ a narrowing conversion is of code that needs consideration.  It is better to have type-safe interfaces than to rely on manual checking of code.

It may help to consider somewhat equivalently, _char *strstr(const char *haystack, const char *needle)_ in C, which leaks the _const_ qualifier. It's _possible_ to cast to put it back, but saying that it's _possible_ to cast isn't the same as saying it's _entirely justified_ that it is necessary to  (as the cast is certainly not _desirable_). In C++, this was split into two separate functions, _char *strstr(char *haystack, const char *needle)_ and _const char *strstr(const char *haystack, const char *needle)_, made possible due to function overloading.

It might well be worth adding an extended types extension to fill this gap. I'm not saying that this is urgent – just that I think we should recognize this as a gap in the interface, and keep this is mind in the future (e.g. if we were perhaps to fold all of the subgroup and workgroup extensions together under a single extension for simplicity).

StuartDBrady: Note that warnings about down-conversions are enabled by default for C++11, so unless the narrowing cast is done _explicitly_, I think the down-conversion might actually cause a diagnostic for C++ for OpenCL.  Yes, explicit casts might be better than implicit casts – but I don't think that means explicit casting is actually _good_ if there's an alternative which does not require the use of casts at all.

PR #792, opened 2022-05-09
Fix __remove_address_space documentation code example

Topotuna:
Previously code example didn't compile because of missing keyword
'typename'. Keyword is now inserted.


AnastasiaStulova (approved):
LGTM! Thanks for this fix!

bashbaug:
Fixes a bug: https://godbolt.org/z/TWd6ccxE6

No reason to wait to merge this I think.

PR #798, opened 2022-05-14
clLinkProgram return value when link fails

SunSerega:
The last paragraph of [clLinkProgram](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clLinkProgram) sais:

> Otherwise [clLinkProgram](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clLinkProgram) returns a NULL program object with an appropriate error in errcode_ret. The application should query the linker status of this program object to check if the link was successful or not. The list of errors that can be returned are:

But I see the contradiction in it because the application cannot query the linker status of the NULL program object.

It was probably meant that `NULL program object` is returned only if `errcode_ret` is neither `CL_SUCCESS` nor `CL_LINK_PROGRAM_FAILURE`. And such behavior can be implied by the text above this paragraph, but...

I tested what is returned when link begins but fails with this program:
```
void p(); // Never implemented
// Some implementations will ignore "p" if it is not used by kernel
kernel void k(int x) { p(); }
```
On 3 different PCs, each with 1 platform of a different vendor:

- On NVidia with up-to-date OpenCL3.0 driver a NULL program object is returned, making it impossible to get linking logs.
- On Intel with up-to-date OpenCL1.2 driver a valid program object is returned, as long as the link has started.
- On AMD with derelict OpenCL2.0 driver `clLinkProgram` behaves either way, depending on which devices I add to the context.

(program objects returned by `clLinkProgram` and passed to `pfn_notify` (when it was not NULL) were equal to each other in all tested cases)

That is all to say, I think the current description of `clLinkProgram` is not clear enough.

---

This pull is a draft because it is only an example of how the description of `clLinkProgram` and be improved.
I can see some problems with these changes:

- These changes cannot be applied to OpenCL3.0 specs, because they invalidate some implementations - like one by NVidia on my PC. I guess this kind of change should go to OpenCL3.1 or something... Though I'm not sure where it would be appropriate to suggest that.

- I made sure not to change the meaning of the spec section (except for what is returned when the link fails), but I'm unsure about some things. For instance:
   > If the linking operation can begin, [clLinkProgram](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clLinkProgram) returns a valid non-zero program object.
   
   This seems to imply that `clLinkProgram` can return a non-zero program object even when `pfn_notify` is not NULL. But this program object would not be usable before `pfn_notify` is called, and would be redundant after.
   
- English is not my native language, so while I have checked my changes with Grammarly, there are probably some not obvious to me issues with my changes.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -6121,7 +6121,7 @@ is thread-safe.
 
 // refError
 
-If the program object is created but the link failed, one of two case occur:
+If the program object is created but the link failed, one of two cases occur:
 
 1. if _pfn_notify_ is `NULL`, then {CL_LINK_PROGRAM_FAILURE} is returned in _errcode_ret_.
 2. if _pfn_notify_ is not `NULL`, then the application should query the linker status of
@@ -6129,7 +6129,7 @@ If the program object is created but the link failed, one of two case occur:
 
 The application can query the log of the created program object using
 {clGetProgramInfo}(_program_, {CL_PROGRAM_BUILD_LOG}, ...)
-to get detailed information when the link failed.
+to get detailed information when the link has failed.
 
 Otherwise, if the program object is not created, {clLinkProgram}
 returns one of the following errors in _errcode_ret_:
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -6085,23 +6085,14 @@ include::{generated}/api/version-notes/clLinkProgram.asciidoc[]
   ** _user_data_ will be passed as an argument when _pfn_notify_ is called.
      _user_data_ can be `NULL`.
 
-If _pfn_notify_ is not `NULL`, {clLinkProgram} does not need to wait for the
-linker to complete, and can return immediately once the linking operation can
-begin.
-Once the linker has completed, the _pfn_notify_ callback function is called
-which returns the program object returned by {clLinkProgram}.
-Any state changes of the program object that result from calling {clLinkProgram}
-(e.g. link status or log) will be observable from this callback function.
-This callback function may be called asynchronously by the OpenCL
-implementation.
-It is the applications responsibility to ensure that the callback function
-is thread-safe.
+The linking operation can begin if the context, list of devices, input
+programs and linker options specified are all valid and appropriate host and
+device resources needed to perform the link are available.
 
-If _pfn_notify_ is `NULL`, {clLinkProgram} does not return until the linker
-has completed.
+Once the linking operation can begin, a valid non-zero program object is created.
 
-{clLinkProgram} creates a new program object which contains the library or
-executable.
+Once the linking operation has been completed and is successful, the program object
+created by {clLinkProgram} contains the valid library or executable.
 The library or executable binary can be queried using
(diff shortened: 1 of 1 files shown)


SunSerega:
I'm sorry if that's impatient, but it's been almost 4 months and no one even commented. Can it be because I made this as a draft? I didn't think that could be a problem because there is another draft... Though that one also seems abandoned.

It seems pulls are actively being worked upon right now. If no one is getting notifications of this because it's a draft, I'll turn this into a proper PR in about a day...

alycm:
Hi @SunSerega, apologies that no one got to this.  Using draft is okay.

Yes, I see the problem you're describing.  The last three sentences in the existing spec are pretty confusing.  Logically they say:

```
If pfn_notify is NULL then succeed or set link error.

If pfn_notify is not NULL then return quickly and provide any link errors via callback.

Otherwise clLinkProgram returns a NULL program object with an appropriate error in errcode_ret.
And also a logically impossible statement about querying the NULL program.
```

But "otherwise" is a confusing word because it's not immediately clear what situations weren't covered by the previous two sentences.  Based on the error codes I think it's clear that it means all the listed situations **except** for the ones resulting in `CL_LINK_PROGRAM_FAILURE`.

Both `clBuildProgram` and `clCompileProgram` also take a `pfn_notify` parameter without having this confusion, so `clLinkProgram` should likely be written more like them.  They describe the logic around deferred errors in the `pfn_notify` parameter description.

You list different behaviour for existing implementations, on a first reading of your proposed wording changes it's not clear to me whether you restrict that, but if you do that would only be possible in a new specification version.

SunSerega:
The difference with `clBuildProgram` and `clCompileProgram` is that `clLinkProgram` creates a new program object.

Can you please also look at the changes I proposed? I put the most effort into not changing the current behavior of  `clLinkProgram`, when straightening it out.
And so I think it should be a better starting point, than trying to apply descriptions of `clBuildProgram`/`clCompileProgram`.

SunSerega:
> if you do that would only be possible in a new specification version

Yes, probably. But it can also be treated as a bug in some implementations, caused by misinterpreted spec.
But that would only be feasible if *fixing* it wouldn't be too hard. In this sense, the biggest concern is NVidia's implementation, because it didn't return the program object in any of my tests, so they may not have any ready code for this.

Can you please ping someone who makes the default OpenCL layer for CUDA?

SunSerega:
> on a first reading of your proposed wording changes it's not clear to me

Well, I guess it's not directly said. Instead:

1. > The linking operation can begin if (...)

1. > Once the linking operation can begin, a valid non-zero program object is created.

1. > Any state changes of the program object created by `clLinkProgram` (e.g. link status or log) will be observable from either return value (...) or passed to `pfn_notify` (...)

So, returning NULL if inputs for `clLinkProgram` are valid for linking to begin is invalid according to spec with my changes.

But, while not ambiguous, I guess it can be made more direct. For instance:

> Once the linking operation can begin, a valid non-zero program object is created **and will be made accessible by the caller of `clLinkProgram` as described below**.

bashbaug:
This is a tricky issue!

I created #1075 for discussion.  Maybe we can decide what we want to clarify first, then we can figure out what we need to change in the spec?

Issue #799, opened 2022-05-14
Interaction between Command Buffers and clSetKernelArg


FreddieWitherden:
Reading through the command buffer extension specification it is not 100% clear how kernels which are added to command buffers interact with clSetKernelArg.  For example, if one records a kernel into a command buffer and then changes a parameter of the kernel (i) before the command buffer is enqueued and (ii) while the command buffer is pending execution?

There are some references to `cl_khr_mutable_dispatch` but it is not clear what role this plays.

Expected outcome: Calling clSetKernelArg should not impact any current or pending executions of a command buffer, but should affect any future enqueues.  Being able to change arguments of a small subset of kernels is extremely useful in HPC simulations (a classic example of a parameter that changes between runs is the simulation time `t`).

EwanC:
Great question, kernel arguments are captured at the time `clCommandNDRangeKernelKHR` is called, and are then
immutable. Calling `clSetKernelArg` in both (i) and (ii) cases have no effect on the already recorded kernel.

However, I do think the ability to modify the configuration of recorded commands between command-buffer enqueues is valuable, and the OpenCL Working Group have ideas of a more explicit way of supporting this functionality. `cl_khr_mutable_dispatch` mentioned refers to an extension the Working Group have not released yet which enables the functionality in case (i). We are intending on releasing that provisionally soon, and this feedback will hopefully accelerate that.

In the meantime, I can add some language clarifying the interaction with `clSetKernelArg`, and also update the section on 
[interaction with other extensions](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_interactions_with_other_extensions_2) to make it more descriptive, rather than just naming extensions which aren't public yet (since the names are in fact no longer correct).

FreddieWitherden:
Thanks.

I also have some comments around relevant areas for future extensions and in ensuring that these are not blocked by the current design.  (Namely the embedability of command buffers inside other buffers, the ability for command buffers to support allocation and deallocation, and the freedom for command buffers to use their own queues internally.)  Is it okay if I create issues for these?

EwanC:
Please do, those all sound like interesting directions.

bashbaug:
Confirming: will #803 fix this issue, or is additional action required?  I don't want to link the PR and have it close this issue prematurely.  Thanks!

bashbaug:
#803 is merged, FYI.

EwanC:
[...]

PR #803, opened 2022-05-23
Expand on command-buffer interactions

EwanC:
Clarifications to the cl_khr_command_buffer extension specification to address feedback from [Issue 799](https://github.com/KhronosGroup/OpenCL-Docs/issues/799). Does not contain changes to the API or intended implementation behaviour.

* Elaborates on the role of the extension as a base which other layered extensions are to build upon to expand functionality. Removing outdated references to particular extension names. When new layered extensions are released, then they can be linked back in here.

* The interaction of kernel commands with `clSetKernelArg` is documented as part of the *kernel* parameter description.

* Remove "mutable" from `cl_ndrange_kernel_command_properties_khr` type description, as other properties could end up being added here in layered extensions, not just those related to mutability.

### DIFF
--- a/ext/cl_khr_command_buffer.asciidoc
+++ b/ext/cl_khr_command_buffer.asciidoc
@@ -131,17 +131,59 @@ the capability is optional to enable optimizations on command-buffer recording.
 
 === Interactions with Other Extensions
 
-All command recording entry-points return a mutable-command handle for
-modifying the command between enqueues of the command-buffer. However,
-functionality for mutating commands is not supported by this
-extension. These are provided for other extensions layered on top to use,
-such as `cl_khr_mutable_dispatch`.
+The introduction of the command-buffer abstraction enables functionality
+beyond what the `cl_khr_command_buffer` extension currently provides, i.e.
+the recording of immutable commands to a single queue which can then be
+executed without commands synchronizing outside the command-buffer. It is
+intended that extra functionality expanding on this will be provided as layered
+extensions on top of `cl_khr_command_buffer`.
 
-Only a single command-queue specified on creation of a command-buffer can
-be used for recording commands to in this `cl_khr_command_buffer` extension.
-The `cl_khr_heterogeneous_replay` extension allows commands to be recorded across
-multiple queues in the same command-buffer, providing replay of heterogeneous
-task graphs.
+Having `cl_khr_command_buffer` as a minimal base specification means that the
+API defines mechanisms for functionality that is not enabled by this extension,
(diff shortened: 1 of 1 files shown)


ext/cl_khr_command_buffer.asciidoc
```diff
@@ -127,17 +127,59 @@ the capability is optional to enable optimizations on command-buffer recording.
...
+but the handles will allow individual commands in a command-buffer to be
+referenced by the user. In particular, the capability for an application to use
+these handles to modify commands between enqueues of a command-buffer is
+envisaged.
+
+==== List of Queues
+
+Only a single command-queue can be associated with a command-buffer in the
+`cl_khr_command_buffer` extension, but the API is designed with the intention
+that a future extension will allows commands to be recorded across multiple
```
bashbaug, line 165: Very minor typo:

```suggestion
that a future extension will allow commands to be recorded across multiple
```

bashbaug (approved):
Updates LGTM - thanks!  We should merge this next week.

jansol:
Should this explicitly mention the (obvious) implication that recording a kernel multiple times with different arguments is allowed?

alycm (approved):
LGTM.

> Should this explicitly mention the (obvious) implication that recording a kernel, changing kernel arguments and then recording that kernel again with the new arguments is valid?

I have no objections to being explicit, but as you say the implication is already obvious.

EwanC:
>Should this explicitly mention the (obvious) implication that recording a kernel, changing kernel arguments and then recording that kernel again with the new arguments is valid?

Added a sentence onto the end of the description of the `kernel` parameter.

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -1530,7 +1572,11 @@ corresponding desired value. The list is terminated with 0. If no properties are
 required, _properties_ may be `NULL`. This extension does not define any
 properties.
 
-_kernel_ A valid kernel object.
+_kernel_ A valid kernel object which **must** have its arguments set. Any
+changes to _kernel_ after calling *clCommandNDRangeKernelKHR*, such as with
+*clSetKernelArg* or *clSetKernelExecInfo*, have no effect on the recorded
+command. If _kernel_ is recorded to a following *clCommandNDRangeKernelKHR*
+command however, then that command will capture the current state of _kernel_.
```
jansol, line 1579:
```suggestion
command however, then that command will capture the updated state of _kernel_.
```

bashbaug:
Merging as discussed, we can still make additional improvements in another PR if needed.  Thanks!

PR #804, opened 2022-05-24
Fix incorrect command-buffer error case

EwanC:
`clEnqueueCommandBufferKHR()` in the cl_khr_command_buffer extension currently has an error case defined as

* `CL_INVALID_COMMAND_QUEUE` if *command_queue* is not a valid command-queue.

However there is no parameter named `command_queue` in `clEnqueueCommandBufferKHR()`, only `queues` and `num_queues`. This patch updates the error condition to reflect this and return the error code if any queue in the `queues` list is not a valid command-queue.


MathiasMagnus (approved):
LGTM too.

PR #807, opened 2022-06-21
An extension that relaxes the constant address space requirement of printf format strings and string args

pjaaskel:
There's an example CPU implementation for the 1.2 address spaces in a [PoCL branch](https://github.com/parmance/pocl/commit/8e9b1002c3952f572515312180251e4f4a5f4314). Looking for feedback and another implementer.

The key motivation for this extension is to minimize the need for modifications of codes using printf when moving between host and device, and to support running CUDA/HIP-based kernels which call printf (that in fact delegates from the device to host printf, whatever that happens to be). 

@AnastasiaStulova @bashbaug @Kerilk

### DIFF
--- a/extensions/cl_ext_relaxed_printf_string_address_space.asciidoc
+++ b/extensions/cl_ext_relaxed_printf_string_address_space.asciidoc
@@ -1,4 +1,4 @@
-cl_pocl_relaxed_printf_address_space
+cl_ext_relaxed_printf_address_space
 ====================================
 
 // This section needs to be after the document title.
@@ -15,7 +15,7 @@ include::../config/attribs.txt[]
 
 == Name Strings
 
-+cl_pocl_relaxed_printf_address_space+
++cl_ext_relaxed_printf_address_space+
 
 == Contact
 
@@ -29,7 +29,7 @@ Brice Videau, Argonne National Laboratory
 
 == Notice
 
-Copyright (c) 2022 Parmance
+Copyright (c) 2022 Parmance and Argonne National Laboratory.
 
 == Status
 
@@ -38,7 +38,7 @@ Early Draft
 == Version
 
 Built On: {docdate} +
-Revision: 1
+Revision: 2
 
 == Dependencies
 
@@ -49,7 +49,7 @@ This extension requires OpenCL 1.2.
 == Overview
 
 The printf as defined OpenCL 1.2 deviates from the C99 printf enough that makes
-using it as a drop-in function for programs originally written in C99 and using
+using it as a drop-in function for programs originally written in C99 that use
 printf sometimes cumbersome.
 
 This extension addresses a major gap in this regard by allowing the format
@@ -67,7 +67,6 @@ arguments can be dynamically produced during the execution time of the kernel.
 Add new function overloads (added to the bottom left cell of Table 6.21): ::
 +
 
-indexterm:[clGetKernelSubGroupInfoKHR]
 [source,c]
 ----
 int printf(global char * restrict format, ...)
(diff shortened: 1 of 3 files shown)


mkinsner:
Thanks @pjaaskel.

Are you planning or aware of any plans for an accompanying SPIR-V extension (OpenCL extended instruction set) to relax the address space requirement in SPIR-V?

mkinsner:
> If I am not mistaken this would be the accompanying SPIRV PR:
https://github.com/KhronosGroup/SPIRV-Registry/pull/148

I somehow missed this.  Exactly what I was wondering - thanks!

extensions/cl_ext_relaxed_printf_string_address_space.asciidoc
```diff
@@ -0,0 +1,123 @@
...
+=== Modify Table 6.21
+
+Add new function overloads (added to the bottom left cell of Table 6.21): ::
++
+
+[source,c]
+----
+int printf(global char * restrict format, ...)
+int printf(local char * restrict format, ...)
+int printf(private char * restrict format, ...)
```
StuartDBrady, line 77: These should have the `const` qualifier, i.e.:

```
int printf(global const char * restrict format, ...)
int printf(local const char * restrict format, ...)
int printf(private const char * restrict format, ...)
```

As a side-note, the existing _constant address-space_ overload is as follows:

```
int printf(constant char * restrict format, ...)
```

It would be good to add an overload for `constant const`:

```
int printf(constant const char * restrict format, ...)
```

Otherwise, adding the `const` qualifier to a string in the `constant` address space may produce a warning, e.g.:
> warning: passing argument 1 of 'printf' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
> note: expected 'constant char *' but argument is of type 'constant const char *'.

(In my own opinion, as this _arguably_ doesn't affect semantics, this we could make this change as a fix to the OpenCL spec change, rather than adding an extra overload in this extension.  If others do not share that opinion, e.g. if there is a stronger affect on semantics than I have understood, then it would seem better to add an extra overload here than to leave this unchanged, though.)

bashbaug:
Mentioned in the September 27th teleconference:

The related Intel extension is here:
https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_non_constant_addrspace_printf.asciidoc

The main difference between this extension and the Intel extension is that this extension explicitly requires support for format strings and strings passed as %s arguments that are dynamic, whereas the Intel extension explicitly requires that these strings remain string literals.

I think we could support the dynamic string behavior, but it would require changes to our current implementation.

extensions/cl_ext_relaxed_printf_string_address_space.asciidoc
bashbaug, line 77: Should we support the generic address space as well if the generic address space is supported ("OpenCL C 2.0 or OpenCL C 3.0 with the __opencl_c_generic_address_space feature")?

pjaaskel:
> The main difference between this extension and the Intel extension is that this extension explicitly requires support for format strings and strings passed as %s arguments that are dynamic, whereas the Intel extension explicitly requires that these strings remain string literals.

I believe OpenCL puts string literals to the 'constant' AS so it would not add anything in practice, if I'm not mistaken? In the proposed extension I just clarified what are the implications of allowing these to originate from other address spaces.

extensions/cl_ext_relaxed_printf_string_address_space.asciidoc
pjaaskel: Yes, it's the second part of the specs: https://github.com/KhronosGroup/OpenCL-Docs/pull/807/files/e95c2be61869a7852c14187edeceee5d3bdec36a#diff-807af2fed1c59e28e26fe56cd246632c5f8d5f809a3dea8a005859b941bc16f5R89

Issue #816, opened 2022-08-16
cl_khr_command_buffer: can a command buffer be finalized multiple times?


bashbaug:
A question came up in a review of the cl_khr_command_buffer emulation layer:

https://github.com/bashbaug/SimpleOpenCLSamples/tree/main/layers/10_cmdbufemu

Is it valid to call clFinalizeCommandBufferKHR to finalize a command buffer multiple times?

Currently, the emulation layer only allows a command buffer to be finalized once (to transition from the "recording" to the "executable" state), the returns an error for subsequent calls, but it is not clear that this is the proper behavior.

Useful links and background:

* The spec [lifecycle diagram](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_add_new_section_section_5_x_1_command_buffer_lifecycle) shows an arrow for "Finalize" going from the "Recording" to "Executable" state, but no other arrows for "Finalize".
* The approximate Vulkan equivalent [vkEndCommandBuffer](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkEndCommandBuffer.html) requires that the command buffer be in the "recording" state.  DX12 is similar.
* There is no defined error code for this condition in the description of [clFinalizeCommandBufferKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_add_new_section_section_5_x_2_enqueuing_a_command_buffer), though.

If there is no good reason to allow clFinalizeCommandBufferKHR to be called multiple times without an error I think it would be best to clarify the spec so finalizing a command buffer multiple times is an error.  The emulation layer currently returns CL_INVALID_OPERATION in this case, but I'm very open to changing this if a different error code makes more sense.

bashbaug:
Discussed in the August 23rd teleconference.  We're trending towards calling it an error to finalize a command buffer twice, probably CL_INVALID_OPERATION.

PR #817, opened 2022-08-24
Error on command-buffer multiple finalization

EwanC:
Action issue [816](https://github.com/KhronosGroup/OpenCL-Docs/issues/816) to make it an error for a command-buffer to be finalized multiple times.

Bumped the patch version of the cl_khr_command_buffer extension as this is a behaviour change, and included contributors who I could recall participating in the teleconference discussion but weren't already listed.


bashbaug (commented):
LGTM!

PR #819, opened 2022-08-31
cl_khr_command_buffer_mutable_dispatch

EwanC:
Add extension spec for provisional extension `cl_khr_command_buffer_mutable_dispatch`, which is layered on top of `cl_khr_command_buffer`.

This extension allows a user to update kernel execution commands recorded to a command-buffer between kernel enqueues.

Added at bottom of `OpenCL_Ext` extension list, rather than next to entry for `cl_khr_command_buffer` to avoid reordering the section numbers of the following extensions.

### DIFF
--- a/ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
+++ b/ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
@@ -1,0 +1,980 @@
+// Copyright 2018-2022 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+[[cl_khr_command_buffer_mutable_dispatch]]
+== Command Buffers - Mutable Dispatch (Provisional)
+
+This extension enables users to modify the configuration of kernel execution
+commands between command-buffer enqueues.
+
+=== General Information
+
+==== Name Strings
+
+`cl_khr_command_buffer_mutable_dispatch`
+
+==== Version History
+
+[cols="1,1,3",options="header",]
+|====
+| *Date*     | *Version* | *Description*
+| 2022-08-31 | 0.9.0     | First assigned version (provisional).
+|====
+
+==== Dependencies
+
+This extension requires the `cl_khr_command_buffer` extension version 0.9.0.
+
+==== Contributors
+
+Ewan Crawford, Codeplay Software Ltd. +
+Gordon Brown, Codeplay Software Ltd. +
+Kenneth Benzie, Codeplay Software Ltd. +
+Alastair Murray, Codeplay Software Ltd. +
+Jack Frankland, Codeplay Software Ltd. +
+Balaji Calidas, Qualcomm Technologies Inc. +
+Joshua Kelly, Qualcomm Technologies, Inc. +
+Kevin Petit, Arm Ltd. +
+Aharon Abramson, Intel. +
+Ben Ashbaugh, Intel. +
+Boaz Ouriel, Intel. +
+Pekka Jääskeläinen, Tampere University +
+Nikhil Joshi, NVIDIA +
+James Price, Google +
+
+=== Overview
+
+The `cl_khr_command_buffer` extension separates command construction from
+enqueue by providing a mechanism to record a set of commands which can then be
+repeatedly enqueued. However, the commands recorded to the command-buffer are
+immutable between enqueues.
+
+`cl_khr_command_buffer_mutable_dispatch` removes this restriction, in particular,
+this extension allows the configuration of a kernel execution command in a
+command-buffer, called a _mutable-dispatch_, to be modified. This allows inputs
+and outputs to the kernel, as well as work-item sizes and offsets, to change
+without having to re-record the entire command sequence in a new command-buffer.
+
+=== Interactions with Other Extensions
+
+The `cl_command_buffer_structure_type_khr` type has been added to this
+extension for the purpose of allowing expansion of mutable functionality in
+future extensions layered on top of `cl_khr_command_buffer_mutable_dispatch`.
+Any parameter that is a structure containing a `void* next` member *must* have
+a value of `next` that is either `NULL`, or is a pointer to a valid structure
+defined by `cl_khr_command_buffer_mutable_dispatch` or an extension layered on
+top. To be a valid structure in the pointer chain the first member of the
+structure *must* be a `cl_command_buffer_structure_type_khr` identifier for the
+structure being iterated through, and the second member a `void* next` pointer
+to the next structure in the chain.
+
+[NOTE]
+====
+This approach is based on structure pointer chains in Vulkan, for more details
+see the "Valid Usage for Structure Pointer Chains" section of the Vulkan
+specification.
+====
+
+This is designed so that another extension layered on
+`cl_khr_command_buffer_mutable_dispatch` could allow modification of commands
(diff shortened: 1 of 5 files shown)


bashbaug (changes requested):
This looks great!

Would you mind adding this extension to the [quick reference page](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_quick_reference) also?

I was going to merge as-is and file an issue so we don't forget, but it looks like this needs a rebase anyhow.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+
+  The list of supported event command types defined by *clGetEventInfo* is used
+  with the matching command.
+
+| `CL_MUTABLE_DISPATCH_PROPERTIES_ARRAY_KHR`
+| `cl_ndrange_kernel_command_properties_khr[]`
+| Return the properties argument specified on _command_ recording with
+  *clCommandNDRangeKernel*.
+
+  If the properties argument specified on creation of _dispatch_ was not
```
alycm, line 648:
```suggestion
  If the properties argument specified on creation of _command_ was not
``

I'm not certain on this change, but I can't see what `dispatch` is referring to otherwise.

File: ext/cl_khr_command_buffer_mutable_dispatch.asciidoc [hunk 2]
```diff
@@ -0,0 +1,980 @@
...
+  `NULL`, the implementation must return the values specified in the
+  properties argument in the same order and without including additional
+  properties.
+
+  If the properties argument specified on creation of _dispatch_ was `NULL`,
```
alycm, line 653:
```suggestion
  If the properties argument specified on creation of _command_ was `NULL`,
```

As above.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+| Return the kernel associated with _command_ when recorded with
+  *clCommandNDRangeKernel*.
+
+  If _command_ was not recorded from a *clCommandNDRangeKernel* command, the
+  implementation must return _param_value_size_ret_ equal to 0, indicating that
+  the value returned in _param_value_ is not valid.
+
+| `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`
+| `cl_uint`
+| Return the number of work-item dimensions specified when _dispatch_ was
```
alycm, line 669:
```suggestion
| Return the number of work-item dimensions specified when _command_ was
```

As above.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+  created.
+
+  If _command_ was not recorded from a *clCommandNDRangeKernel* command, the
+  implementation must return _param_value_size_ret_ equal to 0, indicating that
+  the value returned in _param_value_ is not valid.
+
+| `CL_MUTABLE_DISPATCH_GLOBAL_WORK_OFFSET_KHR`
+| `size_t[]`
+| Return the global work-item offset set on _dispatch_ creation, or from
```
alycm, line 678:
```suggestion
| Return the global work-item offset set on _command_ creation, or from
```

As above.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+  returned by the query `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`. If a global work-item
+  offset was not set, zero is returned for each element in the array.
+
+  If _command_ was not recorded from a *clCommandNDRangeKernel* command, the
+  implementation must return _param_value_size_ret_ equal to 0, indicating that
+  the value returned in _param_value_ is not valid.
+
+| `CL_MUTABLE_DISPATCH_GLOBAL_WORK_SIZE_KHR`
+| `size_t[]`
+| Return the global work-item size set on _dispatch_ creation, or from
```
alycm, line 690:
```suggestion
| Return the global work-item size set on _command_ creation, or from
```

As above.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+  was modified. The output array contains _work_dim_ values, where _work_dim_ is
+  returned by the query `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`.
+
+  If _command_ was not recorded from a *clCommandNDRangeKernel* command, the
+  implementation must return _param_value_size_ret_ equal to 0, indicating that
+  the value returned in _param_value_ is not valid.
+
+| `CL_MUTABLE_DISPATCH_LOCAL_WORK_SIZE_KHR`
+| `size_t[]`
+| Return the local work-item size set on _dispatch_ creation, or from
```
alycm, line 701:
```suggestion
| Return the local work-item size set on _command_ creation, or from
```

As above.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -0,0 +1,980 @@
...
+  If _command_ was not recorded from a *clCommandNDRangeKernel* command, the
+  implementation must return _param_value_size_ret_ equal to 0, indicating that
+  the value returned in _param_value_ is not valid.
+
+| `CL_MUTABLE_DISPATCH_GLOBAL_WORK_SIZE_KHR`
+| `size_t[]`
+| Return the global work-item size set on _dispatch_ creation, or from
+  the most recent update via *clUpdateMutableCommandsKHR* where this value
+  was modified. The output array contains _work_dim_ values, where _work_dim_ is
+  returned by the query `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`.
```
alycm, line 693:
```suggestion
  returned by the query `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`.
  If _command_ was created with a `NULL` value for _global_work_size_
  then zero is returned for each element in the array.
```

Or, alternative wording more consistent with the other table entries:

```suggestion
  returned by the query `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR`.
  If a global work-item size was not set, zero is returned for each element in
  the array.
```

Sorry, this isn't technically editorial, but I don't see what else you would do in this case.

`clCommandNDRangeKernelKHR` inherits the rules for `global_work_size` from `clEnqueueNDRangeKernel`, where in OpenCL 2.0+ 0's are allowed, and NULL is the same as a 0.

Note that wording does not apply to changes to the value by `clUpdateMutableCommandsKHR` because this extension says

> If global_work_size is NULL then the number of global work-items in the dispatch is not changed.

I don't think the difference is a bug, I think it's different default behaviour for different situations.

alycm (approved):
LGTM.

Was doing an editorial review, though spotted one non-editorial point.

All comments made are trivial and can be easily resolved post-merge.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
EwanC: hmm well spotted, I can't think of a way to resolve that ambiguity either. Only that we could provide a note to the reader that such a situation may occur.

bashbaug (approved):

PR #820, opened 2022-09-04
add missing OpenCL 2.1 APIs to reference pages

### LINKED ISSUE #350: Add missing API entries to man/toctail
API entry points from OpenCL 2.2 onwards are missing from `man/toctail`:
* `clSetProgramSpecializationConstant`
* `clSetProgramReleaseCallback`
* `clCreateBufferWithProperties`
* `clCreateImageWithProperties`

Note that `clSetContextDestructorCallback` is *not* missing because that was included in the PR that added it.

bashbaug:
fixes #350 

Adds missing OpenCL 2.1 and 3.0 APIs to the reference pages:

* clCreateBufferWithProperties
* clCreateImageWithProperties
* clSetProgramReleaseCallback
* clSetProgramSpecializationConstant

Note that there isn't a specific page for clCreateBufferWithProperties or clCreateImageWithProperties because these APIs are currently described on the pages for clCreateBuffer and clCreateImage.  Is there a better way to do this?

### DIFF
--- a/man/toctail
+++ b/man/toctail
@@ -50,6 +50,7 @@
                             <li>Buffer Objects
                                 <ul class="Level4">
                                     <li><a href="clCreateBuffer.html" target="pagedisplay">clCreateBuffer</a></li>
+                                    <li><a href="clCreateBuffer.html" target="pagedisplay">clCreateBufferWithProperties</a></li>
                                     <li><a href="clCreateSubBuffer.html" target="pagedisplay">clCreateSubBuffer</a></li>
                                     <li><a href="clEnqueueReadBuffer.html" target="pagedisplay">clEnqueueReadBuffer</a></li>
                                     <li><a href="clEnqueueWriteBuffer.html" target="pagedisplay">clEnqueueWriteBuffer</a></li>
@@ -65,6 +66,7 @@
                             <li>Image Objects
                                 <ul class="Level4">
                                     <li><a href="clCreateImage.html" target="pagedisplay">clCreateImage</a></li>
+                                    <li><a href="clCreateImage.html" target="pagedisplay">clCreateImageWithProperties</a></li>
                                     <li><a href="clEnqueueReadImage.html" target="pagedisplay">clEnqueueReadImage</a></li>
                                     <li><a href="clEnqueueWriteImage.html" target="pagedisplay">clEnqueueWriteImage</a></li>
                                     <li><a href="clEnqueueCopyImage.html" target="pagedisplay">clEnqueueCopyImage</a></li>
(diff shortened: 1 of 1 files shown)


oddhack:
I'm not completely sure what you're asking? This just looks like changes to XML. The refpage generation script uses the 'xrefs=' attribute of the clCreateImage block to create a redirect from [clCreateImageWithProperties.html](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clCreateImageWithProperties.html) in the .htaccess in the registry SDK area (...registry/opencl/sdk/3.0/docs/man/html/.htaccess). However, I think the alias is not showing up in the navigation sidebar. Fixing that would take some scripting work.

oddhack:
You could split the page describing the two APIs into two separate [open] blocks in the source and that would result in two separate refpages, but that does require some restructuring and repetitiveness. We try really hard to avoid repeating ourselves in Vulkan since we eventually, inevitably end up with two non-identical bits of language :-(

bashbaug:
Yeah, I don't think duplicating the descriptions in the spec purely to generate two separate refpages is the right tradeoff.  If this isn't horribly objectionable I'm inclined to keep it as-is, until we can come up with a better solution.

One possibility to consider given the redirects from clCreateXXXWithProperties -> clCreateXXX is to link to the clCreateXXXWithProperties page instead?  Not sure if that's worth the bother though, and it will break linking if anyone builds the reference pages locally.

alycm (approved):
LGTM.

This definitely fixes the core issue of missing links in the navigation tree.  As for alternative approaches I think we're all agreed that anything which introduces repetition would be undesirable.

The only downside of just quietly linking to the joint page is that if I do a web search for "clCreateBufferWithProperties", for example, the top hit is the clCreateBuffer page.  Which is exactly the correct page as it describes both, but perhaps that wouldn't be clear to everyone who is searching.

Neither duplication nor a long double-name page title seem appealing though, so I would be inclined to just make this change.

oddhack:
> The only downside of just quietly linking to the joint page is that if I do a web search for "clCreateBufferWithProperties", for example, the top hit is the clCreateBuffer page. Which is exactly the correct page as it describes both, but perhaps that wouldn't be clear to everyone who is searching.

Perhaps a bigger issue is that the current 3.0 pages do not show up as the top result in this search on either Google or DDG, for me. James can advise on prioritising one similar result over another. We have done some of this for Vulkan specs as well.

If the WG wants to setup symlinks in the refpage tree for aliases, I can propose an MR. We already do it for Vulkan and I'm not sure why it wasn't done here, since the refpage tool pipeline is based on Vulkan's. Likewise with adding aliases to the sidebar navigation.

bashbaug:
> Perhaps a bigger issue is that the current 3.0 pages do not show up as the top result in this search on either Google or DDG, for me. James can advise on prioritising one similar result over another. We have done some of this for Vulkan specs as well.
> 
> If the WG wants to setup symlinks in the refpage tree for aliases, I can propose an MR. We already do it for Vulkan and I'm not sure why it wasn't done here, since the refpage tool pipeline is based on Vulkan's. Likewise with adding aliases to the sidebar navigation.

These both seem like good things to fix.  How do you want to track them so they don't slip through the cracks?  I'd like to merge this PR soon.

oddhack:
I created some tracking issues though I'm a bit backlogged ATM - James may be able to help with the prioritization issue which can be done through Apache config. From my perspective this can be merged immediately.

alycm:
Reviewed and open comments filed as issues, merging.

PR #821, opened 2022-09-05
fix pdf table wrap error in cl_khr_extended_async_copies

### LINKED ISSUE #795: pdf table wrap error in extension spec
Asciidoctor has a problem generating PDF output when a table cell does not fit onto a page, and a table in the OpenCL extension spec is exposing this problem.

The problem can be identified with the OpenCL spec toolchain via an error from optimize-pdf:

```sh
/home/bashbaug/git/OpenCL-Docs/config/optimize-pdf out/pdf/OpenCL_Ext.pdf
   **** Error: File did not complete the page properly and may be damaged.
               Output may be incorrect.
```

The problem appears to be caused by the cl_khr_extended_async_copies spec and I do not see the error if I comment out this spec.

The HTML specs are not affected.  Still, we should fix this problem before the next spec release.

bashbaug:
fixes #795

Fixes a PDF table wrap error in the extension spec description for cl_khr_extended_async_copies.

I'm open to alternative ways to solve this.  This proposed solution factors out some of the common descriptions for the extended async built-in functions into the section description, therefore shortening the descriptions in the table cells themselves.  There should be no changes to the descriptions themselves.

CC @ayalz for awareness.

### DIFF
--- a/ext/cl_khr_extended_async_copies.asciidoc
+++ b/ext/cl_khr_extended_async_copies.asciidoc
@@ -151,7 +151,7 @@ event_t async_work_group_copy_3D3D(
     size_t dst_total_plane_area,
     event_t event)
 ----
-| Perform an async copy of ((_num_elements_per_line_ * _num_lines_) * _num_planes_) elements
+| Perform an async copy of \((_num_elements_per_line_ * _num_lines_) * _num_planes_) elements
 of size _num_bytes_per_element_ from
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM.

bashbaug:
CLA bot seems stuck. I'm going to try closing and re-opening this PR to see if it causes it to become unstuck.

PR #831, opened 2022-09-07
Ensure cl.xml passes validation

### LINKED ISSUE #830: cl.xml does not validate with registry.rnc
cl.xml has a few validation errors, the most common is that the "condition" attribute is not in registry.rnc.  See this comment https://github.com/KhronosGroup/OpenCL-Docs/pull/829#issuecomment-1239389679.

Running the below command shows the issues (may require require `sudo apt install jing`).

```sh
make -C xml validate
```

alycm:
Ensure cl.xml passes validation, and add validation to CI to avoid regressions.

Some issues were fixed by changing cl.xml, some by changing registry.rnc.  I tested that XML changes didn't break anything by ensuring the generated spec doesn't change, and that the re-generated extension headers didn't change.

I tested the CI change by leaving a validation in the XML at first and ensuring that the build failed.

Fixes #830

### DIFF
--- a/xml/registry.rnc
+++ b/xml/registry.rnc
@@ -436,6 +436,7 @@ Extension = element extension {
     attribute type { text } ? ,
     attribute requires { text } ? ,
     attribute requiresCore { text } ? ,
+    attribute condition { text } ? ,
     attribute supported { StringGroup } ? ,
     attribute promotedto { text } ? ,
     attribute deprecatedby { text } ? ,
@@ -445,6 +446,7 @@ Extension = element extension {
     (
         element require {
             attribute api { text } ? ,
+            attribute condition { text } ? ,
             ProfileName ? ,
             ExtensionName ? ,
             FeatureName ? ,
@@ -478,6 +480,7 @@ InterfaceElement =
     Enum |
     element command {
         Name ,
+        attribute requires { text } ? ,
         Comment ?
     }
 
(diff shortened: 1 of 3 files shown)


xml/registry.rnc
```diff
@@ -478,6 +480,7 @@ InterfaceElement =
     Enum |
     element command {
         Name ,
+        attribute requires { text } ? ,
```
alycm, line 483: I flip-flopped on whether this should be a schema change or an XML change.  I landed on schema change because the single use of this pattern does seem valid.

It is for `clEnqueueMigrateMemINTEL` within `cl_intel_unified_shared_memory` which `requires="CL_VERSION_1_2"`.  In other cases it's the entire extension that requires an OpenCL version, but this instance is specified at a finer granularity.

bashbaug: Yeah, I remember discussing various options for `clEnqueueMigrateMemINTEL` and settling on this as the best option, though I can't find the discussion in an issue so it may have been in a PR comment.  I agree this could be a useful pattern going forward for other extensions, as well.

bashbaug (approved):

xml/registry.rnc
SunSerega: Can't see a discussion on this specifically, but [here](https://github.com/KhronosGroup/OpenCL-Headers/pull/65#issuecomment-630501006) you say you settled on this to get things un-stuck.
And the `requires="CL_VERSION_1_2"` didn't exist on `clEnqueueMigrateMemINTEL` before #290 - checked with git blame.

PR #834, opened 2022-09-10
add definition of Valid Object to glossary

### LINKED ISSUE #482: What is the definition of a "valid event"
The phrase "...valid events" is used in every API that accepts an event waitlist, but there is no definition of what a valid event is.

I think a valid event is any event that can be queried by clGetEventInfo - returned by an OpenCL API, is associated with an active context (if applicable), and has not been destroyed (if applicable).

bashbaug:
Adds the definition of a "valid object" to the glossary, with the text described in this comment:

https://github.com/KhronosGroup/OpenCL-Docs/issues/482#issuecomment-795685410

### DIFF
--- a/api/glossary.asciidoc
+++ b/api/glossary.asciidoc
@@ -799,6 +799,21 @@ Unit of execution ::
     host-thread), kernel-instance, host program, work-item or any other
     executable agent that advances the work associated with a program.
 
+Valid Object ::
+    An OpenCL object is considered valid if it meets all of the following
+    criteria:
++
+--
+  * The object was created by a successful call to an OpenCL API function.
+  * The object has a strictly positive application-owned reference count.
+  * The object has not had its backing memory changed outside of normal
+    usage by the OpenCL implementation (e.g. corrupted by the application, a
+    library it uses, the implementation itself, or any other agent that can
+    access the object's backing memory).
+
+An object is only valid in the platform where it was created.
+--
+
 Work-group ::
     A collection of related _work-items_ that execute on a single _compute
     unit_.


jlewis-austin (commented):
This looks good, but I can't find a separate issue to clarify the implementation responsibility to check NULL only, with the other conditions implementation-defined. 

I'm wondering if we need to tackle those changes along with this? Otherwise, removing the ambiguity about what valid means also implies specific requirements on error checking for errors defined in terms of _valid events_, e.g.
> [CL_INVALID_EVENT_WAIT_LIST](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_EVENT_WAIT_LIST) if event_wait_list is NULL and num_events_in_wait_list > 0, or event_wait_list is not NULL and num_events_in_wait_list is 0, or if event objects in event_wait_list are not valid events.

bashbaug:
Hmm, good point, I've taken off the "fixes" comment so merging this PR will not close the issue until this is resolved.

I can add another sentence to the glossary definition if we think this is sufficient, something like:

> Implementations are only required to check for a `NULL` object to test if an object is valid; the behavior for all other invalid objects is implementation-defined.

Or, if we think that more word-smithing is required, we can take the discussion back to the issue.

I do think this addition is useful even if there is still work to be done to define which invalid objects an implementation needs to test for.

jlewis-austin:
@bashbaug I like your idea to add the implementation requirements in the glossary. If discoverability ends up being an issue, it would be nice if we could link usages of "_valid \<objects\>_" to the glossary description--I'm not sure how feasible that is to do, though.

jlewis-austin (approved):
LGTM

alycm (approved):
LGTM.

Issue #805, opened 2022-06-08
Limit Maximum Number of Function Arguments to 255


bashbaug:
See discussion in: https://github.com/KhronosGroup/OpenCL-CTS/pull/1372

We need to update the OpenCL specification to restrict the maximum number of kernel arguments to 255 if it is not restricted to a smaller value (via e.g. the minimum supported value for CL_DEVICE_MAX_PARAMETER_SIZE). This will most likely be done as an update to the description of CL_DEVICE_MAX_PARAMETER_SIZE and as a TBD update to the OpenCL C spec.

bashbaug:
PR https://github.com/KhronosGroup/OpenCL-Docs/pull/835 updates the API spec to say that the maximum number of kernel arguments is 255.

Is this sufficient, or should we update the OpenCL C spec also?

alycm:
The [universal limits](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_universal_limits) in the SPIR-V specification are a minimum limit:

> These quantities are minimum limits for all implementations and validators. Implementations are allowed to support larger quantities. Client APIs may impose larger minimums.

So while we can't assume a SPIR-V implementation will support more than 255, it's not set in stone that that must be the limit.  I'm not suggesting that OpenCL should set a higher minimum, but there's nothing in the SPIR-V encoding to limit it beyond the 32-bit word-size.  Ridiculous program compiled to SPIR-V to make the point [here](https://godbolt.org/z/jYThGG3Ej).  However, in trying to think how else to handle it I'm just convincing myself of the need for https://github.com/KhronosGroup/OpenCL-Docs/issues/806.

We could use cop-out wording with "implementation defined", but we use that very rarely. E.g:

> Support for passing more than 255 arguments to a kernel is implementation defined.

alycm:
For the case of compiling an OpenCL C kernel to SPIR-V offline the SPIR-V consumer may be an unknown quantity and so respecting the minimums makes sense.  Strictly speaking that would need an update to the OpenCL C specification too.

Also, the API specification change in https://github.com/KhronosGroup/OpenCL-Docs/pull/835 technically only affects enqueued kernels, the original SPIR-V concern would apply to functions and function calls inside a program as well.  Again, strictly speaking that would need an update to the OpenCL C specification.

bashbaug:
[...]

Issue #806, opened 2022-06-08
add a query for the maximum number of function arguments


bashbaug:
See discussion in: https://github.com/KhronosGroup/OpenCL-CTS/pull/1372

See also: https://github.com/KhronosGroup/OpenCL-Docs/issues/805

As a longer-term solution, rather than defining a hard-coded limit of 255 function arguments, we could add a new device query for the maximum number of supported function arguments.  This could be done as an extension or in a new version of the API.

PR #835, opened 2022-09-10
limit the number of kernel arguments to 255

bashbaug:
This is a partial fix for #805.

Updates the API spec to say that the maximum number of kernel arguments is 255, both for the full and embedded profiles.

### DIFF
--- a/api/embedded_profile.asciidoc
+++ b/api/embedded_profile.asciidoc
@@ -220,6 +220,8 @@ Device Queries>> table.
       | Max size in bytes of all arguments that can be passed to a kernel.
         The minimum value is 256 bytes for devices that are not of type
         {CL_DEVICE_TYPE_CUSTOM}.
+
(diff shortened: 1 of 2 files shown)


alycm (approved):
I've been think on this in #805, but the wording here is fine.  If nothing more is said by tomorrow I'll merge this and we can fix it properly in #806.

alycm:
Merging this, discussion can continue in #805.

karolherbst:
Can we revert this? SPIR-V doesn't mandate 255 arguments. It only says that _all_ implementations have to accept at least 255. But in OpenCL we could allow and require more depending on e.g. `CL_DEVICE_MAX_PARAMETER_SIZE`. 

And we should probably just add it to the env spec, that the amount of function parameters need to be high enough to allow CL_DEVICE_MAX_PARAMETER_SIZE arguments.

PR #836, opened 2022-09-10
clarify device enqueue clk_event_t behavior

### LINKED ISSUE #307: Device Enqueue clk_event_t Behavior
There are several instances in the OpenCL C spec that includes text similar to:

> _event_ must be an event returned by enqueue_kernel or enqueue_marker or a user event.

This isn't quite sufficient though, since the event may have originally been created by one of these functions and subsequently deleted.

Suggest changing this text instead to say:

> Behavior is undefined if _event_ is not a valid event.

Note that there is an `is_valid_event` function to test whether an event is valid.

bashbaug:
fixes #307

Clarifies that behavior is undefined if an invalid device enqueue `clk_event_t` is passed to **retain_event**, **release_event**, **set_user_event_status**, or **capture_event_profiling_info**.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -10800,17 +10800,15 @@ events.
 | *Built-in Function* | *Description*
 
 | void *retain_event*(clk_event_t _event_)
-    | Increments the event reference count. _event_ must be an event
-      returned by enqueue_kernel or enqueue_marker or a user event.
+    | Increments the event reference count.
+      Behavior is undefined if _event_ is not a valid event.
 | void *release_event*(clk_event_t _event_)
     | Decrements the event reference count.
       The event object is deleted once the event reference count is zero,
       the specific command identified by this event has completed (or
-      terminated) and there are no commands in any device command queue that
+      terminated), and there are no commands in any device command queue that
       require a wait for this event to complete.
-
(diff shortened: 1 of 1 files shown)


OpenCL_C.txt
```diff
@@ -10822,20 +10820,21 @@ events.
...
    a| Captures the profiling information for functions that are enqueued as
       commands.
-      The specific function being referred to is: enqueue_kernel.
       These enqueued commands are identified by unique event objects.
       The profiling information will be available in _value_ once the
       command identified by _event_ has completed.
 
-_event_ must be an event returned by enqueue_kernel.
+      Behavior is undefined if _event_ is not a valid event returned by
+      *enqueue_kernel*.
```
alycm, line 10837:
```suggestion
Behavior is undefined if _event_ is not a valid event returned by
*enqueue_kernel*.
```

Having this sentence indented causes it to be rendered in a box because of the `a` [cell specifier](https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/) on the cell.

Removing the specifier breaks the `NOTE` at the bottom of the cell.

alycm (approved):
Textually good, just a trivial formatting issue.

PR #842, opened 2022-09-14
document implementation requirements for invalid object checking

bashbaug:
This is a follow-on to PR #834, and a possible fix for #482.

This specifies the implementation requirements to test for invalid objects:

> An OpenCL implementation must check for a `NULL` object to determine if an object is valid.  The behavior for all other invalid objects is implementation-defined.

It also links the error code descriptions to the "valid object" glossary entry, to (hopefully) make it easier to navigate from an invalid object error code in an API description to the error code description then to the valid object definition.

### DIFF
--- a/api/appendix_f.asciidoc
+++ b/api/appendix_f.asciidoc
@@ -98,7 +98,7 @@ include::{generated}/api/version-notes/CL_INVALID_BUILD_OPTIONS.asciidoc[]
 | {CL_INVALID_COMMAND_QUEUE_anchor}
 
 include::{generated}/api/version-notes/CL_INVALID_COMMAND_QUEUE.asciidoc[]
-| Returned when the specified command queue is not a valid command queue.
+| Returned when the specified command queue is not a <<valid-object-definition,valid command queue>>.
 
 | {CL_INVALID_COMPILER_OPTIONS_anchor}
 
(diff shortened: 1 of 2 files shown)


alycm (approved):
LGTM.

api/appendix_f.asciidoc
```diff
@@ -225,12 +225,12 @@ include::{generated}/api/version-notes/CL_INVALID_PIPE_SIZE.asciidoc[]
...
 
 include::{generated}/api/version-notes/CL_INVALID_PLATFORM.asciidoc[]
-| Returned when the specified platform is not a valid platform.
+| Returned when the specified platform is not a <<valid-object-definition,valid platform>>.
 
 | {CL_INVALID_PROGRAM_anchor}
 
 include::{generated}/api/version-notes/CL_INVALID_PROGRAM.asciidoc[]
-| Returned when a specified program is not a valid program object.
+| Returned when a specified program is not a <<valid-object-definition,valid program object>>.
```
jlewis-austin, line 233: This one's a little odd because clGetPlatformInfo and clGetDeviceIDs overload the platform argument to allow NULL, and specify the behavior as implementation-defined. The error description for CL_INVALID_PLATFORM conflicts a bit since it implies you have to return an error for NULL platforms. 
Nothing to hold up this change, but we might want to think about how to resolve that.

jlewis-austin (approved):
LGTM, thanks! One minor comment that can be addressed later

bashbaug:
Thanks, I've copied your comment to the original issue so it doesn't slip through the cracks.  Merging!

PR #846, opened 2022-09-18
Add all missing extensions to the extensions document

kpet:
- Remove anchor in cl_intel_mem_alloc_buffer_location that had the same name as an anchor in cl_intel_unified_shared_memory so both can be included in a parent document without warning.
- Remove executable file permissions on specification documents that had them.


kpet:
Moving forward we should probably:
- Generate the list from the XML
- Merge this document with the KHR extensions document

extensions/cl_intel_mem_alloc_buffer_location.asciidoc
```diff
@@ -137,7 +137,6 @@ Modifications to the OpenCL API Specification
 | Identifies the ID of global memory partition to which the memory should be allocated. The range of legal values is implementation defined. If the value is not valid, or the implementation is unable to allocate memory in the requested memory type, an `CL_INVALID_PROPERTY` error will be emitted.
 |====
 
-[[cl_mem_info_intel]]
```
bashbaug, line 140: Mostly a note for myself - I initially thought it would be better to rename this anchor vs. removing it entirely, but it isn't used anywhere in the extension document so it's OK to remove it.

bashbaug (approved):
LGTM!

bashbaug:
+1 for generating the list of extensions from the XML file vs. building it manually.  With the manual process it's too easy to forget to add the extension to the extensions document.

If we want to do this a couple of things we'll need to figure out are:

1. What should we do for older extensions that are still written in plaintext and haven't been ported to asciidoctor?
2. Are there some vendor extensions that are in the XML file but do not have a published extension specification?  If so, we'll need to handle this case too, e.g. by generating or otherwise routing to a placeholder.

extensions/cl_intel_mem_alloc_buffer_location.asciidoc
bashbaug: Thinking about this a bit more: while it's OK to remove the anchor in this particular case, we're likely to have an issue in the future where multiple extension documents legitimately want to use the same anchor.  Should we suggest (in the extension template, perhaps) to prefix or suffix anchor names with the extension name to avoid collisions?  Or is this not something we need to worry about right now?

kpet: Agree we will need a solution to this problem going forward. I've added a task on the newly created issue to track automatic generation (https://github.com/KhronosGroup/OpenCL-Docs/issues/848). Prefixing anchor names with the extension name sounds like a good direction. I can rename both conflicting anchors in this PR if you want but I don't really want to rename all extension anchors here :).

kpet:
All good questions, thanks! I've listed them as tasks on https://github.com/KhronosGroup/OpenCL-Docs/issues/848 which I've also linked to this PR.

extensions/cl_intel_mem_alloc_buffer_location.asciidoc
bashbaug: Yeah, no need to change anything here, this PR is good to go.  I'll resolve this conversation so there is no confusion :).

bashbaug:
Do you want to merge this as-is or do you want to finish some of the tasks in #848 first?  To be clear, I'm fine either way, I just don't want to merge this too soon.  Thanks!

kpet:
I think I'd rather merge this one as a first step. It's useful on its own and I'm not sure I'd be able to spend time on #848 in the near future. Thanks!

PR #850, opened 2022-09-28
New Command-buffer query for supported queue properties

EwanC:
This change introduces a new device query related to the command-buffer extension - `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR`. This is different from `CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR`, as we want to convey to the user that an implementation *allows* using a queue property with a command-buffer, but it is not *mandatory* to use the property with a command-buffer.

This mechanism supersedes reporting queue related values from the `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR` query. The flaw with `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR` is that it contains bits explicitly added by the command-buffer extension for reporting support for queue properties. This is a brittle design, as any new queue property added in future would need to have a new bit added here in the command-buffer extension to report support when used with command-buffers.

Instead, a better design is to have a new query reporting queue properties supported, `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR`, and keeping `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR` for capabilities unrelated to the command-queue properties.

The `CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR` use-case can now be covered by returning `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` from `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR`, so it is removed.

Make `CL_QUEUE_PROFILING_ENABLE`  the mandated minimum capability reported from  `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR`, to keep existing command-buffer extension requirement to implement profiling, which is inline with minimum requirements for host queues.

Related OpenCL repo PRs:
* https://github.com/KhronosGroup/OpenCL-Headers/pull/265
* https://github.com/KhronosGroup/OpenCL-CLHPP/pull/307
* https://github.com/KhronosGroup/OpenCL-CTS/pull/2101

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1330,7 +1330,7 @@ server's OpenCL/api-docs repository.
         <enum bitpos="0"            name="CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR"/>
         <enum bitpos="1"            name="CL_COMMAND_BUFFER_CAPABILITY_DEVICE_SIDE_ENQUEUE_KHR"/>
         <enum bitpos="2"            name="CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR"/>
-        <enum bitpos="3"            name="CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR"/>
+            <unused start="3" end="3" comment="Available for future use"/>
         <enum bitpos="4"            name="CL_COMMAND_BUFFER_CAPABILITY_MULTIPLE_QUEUE_KHR"/>
             <unused start="5" end="6" comment="Used by future command-buffer extensions"/>
             <unused start="6" end="31"/>
@@ -1780,7 +1780,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x1297"             name="CL_COMMAND_BUFFER_STATE_KHR"/>
         <enum value="0x1298"             name="CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR"/>
         <enum value="0x1299"             name="CL_COMMAND_BUFFER_CONTEXT_KHR"/>
-            <unused start="0x129A" end="0x129F" comment="Available to use"/>
+        <enum value="0x129A"             name="CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR"/>
+            <unused start="0x129B" end="0x129F" comment="Available to use"/>
         <enum value="0x12A0"             name="CL_MUTABLE_COMMAND_COMMAND_QUEUE_KHR"/>
         <enum value="0x12A1"             name="CL_MUTABLE_COMMAND_COMMAND_BUFFER_KHR"/>
         <enum value="0x12A2"             name="CL_MUTABLE_COMMAND_PROPERTIES_ARRAY_KHR"/>
@@ -7186,7 +7187,7 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
-        <extension name="cl_khr_command_buffer" revision="0.9.5" supported="opencl" depends="CL_VERSION_1_2" ratified="opencl" provisional="true">
+        <extension name="cl_khr_command_buffer" revision="0.9.6" supported="opencl" depends="CL_VERSION_1_2" ratified="opencl" provisional="true">
             <require>
                 <type name="CL/cl.h"/>
             </require>
@@ -7203,13 +7204,13 @@ server's OpenCL/api-docs repository.
             </require>
             <require comment="cl_device_info">
                 <enum name="CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR"/>
+                <enum name="CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR"/>
                 <enum name="CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR"/>
             </require>
             <require comment="cl_device_command_buffer_capabilities_khr - bitfield">
                 <enum name="CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR"/>
                 <enum name="CL_COMMAND_BUFFER_CAPABILITY_DEVICE_SIDE_ENQUEUE_KHR"/>
                 <enum name="CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR"/>
-                <enum name="CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR"/>
             </require>
             <require comment="cl_command_buffer_properties_khr">
                <enum name="CL_COMMAND_BUFFER_FLAGS_KHR"/>
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -4,7 +4,7 @@
 include::{generated}/meta/{refprefix}cl_khr_command_buffer.txt[]
 
 // *Revision*::
-//    0.9.5
+//    0.9.6
 // *Extension and Version Dependencies*::
 //     This extension requires OpenCL 1.2 or later.
 //     Buffering of SVM commands requires OpenCL 2.0 or later.
@@ -12,7 +12,7 @@ include::{generated}/meta/{refprefix}cl_khr_command_buffer.txt[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2024-07-24
+    2024-10-02
 *IP Status*::
     No known IP claims.
 *Contributors*::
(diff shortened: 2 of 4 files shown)


ext/cl_khr_command_buffer.asciidoc
```diff
@@ -452,8 +453,12 @@ Add to *Table 5*, _Device Queries_, of section 4.2:
...
   command-buffers having a <<pending_count, Pending Count>> that exceeds 1.
 
-  `CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR` Device supports the ability
-  to record command-buffers to out-of-order command-queues.
+| `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR`
+| `cl_command_queue_properties`
+| Bitmask of the supported properties with which a command-queue may be created
+  to allow a command-buffer to be executed on it. It is invalid for a
+  command-queue to be created with a property not reported and still be
+  compatible with command-buffer execution.
```
bashbaug, line 461: This mechanism will work fine for command-queue properties that are expressed via the `cl_command_queue_properties` bitfield but it won't work as well for command-queue properties that are expressed via `cl_queue_properties` property-value pairs and passed to `clCreateCommandQueueWithProperties`.  Are we OK with this?

Regardless, are there some command-queue properties that must be supported by all implementations?  For example, would we expect all implementations to support `CL_QUEUE_PROFILING_ENABLE` because this is required by [`CL_DEVICE_QUEUE_ON_HOST_PROPERTIES`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_QUEUE_ON_HOST_PROPERTIES)?

EwanC: > This mechanism will work fine for command-queue properties that are expressed via the cl_command_queue_properties bitfield but it won't work as well for command-queue properties that are expressed via cl_queue_properties property-value pairs and passed to clCreateCommandQueueWithProperties. Are we OK with this?

That's a good point, we should check that the WG is OK with this. We already return `cl_command_queue_properties` for `CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR`, so the spec has this limitation in another place. We could return a `cl_queue_properties`  from these queries rather than `cl_command_queue_properties`, but it would make the extension incompatible with 1.2 implementations, which would be a shame.

I feel like this is limitation is okay, as we can still add a bespoke `cl_device_command_buffer_capabilities_khr` capability for a `cl_queue_properties ` pair that can't be represented by a property if the situation comes up. So I don't think this changes makes things any worse for this scenario.


> Regardless, are there some command-queue properties that must be supported by all implementations? For example, would we expect all implementations to support CL_QUEUE_PROFILING_ENABLE because this is required by CL_DEVICE_QUEUE_ON_HOST_PROPERTIES?

This update would indeed remove any baseline properties that a user could rely on across all implementations. We are actually finding it non-trivial to implement command-buffer profiling in ComputeAorta properly without the profiling information from our counters being overwritten by subsequent enqueues of the same command-buffer. So would not report support for `CL_QUEUE_PROFILING_ENABLE` from `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR` until we ironed out those issues if the spec made that possible.

bashbaug: Selfishly, having mandatory support for event profiling is really nice because it enables standard and portable profilers like the OpenCL intercept layer.  I suppose it's worth observing though that profiling command buffers is one of the few features that is not currently supported by the command buffer emulation layer, so I can certainly appreciate some of the complexity.

Note that the problem for profilers may be even worse than it originally appears depending what it means if a command buffer does not support profiling: At least for the OpenCL intercept layer, when device profiling is enabled **all** command queues are created with profiling enabled.  If command buffers do not support profiling would this mean that these command queues wouldn't be able to execute command buffers at all?  If so, that would be a pretty painful restriction and it'd be tricky to work around.

FWIW, here is how I was thinking of doing event profiling in the command buffer emulation layer if this ends up being required.  This is the general case that works for out-of-order queues, but it may be simplified slightly for an in-order queue, or made a little more complex to use a marker instead of a barrier:

```c
// sequence to emulate a call to:
// clEnqueueCommandBuffer(1, &queue, cmdbuf, num_events, wait_list, &event)

cl_event start;
clEnqueueBarrierWithWaitList(queue, num_events, wait_list, &start);

// execute all of the commands in cmdbuf here:
// clEnqueueSomeCommand(queue, ...);

clEnqueueBarrierWithWaitList(queue, 0, NULL, &event);

// for the purposes of event profiling:
// * use the "start" event for QUEUED, SUBMIT, and START
// * use the "end" event for END and COMPLETE
```

This doesn't seem too bad?

EwanC: hmm I see the issue with the OpenCL intercept layer, I think we should specify that `CL_QUEUE_PROFILING_ENABLE` must be reported then, which preserves the current command-buffer spec behaviour. I can push a commit to update this PR to that.

Your proposed emulation layer implementation makes sense. We do something similar with the command-buffer objects
in our underlying API, it's just propagating the logic for working with the equivalent of start/end events that gets tricky. It's definitely achievable though, so if mandating profiling support improves the user experience then we should go with that.

bashbaug:
I lost track of this one, my apologies.  Is it ready to go?

EwanC:
> I lost track of this one, my apologies. Is it ready to go?

No worries, I've deliberately been holding back from adding it to the WG agenda because of the Mobica CTS project adding command-buffer tests. I think doing an API change in the middle of that work is a complication we should avoid. So once that's done, I think we could make this API change and I can make the according CTS update myself to reflect it.

xml/cl.xml
```diff
@@ -1727,7 +1727,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x1296"             name="CL_COMMAND_BUFFER_REFERENCE_COUNT_KHR"/>
         <enum value="0x1297"             name="CL_COMMAND_BUFFER_STATE_KHR"/>
         <enum value="0x1298"             name="CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR"/>
-            <unused start="0x1299" comment="Reserved for MR199"/>
+        <enum value="0x1299"             name="CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR"/>
```
EwanC, line 1730: Note that PR https://github.com/KhronosGroup/OpenCL-Docs/pull/899 is now also trying to claim this enum

aharon-abramson (approved):

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -457,8 +458,14 @@ Add to *Table 5*, _Device Queries_, of section 4.2:
...
   command-buffers having a <<pending_count, Pending Count>> that exceeds 1.
 
-  {CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR} Device supports the ability
-  to record command-buffers to out-of-order command-queues.
+| {CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR}
+| {cl_command_queue_properties_TYPE}
+| Bitmask of the supported properties with which a command-queue may be created
+  to allow a command-buffer to be executed on it. It is invalid for a
+  command-queue to be created with a property not reported and still be
+  compatible with command-buffer execution.
```
bcalidas, line 466: Are the queue properties for executing a command buffer mandated to be the same as the queue properties while recording the command buffer.

EwanC: Yes we currently have this [wording](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#compatible) in the spec about being *compatible*
> Command-buffers are created using an ordered list of command-queues that commands are recorded to and execute on by default. These command-queues can be replaced on command-buffer enqueue with different command-queues, provided for each element in the replacement list the substitute command-queue is compatible with the command-queue used on command-buffer creation. Where a **compatible** command-queue is defined as a command-queue with identical properties targeting the same device and in the same OpenCL context.

And then the actual error condition to `clEnqueueCommandBufferKHR` is
>CL_INCOMPATIBLE_COMMAND_QUEUE_KHR if any element of queues is not compatible with the command-queue set on command_buffer creation at the same list index.


I've ignored that `clSetCommandQueueProperty` exists because its been depreciated since OpenCL 1.1, as that could change a command-queue's properties after creation. Maybe I should put a note about not using that.

bcalidas:
Looks ok

kpet (approved):
LGTM

bashbaug:
Discussed in the December 10th teleconference.  We will convert this PR to a draft for now, and will create a separate PR that includes both the new query for supported queue properties and the queue compatibility relaxations described in #1142.

EwanC:
As discussed on the Jan 7th teleconference I've taken this PR out-of-draft, and taken all the associated ones linked in the description out of draft too

bashbaug:
Merging as discussed in the January 14th teleconference.

PR #851, opened 2022-10-04
cl_loader_layers update

Kerilk:
This is a pull request to add new info query to the cl_loader_layer `clGetLayerInfo` API.
This is a prerequisite of https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/175

### DIFF
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -23,6 +23,10 @@ interception layers (Layer ICDs) for OpenCL.
 | 2020-11-04 | 1.0.0     | First assigned version.
 |====
 
+==== Contributors
+
+Brice Videau, Argonne National Laboratory
+
 [[cl_loader_layers-new-procedures-and-functions]]
 === New Procedures and Functions
 
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -54,12 +54,6 @@ cl_int clInitLayer(cl_uint                 num_entries,
 ----
 typedef cl_uint cl_layer_info;
 typedef cl_uint cl_layer_api_version;
-typedef struct _cl_layer_option {
-    const char *name;
-    const char *value;
-    const char *default_value;
-    const char *description;
-} cl_layer_option;
 ----
 
 [[cl_loader_layers-new-enums]]
@@ -71,7 +65,6 @@ Accepted as _param_name_ to the function *clGetLayerInfo*:
 ----
 #define CL_LAYER_API_VERSION        0x4240
 #define CL_LAYER_NAME               0x4241
-#define CL_LAYER_OPTIONS            0x4242
 ----
 
 
@@ -138,12 +131,6 @@ in the <<layer-queries-table, Layer Queries>> table.
 | Layer Info | Return Type | Description
 | *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
 | *CL_LAYER_NAME* | char[] | OpenCL Layer Name
-| *CL_LAYER_OPTIONS* | cl_layer_option[]
-  | OpenCL Layer Options.
-    The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
-    Layers can, but are not required to, use the requested storage for
-    backing memory. Strings contained in the `cl_layer_option` structure
-    are `NULL` terminated if provided.
 |====
 
 *clGetLayerInfo* returns *CL_SUCCESS* if the function is executed
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -141,7 +141,7 @@ in the <<layer-queries-table, Layer Queries>> table.
 | *CL_LAYER_OPTIONS* | cl_layer_option[]
   | OpenCL Layer Options.
     The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
-    Layers can, but are not required, to used the requested storage for
+    Layers can, but are not required to, use the requested storage for
     backing memory.
 |====
 
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -64,6 +64,7 @@ Accepted as _param_name_ to the function *clGetLayerInfo*:
 [source,opencl]
 ----
 #define CL_LAYER_API_VERSION        0x4240
+#define CL_LAYER_NAME               0x4241
 ----
 
 
@@ -129,6 +130,7 @@ in the <<layer-queries-table, Layer Queries>> table.
 |====
 | Layer Info | Return Type | Description
 | *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
+| *CL_LAYER_NAME* | char[] | OpenCL Layer Name
 |====
 
 *clGetLayerInfo* returns *CL_SUCCESS* if the function is executed
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -142,7 +142,8 @@ in the <<layer-queries-table, Layer Queries>> table.
   | OpenCL Layer Options.
     The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
     Layers can, but are not required to, use the requested storage for
-    backing memory.
+    backing memory. Strings contained in the `cl_layer_option` structure
+    are `NULL` terminated if provided.
 |====
 
 *clGetLayerInfo* returns *CL_SUCCESS* if the function is executed
--- a/ext/cl_loader_layers.asciidoc
+++ b/ext/cl_loader_layers.asciidoc
@@ -54,6 +54,12 @@ cl_int clInitLayer(cl_uint                 num_entries,
 ----
 typedef cl_uint cl_layer_info;
 typedef cl_uint cl_layer_api_version;
+typedef struct _cl_layer_option {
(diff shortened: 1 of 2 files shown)


ext/cl_loader_layers.asciidoc
```diff
@@ -125,6 +137,12 @@ in the <<layer-queries-table, Layer Queries>> table.
 |====
 | Layer Info | Return Type | Description
 | *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
+| *CL_LAYER_NAME* | char[] | OpenCL Layer Name
+| *CL_LAYER_OPTIONS* | cl_layer_option[]
+  | OpenCL Layer Options.
+    The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
+    Layers can, but are not required, to used the requested storage for
+    backing memory.
```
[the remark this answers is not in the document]
Kerilk: Yes, thanks for catching that!

ext/cl_loader_layers.asciidoc
```diff
@@ -125,6 +137,12 @@ in the <<layer-queries-table, Layer Queries>> table.
 |====
 | Layer Info | Return Type | Description
 | *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
+| *CL_LAYER_NAME* | char[] | OpenCL Layer Name
+| *CL_LAYER_OPTIONS* | cl_layer_option[]
+  | OpenCL Layer Options.
+    The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
+    Layers can, but are not required to, use the requested storage for
+    backing memory.
```
bashbaug, line 145: I think we need to document a little more precisely what memory the caller is intended to allocate (and free!) and which memory is owned by the implementation and hence should not be freed.  It might be useful to have some psueudocode showing how to query and print these layer options.

Additionally, I think a caller should be able to compute the number of supported options via `param_value_size_ret`.  Is this the case?  If so, do we need to have a special `{NULL, NULL, NULL, NULL}` entry to denote the list terminator?

ext/cl_loader_layers.asciidoc
```diff
@@ -50,6 +54,12 @@ cl_int clInitLayer(cl_uint                 num_entries,
 ----
 typedef cl_uint cl_layer_info;
 typedef cl_uint cl_layer_api_version;
+typedef struct _cl_layer_option {
+    const char *name;
+    const char *value;
+    const char *default_value;
+    const char *description;
```
bashbaug, line 61: Presumably these are `NUL` terminated strings?  It might be good to say this explicitly.

bashbaug (commented):
This looks interesting!

I have a few high-level questions as well:

* Should we say anything in the extension overview to describe expected usage for layer options, or is it best to leave this open-ended for each layer?
* Do we need to bump the layer version because we are adding new layer queries?

ext/cl_loader_layers.asciidoc
Kerilk: The idea here is that the layer is able to request either a partial allocation (just the NULL terminated array), a full allocation (the array plus backing memory for the strings), or anything in-between. The user is only required to free the contiguous buffer they allocated. Given that the layer can request space for more than the array, the size requested by the layer cannot be used to determine the array length. We could also add a query to get the option count.

ext/cl_loader_layers.asciidoc
Kerilk: Indeed, will add it.

Kerilk:
Thanks for the review @bashbaug .

>     * Should we say anything in the extension overview to describe expected usage for layer options, or is it best to leave this open-ended for each layer?

I am not sure I have a lot of idea here. We could look at what @MathiasMagnus did for the layer options and see if patterns emerge, or maybe best practices? We could mandate options to control output, so users could expect uniform usage between layers.

>     * Do we need to bump the layer version because we are adding new layer queries?

I thought not, as the API stays the same and the loader / cllayerinfo can work without those being implemented. But I am very open to change on this.

ext/cl_loader_layers.asciidoc
Kerilk, line 145:
```suggestion
    backing memory. Strings contained in the `cl_layer_option` structure are `NULL` terminated if provided.
```

ext/cl_loader_layers.asciidoc
Kerilk: Done in https://github.com/KhronosGroup/OpenCL-Docs/pull/851/commits/39686e806dc0a1d27a5938de46846474785a30a9

ext/cl_loader_layers.asciidoc
```diff
@@ -125,6 +137,13 @@ in the <<layer-queries-table, Layer Queries>> table.
 |====
 | Layer Info | Return Type | Description
 | *CL_LAYER_API_VERSION* | cl_layer_api_version | OpenCL Layer API version
+| *CL_LAYER_NAME* | char[] | OpenCL Layer Name
+| *CL_LAYER_OPTIONS* | cl_layer_option[]
+  | OpenCL Layer Options.
+    The list is terminated by a `{NULL, NULL, NULL, NULL}` entry.
+    Layers can, but are not required to, use the requested storage for
+    backing memory. Strings contained in the `cl_layer_option` structure
+    are `NULL` terminated if provided.
```
Kerilk, line 146:
```suggestion
    are `NULL` terminated if provided. Users are required to allocate
    at least the requested amount of memory, and should only free
    this single allocation when they are done using the options.
    Referencing the strings after the user memory has been freed is
    undefined behavior, as the layer may have used the user
    provided allocation for backing storage.
```

Kerilk: See here for the expected usage:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/ad9dc9dfc43d25e312d4bfce7536b4316800960f/loader/cllayerinfo.c#L123-L144
```C
    result = p_clGetLayerInfo(CL_LAYER_OPTIONS, 0, NULL, &sz);
    if (CL_SUCCESS == result)
    {
        cl_layer_option *options = (cl_layer_option *)malloc(sz);
        if (options)
        {
            result = p_clGetLayerInfo(CL_LAYER_OPTIONS, sz, options, NULL);
            if (CL_SUCCESS == result)
            {
                size_t i = 0;
                printf("\tCL_LAYER_OPTIONS:\n");
                while (options[i].name)
                {
                    if (options[i].description)
                        printf("\t\t# %s\n", options[i].description);
                    printf("\t\t%s = %s # (default: %s)\n", options[i].name, options[i].value, options[i].default_value);
                    i++;
                }
            }
            free(options);
        }
    }
```
It should be used as any other OpenCL Info query. @bashbaug if you have a better way to formulate it than my clumsy attempt above, I am eager for suggestions.

ext/cl_loader_layers.asciidoc
bashbaug: This makes sense to me - thanks!

ext/cl_loader_layers.asciidoc
bashbaug: Sorry for the very slow reply.

I frequently use a pattern like this one - the same pattern is also used by the [C++ bindings](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L1034) for queries that return an array (aka vector) of values:

```c++
        size_t  sizeInBytes = 0;
        clGetPlatformInfo(
            platformId,
            CL_PLATFORM_EXTENSIONS_WITH_VERSION,
            0,
            NULL,
            &sizeInBytes );

        size_t  numExtensions = sizeInBytes / sizeof(cl_name_version);
        std::vector<cl_name_version>  extensions{ numExtensions };
        clGetPlatformInfo(
            platformId,
            CL_PLATFORM_EXTENSIONS_WITH_VERSION,
            sizeInBytes,
            extensions.data(),
            NULL );
```

If the queried "size in bytes" can contain the backing storage for the string also then this pattern won't work.  Are we OK with that?

Kerilk: Yes, you are correct and this is why the array is NULL terminated. Considering current options values could be arbitrary, the layers would be hard pressed to guarantee their existence in static memory.

ext/cl_loader_layers.asciidoc
```diff
@@ -50,6 +54,12 @@ cl_int clInitLayer(cl_uint                 num_entries,
 ----
 typedef cl_uint cl_layer_info;
 typedef cl_uint cl_layer_api_version;
+typedef struct _cl_layer_option {
+    const char *name;
+    const char *value;
+    const char *default_value;
+    const char *description;
+} cl_layer_option;
```
Kerilk, line 62:
```suggestion
```
Remove options support for now as suggested by the working group.

ext/cl_loader_layers.asciidoc
```diff
@@ -60,6 +70,8 @@ Accepted as _param_name_ to the function *clGetLayerInfo*:
 [source,opencl]
 ----
 #define CL_LAYER_API_VERSION        0x4240
+#define CL_LAYER_NAME               0x4241
+#define CL_LAYER_OPTIONS            0x4242
```
Kerilk, line 74:
```suggestion
```
Remove options support for now as suggested by the working group.

ext/cl_loader_layers.asciidoc
Kerilk, line 146:
```suggestion
```
Remove options support for now as suggested by the working group.

bashbaug (approved):
LGTM!  Merging as discussed in the January 31st teleconference.

PR #858, opened 2022-10-25
Allocate enums for upcoming Intel memory property extension

mkinsner:


xml/cl.xml
```diff
@@ -7203,5 +7205,11 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_IMAGE_REQUIREMENTS_SLICE_PITCH_ALIGNMENT_EXT"/>
             </require>
         </extension>
+        <extension name="cl_intel_create_mem_object_properties" supported="opencl">
+            <require>
```
bashbaug, line 7209: Very minor: consider adding a `comment` here to describe how these enums are used - I think they should probably be tagged `cl_mem_properties`?  The comment gets included by the header generation script and helps to self-document the extension headers.

bashbaug (commented):
LGTM, I verified the XML changes are correct by examining the generated headers.

I left one minor comment, probably should have a non-Intel reviewer?

xml/cl.xml
mkinsner: Good point.  Added.

PR #860, opened 2022-11-17
Fix description of the half precision math functions

Melirius:
Likewise in already merged https://github.com/KhronosGroup/SYCL-Docs/pull/306

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -4926,7 +4926,7 @@ The following table describes the following functions:
   * A subset of functions from <<table-builtin-math>> that are defined with
     the half_ prefix .
     These functions are implemented with a minimum of 10-bits of accuracy,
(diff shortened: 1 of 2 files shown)


bashbaug:
LGTM, but I think we need to update the OpenCL C spec as well:
https://github.com/KhronosGroup/OpenCL-Docs/blob/a897dfb2e238e7e0b623dd9b774e67a922fe52dc/OpenCL_C.txt#L4928

I did a quick check through the SPIR-V environment spec and I didn't see anything similar, so I think it's OK, but it'd be nice to have a second pair of eyes to be sure I didn't miss it.  Thanks!

Melirius:
Updated the OpenCL C spec.

bashbaug (approved):
Thank you for the fixes!

bashbaug:
Merging, fixes make sense and this will align the OpenCL and SYCL specs.

PR #863, opened 2022-12-04
A couple of fixes to clCreateSemaphoreWithPropertiesKHR

kpet:
- Remove an error case for CL_INVALID_VALUE. Invalid properties or property values are already covered by CL_INVALID_PROPERTY.
- State clearly that CL_SEMAPHORE_TYPE_KHR must be pass when creating a semaphore in the table describing the property. This was previously just documented in the description for CL_INVALID_VALUE.


bashbaug (approved):
LGTM.

At some point we may want to expand this table so it includes a column for the "default value" if the proprety is omitted, or no default value when the property is required, but this is a nice improvement the way the table is currently formatted.

bashbaug:
Merging as discussed in the December 6th teleconference.

PR #867, opened 2022-12-13
cl_khr_defined_builtin_kernels

pjaaskel:
First WiP draft of a defined BiKs extension.

See: https://github.com/KhronosGroup/OpenCL-Registry/issues/131 for the background. [HTML rendered here](https://htmlpreview.github.io/?https://raw.githubusercontent.com/KhronosGroup/OpenCL-Docs/a40b2f456f3feb192a54b72e35140c5699482e97/ext/cl_khr_defined_builtin_kernels.html).

Any input appreciated!

### DIFF
--- a/ext/cl_khr_defined_builtin_kernels.asciidoc
+++ b/ext/cl_khr_defined_builtin_kernels.asciidoc
@@ -1,0 +1,330 @@
+// Copyright 2018-2022 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+= cl_khr_defined_builtin_kernels =
+
+:source-highlighter: coderay
+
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
The documentation for `khr_blas_gemm_float` seems off.  For GEMM we need three dimensions, but only two appear to be specified through `NDRange`.  Also, the leading dimension arguments do not make sense as these are separate to row/column major (which should be its own argument).  Also, the arguments should probably be size_t to allow for large multiplications.

pjaaskel:
Thanks for the feedback. The current GEMM BiK is supposed to be an example for the general concept of the extension, but it will definitely be one of the first ones to "freeze" in the registry. We'll take your input in account while shaping it.

pjaaskel:
Continuing the discussion in the [updated PR](https://github.com/KhronosGroup/OpenCL-Docs/pull/1007).

PR #869, opened 2023-01-16
Add missing clBuildProgram call to command-buffer extension samples

EwanC:
The sample code in the `cl_khr_command_buffer` and `cl_khr_command_buffer_mutable_dispatch` extensions was missing a call to `clBuildProgram` which would be required for the sample to work.

Discovered when looking at some application code @jansol had written.


bashbaug (approved):
Nice catch!

PR #878, opened 2023-01-31
update the spec change log with changes from v3.0.12

bashbaug:
This includes all of the spec changes so far.

If we merge any other changes before release they will need to be added to the change log.

Diff: https://github.com/KhronosGroup/OpenCL-Docs/compare/v3.0.12...main


alycm (approved):
LGTM.

bashbaug:
Need to update the change log to describe the non-linear tiling issue in the cl_khr_external_memory.

bashbaug:
I added the issue description to the change log and this is ready to go now.

PR #879, opened 2023-01-31
clarify a clCreateBuffer with SVM pointer error condition

### LINKED ISSUE #875: Clarify on error code on clCreateBuffer failure with SVM buffer
The spec mentions below on the buffer allocation with a *host_ptr* from `clSVMAlloc`.

> If [clCreateBuffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBuffer) or [clCreateBufferWithProperties](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBufferWithProperties) is called with a pointer returned by [clSVMAlloc](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSVMAlloc) as its host_ptr argument, and [CL_MEM_USE_HOST_PTR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_USE_HOST_PTR) is set in its flags argument, [clCreateBuffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBuffer) or [clCreateBufferWithProperties](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBufferWithProperties) will succeed and return a valid non-zero buffer object as long as the size argument is no larger than the size argument passed in the original [clSVMAlloc](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSVMAlloc) call

In the case of an invalid *size* argument being used between `clSVMAlloc` and `clCreateBuffer`, the spec is little unclear on what would be the error code that needs to be returned. `CL_INVALID_BUFFER_SIZE` seems to be a reasonable choice but then again the description for `CL_INVALID_BUFFER_SIZE` is very specific on the values of `size` for error scenario

> [CL_INVALID_BUFFER_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_BUFFER_SIZE) if size is 0 or if size is greater than [CL_DEVICE_MAX_MEM_ALLOC_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_MEM_ALLOC_SIZE) for all devices in context.

Is `CL_INVALID_BUFFER_SIZE` a reasonable option for error code here, if it is can we add that to the spec.

bashbaug:
Fixes #875.

Clarifies that `CL_INVALID_BUFFER_SIZE` should be returned if `clCreateBuffer` is passed a pointer returned by `clSVMAlloc` as its _host_ptr_ and the size of the buffer exceeds the size of the SVM allocation.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -518,8 +518,10 @@ returned in _errcode_ret_:
     once.
   * {CL_INVALID_VALUE} if values specified in _flags_ are not valid as defined
     in the <<memory-flags-table,Memory Flags>> table.
-  * {CL_INVALID_BUFFER_SIZE} if _size_ is 0 or if _size_ is greater than
-    {CL_DEVICE_MAX_MEM_ALLOC_SIZE} for all devices in _context_.
+  * {CL_INVALID_BUFFER_SIZE} if _size_ is 0, or if _size_ is greater than
+    {CL_DEVICE_MAX_MEM_ALLOC_SIZE} for all devices in _context_, or if
+    {CL_MEM_USE_HOST_PTR} is set in _flags_ and _host_ptr_ is a pointer returned by
+    {clSVMAlloc} and _size_ is greater than the size passed to {clSVMAlloc}.
   * {CL_INVALID_HOST_PTR} if _host_ptr_ is `NULL` and {CL_MEM_USE_HOST_PTR} or
     {CL_MEM_COPY_HOST_PTR} are set in _flags_ or if _host_ptr_ is not `NULL`
     but {CL_MEM_COPY_HOST_PTR} or {CL_MEM_USE_HOST_PTR} are not set in _flags_.


bashbaug:
Possible fix - still needs agreement by the working group.

alycm (approved):
Text looks good to me and matches discussion in https://github.com/KhronosGroup/OpenCL-Docs/issues/875, however there was a request on call 2023-01-31 for some to have a little time to check their current behaviour before merging.

tgowtham:
@bashbaug thanks for creating this. Looks good to me.

kpet (approved):
LGTM, we should add a negative test to the CTS.

kpet:
I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/1701

bashbaug:
Merging as discussed on the April 25th teleconference.

PR #880, opened 2023-01-31
update the spec copyright dates to 2023

bashbaug:
Copyright dates that get included in the specs to 2023.

Note, this doesn't update all of the spec source copyright dates to 2023.  I'll do that in a separate PR.


alycm (approved):
LGTM, merging immediately as obvious.

PR #881, opened 2023-01-31
update the spec source copyright dates to 2023

bashbaug:
Updates the spec source copyright dates to 2023 using the script from https://github.com/KhronosGroup/OpenCL-Docs/pull/757.

Note: This PR includes the spec source from several vendor-specific extensions.  I can break those off into a separate PR, if desired.


alycm (approved):
LGTM.  I don't see any reason to split out vendor extensions (even if I've done so myself in the past).

PR #882, opened 2023-02-06
Update semaphore wait and signal rules in cl_khr_sempahore

lakshmih:
Addresses issue #883 

Clarify behavior of clEnqueueSignalSemaphore and clEnqueueWaitSemaphore commands.
Explain when semaphores can be safely re-used with both signal and wait commands.


bashbaug (approved):
Merging as discussed in the April 11th teleconference.

PR #885, opened 2023-02-23
Remove command-buffer Invalid state

EwanC:
It was discovered during cl_khr_command_buffer layered extension review that the [Invalid command-buffer state](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_add_new_section_section_5_x_1_command_buffer_lifecycle) is undesirable.

An approach more in-keeping with the OpenCL philosophy is for an OpenCL object to be able to update the reference count of the OpenCL objects it uses. Keeping those objects alive for its lifetime, rather than having a specific validity check. This change specifies that command-buffers match that behaviour. Any validity checking further than that may be expensive, and should not be mandated.

There is also a clarification in this change that interleaving queue submissions with command recording to the same queue is valid.

Tagging @bcalidas @bashbaug @Kerilk who commented on the original issue


ext/cl_khr_command_buffer.asciidoc
```diff
@@ -490,6 +490,19 @@ with the command-queue used on command-buffer creation. Where a _compatible_
...
+While constructing a command-buffer it is valid for the user to interleave calls
+to the same queue which create commands, such as {clCommandNDRangeKernelKHR}, with
+queue submission calls, such as {clEnqueueNDRangeKernel} or
+{clEnqueueCommandBufferKHR}. That is, there is no effect on queue state from
+recording commands. The purpose of the queue parameter is to define the device
+and properties of the command, which are constant queries on the queue object.
+
+A command-buffer object should increment the reference count of OpenCL objects
+such as queues, buffers, images, and kernels referenced in commands recorded to
+the command-buffer. This enables correct behaviour of the command-buffer when the
```
bashbaug, line 502: Minor: we've been using the US English spelling elsewhere in the spec.

```suggestion
the command-buffer. This enables correct behavior of the command-buffer when the
```

bashbaug (approved):

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -490,6 +490,19 @@ with the command-queue used on command-buffer creation. Where a _compatible_
...
+objects it uses have been released. On destruction of the command-buffer it should
+decrement these reference counts, allowing the referenced objects to be freed if
+appropriate.
```
bashbaug, line 505: Actually one more comment with a few questions about this paragraph in general:

* Is this intended to increment application-visible reference count, or the internal reference count, or is either fine?  I know this is a complicated issue so I'm OK leaving this ambiguous if we need to.
* Should some of these "shoulds" turn into "musts"?
* I don't think the command-buffer should increment the reference count of any kernel arguments set on kernels recorded into the command-buffer (though it should retain or otherwise clone the kernel itself).  Do you agree?  If so, is it worth adding some non-normative text warning about this, similar to clSetKernelArg?

bashbaug: Other parts of the spec talk about "attached" objects, e.g.

> After the device reference count becomes zero and all the objects attached to device (such as command-queues) are released, the device object is deleted.

I wonder if we should use the same term here, in which case the reference counting would fall out naturally?

> Implementations may also modify the reference count, e.g. to track attached objects...

The other term used in the spec is "associated with", but "attach" seems stronger and more appropriate for this use.

ext/cl_khr_command_buffer.asciidoc
EwanC: Done, there were also a couple places in the mutable-dispatch spec where "behaviour" had crept in that I've updated

ext/cl_khr_command_buffer.asciidoc
EwanC: ah, I didn't spot the consistent use of "attached". Rephrased this paragraph to use the term "attached" throughout

bashbaug: Thanks!  I like the updated wording.

aharon-abramson (approved):

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -492,6 +492,26 @@ with the command-queue used on command-buffer creation. Where a _compatible_
...
+{clEnqueueCommandBufferKHR}. That is, there is no effect on queue state from
+recording commands. The purpose of the queue parameter is to define the device
+and properties of the command, which are constant queries on the queue object.
+
+A command-buffer object should increment the reference count of attached OpenCL
+objects such as queues, buffers, images, and kernels referenced in commands
+recorded to the command-buffer. This enables correct behavior of the
+command-buffer when its attached objects have been released. On destruction of
+the command-buffer it should decrement these reference counts, allowing the
+attached objects to be freed if appropriate.
```
bcalidas, line 507: Is this referring to commands such as clEnqueueCopyBuffer?

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -492,6 +492,26 @@ with the command-queue used on command-buffer creation. Where a _compatible_
...
+
+[NOTE]
+====
+A command-buffer object does not update the reference count of objects set as
+arguments on kernels recorded into the command-buffer. This is consistent with
+the reference counting behavior of {clSetKernelArg}.
```
bcalidas, line 513: It feels odd that arguments to kernels are not refcounted but other objects used in commands are. We should be consistent in our approach.

bashbaug: > My perspective on this was that we are updating the ref counts on the OpenCL objects that are passed to the command creation entry-points, and for clCommandNDRangeKernelKHR that is a cl_kernel, so we're updating the ref count of the cl_kernel object.

This was the rationale I was thinking of also, and I like the symmetry with clSetKernelArg.

Note, now that we have memory object destructor callbacks it's probably safer to retain references to memory objects then it was back in the old OpenCL 1.0 days.  If we decide to "fix" this though, I think we should do so consistently, in some future version of OpenCL.

(Selfishly, it would also be really hard or impossible to retain kernel arguments in my command buffer emulation layer, though if the kernel held reference counts to its kernel arguments then this should "just work".)

bashbaug:
@bcalidas are you ok merging this as-is or would you like a little more time?

bcalidas:
Changes removing invalid state look fine. I think the kernel arg ref counting needs more discussion. I'm specifically thinking about what happens when the command buffer is executing and one of the referenced arguments ( a buffer for example ) is released.
Is it expected that the implementation will ensure correct execution of the command buffer in this situation.

bcalidas:
We had some more internal discussion on the arg ref counting. As per the OpenCL spec, "After the memobj reference count becomes zero and commands queued for execution on a
command-queue(s) that use memobj have finished, the memory object is deleted".  This means that an application can release a mem object without worrying about any undesirable effects on commands that are currently executing.

However, this guarantee is expensive for a driver to implement. With a recording that could contain hundreds of commands and thousands of arguments, the cost of this guarantee goes up astronomically. The overhead could so be high as to erase many of the advantages of command buffers.

Our recommendation is to have the command buffer hold a reference count to the objects that are referenced as arguments. 
The other option would be to make the application responsible for ensuring that command buffer execution completes prior to Releasing the object. We feel that this option is not desirable since it is not consistent with what applications typically expect from implementations.

bashbaug:
> The other option would be to make the application responsible for ensuring that command buffer execution completes prior to Releasing the object. We feel that this option is not desirable since it is not consistent with what applications typically expect from implementations.

I think we should consider this option more carefully.  Rationale:

1. It's simpler and more efficient for implementations.
2. It will be needed for SVM/USM-based kernel arguments anyhow, where memory allocations are set using pointers and do not support reference counting.
3. It really isn't that different from non-command buffer kernel argument behavior, because kernel objects explicitly do not update reference counts for their kernel arguments.

For (3) specifically the case I am considering is:

```c
clSetKernelArg(my_kernel, 0, sizeof(my_buffer), &my_buffer);
clEnqueueNDRangeKernel(my_queue, my_kernel, ...);
clReleaseMemObject(my_buffer);  // Say this decrements the reference count to zero.
                                // This is OK and safe even if my_kernel is still executing.
clFinish(my_queue);  // This is OK too - now my_kernel must be done executing.
clEnqueueNDRangeKernel(my_queue, my_kernel, ...);  // This is NOT safe, since my_buffer may (should) be deleted.
```

bcalidas:
We are open to making the application responsible for the life-cycle of objects referenced in the command buffer. Although this would need careful messaging, it's much better than moving the burden of ensuring safety to the implementation. The static ref-counting is another option, but if other vendors have concerns, we are ok with not using this option.

bcalidas:
We would like the following warning to be added to the extension text as part of the note on reference count on kernel arguments.  "Applications should ensure that objects passed as arguments to kernels recorded to the command buffer are not deleted  until the command buffer has completed execution.  Undefined behavior may result from the failure to follow this requirement".

With the addition of this note, we are fine with the spec update.

EwanC:
> We would like the following warning to be added to the extension text as part of the note on reference count on kernel arguments. "Applications should ensure that objects passed as arguments to kernels recorded to the command buffer are not deleted until the command buffer has completed execution. Undefined behavior may result from the failure to follow this requirement".
>
> With the addition of this note, we are fine with the spec update.


Pushed a [commit](https://github.com/KhronosGroup/OpenCL-Docs/pull/885/commits/e305c7984dadc3654e53318a74963603fc74dde9) expanding on this note with guidance on safe usage.

I wasn't sure that the wording "until the command buffer has completed execution." covers all cases, as a command-buffer can be executed multiple times and also cloned by the multi-device extension. So I changed the advice to be to free objects after the command-buffer has been released.

Additionally, I added a couple sentence about the mutable-dispatch extension, since that could replace all the kernel arguments where an object is used, in which case I think it would be safe to delete the object.
 
This is full note is now, but please say if you prefer the original more concise version, or some other edit.

image

bcalidas:
The note looks ok. Had a question about whether a reference to the mutable dispatch extension is necessary and/or appropriate.
Worth a quick discussion.

EwanC:
> The note looks ok. Had a question about whether a reference to the mutable dispatch extension is necessary and/or appropriate. Worth a quick discussion.

As discussed on the WG call I've split the note information relating to mutable-dispatch out into the layered spec, and linked to it from the base command-buffer specification - https://github.com/KhronosGroup/OpenCL-Docs/pull/885/commits/5d4361f742812bec60180239befc60bbf937b9c8

bcalidas:
Latest update looks good.

bashbaug:
We have two GitHub approvals and an approval comment from @bcalidas - I think this one is ready to merge?

EwanC:
> We have two GitHub approvals and an approval comment from @bcalidas - I think this one is ready to merge?

Yes I believe so, I'll click the button

PR #886, opened 2023-02-24
Add definition of cl_icdl extension.

Kerilk:
This is the formal definition of the cl_icdl extension that adds info queries to an OpenCL ICD loader.
The PR for the headers is here:
https://github.com/KhronosGroup/OpenCL-Headers/pull/214

Once those two are merged, we can update the loader and remove the temporary definition there.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -5516,9 +5516,9 @@ server's OpenCL/api-docs repository.
                 <command name="clInitLayer"/>
             </require>
         </extension>
-        <extension name="cl_icdl" supported="opencl">
+        <extension name="cl_loader_info" supported="opencl">
             <require>
-                <type name="CL/cl_icdl.h"/>
+                <type name="CL/cl.h"/>
             </require>
             <require>
                 <type name="cl_icdl_info"/>
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -223,6 +223,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mem_alloc_flags_img</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_layer_info</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_layer_api_version</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_icdl_info</name>;</type>
         <type category="define">typedef struct _cl_icd_dispatch       <name>cl_icd_dispatch</name>;</type>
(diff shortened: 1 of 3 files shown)


xml/cl.xml
```diff
@@ -5501,6 +5516,23 @@ server's OpenCL/api-docs repository.
...
             </require>
         </extension>
+        <extension name="cl_icdl" supported="opencl">
+            <require>
+                <type name="CL/cl_icdl.h"/>
+            </require>
+            <require>
+                <type name="cl_icdl_info"/>
+            </require>
+            <require comment="cl_icdl_info"/>
```
Kerilk, line 5526:
```suggestion
            <require comment="cl_icdl_info">
```

bashbaug (commented):
This mostly looks really good.  I have a few questions though:

1. Have you thought about how to publish this extension?  Some possibilities:
    * Don't publish it at all, just put it in this repo as an asciidoc document.
    * Include it in the extension spec.
    * Build it as a standalone HTML doc and publish it via the registry like an EXT or vendor extension.

2. Have you investigated using the generated asciidoctor dictionaries to help ensure a consistent look-and-feel in this extension spec?  We probably wouldn't want to do this if we just left it as an asciidoc document since the GitHub rendering doesn't look as nice, but if we're intending to publish it as an HTML spec I'd strongly recommend at least looking into this.

bashbaug:
I'm totally open to alternatives but here's one possibility:

Treat this extension like any EXT or vendor extension.  This means:

1. Publish it on the OpenCL registry as a standalone HTML doc.  Might require a small bit of refactoring to follow the format of other EXT and vendor extensions - the [extension template](https://github.com/KhronosGroup/OpenCL-Docs/blob/main/extensions/cl_extension_template.asciidoc) should hopefully be useful.
2. Include the definitions for this extension in `cl_ext.h`, like other extensions.  Rationale: it's a user-facing extension just like any other, and this way we won't need to manage a new file.
3. We'll need a good name for the extension since it will show up in the headers and have a link on the registry even if it doesn't show up in any queryable extension strings.  `cl_loader_info`, similar to `cl_loader_layers`?  `cl_icd_loader_info`?  `cl_icdl_info`?

Maybe we can discuss this as an open in tomorrow's teleconference?

Kerilk:
@bashbaug Thanks for the valuable input. I gave it a shot, what do you think?
I still need to investigate the asciidoctors dictionaries.

Kerilk:
Tagging @vdanjean a he is credited in the extension author list.

bashbaug (approved):
LGTM!

I checked and the header generation scripts "just worked" for this extension, pretty cool.

I'll leave this open for a couple more days just in case there are other comments, otherwise I'll merge first thing next week.

bashbaug:
Merging as discussed in the March 28th teleconference.

PR #895, opened 2023-03-22
Fix clInitLayer parameter name

Kerilk:
This fixes an issue where the `layer_dispatch_ret` parameter of `clInitLayer` was incorrectly named `layer_dispatch` in some part of the specification and in the xml, but not in the Headers. This is a return parameter.


bashbaug (approved):
LGTM - I've checked that the headers use `layer_dispatch_ret`.

Do you want to fix this place in the spec also?

https://github.com/KhronosGroup/OpenCL-Docs/blob/main/ext/cl_loader_layers.asciidoc#new-procedures-and-functions

The body of the spec uses `layer_dispatch_ret` but the overview still uses `layer_dispatch`.

bashbaug:
Oops, my fault, for some reason I only saw the XML change.  All good now, merging!

Kerilk:
Thanks for the quick review!

PR #897, opened 2023-03-30
Move semaphore sharing error condition to cl_khr_external_semaphore

kpet:
Semaphore sharing is not defined by the cl_khr_semaphore specification.

### DIFF
--- a/ext/cl_khr_external_semaphore.asciidoc
+++ b/ext/cl_khr_external_semaphore.asciidoc
@@ -235,6 +235,10 @@ Add to the list of supported _param_names_ by {clGetSemaphoreInfoKHR}:
         semaphore does not support any handle types for exporting.
 |====
 
+Add to the list of error conditions for {clEnqueueWaitSemaphoresKHR} and {clEnqueueSignalSemaphoresKHR}:
+
+* {CL_INVALID_COMMAND_QUEUE} if one or more of _sema_objects_ can not be shared with the device associated with _command_queue_.
+
 === Exporting semaphore external handles
 
 To export an external handle from a semaphore, call the function
--- a/ext/cl_khr_semaphore.asciidoc
+++ b/ext/cl_khr_semaphore.asciidoc
@@ -316,8 +316,7 @@ Otherwise, it returns one of the following errors:
 * {CL_INVALID_COMMAND_QUEUE}
 ** if _command_queue_ is not a valid command-queue, or
 ** if the device associated with _command_queue_ is not same as one of the devices specified by {CL_DEVICE_HANDLE_LIST_KHR} at the time of creating one or more of _sema_objects_, or
-** if one or more of _sema_objects_ belong to a context that does not contain a device associated with_command_queue_, or
-** if one or more of _sema_objects_ can not be shared with the device associated with _command_queue_.
+** if one or more of _sema_objects_ belong to a context that does not contain a device associated with_command_queue_.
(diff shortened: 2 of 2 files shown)


ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -235,6 +235,10 @@ Add to the list of supported _param_names_ by {clGetSemaphoreInfoKHR}:
         semaphore does not support any handle types for exporting.
 |====
 
+Add to the list of error conditions for {clEnqueueWaitSemaphoresKHR} and {clEnqueueSignalSemaphoresKHR}:
+
+* {CL_INVALID_COMMAND_QUEUE} if one or more of _sema_objects_ can not be shared with the device associated with _command_queue_.
```
bashbaug, line 240: Good catch, the current description is definitely in the wrong place.

I do wonder though: what this condition is intended to check for?  Under which circumstances would we return an error because "one or more of _sema_objects_ can not be shared with the device associated with _command_queue_" where we wouldn't return an error otherwise?  In other words, is this condition redundant?

There may be similar redundancy for this error condition for external memory also, for `clEnqueueAcquireExternalMemObjectsKHR` and `clEnqueueReleaseExternalMemObjectsKHR`.

bashbaug: @nikhiljnv and @bcalidas any thoughts?  If we're not sure what this condition is intended to cover I'm inclined to just remove it, both for semaphores and for external memory.

bcalidas: My interpretation is that this condition refers to the following case.

A semaphore S  is created against device A
The application creates Context X which includes device A and device B. 
The application creates CommandQueue Y which is associated with device B
The application calls EnqueueSignalSemaphore or EnqueueWaitSemaphore on semaphore S against command queue Y.

kpet: Isn't this covered by the following condition already?

> CL_INVALID_COMMAND_QUEUE if the device associated with command_queue is not same as one of the devices specified by CL_DEVICE_HANDLE_LIST_KHR at the time of creating one or more of sema_objects, or

nikhiljnv: IIRC, this was added before introducing CL_DEVICE_HANDLE_LIST_KHR and was intended to cover the case where CL_DEVICE_HANDLE_LIST_KHR is not specified during semaphore creation with a context having more than one devices and if semaphores may not be shared across devices. 

I guess most other cases should be covered by  error conditions below -

{CL_INVALID_COMMAND_QUEUE}  if the device associated with command_queue is not same as one of the devices specified by {CL_DEVICE_HANDLE_LIST_KHR} at the time of creating one or more of sema_objects

{CL_INVALID_CONTEXT} if the context associated with command_queue and any of the semaphore objects in sema_objects are not the same or if the context associated with command_queue and that associated with events in event_wait_list are not the same.

    if one or more of sema_objects belong to a context that does not contain a device associated with_command_queue_.

bashbaug (approved):
Merging as discussed in the April 11th teleconference.

We will likely remove this error condition in a subsequent PR.

ext/cl_khr_external_semaphore.asciidoc
bashbaug: Resolving - we will fix this with a subsequent PR.

Issue #898, opened 2023-03-31
Missing `CL_COMMAND_BUFFER_CONTEXT_KHR` query


kpet:
All API objects support a query to obtain the context they were created within but command buffers currently do not. Is this intentional?

EwanC:
I think this is probably an oversight, I can't remember any particular reason for omitting it. A command-buffer doesn't take a `cl_context` on creation, but neither does a `cl_kernel` and you can still query the context directly from `clGetKernelInfo`. So agree we should provide a `CL_COMMAND_BUFFER_CONTEXT_KHR` context query.

kpet:
Yeah, it can be done via the command queue query but I agree that we should add a direct query. Symmetry with other objects is desirable. Thanks!

PR #899, opened 2023-03-31
Add command-buffer context query

EwanC:
Introduce query `CL_COMMAND_BUFFER_CONTEXT_KHR` for the `cl_context` associated with a command-buffer.

This PR uses enum value `0x1299` which is also attempting to be used by another command-buffer PR https://github.com/KhronosGroup/OpenCL-Docs/pull/850 However, that PR still needs reviewed, so would prefer to use that enum here as `0x1299` is contiguous with the other command-buffer query enums, and I can find a new value for PR #850

Closes https://github.com/KhronosGroup/OpenCL-Docs/issues/898


kpet (approved):
LGTM, thanks! We should capture work to update the CTS tests as well.

PR #901, opened 2023-04-05
Fix typo in cl_ext_image_from_buffer specification

kpet:


bashbaug (approved):

bashbaug:
I totally understand if you don't want to do this right now, but switching to the asciidoctor dictionary atttribute `{clGetImageRequirementsInfoEXT}` would catch issues like these.

PR #902, opened 2023-04-07
hyphenation changes for consistency

bashbaug:
See #474 

Couple of notes:

* We had more usage of `subgroup` than `sub-group` so I've gone with that for now, though I actually think `sub-group` is more appropriate.  There is going to be some awkwardness no matter what we choose since the Vulkan spec uses `subgroup` and the SYCL spec uses `sub-group`.
* I didn't think any of the others were as controversial, but if there is a strong preference for a different usage instead just let me know.  We should be consistent now at least!
* I updated a few vendor extensions also (from Arm, Imagination, and Intel), though I'm fine reverting these changes if these vendors would prefer to leave their extensions unchanged.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -11072,9 +11072,9 @@ footnote:[{fn-int64-supported}], `ulong`, `half` footnote:[{fn-half-supported}],
 
 The *<op>* in *sub_group_reduce_<op>*, *sub_group_scan_inclusive_<op>* and *sub_group_scan_exclusive_<op>* defines the operator and can be *add*, *min* or *max*.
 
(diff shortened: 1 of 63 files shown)


alycm:
On subgroup vs sub-group:  The Vulkan spec may use subgroup, but it also uses workgroup whereas the OpenCL and SYCL specs use work-group.

bashbaug:
> The Vulkan spec may use subgroup, but it also uses workgroup [...]

Good point.  I think I'm going to bite the bullet and switch everything to sub-group consistently.  I'll get this done before tomorrow's call.  If we don't like this for whatever reason it'll all be in a separate commit so easy to roll back.  Thanks!

alycm (approved):
LGTM.

A lot of procedural fixes, and some contextual fixes around those.

Checked in the CI log that subgroupFunctions.html refpage is now subGroupFunctions.html (just something I didn't 100% understand when I saw the change).

Not clicking merge so as to avoid causing problems for current release, happy for this to merge before or straight after.

PR #903, opened 2023-04-11
Remove redundant error condition in cl_khr_external_semaphore

kpet:
cl_khr_semaphore has the following condition:

* {CL_INVALID_COMMAND_QUEUE} if the device associated with _command_queue_ is not same as one of the devices specified by {CL_DEVICE_HANDLE_LIST_KHR} at the time of creating one or more of _sema_objects_

cl_khr_external_semaphore has the following condition:

* {CL_INVALID_DEVICE} if one or more devices identified by properties {CL_DEVICE_HANDLE_LIST_KHR} can not import the requested external semaphore handle type.


kpet:
Also see discussion on https://github.com/KhronosGroup/OpenCL-Docs/pull/897

bashbaug (approved):
LGTM!

PR #905, opened 2023-04-14
cl_intel_unified_shared_memory version 1.1

bashbaug:
Updates the cl_intel_unified_shared_memory extension to version 1.1:

* Version 1.1 relaxes error conditions for `clSetKernelArgMemPointerINTEL` and `clSetKernelExecInfo(CL_KERNEL_EXEC_INFO_USM_PTRS_INTEL)` to allow passing any arbitrary pointer to a kernel even without system USM.  It is only a programming error (specifically, undefined behavior) to dereference an arbitrary pointer in a kernel.
* Also adds a few issues for questions that have been raised in the last couple of months.

I'll probably keep this a draft PR for a bit, to provide opportunity for additional commentary and while we implement this behavior.

CC: @gmlueck


bashbaug:
This change is now ready for review and merging.

MichalMrozek (approved):
lgtm

extensions/cl_intel_unified_shared_memory.asciidoc
```diff
@@ -52,7 +52,7 @@ Shipping
 == Version
 
 Built On: {docdate} +
-Revision: 1.0.0
+Revision: 1.1.0
```
EwanC, line 55: Since we're bumping the revision, this is a change to set `revision` in the XML https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml#L6699

bashbaug: Good catch, thanks!

PR #906, opened 2023-04-15
cl_intel_subgroups_short version 1.1

bashbaug:
Updates the cl_intel_subgroups_short extension to version 1.1.

Version 1.1 adds support for 16-wide vector loads and stores.

### DIFF
--- a/extensions/cl_intel_subgroups_short.asciidoc
+++ b/extensions/cl_intel_subgroups_short.asciidoc
@@ -1,7 +1,9 @@
 :data-uri:
 :sectanchors:
 :icons: font
-:source-highlighter: coderay
+:source-highlighter: rouge
+:source-language: opencl
+:rouge-style: opencl.spec
 
 = cl_intel_subgroups_short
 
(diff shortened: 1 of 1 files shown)


extensions/cl_intel_subgroups_short.asciidoc
```diff
@@ -533,10 +549,11 @@ None.
...
 [options="header"]
 |========================================
-|Rev|Date|Author|Changes
-|1|2016-10-20|Ben Ashbaugh|*First public revision.*
-|2|2018-11-15|Ben Ashbaugh|Conversion to asciidoc.
-|3|2019-09-17|Ben Ashbaugh|Added vec3 types for shuffles and asciidoctor formatting fixes.
+|Version|Date|Author|Changes
+|rev 1|2016-10-20|Ben Ashbaugh|*First public revision.*
+|rev 2|2018-11-15|Ben Ashbaugh|Conversion to asciidoc.
+|rev 3|2019-09-17|Ben Ashbaugh|Added vec3 types for shuffles and asciidoctor formatting fixes.
```
kpet, line 555: Do you want to specify what version gets reported for rev1-3? I guess there are implementations in the field that don't support versioning at all though.

bashbaug: Good point - I added a line to the table to say that v1.0.0 is equivalent to rev3.

Issue #883, opened 2023-02-06
Clarify binary semaphore reset behavior for cl_khr_semaphore.


bcalidas:
Binary semaphores must be in the correct state before they can be signaled or waited on. Clarify semaphore state transitions and when semaphores may be safely re-signaled or re-waited on.

bcalidas:
At a high level binary semaphores must be in the correct state  for use in signal and wait commands. Semaphores can be in one of the following states -

- Reset state 
- Pending signal state
- Pending wait state
- Signaled state.

Semaphores are in the reset state when created . The other states can be explained as follows.

- A semaphore of the type CL_SEMAPHORE_TYPE_BINARY_KHR will be in the pending signal state when a clEnqueueSignalSemaphoreKHR command that references them is in the ready state or running state.
- A semaphore of the type CL_SEMAPHORE_TYPE_BINARY_KHR will be in the signaled state when a clEnqueueSignalSemaphoreKHR command that references them has reached the ended state or the complete state.
- A semaphore of the type CL_SEMAPHORE_TYPE_BINARY_KHR will be in the pending wait state when a clEnqueueWaitSemaphoreKHR command that references them is in the ready state or the running state.
- A semaphore of the type CL_SEMAPHORE_TYPE_BINARY_KHR will return to  the reset state when a clEnqueueWaitSemaphoreKHR command that references them has reached the ended state or the complete state.

bcalidas:
When the clEnqueueWaitSemaphoreKHR command called on one or more semaphore objects of the type CL_SEMAPHORE_TYPE_BINARY_KHR, has reached the ended state, the corresponding semaphore objects are unsignaled and their state is reset. 
- A subsequent clEnqueueWaitSemaphoreKHR command on these semaphores may not transition to ready state without the semaphores having first transitioned to either a pending signal or signaled state.  Violation of this rule may lead to implementation-defined behavior.

- When the clEnqueueSignalSemaphoreKHR command called on one or more semaphore objects of the type CL_SEMAPHORE_TYPE_BINARY_KHR, has reached the ended state, the corresponding semaphore objects are signaled, and their state is set to signaled. 
- 
- A subsequent clEnqueueSignalSemaphoreKHR command on these semaphores may not transtion to ready state without the semaphores having first transitioned to either a pending wait or reset state. A Violation of this rule may lead to implementation-defined behavior.

bashbaug:
[...]

PR #910, opened 2023-04-17
update the spec change log with changes from v3.0.13

bashbaug:
Diff: https://github.com/KhronosGroup/OpenCL-Docs/compare/v3.0.13...main

### DIFF
--- a/api/appendix_e.asciidoc
+++ b/api/appendix_e.asciidoc
@@ -561,3 +561,12 @@ Changes from *v3.0.12*:
   * Added missing calls to {clBuildProgram} to the `cl_khr_command_buffer` and `cl_khr_command_buffer_mutable_dispatch` sample code.
   * Fixed a copy-paste error in the extensions quick reference appendix.
   * Fixed typos and improved formatting consistency in the extensions spec.
+
+Changes from *v3.0.13*:
+
+  * Corrected the precision for `cross` and `dot` to be based on `HALF_EPSILON` in `cl_khr_fp16`.
+  * Added a context query for command-buffers to `cl_khr_command_buffer`.
+  * Updated the semaphore wait and signal rules in `cl_khr_semaphore`.
+  * Removed redundant error conditions from `cl_khr_external_semaphore` and `cl_khr_external_memory`.
+  * Added new extension:
+      ** `cl_khr_command_buffer_multi_device` (provisional)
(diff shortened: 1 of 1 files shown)


api/appendix_e.asciidoc
```diff
@@ -561,3 +561,12 @@ Changes from *v3.0.12*:
   * Added missing calls to {clBuildProgram} to the `cl_khr_command_buffer` and `cl_khr_command_buffer_mutable_dispatch` sample code.
   * Fixed a copy-paste error in the extensions quick reference appendix.
   * Fixed typos and improved formatting consistency in the extensions spec.
+
+Changes from *v3.0.13*:
+
+  * Corrected the precision for `cross` and `dot` to be based on `HALF_EPSILON` in `cl_khr_fp16`.
+  * Added a context query for command-buffers to `cl_khr_command_buffer`.
+  * Updated the semaphore wait and signal rules in `cl_khr_semaphore`.
```
kpet, line 569: I'm wondering whether we shouldn't say something a bit more descriptive here. Here's a straw-man:

> Changed the behavior of wait and signal enqueue commands in `cl_khr_semaphore`; they may now block under some circumstances (see KhronosGroup/OpenCL-Docs#883 for background).

I _think_ this would be the first occurrence of a Github issue link in the spec but IMO that'd be a good habit to get into. It'd make looking at historical discussions a bit easier than having to search through the git history (which admittedly isn't typically that hard).

bashbaug: Interesting idea, I added GitHub PRs for each of the items in the change log.  I didn't actually make them links, though I suppose we could do that too.  Let me know what you think.

Note, I added a bit more description to the semaphore item too, but I didn't say "blocking" because I think it's too easy to misunderstand in isolation.

alycm (approved):
LTGM, happy to have GitHub issues referenced.

bashbaug:
I'm going to merge this as-is.  @kpet, if you have additional suggestions I'll be happy to make them in a subsequent PR.

api/appendix_e.asciidoc
kpet: I think links would be better for people for whom it wouldn't be obvious that KhronosGroup/OpenCL-Docs is the project the number are valid for. Also makes it easier to refer to PRs and Issues. Regarding the semaphore change I was more after something that would communicate the gist of the change in one line (which is hard to do! :)). We're talking polishing here.

kpet:
Fine by me given the time constraints.

api/appendix_e.asciidoc
bashbaug: I added https://github.com/KhronosGroup/OpenCL-Docs/issues/911 so we remember to add the PR and issue links.  Thanks!

PR #914, opened 2023-04-24
Reserve enums used by cl_ext_yuv_images

kpet:
Covers all enums currently used by #722.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -900,6 +900,7 @@ server's OpenCL/api-docs repository.
         <enum bitpos="2"            name="CL_MAP_WRITE_INVALIDATE_REGION"/>
             <unused start="3" end="31"/>
             <unused start="32" end="33" comment="Reserved for cl_ext_yuv_images #722"/>
+            <unused start="34" end="63"/>
     </enums>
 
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -899,6 +899,7 @@ server's OpenCL/api-docs repository.
         <enum bitpos="1"            name="CL_MAP_WRITE"/>
         <enum bitpos="2"            name="CL_MAP_WRITE_INVALIDATE_REGION"/>
             <unused start="3" end="31"/>
+            <unused start="32" end="33" comment="Reserved for cl_ext_yuv_images #722"/>
```
alycm, line 902: If we're going beyond 32-bits should there be an extra line saying

```c
<unused start="34" end="63"/>
```

`cl_map_flags` is `cl_bitfield`, which is `cl_ulong`, so 64-bits are available.

alycm (approved):
1 small question, fine in principle.

xml/cl.xml
kpet: That makes sense. I've added the missing unused range.

Jeremy-Kemp (approved):
LGTM!

kpet:
Merging with two approvals.

Issue #843, opened 2022-09-14
Interaction between mutable-dispatch and USM extensions


EwanC:
The cl_khr_command_buffer_mutable_dispatch extension defines a mechanism for updating SVM arguments to command-buffer kernel commands, however there is no mention of the interaction with [USM](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_unified_shared_memory.html) kernel arguments.

It was initially considered that updating USM arguments could be done through the SVM mechanism - `num_svm_args` / `arg_svm_list`. However, this will only work if either a) a vendor doesn't support SVM, so can unambiguously treat user passed values here as USM, or b) a vendor implements SVM and USM using the same underlying mechanism so the pointers are interchangeable. E.g passing a USM pointer to `clSetKernelArgSVMPointer` happens to work as an implementation detail.

Although these two scenarios cover are applicable to the two known vendors of USM a) Codeplay and b) Intel, it would be valid for another vendor in the future to implement both SVM and USM in a way that the pointers are not interchangeable. Therefore, we should have an explicit mechanism for updating USM args. I can think of two possible approaches: 

1) Modify cl_khr_command_buffer_mutable_dispatch extension to define USM fields similar to `num_svm_args` / `arg_svm_list` but for USM.
   
2) Add an "Interaction with Other Extensions" section to the USM extension to take advantage of the pointer chaining mechanism in cl_khr_command_buffer_mutable_dispatch, e.g:
   ```c
   // Pointer chained from cl_mutable_dispatch_config_khr
   typedef struct {
    cl_command_buffer_structure_type_khr type;
    const void* next;
	cl_uint num_usm_args;
	const cl_mutable_dispatch_arg_khr* arg_usm_list;
   } cl_mutable_dispatch_usm_arg_intel;
   
   // Value for cl_command_buffer_structure_type_khr which would need reserved in the XML
   #define CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_USM_ARG_CONFIG_INTEL 2      
   ```

EwanC:
Since this issue was created a draft of [cl_khr_unified_svm](https://github.com/KhronosGroup/OpenCL-Docs/pull/1282) has been created.

[...]

PR #915, opened 2023-04-26
command_buffer: Add clCommandSVMMemcpy and clCommandSVMMemFill

pjaaskel:
This extends the command buffer extension with two SVM commands.


EwanC (commented):
Have some minor editorial feedback, but don't have a problem adding these optional SVM entry-points. Although we don't support SVM, so can't provide in depth implementation coverage.

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -1516,6 +1523,174 @@ New errors:
...
+the same context from which _command_queue_ was created.
+Otherwise the behavior is undefined.
+
+_pattern_ is a pointer to the data pattern of size _pattern_size_ in bytes.
+_pattern_ will be used to fill a region in _buffer_ starting at _svm_ptr_
+and is _size_ bytes in size.
+The data pattern must be a scalar or vector integer or floating-point data
+type supported by OpenCL as described in <<scalar-data-types, Shared
+Application Scalar Data Types>> and <<vector-data-types, Supported
+Application Vector Data Types>>.
```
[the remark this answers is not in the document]
pjaaskel: I took the lazy way out, and just removed the reference. All in all, it would be nice to remove the duplication of API descriptions here due to the "API copy".

EwanC (approved):

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -1516,6 +1523,172 @@ New errors:
...
+
+_command_queue_ Specifies the command-queue the command will be recorded to.
+Parameter is unused by this extension as only a single command-queue is
+supported and **must** be `NULL`.
+
+_dst_ptr_ is the pointer to a host or SVM memory allocation where data is
+copied to.
+
+_src_ptr_ is the pointer to a host or SVM memory allocation where data is
+copied from.
```
bashbaug, line 1543: I think we should think carefully whether we want to support memcpys to or from a host allocation, at least when the device executing the command buffer does not support system SVM.  Supporting arbitrary host allocations could be expensive for some devices and IIRC this was one of the reasons why we do not allow equivalents of `clEnqueueReadBuffer` or `clEnqueueWriteImage` to be recorded in a command buffer.

Can we restrict this to SVM memory allocations only, unless the device supports system SVM?

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -1516,6 +1523,172 @@ New errors:
...
+Records a command to fill a region in SVM with a pattern of a given pattern size.
+
+_command_buffer_ Refers to a valid command-buffer object.
+
+_command_queue_ Specifies the command-queue the command will be recorded to.
+Parameter is unused by this extension as only a single command-queue is
+supported and **must** be `NULL`.
+
+_svm_ptr_ is a pointer to a memory region that will be filled with
+_pattern_. It must be aligned to _pattern_size_ bytes.
```
bashbaug, line 1616: Similar to my comment for the SVM memcpy, can this be any arbitrary host pointer, or must it be an SVM allocation if the device executing the command buffer does not support system SVM?

bashbaug (commented):
These should be pretty easy additions to the command buffer emulation layer if we decide to do this.

I would be interested in adding support for UVM as well, perhaps as a layered extension.

EwanC:
> I would be interested in adding support for UVM as well, perhaps as a layered extension.

I also thought about USM support when I saw this PR, I think support could be useful in the context of SYCL-Graphs where you can capture SYCL USM memcpy/memfill nodes, and want to propagate that down to commands in the underlying command-buffer representation.

A layered vendor extension for this makes sense to me if there are dependencies on both a vendor & KHR extension. We also have an [existing ticket](https://github.com/KhronosGroup/OpenCL-Docs/issues/843) for the interaction of USM with `cl_khr_command_buffer_mutable_dispatch`, which a new layered extension could solve.

ext/cl_khr_command_buffer.asciidoc
EwanC: That's a good point, supporting host memory was problematic for one of our customers, so we didn't want equivalents. Omitting the entry-points also reduced the scope of the extension so that we could release something publicly earlier.

If I recall, rather than omitting the use of host memory in command-buffers @pjaaskel asked for it to be optional. Maybe that's something we want to revisit before the extension moves from provisional to final. Extension could have a device query to report if you can use host memory in SVM memcpy/memfill and also record `clEnqueueReadBuffer` type commands.

pjaaskel: Good points. Yes, the capabilities of SVM commands that can be buffered should naturally match to the level of SVM supported.

I planned to submit host interaction and the read/write buffer commands as separate layered extensions on top, but naturally making them optional parts of the extension would be simpler. We try to fulfil the needs of CUDA/HIP graph with these enhancements. It's ongoing work.

pjaaskel:
Thanks! I added the mention of System SVM as suggested by Ben.

bashbaug (approved):

aharon-abramson (approved):

franz:
related OpenCL-CTS tests are [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/1821)

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -378,6 +380,11 @@ cl_int clGetCommandBufferInfoKHR(
     size_t* param_value_size_ret);
 ----
 
+The following SVM entry points are included only from version 2.0, adding selected commands
+from Section 5.6:
+
+include::{generated}/api/protos/clCommandSVMMemcpyKHR.txt[]
+include::{generated}/api/protos/clCommandSVMMemFillKHR.txt[]
```
bashbaug, line 387: I'm getting an asciidoctor warning building the specs that I think is caused by these two lines:

```
asciidoctor: WARNING: generated/api/protos/clCommandSVMMemcpyKHR.txt: line 4: id assigned to block already in use: clCommandSVMMemcpyKHR
asciidoctor: WARNING: generated/api/protos/clCommandSVMMemFillKHR.txt: line 4: id assigned to block already in use: clCommandSVMMemFillKHR
```

Basically, these generated snippets include a "label" for the API entry points to this means that they can only be included once.  If we want to support including them multiple times - here, for example, and in the description of each API - then we'll need to do something to the spec tooling to allow this.

For now I would recommend just describing each API "manually", as is done above.

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -378,6 +380,11 @@ cl_int clGetCommandBufferInfoKHR(
     size_t* param_value_size_ret);
 ----
 
+The following SVM entry points are included only from version 2.0, adding selected commands
```
bashbaug, line 383: Is "from version 2.0" the right description?  Or, should this be "from version 0.9.4"?

ext/cl_khr_command_buffer.asciidoc
```diff
@@ -1539,6 +1546,173 @@ New errors:
...
+complete. If the _sync_point_wait_list_ and the _sync_point_ arguments are not
+`NULL`, the _sync_point_ argument should not refer to an element of the
+_sync_point_wait_list_ array.
+
+_mutable_handle_ Returns a handle to the command. Handle is unused by
+this extension and must be passed as `NULL`.
+
+{clCommandSVMMemcpyKHR} returns {CL_SUCCESS} if the function is executed
+successfully. Otherwise, it returns the errors defined by
+{clEnqueueSVMMemcpy} except:
```
bashbaug, line 1597: Do we need to say anything about an OpenCL 3.0 device that does not support SVM here, or is this one of the errors that was inherited by `clEnqueueSVMMemcpy`?

For reference, I believe the relevant error is:

> [CL_INVALID_OPERATION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_OPERATION) if the device associated with command queue does not support SVM.

Same comment also applies to the new SVM fill API.

pjaaskel: I believe we don't need to explicitly state it, but it should be inherited from the 3.0 spec error description.

ext/cl_khr_command_buffer.asciidoc
pjaaskel: Fixed.

ext/cl_khr_command_buffer.asciidoc
pjaaskel: Clarified. It now mentions both the min OpenCL version as well as the minimum ext version.

PR #922, opened 2023-05-04
cl_khr_external_memory: add error when CL_DEVICE_HANDLE_LIST_KHR is specified without an external memory handle

### LINKED ISSUE #916: Behaviour of `clCreate{Buffer,Image}WithProperties` when `CL_DEVICE_HANDLE_LIST_KHR` is provided but no external memory handle is
`CL_DEVICE_HANDLE_LIST_KHR` can be passed to `clCreateBufferWithProperties` or `clCreateImageWithProperties` to restrict which devices in the context import the external memory described by an external memory handle. The specification does not however state what happens when `CL_DEVICE_HANDLE_LIST_KHR` is provided (especially if the list does not contain all the devices that are part of the context) but no external memory handle is provided. A few possible options:
1. Ignore the device list.
2. Restrict which devices can use the memory object even though it wouldn't be imported (feels weird to introduce this with `cl_khr_external_memory`).
3. Report an error. We could add a rule stating that an external memory handle must be provided when `CL_DEVICE_HANDLE_LIST_KHR` is and require implementations to report `CL_INVALID_PROPERTY` otherwise.
4. Report an error when the device list does not contain all the devices that are in the context.

I think I have a preference for option 3. Thoughts?

kpet:
Fixes #916

### DIFF
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -30,6 +30,7 @@ Other related extensions define specific external memory types that may be impor
 [cols="1,1,3",options="header",]
 |====
 | *Date*     | *Version* | *Description*
+| 2023-05-04 | 0.9.1     | {CL_DEVICE_HANDLE_LIST_KHR} cannot be specified without an external memory handle (provisional).
 | 2021-09-10 | 0.9.0     | Initial version (provisional).
 |====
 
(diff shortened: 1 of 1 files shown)


bashbaug (approved):

bashbaug:
Merging as discussed on the May 9th teleconference.

PR #923, opened 2023-05-09
Clarify the definition of prerequisites to commands

kpet:
- Reword the first source of prerequisites so the wording is symmetrical with respect to the others (i.e. the first/second/third ...).
- Broaden the first source of prerequisites to cover all implicit dependencies and provide an exhaustive list of how they arise: either because of barriers or because of ordering in in-order command-queues.

Change-Id: Ic464066261fe13756347bafb4878cd6ffb5a8427

### DIFF
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -198,10 +198,12 @@ A command submitted to a device will not launch until prerequisites that
 constrain the order of commands have been resolved.
 These prerequisites have three sources:
 
-  * They may arise from commands submitted to a command-queue that constrain
-    the order in which commands are launched.
-    For example, commands that follow a command-queue barrier will not
-    launch until all commands prior to the barrier are complete.
+  * The first source of prerequisites is implicit dependencies between commands
+    enqueued to the same command-queue which arise as follows:
+  ** Commands enqueued after a command-queue barrier have this barrier command
+     a prerequisite.
+  ** Commands enqueued in an in-order command-queue have the command enqueued
+     before them as a prerequisite.
   * The second source of prerequisites is dependencies between commands
     expressed through events.
     A command may include an optional list of events.
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -200,8 +200,8 @@ These prerequisites have three sources:
 
   * The first source of prerequisites is implicit dependencies between commands
     enqueued to the same command-queue which arise as follows:
-  ** Commands enqueued after a command-queue barrier have this barrier command
-     a prerequisite.
+  ** Commands enqueued after a command-queue barrier have the preceding barrier command
+      as a prerequisite.
   ** Commands enqueued in an in-order command-queue have the command enqueued
      before them as a prerequisite.
   * The second source of prerequisites is dependencies between commands


api/opencl_architecture.asciidoc
```diff
@@ -199,10 +199,12 @@ A command submitted to a device will not launch until prerequisites that
...
 These prerequisites have three sources:
 
-  * They may arise from commands submitted to a command-queue that constrain
-    the order in which commands are launched.
-    For example, commands that follow a command-queue barrier will not
-    launch until all commands prior to the barrier are complete.
+  * The first source of prerequisites is implicit dependencies between commands
+    enqueued to the same command-queue which arise as follows:
+  ** Commands enqueued after a command-queue barrier have all commands enqueued
+     before the barrier as prerequisites.
```
bashbaug, line 205: This is one of the more confusing parts of the OpenCL spec, and this isn't common, but strictly speaking I don't think this is true: if a barrier is in an out-of-order queue, and if the barrier has an explicit event wait list, then the barrier will only wait for the commands described by the event wait list commands to complete, and subsequent commands will only wait for the barrier to complete.  This means that other commands enqueued before the barrier, but that the barrier did not wait on, may still be executing.

The easiest way to describe this may be to describe dependency on the barrier itself, not on the commands before the barrier, something like:

> * Commands enqueued after a command-queue barrier have an implicit dependency on the command-queue barrier as a prerequesite.

Or, slightly more verbose:

> * Commands enqueued after a command-queue barrier have an implicit dependency on the completion of the preceding command-queue barrier as a prerequesite.

api/opencl_architecture.asciidoc
```diff
@@ -199,10 +199,12 @@ A command submitted to a device will not launch until prerequisites that
...
+  ** Commands enqueued in an in-order command-queue have all commands enqueued before
+     them as prerequisites.
```
bashbaug, line 207: I'm fine with this as-is and I think this is correct, but I've always found in-order queues easier to think of using ideas from  mathematical induction, and simpliy thinking about an implicit dependency on the previous command.  Logically this means that each command has an implicit dependency on all previous commands also, but if you draw this as a picture (really, a DAG) then it's the difference between one dependence arrow between each command (node in the graph) vs. N dependence arrows between commands (one connecting to each previous graph node).

I'd at least consider rephrasing to something like:

> * Commands in an in-order queue have an implicit dependency on the completion of the preceding command as a prerequisite.

api/opencl_architecture.asciidoc
kpet: I agree with your description of what the behaviour should be. This was meant to cover barriers **with no explicit prerequisites**, good catch. I've updated the description.

I don't think your proposed "dependency on the barrier" wording (paraphrasing) works well. We'd still need to describe what a barrier depends on and that is only done in the description for `clEnqueueBarrierWithWaitList` and not in terms of prerequisites.

api/opencl_architecture.asciidoc
kpet: I like the idea. This could come in handy for an extension I have on the back burner. I've updated the description but used a different wording.

api/opencl_architecture.asciidoc
bashbaug: I'm fine with the updated text and this is a definite improvement over what is currently in the spec, but FWIW I do quite like the "dependency on the barrier" wording.  Consider a case like the following (pseudocode):

```c
clEnqueueNDRangeKernel(A);
clEnqueueNDRangeKernel(B);
clEnqueueNDRangeKernel(C);
clEnqueueBarrier(wait on {A, B});
clEnqueueNDRangeKernel(D);
```

Do we agree:

1. Kernels A, B, and C have no implicit or explicit prerequisites, so they may execute anytime and in any order, and may overlap execution.
2. The barrier will not complete until kernels A and B are complete because of the explicit dependencies specified by the event wait list.
3. Kernel D has the barrier as an implicit prerequisite, hence a transitive dependency on kernels A and B, even though it has no explicit prerequisites.  There is no dependency between kernels C and D though, so C and D may overlap execution.

If so, I think (2) is covered by "the second source of prerequisites", but I'm not sure how (3) would be covered by the text in this PR.

kpet: 1. Yes.
2. Yes.
3. Yes.

I've returned to this with a fresher mind and definitely agree your suggestion of describing implicit prerequisites in terms on dependencies on barriers is better. I've tweaked the PR to reflect that which I believe covers your case 3.
 PTAL.

api/opencl_architecture.asciidoc
```diff
@@ -198,10 +198,12 @@ A command submitted to a device will not launch until prerequisites that
...
 These prerequisites have three sources:
 
-  * They may arise from commands submitted to a command-queue that constrain
-    the order in which commands are launched.
-    For example, commands that follow a command-queue barrier will not
-    launch until all commands prior to the barrier are complete.
+  * The first source of prerequisites is implicit dependencies between commands
+    enqueued to the same command-queue which arise as follows:
+  ** Commands enqueued after a command-queue barrier have this barrier command
+     a prerequisite.
```
EwanC, line 204: To clarify, i'm reading this as *all* barrier commands in a command-queue are a prerequisite of the new command.
i.e. if you had the following
A. clEnqueueBarrier()
B. clEnqueueBarrier()
C. clEnqueueNDRangeKernel()

Then the dependency graph would be:
```
A 
| \
B  |
| /
C
```

Rather than if a prerequisite is just the last barrier added to the queue

```
A
|
B
|
C
```

There's no difference in the command scheduling, so this is nitpicking a bit.

EwanC (approved):

api/opencl_architecture.asciidoc
bashbaug: I agree, though I don't think the difference is observable.  In the first case C has an explicit dependency on A and B, and in the second case there is an explicit dependency on B and an implicit (transitive) dependency on B.  Since there's no way to query the dependencies for a command, I don't think it really matters which interpretation gets used.

FWIW, I think the mental model is simpler when it just considers the last barrier before the command (so, the second interpretation).  If we want to describe it to remove ambiguity and to nudge towards this mental model I think we could adopt the "slightly more verbose" version I suggested in a comment above, but I don't think this is strictly necessary.

bashbaug (approved):

api/opencl_architecture.asciidoc
kpet, line 204:
```suggestion
  ** Commands enqueued after a command-queue barrier have the preceding barrier command
      as a prerequisite.
```

kpet: It shouldn't matter but better not to use something ambiguous. I've changed the wording to say "the preceding barrier command" instead of "this barrier command" and added the missing "as".

Issue #924, opened 2023-05-15
CL_DEVICE_COMMAND_BUFFER_NUM_SYNC_DEVICES_KHR not consistent with CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR


aharon-abramson:
Regarding [this](https://github.com/KhronosGroup/OpenCL-Docs/blob/41fefb4b58cf2f046ca0f15a1853505823fbdb36/ext/cl_khr_command_buffer_multi_device.asciidoc?plain=1#L182), the description of CL_DEVICE_COMMAND_BUFFER_NUM_SYNC_DEVICES_KHR isn't consistent with CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR: whereas the latter includes sub-devices, the former doesn't.
I believe this isn't intentional.

EwanC:
Good spot, created https://github.com/KhronosGroup/OpenCL-Docs/pull/925 the rectify this.

EwanC:
Closing this issue as done now that https://github.com/KhronosGroup/OpenCL-Docs/pull/925 is merged

PR #925, opened 2023-05-16
Consistent sub-device/root-device definition of CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR

EwanC:
Currently, `CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR` mentions sub-devices in the reported list for the query, but `CL_DEVICE_COMMAND_BUFFER_NUM_SYNC_DEVICES_KHR` does not. 

Actions issue https://github.com/KhronosGroup/OpenCL-Docs/issues/924 to make these consistent.

### DIFF
--- a/ext/cl_khr_command_buffer_multi_device.asciidoc
+++ b/ext/cl_khr_command_buffer_multi_device.asciidoc
@@ -179,15 +179,17 @@ of section 4.2. Also, add additional text to the
 
 | {CL_DEVICE_COMMAND_BUFFER_NUM_SYNC_DEVICES_KHR}
 | {cl_uint_TYPE}
-| Return the number of devices and sub-devices listed in
+| Return the number of root devices listed in
   {CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR} that _device_ can use device-side
   synchronization with.
 
 | {CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR}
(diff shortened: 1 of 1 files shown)


aharon-abramson (approved):

bashbaug:
I actually wonder if the right thing to do here might be to remove any notion of sub-devices from these queries, and to only return root devices instead.  Here's why:  Since these are `clGetDeviceInfo` queries there is no "context" that gets passed, so it's not clear what set of sub-devices should be considered for this query if sub-devices are included.  I guess this could be the set of "all sub-devices that have been created so far" or similar, but this is a little strange and I'm not aware of any other queries that behave similarly.

Should sub-devices be considered, or is it OK if these queries only return root devices instead?

aharon-abramson:
I agree with Ben: I think root devices should be enough.

EwanC:
Good point, you could indeed call `clGetDeviceInfo` twice with the same parameters and get different results if you created a sub-device in-between those calls, which may be surprising, and is at least inconsistent.

Will update this PR to just be root devices, the user can always call the device info query again passing a sub-device.

EwanC:
> Will update this PR to just be root devices, the user can always call the device info query again passing a sub-device.

Pushed an update. I also had to update some error wording, otherwise a user could create a command-buffer with sub-devices and it'd be an error since only root devices are returned by the query.

bashbaug (approved):

Issue #917, opened 2023-05-02
[XML] Version condition mismatch on `CL_DEPTH`


SunSerega:
```
            <require condition="!defined(CL_VERSION_1_2)" comment="cl_channel_order - defined in CL.h for OpenCL 1.2 (?) and newer">
                <enum name="CL_DEPTH"/>
            </require>
```
But above, `CL_DEPTH` is added as part of the:
```
    <feature api="opencl" name="CL_VERSION_2_0" number="2.0" comment="OpenCL core API interface definitions">
```
Is it supposed to be so?

bashbaug:
Note, there is some additional background here: https://github.com/KhronosGroup/OpenCL-Docs/issues/284 and here: https://github.com/KhronosGroup/OpenCL-Docs/issues/284.

Also here: https://github.com/KhronosGroup/OpenCL-Headers/issues/117

I think this problem really stems from the OpenCL headers and specifically cl.h, where we have:

```c
#ifdef CL_VERSION_1_2
#define CL_DEPTH                                    0x10BD
#define CL_DEPTH_STENCIL                            0x10BE
#endif
```

IMHO there are two bugs here:
1. `CL_DEPTH` is in the core specification but it wasn't until added until OpenCL 2.0, so it should be guarded by `CL_VERSION_2_0` and not `CL_VERSION_1_2`.
2. `CL_DEPTH_STENCIL` was never in the core specification (it's part of `cl_khr_gl_depth_images`), so it shouldn't be in cl.h at all.

We could fix this but there's a small chance it could break shipping code.  For example, if an application was building for OpenCL 1.2, was including cl.h but not cl_ext.h (for `CL_DEPTH`) or cl_gl.h (for `CL_DEPTH_STENCIL`), and was using `CL_DEPTH` or `CL_DEPTH_STENCIL`, code could stop building.

Admittedly the chances of this happening seems small.  Should we just fix this?  Is the current behavior causing a problem or is it just confusing?

SunSerega:
> Note, there is some additional background here

You pasted the same issue link twice... I guess you meant associated PR?

> Is the current behavior causing a problem or is it just confusing?

It didn't cause any problems for me personally.
Found while looking for other info - thought should probably report it.

bashbaug:
> You pasted the same issue link twice... I guess you meant associated PR?

Oops, thanks!  Fixed.

> Found while looking for other info - thought should probably report it.

Thank you, much appreciated.

[...]

PR #926, opened 2023-05-18
Сonsistent enum types

SunSerega:
### History

I first tried to rectify the idea of #779.

But there are enum types, like `cl_mem_flags` and `cl_svm_mem_flags`, which have their value sets intersecting. I don't know how to describe that using `<enums name="..."`, because `<enum>` tags are unique and sorted by value.

This is, however, already described in `<require>` tags:
```
        <require comment="cl_mem_flags and cl_svm_mem_flags - bitfield">
            <enum name="CL_MEM_READ_WRITE"/>
            <enum name="CL_MEM_WRITE_ONLY"/>
            <enum name="CL_MEM_READ_ONLY"/>
            <enum name="CL_MEM_USE_HOST_PTR"/>
            <enum name="CL_MEM_ALLOC_HOST_PTR"/>
            <enum name="CL_MEM_COPY_HOST_PTR"/>
        </require>
```
And after I've checked all of it, I can say - `comment` attributes actually more consistent. Even despite being a horrible abuse and a few cases like:
```
            <require condition="!defined(CL_VERSION_1_2)" comment="cl_device_info - defined in CL.h for OpenCL 1.2 and newer">
```
Making it hard to parse out the enum type name...
(this is what caused me to find #917)

---

### This PR

So I've thrown away most changes #779 (my version of it) did to to `<enums>` tags for this PR and focused on replacing the `comment` attribute with something consistent:
```
        <require group="cl_device_type" etype="bitfield">
```
And also on using the defined enum types where they should be (like structs).

The `etype` is supposed to be `enum` or `bitfield`, where `enum` can be omitted, but can also be specified explicitly.
(I'll reflect this in `registry.rnc`, but would like to hear some feedback first)
I didn't find any cases where explicit `etype="enum"` would be useful, but I have some ideas concerning other changes I'd like to make further PRs for:

---

### Other changes

Most notable of them is info from state tables, they are [here](https://github.com/SunSerega/OpenCL-Docs/tree/changes/clear) ([current commit](https://github.com/SunSerega/OpenCL-Docs/commit/d4d1ac4f8c57aa709d9d4d511d36d21c3c674292)) ([compare](https://github.com/KhronosGroup/OpenCL-Docs/compare/main...SunSerega:OpenCL-Docs:changes/clear)).
But only after translating this from my internal binding fixers to `etype="obj info"` and `etype="property list"` did I think that there are much better ways to express this in XML. Well, I'd like to hear what others think before Iimit ideas to what I'm thinking of replacing it with.
Regardless, that branch is currently a mess, only good as proof of concept, because I'm currently using it for my bindings.

Other changes that probably should be in separate PR are the rename `clCommandExecutionStatus`=>`cl_command_execution_status` (for consistency of naming convention) and the addition of `cl_error_code` as a proper enum type, instead of `Error codes` comments.
But it felt awkward to separate because that would leave some stray `comment` attribute abuse cases in this PR.

Also, if anyone thinks there is value in how I merged main into #779 - it can be found [here](https://github.com/SunSerega/OpenCL-Docs/tree/changes/original) ([current commit](https://github.com/SunSerega/OpenCL-Docs/commit/3c65fe3b790a4bc747dec416c1fa501f38f73671)) ([compare](https://github.com/KhronosGroup/OpenCL-Docs/compare/main...SunSerega:OpenCL-Docs:changes/original)), in an even bigger mess of changes.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -139,6 +139,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_uint</type>          <name>cl_queue_priority_khr</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_queue_throttle_khr</name>;</type>
         <type category="define">typedef <type>intptr_t</type>         <name>cl_import_properties_arm</name>;</type>
+        <type category="define">typedef <type>intptr_t</type>         <name>cl_import_type_arm</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_svm_mem_flags_arm</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_kernel_exec_info_arm</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_svm_capabilities_arm</name>;</type>
@@ -171,6 +172,8 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>intptr_t</type>         <name>cl_device_partition_property</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_affinity_domain</name>;</type>
         <type category="define">typedef <type>intptr_t</type>         <name>cl_context_properties</name>;</type>
+        <type category="define">typedef <type>intptr_t</type>         <name>cl_context_property_diagnostics_level</name>;</type>
+        <type category="define">typedef <type>intptr_t</type>         <name>cl_context_memory_initialize_khr</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_context_info</name>;</type>
         <type category="define">typedef <type>cl_properties</type>    <name>cl_queue_properties</name>;</type>
         <type category="define">typedef <type>cl_properties</type>    <name>cl_queue_properties_khr</name>;</type>
@@ -251,21 +254,33 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_fp_atomic_capabilities_ext</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_image_requirements_info_ext</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_platform_command_buffer_capabilities_khr</name>;</type>
-
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_affinity_domain_ext</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_command_termination_reason_arm</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_host_cache_policy</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_allocation_type</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_mb_block_type</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_subpixel_mode</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_sad_adjust_mode</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_search_path_type</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_device_me_version</name>;</type>
+        <type category="define">typedef <type>cl_uint</type>          <name>cl_device_avc_me_version</name>;</type>
+        <type category="define">typedef <type>cl_int</type>           <name>cl_command_execution_status</name>;</type>
+        <type category="define">typedef <type>cl_int</type>           <name>cl_error_code</name>;</type>
+        
             <comment>Structure types</comment>
         <type category="struct" name="cl_dx9_surface_info_khr">
             <member><type>IDirect3DSurface9</type>*  <name>resource</name></member>
             <member><type>HANDLE</type>              <name>shared_handle</name></member>
         </type>
         <type category="struct" name="cl_motion_estimation_desc_intel">
-            <member><type>cl_uint</type>             <name>mb_block_type</name></member>
-            <member><type>cl_uint</type>             <name>subpixel_mode</name></member>
-            <member><type>cl_uint</type>             <name>sad_adjust_mode</name></member>
-            <member><type>cl_uint</type>             <name>search_path_type</name></member>
+            <member><type>cl_mb_block_type</type>    <name>mb_block_type</name></member>
+            <member><type>cl_subpixel_mode</type>    <name>subpixel_mode</name></member>
+            <member><type>cl_sad_adjust_mode</type>  <name>sad_adjust_mode</name></member>
+            <member><type>cl_search_path_type</type> <name>search_path_type</name></member>
         </type>
         <type category="struct" name="cl_mem_ext_host_ptr">
-            <member><type>cl_uint</type>             <name>allocation_type</name></member>
-            <member><type>cl_uint</type>             <name>host_cache_policy</name></member>
+            <member><type>cl_allocation_type</type>  <name>allocation_type</name></member>
+            <member><type>cl_host_cache_policy</type><name>host_cache_policy</name></member>
         </type>
         <type category="struct" name="cl_mem_ion_host_ptr">
             <member><type>cl_mem_ext_host_ptr</type> <name>ext_host_ptr</name></member>
@@ -728,7 +743,7 @@ server's OpenCL/api-docs repository.
             <unused start="-4" end="-9999"/>
     </enums>
 
-    <enums name="clCommandExecutionStatus" vendor="Khronos">
+    <enums name="cl_command_execution_status" vendor="Khronos">
         <enum value="0x0"           name="CL_COMPLETE"/>
         <enum value="0x1"           name="CL_RUNNING"/>
         <enum value="0x2"           name="CL_SUBMITTED"/>
@@ -957,7 +972,7 @@ server's OpenCL/api-docs repository.
             <unused start="0x8" end="0x80000000"/>
     </enums>
 
-    <enums name="cl_arm_device_svm_capabilities.flags" vendor="ARM" type="bitmask">
+    <enums name="cl_device_svm_capabilities_arm" vendor="ARM" type="bitmask">
         <enum bitpos="0"            name="CL_DEVICE_SVM_COARSE_GRAIN_BUFFER_ARM"/>
         <enum bitpos="1"            name="CL_DEVICE_SVM_FINE_GRAIN_BUFFER_ARM"/>
         <enum bitpos="2"            name="CL_DEVICE_SVM_FINE_GRAIN_SYSTEM_ARM"/>
@@ -971,7 +986,7 @@ server's OpenCL/api-docs repository.
             <unused start="0x4" end="0x80000000"/>
     </enums>
 
-    <enums name="cl_intel_driver_diagnostics" vendor="Intel" type="bitmask">
+    <enums name="cl_context_property_diagnostics_level" comment="From cl_intel_driver_diagnostics" vendor="Intel" type="bitmask">
         <enum value="0xff"          name="CL_CONTEXT_DIAGNOSTICS_LEVEL_ALL_INTEL"/>
         <enum bitpos="0"            name="CL_CONTEXT_DIAGNOSTICS_LEVEL_GOOD_INTEL"/>
         <enum bitpos="1"            name="CL_CONTEXT_DIAGNOSTICS_LEVEL_BAD_INTEL"/>
@@ -1302,7 +1317,7 @@ server's OpenCL/api-docs repository.
             <unused start="3" end="31"/>
     </enums>
 
-    <enums name="cl_device_command_buffer_capabilities_khr " vendor="Khronos" type="bitmask">
+    <enums name="cl_device_command_buffer_capabilities_khr" vendor="Khronos" type="bitmask">
         <enum bitpos="0"            name="CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR"/>
         <enum bitpos="1"            name="CL_COMMAND_BUFFER_CAPABILITY_DEVICE_SIDE_ENQUEUE_KHR"/>
         <enum bitpos="2"            name="CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR"/>
@@ -2283,7 +2298,7 @@ server's OpenCL/api-docs repository.
     <!-- SECTION: CL command definitions. (TBD) -->
     <commands>
         <command suffix="CL_API_SUFFIX__VERSION_1_0">
-            <proto><type>cl_int</type>                     <name>clGetDeviceIDsFromD3D10KHR</name></proto>
+            <proto><type>cl_error_code</type>              <name>clGetDeviceIDsFromD3D10KHR</name></proto>
             <param><type>cl_platform_id</type>             <name>platform</name></param>
             <param><type>cl_d3d10_device_source_khr</type> <name>d3d_device_source</name></param>
             <param><type>void</type>*                      <name>d3d_object</name></param>
@@ -2297,7 +2312,7 @@ server's OpenCL/api-docs repository.
             <param><type>cl_context</type>                 <name>context</name></param>
             <param><type>cl_mem_flags</type>               <name>flags</name></param>
             <param><type>ID3D10Buffer</type>*              <name>resource</name></param>
-            <param><type>cl_int</type>*                    <name>errcode_ret</name></param>
+            <param><type>cl_error_code</type>*             <name>errcode_ret</name></param>
         </command>
         <command suffix="CL_API_SUFFIX__VERSION_1_0">
             <proto><type>cl_mem</type>                     <name>clCreateFromD3D10Texture2DKHR</name></proto>
@@ -2305,7 +2320,7 @@ server's OpenCL/api-docs repository.
             <param><type>cl_mem_flags</type>               <name>flags</name></param>
             <param><type>ID3D10Texture2D</type>*           <name>resource</name></param>
             <param><type>UINT</type>                       <name>subresource</name></param>
-            <param><type>cl_int</type>*                    <name>errcode_ret</name></param>
+            <param><type>cl_error_code</type>*             <name>errcode_ret</name></param>
         </command>
         <command suffix="CL_API_SUFFIX__VERSION_1_0">
             <proto><type>cl_mem</type>                     <name>clCreateFromD3D10Texture3DKHR</name></proto>
@@ -2313,10 +2328,10 @@ server's OpenCL/api-docs repository.
             <param><type>cl_mem_flags</type>               <name>flags</name></param>
             <param><type>ID3D10Texture3D</type>*           <name>resource</name></param>
             <param><type>UINT</type>                       <name>subresource</name></param>
-            <param><type>cl_int</type>*                    <name>errcode_ret</name></param>
+            <param><type>cl_error_code</type>*             <name>errcode_ret</name></param>
         </command>
         <command suffix="CL_API_SUFFIX__VERSION_1_0">
-            <proto><type>cl_int</type>                     <name>clEnqueueAcquireD3D10ObjectsKHR</name></proto>
+            <proto><type>cl_error_code</type>              <name>clEnqueueAcquireD3D10ObjectsKHR</name></proto>
             <param><type>cl_command_queue</type>           <name>command_queue</name></param>
             <param><type>cl_uint</type>                    <name>num_objects</name></param>
             <param>const <type>cl_mem</type>*              <name>mem_objects</name></param>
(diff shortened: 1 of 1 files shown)


SunSerega:
Ah, also forgot to mention `etype="constants"` and `etype="OpenCL-C only"`.

- `constants` are intrinsic values, that are not actually grouped (and don't have their own enum type)

- Meanwhile `OpenCL-C only` is a hotfix, because I don't know what to do with them:
Those enums are not actually part of OpenCL API. They only exist in kernels, but 2 intel motion estimation extensions added them to `cl.xml`...
Probably don't want to lose this info, but also it doesn't seem consistent with other extensions.

bashbaug:
Can you say a bit more how you expect these changes to be used?  As-written they will require some updates to some of the tooling based on the XML file (the generated headers come to mind), but I don't mind making these updates if this will be useful.

Changing from `cl_int` to `cl_error_code` is going to be a pretty big change and is likely to be the most controversial, so if you can do without that one (or at least move it to a separate PR) that would probably help to get these changes in faster.  Most of the others look to be XML cleanup that might affect tooling but not the actual API specification, though I haven't gone through everything in detail.

As an aside, it's unfortunate that #779 never got across the finish line; I think it was close and had a lot of nice improvements.

Thanks!

SunSerega:
> Can you say a bit more how you expect these changes to be used?

In high-level languages, it is a good practice to only let access enums through their type name. But for that, a consistent mapping (enum=>type) is needed.

Before, I primarily used info from `<enums>` blocks.
But `<enums name="cl_device_info">` contains many enums from other types, so I then used info from `<require>` blocks to change mappings.
That wasn't a very good approach, because I only checked if the enum type name is one of the words in the `comment` attribute.
And so it still left many broken enum types, which I fixed using my own procedures.
For instance, the `<require>` block for `cl_svm_mem_flags` stole some enums from `cl_mem_flags`.
So I had to manually dup those enums, making sure they would be both in `cl_mem_flags` and `cl_svm_mem_flags`. Here are the instructions I made for my fixers:
https://github.com/SunSerega/POCGL/blob/27a1b7eb288d1822a8b95c0d301fc173b157139a/Packing/Template/OpenCL/Fixers/Enums/MissingDuplicates.dat

I'm currently looking at what info I can share by moving it to common XML, so others can use it without jumping through the same number of hoops.
And as a bonus, there are many chances for me to notice what I didn't fix, by using git diff everywhere I can, forcing myself to more consciously think about what changed between the old ways and new XML info.

> Changing from cl_int to cl_error_code is going to be a pretty big change and is likely to be the most controversial, so if you can do without that one (or at least move it to a separate PR) that would probably help to get these changes in faster

I mentioned that briefly in the PR description.

But I also think the `cl_int`=>`cl_error_code` change is not as scary as it first looks.
On the side of C/C++, this is nothing but a visual change because of how `typedef` works.

And on the side of other bindings, replacing `cl_error_code` back with `cl_int` should be trivial. Or at least much less painful than replacing `cl_int` with `cl_error_code` - because of functions like `clCreateProgramWithBinary` with multiple error-returning parameters as well as functions like `clSVMAlloc` that do not return errors.

Then again, even if changing back is not trivial, the wonderful thing about using git here is that people can just stick to the previous commit, not pulling more changes until they can process them.
From my experience, even XML files of OpenGL, which is barely updated nowadays, tend to once in a while break my binding code-generators in the most unexpected ways. And it's only worse if the code consuming the XML just works with new changes because it likely means something wrong has happened and wasn't caught by any sanity checks.
So IMO being able to use older commits until ready and to manually check what changed and how it affects things - is already a must when using these XML files.
While on the good side - making change more in the face would result in more attention from XML consumers, so they would also think about how they can make use of the rest of the changes.

That said, separating changes to a separate PR is easy if you think it would still be better to do this in multiple steps.
I can also specifically move changes to `<commands>` to a separate PR while leaving `<require group="cl_error_code">` in this PR.

> Most of the others look to be XML cleanup that might affect tooling but not the actual API specification

Yep, I checked everything with the core and extension specs. Where it didn't make sense - I still added it to XML as it is in spec and made a note for myself to discuss it later by making issues like #932.

SunSerega:
I've split off changes to the `<commands>` tag, among other reasons - it would allow me more concurrency in submitting PRs. Thought I would've liked a bit more of a discussion here...

Also added the new attributes to the schema, so this is now ready for review.

SunSerega:
Also, I noticed that [here](https://github.com/KhronosGroup/OpenCL-Docs/pull/931/files#diff-5f7f7e7042413bd5c24a2d20c8d2d299c93f0dce6807cc356cb4f9a38c6becccR48) the enum group type was marked in a special way in the .asciidoc file when added to XML.
Should I also do that for this PR?

SunSerega:
@bashbaug can you please look at this PR again?

In particular, the test fails now, but it seems to have nothing to do with my changes. Maybe some of the recently merged PRs conflict with each other?

bashbaug:
> In particular, the test fails now, but it seems to have nothing to do with my changes. Maybe some of the recently merged PRs conflict with each other?

I think this is a merge-like issue related to #956.  Can you try updating `CL_DEVICE_HANDLE_LIST_KHR` on line 7000?  I think this will get past the build error:

```diff
diff --git a/xml/cl.xml b/xml/cl.xml
index 3bd3a24..8991c2b 100644
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -6997,7 +6997,7 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_SEMAPHORE_PROPERTIES_KHR"/>
                 <enum name="CL_SEMAPHORE_PAYLOAD_KHR"/>
                 <enum name="CL_SEMAPHORE_TYPE_KHR"/>
-                <enum name="CL_DEVICE_HANDLE_LIST_KHR"/>
+                <enum name="CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR"/>
             </require>
             <require group="cl_semaphore_properties_khr">
                 <enum name="CL_SEMAPHORE_TYPE_KHR"/>
```

SunSerega:
Ah, you're right. I fixed this when merging into my custom branch (all my PRs and some more) and was sure I propagated it back, but apparently, I didn't.

SunSerega:
Does it maybe make sense to create an issue like how you did for #798 to facilitate discussion?
I have other improvements I'd like to propose, which are blocked by this.

And the conflicts with my `custom` branch (containing everything I'd like to make PRs for, eventually), while manageable - appear very often.
Before trying to share my metadata with other XML consumers through this repo, I was just storing it in local files of my bindings' repo, and so I didn't have any conflicts. Only warnings when something not covered by my metadata appeared.
I moved to the `custom` branch to stress-test my changes on myself and ensure I didn't miss anything, but I didn't expect to be solving constant conflicts for almost a year...

What can I do to help this move forward?

bashbaug:
> Does it maybe make sense to create an issue like how you did for #798 to facilitate discussion?

Yes, I think this would be helpful, +1.

As for moving things forward, can this be broken into several smaller PRs rather than one big one?  I'd imagine there are some changes that are less contentious that we could merge in the very near future, then we can focus on the changes that are more complicated.

SunSerega:
Alright: #1145

> As for moving things forward, can this be broken into several smaller PRs rather than one big one?

I already separated, what you asked me to, [here](https://github.com/KhronosGroup/OpenCL-Docs/pull/926#issuecomment-1615851108).
I think the changes are pretty indivisible now.
But I'd separate it further if you have any suggestions.

SunSerega:
Do you have any updates on this?

I've implemented all the requests/suggestions. Please let me know if you want me to change the issue or somehow split this PR even further.

And as I understand, to move forward, we need to mention people who might want to see this, right?

bashbaug:
Hello, here are a few suggestions to break this PR up further.  I don't think you necessarily need to do all of these, but I think a few of the changes are likely to be more contentious than others, and for those it'd be best to put them in a separate PR:

1. Addition of the `cl_error_code` type.  Is this even needed right now?  It doesn't seem to be used.  Regardless, I think this is likely to be the most contentious.  Note, some form of an error code "group" is probably fine, since this mostly seems like an XML convention.
2. Addition of new named types in the core specification, e.g. `cl_command_execution_status` (are there others?).  I'm hesitant to add these to the XML file and the headers, certainly not without corrosponding spec updates.
3. Addition of new named types for extensions, e.g. `cl_import_type_arm`.  I'm hesitant to add these to the XML file and the headers without extension spec updates, also.  Note, assuming we decide to do this, these new types will also need proper suffixes, e.g. `cl_context_property_diagnostics_level_intel` rather than `cl_context_property_diagnostics_level`.
4. Updates to the XML schema for "etype" and "group".  These generally seem reasonable and are not likely to be contentious.
    * Aside: If the "OpenCL-C only" enums really are specific to OpenCL C then I think the right course of action is to remove them from the XML file and the headers, since we don't currently include OpenCL C information in the XML file.
5. Any other XML fixes and corrections - should be non-contentious.

Maybe start with (4) and (5) and let's go from there?

How important are (1), (2), and (3)?

SunSerega:
Ok, I've fully removed everything you mentioned from (1) to (3), leaving only (4) and (5)
It kinda sounds like you wanted me to also separate (4) and (5) into 2 new PRs, but that doesn't really make sense to me since they are dependent on each other...

---

> How important are (1), (2), and (3)?

(1) is extremely important because it fixes a machine-readability problem with XML:
Currently, it's impossible to automatically determine which parameters/return values are meant to represent the error code.
Many functions don't return an error code in any way. `clCreateProgramWithBinary` returns error codes in 2 parameters at once.

(2) is similar, tho arguably less painful to fix on a per-binding basis. It's only passed into `clSetUserEventStatus` and `clSetEventCallback`, accepted as an argument to `clSetEventCallback`, and returned from `clGetEventInfo`. But there is nothing in XML to automatically find these 4 uses.

Basically, my main argument is that it's very easy to go from `cl_error_code` back to `cl_int` (1-to-1 type swap) but impossible to automatically go the other way, so it's more correct for common XML used by many bindings to have this separation.

But also relevant to this PR - (1) and (2) are both groups of enums, and their current definition in XML is the most awkward to parse of all groups...
So, to make all enum types consistent - we need a consistent type name for (1) and (2).

The importance of (3) is... Well, so far, I thought of it as just minor fixes that are better to have than not.
But I see that from your perspective, they are not minor, so let's split (3) even further:

---

> 3. Addition of new named types for extensions, e.g. cl_import_type_arm. I'm hesitant to add these to the XML file and the headers without extension spec updates, also.

Well, the `cl_arm_import_memory` spec says this:
```
      Valid <properties> are:

        * CL_IMPORT_TYPE_ARM
        * CL_IMPORT_TYPE_PROTECTED_ARM
        * CL_IMPORT_DMA_BUF_DATA_CONSISTENCY_WITH_HOST_ARM
        * CL_IMPORT_ANDROID_HARDWARE_BUFFER_PLANE_INDEX_ARM
        * CL_IMPORT_ANDROID_HARDWARE_BUFFER_LAYER_INDEX_ARM

      Valid values for CL_IMPORT_TYPE_ARM are:

        * CL_IMPORT_TYPE_HOST_ARM - this is the default
        * CL_IMPORT_TYPE_DMA_BUF_ARM
        * CL_IMPORT_TYPE_ANDROID_HARDWARE_BUFFER_ARM
```
And it seems to be a convention to name the enum group types the same as the `_properties`/`_info` enum they are bound to (So `CL_IMPORT_TYPE_ARM` => `cl_import_type_arm`)

And I think it's much better to always separate such groups of enums. Especially in the case of bitfields because they are the only items in the `<properties>` array that can be bit-combined.

This kind of info is necessary to generate fool-proofed high-lvl wrappers.

But I now see that the extension spec defines other enum types explicitly, yet for this one, there is only this mention I've shown above.
I guess the main problem here is a spec inconsistency. And it should be fixed separately from this PR...

Same for:
- `cl_affinity_domain_ext` from `cl_ext_device_fission`
- `cl_device_me_version` from `cl_intel_advanced_motion_estimation`
- `cl_device_avc_me_version` from `cl_intel_device_side_avc_motion_estimation`

Do you have any objections/suggestions for adding enum types to these 4 extensions?

Similar but not so clear cut:
- All fields of `cl_motion_estimation_desc_intel` struct from `cl_intel_motion_estimation`: #1261
- All fields of `cl_mem_ext_host_ptr` struct from `cl_qcom_ext_host_ptr`

On the other hand, this type is actually defined in the spec but inconsistent in XML:
- `cl_context_property_diagnostics_level` from `cl_intel_driver_diagnostics`: #1260

I've removed all types above from this PR.
But the `cl_command_termination_reason_arm` type was already defined in spec and partially in XML, just not in `<type>` tags, so I kept it in this PR.

> Aside: If the "OpenCL-C only" enums really are specific to OpenCL C then I think the right course of action is to remove them from the XML file and the headers, since we don't currently include OpenCL C information in the XML file.

Yes, I'm pretty sure I checked all of them when initially making this PR - and found that none are relevant outside of OpenCL-C.

I don't mind removing them from XML, but then we would be losing info. These enums are already neatly grouped, though I didn't find canonical names for those groups.

Generally, what do you think about eventually properly adding XML for OpenCL-C side info to this repo?
Maybe start with only the KHR stuff like `cl_khr_subgroup_rotate`...

bashbaug:
Thanks for separating out the changes and explaining the rationale for (1), (2), and (3)!

Let's leave this more-or-less as-is for now, since it's already a pretty lengthy PR.  Once we get this in we can start thinking about additional changes.  Note: there is a merge conflict now, though I suspect it'll be fairly easy to resolve.

xml/cl.xml
```diff
@@ -5583,15 +5607,15 @@ server's OpenCL/api-docs repository.
             <require>
                 <type name="CL/cl.h"/>
             </require>
-            <require condition="!defined(CL_VERSION_1_2)" comment="cl_device_info - defined in CL.h for OpenCL 1.2 and newer">
+            <require group="cl_device_info" condition="!defined(CL_VERSION_1_2)">
```
bashbaug, line 5610: I think we lost some important information here.  Consider:

```suggestion
            <require group="cl_device_info" condition="!defined(CL_VERSION_1_2)" comment="defined in CL.h for OpenCL 1.2 and newer">
```

xml/cl.xml
```diff
@@ -7412,30 +7449,31 @@ server's OpenCL/api-docs repository.
...
             </require>
-            <require comment="cl_platform_command_buffer_capabilities_khr - bitfield">
+            <require group="cl_platform_command_buffer_capabilities_khr" etype="bitfield">
                 <enum name="CL_COMMAND_BUFFER_PLATFORM_UNIVERSAL_SYNC_KHR"/>
                 <enum name="CL_COMMAND_BUFFER_PLATFORM_REMAP_QUEUES_KHR"/>
                 <enum name="CL_COMMAND_BUFFER_PLATFORM_AUTOMATIC_REMAP_KHR"/>
             </require>
-            <require comment="cl_device_info">
+            <require group="cl_device_info">
+                <enum name="CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR"/>
```
bashbaug, line 7461: I don't think this is correct - `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR` was added by `cl_khr_command_buffer` and is not part of `cl_khr_command_buffer_multi_device`.

```suggestion
            <require group="cl_device_info">
```

xml/cl.xml
```diff
@@ -5864,15 +5884,17 @@ server's OpenCL/api-docs repository.
...
-            <require comment="cl_device_info">
+            <require group="cl_device_info">
                 <enum name="CL_DEVICE_EXT_MEM_PADDING_IN_BYTES_QCOM"/>
                 <enum name="CL_DEVICE_PAGE_SIZE_QCOM"/>
             </require>
-            <require comment="cl_image_pitch_info_qcom">
+            <require group="cl_image_pitch_info_qcom">
+                <enum name="CL_IMAGE_ROW_PITCH"/>
                 <enum name="CL_IMAGE_ROW_ALIGNMENT_QCOM"/>
+                <enum name="CL_IMAGE_SLICE_PITCH"/>
```
bashbaug, line 5897: This isn't correct either.  `CL_IMAGE_ROW_PITCH` and `CL_IMAGE_SLICE_PITCH` are core enums are are not added by this extension.

```suggestion
            <require group="cl_image_pitch_info_qcom">
                <enum name="CL_IMAGE_ROW_ALIGNMENT_QCOM"/>
```

xml/cl.xml
```diff
@@ -6854,18 +6882,21 @@ server's OpenCL/api-docs repository.
             <require>
                 <type name="CL/cl.h"/>
             </require>
-            <require comment="cl_mem_properties">
+            <require>
+                <type name="cl_command_termination_reason_arm"/>
+            </require>
```
bashbaug, line 6887: I suspect this is a typo or merge error?  This type is not part of this extension.

```suggestion
```

xml/cl.xml
```diff
@@ -6644,46 +6671,47 @@ server's OpenCL/api-docs repository.
...
                 <enum name="CL_MEM_ALLOC_FLAGS_INTEL"/>
             </require>
-            <require comment="cl_mem_alloc_flags_intel - bitfield">
+            <require group="cl_mem_alloc_flags_intel" etype="bitfield">
                 <enum name="CL_MEM_ALLOC_WRITE_COMBINED_INTEL"/>
                 <enum name="CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL"/>
                 <enum name="CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL"/>
             </require>
-            <require comment="cl_mem_alloc_info_intel">
+            <require group="cl_mem_info">
```
bashbaug, line 6695: This should have an "intel" suffix.

```suggestion
            <require group="cl_mem_info_intel">
```

xml/cl.xml
```diff
@@ -7290,11 +7327,11 @@ server's OpenCL/api-docs repository.
             <require>
                 <type name="CL/cl.h"/>
             </require>
-            <require comment="clGetEventInfo response when param_name is CL_EVENT_COMMAND_TYPE">
+            <require group="cl_command_type" comment="clGetEventInfo response when param_name is CL_EVENT_COMMAND_TYPE">
```
bashbaug, line 7330: I don't think the comment is adding much here, and it's inconsistent with the other enum blocks.  I'd be fine removing it.

```suggestion
            <require group="cl_command_type">
```

bashbaug (commented):
Thanks for doing this - there are a lot of nice updates here!

I updated the extension header generation scripts for these changes, which prompted most of these comments.  If you're curious: https://github.com/KhronosGroup/OpenCL-Headers/pull/263

It would be good to have somebody from Arm and perhaps Qualcomm review these changes also.

xml/cl.xml
SunSerega: Hmm, but these are newly added to `cl_image_pitch_info_qcom` otherwise:
```
Table 5.XXX

    List of supported param_names by clGetDeviceImageInfoQCOM

        cl_image_pitch_info_qcom       Return Type   Info returned in
                                                     param_value

        CL_IMAGE_ROW_PITCH             cl_uint       Returns the image row pitch
                                                     supported by this device

        CL_IMAGE_ROW_ALIGNMENT_QCOM    cl_uint       Returns the image row pitch
                                                     alignment supported by this
                                                     device

        CL_IMAGE_SLICE_PITCH           cl_uint       Returns the image slice
                                                     pitch supported by this
                                                     device

        CL_IMAGE_SLICE_ALIGNMENT_QCOM  cl_uint       Returns the image slice
                                                     pitch alignment supported
                                                     by this device
```
I think I already saw multiple extensions adding the same enums - in case either one of them is supported. Is it actually wrong for extension to add relevant enums again, even if they are in core?

xml/cl.xml
SunSerega: I'm going to just commit this, but can you please describe in more detail, why this separation between `cl_mem_info`/`cl_mem_info_intel` (and other suffixed enum type names) exists?

xml/cl.xml
SunSerega: I think it should be part of the next extension (`cl_arm_controlled_kernel_termination`), rather than just deleted... I probably got tunnel vision and didn't see a separation between the 2 extensions when I did this.

xml/cl.xml
SunSerega: I can't seem to find why I added it here... Well, let me just commit this and see what might break.

SunSerega:
I don't agree with that one thing about `cl_image_pitch_info_qcom`, but otherwise, all suggestions are done.

Your header generation script looks good. I should note that all `<require>` tags with enums inside are supposed to either have a `group` or `etype` (and I'm checking for that in my bindings), but I think that doesn't yet hold with this PR alone... Mainly because of error codes.

> It would be good to have somebody from Arm and perhaps Qualcomm review these changes also.

Would you like to go ahead and ping them? I can't do this myself because I don't know who the relevant people are. But I'm fine waiting for a bit longer if the cause is quality.

xml/cl.xml
SunSerega: Ah, in this case, these 2 are actually different groups of enums for different commands...

xml/cl.xml
bashbaug: I think we may have a different idea what should be included in the XML file.  To be honest, I'm not sure if we have this written down anywhere (assuming my understanding is correct); maybe @oddhack can help.

My understanding is that the XML file is describing the types, enums, APIs, etc. that are *added* by a new extension, on top of whatever is supported in the core API.  If two extensions happen to add the same new enum (or type, or API), and neither extension has a dependency on the other extension, that's OK and both extensions can require the same new enum, because extensions are intended to stand on their own.  An extension would never require a core API enum explicitly though - instead, this requirement would be indicated by requiring a specific API version.

In the case above, `cl_image_pitch_info_qcom` can accept the `CL_IMAGE_ROW_PITCH` and `CL_IMAGE_SLICE_PITCH` enums, but these enums already exist in the core API and are not added by the extension.  Therefore, I don't think they should be required by the extension, just like e.g. the `cl_uint` type is not required by the extension.  Does this make sense?

SunSerega: Well, deciding if something should exist in XML without the context is indeed arbitrary.
But in this case, the reason for this info to exist in XML is the same as the reason for this PR: The consistent grouping of enums.

The extension adds functionality to these core enums (attached `cl_uint` return type, which I'm planning to cover in a future PR), and for that, also adds them to the `cl_image_pitch_info_qcom` enum group (which this PR's purpose covers). I don't know a better way to describe the fact that this extension adds 2 core enums to its new group.

Meanwhile, my argument about other extensions adding the same enum is more about the fact that we cannot break anything this way - since any code should already be capable of handling this.
Actually, a better example should be the `cl_mem_flags` and `cl_svm_mem_flags` groups, which add some of the same enums in the 1.0 version of the core in this PR.

PR #927, opened 2023-05-18
Document NV error code

SunSerega:
It feels a bit backward to add like this, but this error code is used in production for a while but isn't defined in XML.

The name `CL_KERNEL_ILLEGAL_BUFFER_READ_WRITE_NV`... Seem to not fully describe the situation, I think I've seen it thrown in other cases.
The only thing I'm sure about is that this error is only thrown while the kernel is executing (never while compiling/linking).

I think I found this name on some random forum a while ago.
Back then I force-added it to my bindings without thinking, to get human-readable errors while testing my code.
But I can imagine, this would also be useful the same way to others...

Do I need to ping someone from NV?
Also, any naming suggestions?

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -570,6 +570,7 @@ server's OpenCL/api-docs repository.
         <enum value="-70"           name="CL_INVALID_DEVICE_QUEUE"/>
         <enum value="-71"           name="CL_INVALID_SPEC_ID"/>
         <enum value="-72"           name="CL_MAX_SIZE_RESTRICTION_EXCEEDED"/>
+        <enum value="-9999"         name="CL_KERNEL_ILLEGAL_BUFFER_READ_WRITE_NV"/>
             <unused start="-73" end="-999" comment="Reserved for Khronos"/>
     </enums>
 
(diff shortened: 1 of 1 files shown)


nikhiljnv:
>> The name CL_KERNEL_ILLEGAL_BUFFER_READ_WRITE_NV... Seem to not fully describe the situation, I think I've seen it thrown in other cases.

Yes, this is a somewhat generic error and can mean one of many things.
If at all we want to name it, I would be more comfortable naming it more generically something like "CL_KERNEL_ERROR_UNKNOWN_NV" or "CL_KERNEL_EXECUTION_ERROR_NV".

A couple of options -

1. We can reuse existing error-codes like CL_OUT_OF_RESOURCES or CL_INVALID_OPERATION. This would require a driver change and possibly change in error code returned for existing apps/use-cases.
2. Add a new error code as mentioned above with this enum value so that applications can continue to see the same error value that can be mapped to NV specific CL error code.

SunSerega:
Well, confirmation from someone from NV that the error code is more general - is already great.
I've renamed the current enum to match the current behavior.

But I still think it's best to review where this error is thrown and add a few new, more specific enums.
Meanwhile the `-9999` code can be left for general errors, that cannot be distinguished.

PR #931, opened 2023-05-22
adds type and enums for cl_khr_initialize_memory

### LINKED ISSUE #872: missing extension types and enums in the XML file
Found while working on #825.  We are missing extension types and enums for:

cl_khr_initialize_memory:

* cl_context_memory_initialize_khr
* CL_CONTEXT_MEMORY_INITIALIZE_LOCAL_KHR
* CL_CONTEXT_MEMORY_INITIALIZE_PRIVATE_KHR

### LINKED ISSUE #921: Extension `cl_khr_initialize_memory` is missing enums
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_initialize_memory-additions-to-chapter-4

- `CL_CONTEXT_MEMORY_INITIALIZE_LOCAL_KHR`
- `CL_CONTEXT_MEMORY_INITIALIZE_PRIVATE_KHR`

These are not defined in XML or headers.
I can't find their exact values anywhere.

bashbaug:
Fixes #872 
Fixes #921 

Adds the extension type and enums for `cl_khr_initialize_memory`.  Specifically, adds:

* `cl_context_memory_initialize_khr` as a bitfield type
* `CL_CONTEXT_MEMORY_INITIALIZE_LOCAL_KHR` and `CL_CONTEXT_MEMORY_INITIALIZE_PRIVATE_KHR` as bits for this bitfield type

I couldn't find any place that defined how this type should work or what the bitfield bits should be so I took my best guess based on other code on GitHub.

Note that there is a weird potential mismatch between `cl_context_properties`, which is currently an `intptr_t` and hence possibly something other than 64 bites, and the new extension type, which is a `cl_bitfield` and hence unconditionally 64 bits.  I don't think this will be a problem, but calling it out for completeness.  See also: https://github.com/KhronosGroup/OpenCL-Docs/issues/323

We will want to regenerate headers after this changes are merged.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -207,6 +207,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_uint</type>          <name>cl_profiling_info</name>;</type>
         <type category="define">typedef <type>cl_properties</type>    <name>cl_sampler_properties</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_kernel_exec_info</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_context_memory_initialize_khr</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_terminate_capability_khr</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_unified_shared_memory_capabilities_intel</name>;</type>
         <type category="define">typedef <type>cl_properties</type>    <name>cl_mem_properties_intel</name>;</type>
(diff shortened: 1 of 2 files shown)


xml/cl.xml
```diff
@@ -5590,9 +5597,16 @@ server's OpenCL/api-docs repository.
...
                 <type name="CL/cl.h"/>
             </require>
-            <require comment="Interop tokens">
+            <require>
+                <type name="cl_context_memory_initialize_khr"/>
+            </require>
+            <require comment="cl_context_properties">
                 <enum name="CL_CONTEXT_MEMORY_INITIALIZE_KHR"/>
             </require>
+            <require comment="cl_device_terminate_capability_khr">
```
SunSerega, line 5606: Should probably be cl_context_memory_initialize_khr instead of cl_device_terminate_capability_khr?

xml/cl.xml
```diff
@@ -834,6 +835,12 @@ server's OpenCL/api-docs repository.
         <enum bitpos="31"           name="CL_QUEUE_THREAD_LOCAL_EXEC_ENABLE_INTEL"/>
     </enums>
 
+    <enums name="cl_context_memory_initialize_khr" vendor="Khronos" type="bitmask">
+        <enum bitpos="0"            name="CL_CONTEXT_MEMORY_INITIALIZE_LOCAL_KHR"/>
+        <enum bitpos="1"            name="CL_CONTEXT_MEMORY_INITIALIZE_PRIVATE_KHR"/>
+            <unused start="1" end="31"/>
```
SunSerega, line 841:
```suggestion
            <unused start="2" end="31"/>
```
Also maybe comment here to never extend this bitfield to the upper 32 bits?

xml/cl.xml
bashbaug: Yes, thank you - too much copy-paste!

xml/cl.xml
bashbaug: Fixed - I added the comment too.

SunSerega (approved):
I thought about comment-tag or comment-attribute...
But this might be even better, at least because multiline.
And anyway, this info is only for XML writers, not consumers.

alycm (approved):
LGTM.

The `cl_bitmask` vs `intptr_t` potential size mismatch is unfortunate, but using anything other than `cl_bitmask` for the type would also be unfortunate given it is a bitmask.

Bitmask values and order are the logical choice.

bashbaug:

PR #934, opened 2023-05-23
fix a few enum typos where underscores were replaced by spaces

bashbaug:
Not sure how or when this happened, but it looks like several enums had spaces instead of underscores, so `CL SUCCESS` instead of `CL_SUCCESS`.  Fixed all of these to use underscores instead, and to use asciidoctor dictionaries where appropriate.


SunSerega:
I also see `CL DEPTH_STENCIL` in 2 places.

alycm (approved):
LGTM, just going to merge immediately as obvious fixes.

PR #936, opened 2023-05-24
format the cl_khr_fp64 extension name manually

bashbaug:
Looks like my search-and-replace macros unintentionally picked up the `cl_khr_fp64` extension and tried to use an asciidoctor dictionary for it even though we haven't added dictionaries for extension names.  Format it manually, instead.


alycm (approved):
LGTM, merging immediately as an obvious fix.

PR #937, opened 2023-05-30
fix a few numerical compliance bugs in the SPIR-V environment spec

bashbaug:
Fixes internal issue 331.

* There is no dot extended instruction, use OpDot instead.
* Use HALF_EPSLION only for fp16, not for fp64.

### DIFF
--- a/env/numerical_compliance.asciidoc
+++ b/env/numerical_compliance.asciidoc
@@ -165,9 +165,9 @@ there was sufficient range, meets ULP error tolerance.
 
 ==== ULP Values for Math Instructions - Full Profile
 
-The ULP Values for Math Instructions table below describes the minimum
-accuracy of floating-point math arithmetic instructions for full profile
-devices given as ULP values.
+The ULP Values for Math Instructions - Full Profile table below
+describes the minimum accuracy of floating-point math arithmetic instructions
+given as ULP values for the full profile.
 
 [[ulp_values_for_math_instructions]]
 .ULP Values for Math Instructions - Full Profile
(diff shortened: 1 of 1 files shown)


env/numerical_compliance.asciidoc
```diff
@@ -304,10 +304,10 @@ devices given as ULP values.
 
 // n + n-1  Number of operations from n multiples and (n-1) additions
 // 2n - 1
-| *OpExtInst* *dot*
-| absolute error tolerance of 'max * max * (2n - 1) * HALF_EPSILON', for vector width _n_ and maximum input operand magnitude _max_ across all vector components
+| *OpDot*
```
gnl21, line 307: The other instructions seem to be ordered with core instructions before extended instructions, so maybe move this to after OpFDiv?

gnl21 (commented):
Apart from the one minor change that I've suggested, this looks good to me.

env/numerical_compliance.asciidoc
bashbaug: I went back and forth several times and ultimately decided to keep it in the same place.  This roughly means it matches the tables in the OpenCL C spec, and rightly keeps the instructions alphabetized.  I'm flexible though - if more people think it makes sense to list the core instructions first I don't mind moving it.

bashbaug:
> it turns out that the following opcodes are missing from there

I think I fixed all of these now.

Note, it's not entirely clear what the required accuracy should be for the fp64 versions of these opcodes in the embedded profile.  For now, I've made them all "implementation defined", since they all seem to be "implementation defined" for fp16 and fp32, and I don't see anywhere else where we have a defined accuracy for fp64 and an implementation defined accuracy for fp16 of fp32.

alycm:

PR #938, opened 2023-06-09
cl_khr_external_semaphore: Clarify language

lakshmih:
Refined the cl_khr_external_semaphore spec. Removed references to permanence which appear to have been leveraged from the Vulkan spec but don’t apply to the OpenCL spec in its current form.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -13021,7 +13021,6 @@ ifdef::cl_khr_external_semaphore_sync_fd[]
 
 include::{generated}/api/version-notes/CL_SEMAPHORE_HANDLE_SYNC_FD_KHR.asciidoc[]
   | Copy
-    | Temporary
 endif::cl_khr_external_semaphore_sync_fd[]
 |====
 
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -12866,6 +12866,10 @@ ifdef::cl_khr_external_semaphore[]
 
 [open,refpage='clGetSemaphoreHandleForTypeKHR',desc='Export external handle from a semaphore',type='protos']
 --
+
+Export operations have the same transference as the specified handle type's import operations. Additionally, exporting a semaphore payload to a handle with copy transference has the same side effects on the source semaphore's payload as executing a semaphore wait operation.
+
+Please refer to handle specific documentation for more details on transference requirements per handle type.
 To export an external handle from a semaphore, call the function
 
 include::{generated}/api/protos/clGetSemaphoreHandleForTypeKHR.txt[]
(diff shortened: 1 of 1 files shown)


bashbaug:
Leaving this open until the discussion around re-importing semaphore handles is closed.

nikhiljnv:
@lakshmih If we still need this PR, we need to resolve conflicts at TOT.

api/opencl_runtime_layer.asciidoc
```diff
@@ -13034,17 +13004,16 @@ descriptor when exporting a {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR} from a
 
 endif::cl_khr_external_semaphore_sync_fd[]
 
-.Transference and Permanence Properties for File Descriptor Handles
-[width="100%",cols="60%,<20%,<20%",options="header"]
+.Transference Properties for File Descriptor Handles
+[width="100%",cols="60%,<40%",options="header"]
 |====
-| Handle Type | Transference | Permanence
+| Handle Type | Transference
```
bashbaug, line 13010: Since we're modifying the number of table columns I think we also need to update the table contents for `CL_SEMAPHORE_HANDLE_SYNC_FD_KHR` below.

bashbaug (commented):
This mostly looks good to me, and I really like the updated description, nice work!

If it helps other reviewers, the relevant Vulkan control for temporary vs. permanent import is `VK_SEMAPHORE_IMPORT_TEMPORARY_BIT_KHR`.  Since there is no OpenCL equivalent, I agree we should not make any references to application control of temporary or permanent imports.

Do we still need to introduce the concept of a temporary import though, since SYNC_FD imports are temporary?  Or, is this also not applicable to OpenCL, because we only support creating a semaphore from an external handle, and do not support importing an external handle into an already created semaphore?

(Note: clReImportSemaphoreSyncFdKHR is specified to behave as-if the semaphore object was destroyed and re-created with the new handle, so conceptually it is still creating a semaphore.)

bcalidas:
As mentioned in an earlier comment, OpenCL does not importing  an external handle into an already created semaphore. This avoids the need for introducing the concept of temporary handle  import.

api/opencl_runtime_layer.asciidoc
```diff
@@ -12919,53 +12923,19 @@ Otherwise, it returns one of the following errors:
...
+handle types with reference transference, importing a payload to a semaphore adds the semaphore to
+the set of all semaphores sharing that payload. This set includes the semaphore from which the payload
+was exported. Semaphore signaling and waiting operations performed on any semaphore in the set must
+behave as if the set were a single semaphore. Importing a payload using handle types with copy
+transference creates a duplicate copy of the payload at the time of import, but makes no further
+reference to it. Semaphore signaling and waiting operations performed on the target of copy imports
+must not affect any other semaphore or payload.
+
+Please refer to handle specific documentation for more details on transference requirements per
+handle type.
```
kpet, line 12938: This feels like a good step forward. I'm wondering whether we should mention re-import commands in this paragraph. In the end the command was made specific to sync_fd while this paragraph is documents generic bits of behaviour.

kpet (approved):

paulfradgley (approved):
Changes look good to me.

bashbaug:
Merging as discussed in the July 9th teleconference.

Issue #888, opened 2023-03-07
Add Semaphore Re-Import mechanism to cl_khr_external_semaphore_sync_fd


bcalidas:
Sync fd semaphores must be in pending signal state before the fd  can be exported.  This means that in common usage across OpenCL and an external API, the semaphore must be continously released and a new semaphore Created.

Proposing a command to re-import semaphores for the Sync Fd handle type. This matches an API command that Vulkan has as well.

cl_int clImportSemaphoreSyncFdKHR(
    cl_semaphore_khr sema_object,
    int fd);

This switches the usage pattern from something like 

//Vulkan
CreateSyncFDSemaphore
    Submit Work
SignalSemaphore
Export Sync Fd

//OpenCL
CreateSemaphore from imported Sync Fd
Wait Semaphore 
      Submit Work
Release Semaphore

//Vulkan
       Submit Wortk
Signal Semaphore 
Export Sync Fd

//OpenCL
CreateSemaphore from imported Sync Fd
Wait Semaphore 
      Submit Work
Release Semaphore
......

To

//Vulkan
CreateSyncFDSemaphore
    Submit Work
SignalSemaphore
Export Sync Fd

//OpenCL
CreateSemaphore from imported Sync Fd
Wait Semaphore 
      Submit Work

//Vulkan
       Submit Work
Signal Semaphore 
Export Sync Fd

//OpenCL
Import Semaphore
Wait Semaphore 
      Submit Work

bcalidas:
We considered re-import via ClCreateSemaporeKHR. However that could create confusion between the use of clCreateSemapore to create a  new semaphore and the use of clCreateSemaphore for re-import. We'll also have to find a way to pass in the current cl_semaphore_khr to clCreatesemaphore which would then have to return it. The overall flow seems very odd. The main objective of clImportSemaphoreSyncFdKHR() is to allow OpenCL to use a single cl_semapore_khr object across multiple frames.

paulfradgley:
Simplifies a common use-case without adding much complexity to the API.
Seem good to me.

nikhiljnv:
The spec PR is merged as part of #939.
CTS PR is here - https://github.com/KhronosGroup/OpenCL-CTS/pull/1854
OpenCL Header PR is here - https://github.com/KhronosGroup/OpenCL-Headers/pull/237

Is there anything left to track on this issue? Or can this be closed? We can continue to track CTS and Headers changes on above PRs.

[...]

PR #939, opened 2023-06-09
cl_khr_external_semaphore_khr: semaphore re-import

lakshmih:
Add clImportSemaphoreSyncFD API call.  Sync fd semaphores must re-import the sync_fd after every wait.  Add an API call to make this re-import possible, without creating a new OpenCL semaphores.  See issue #888.

### DIFF
--- a/ext/cl_khr_external_semaphore.asciidoc
+++ b/ext/cl_khr_external_semaphore.asciidoc
@@ -339,7 +339,7 @@ imported payload, the semaphore's prior permanent payload will be restored.
 Please refer to handle specific specifications for more details on transference and
 permanence requirements specific to handle type.
 
-To import a handle of type {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR} into an existing semaphore, call the function
+To import a handle of type _handle_type_ into an existing semaphore, call the function
 
 include::{generated}/api/protos/clImportSemaphoreKHR.txt[]
 
--- a/ext/cl_khr_external_semaphore.asciidoc
+++ b/ext/cl_khr_external_semaphore.asciidoc
@@ -524,11 +524,11 @@ int fd = getFdForExternalSemaphore();
 // Create clSema of type cl_semaphore_khr usable only on device 1
 // assuming the semaphore was imported from the same device.
 cl_semaphore_properties_khr sema_props[] = {
-    (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_KHR, 
+    (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_KHR,
     (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_BINARY_KHR,
-    (cl_semaphore_properties_khr)CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR, 
+    (cl_semaphore_properties_khr)CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR,
     (cl_semaphore_properties_khr)fd,
-    (cl_semaphore_properties_khr)CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR, 
+    (cl_semaphore_properties_khr)CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR,
     (cl_semaphore_properties_khr)devices[1],
     CL_SEMAPHORE_DEVICE_HANDLE_LIST_END_KHR,
     0
@@ -596,9 +596,9 @@ clCreateContext(..., 2, devices, ...);
 
 // Create clSema of type cl_semaphore_khr usable only on device 1
 cl_semaphore_properties_khr sema_props[] = {
-    (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_KHR, 
+    (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_KHR,
     (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_BINARY_KHR,
-    (cl_semaphore_properties_khr)CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR, 
+    (cl_semaphore_properties_khr)CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR,
     (cl_semaphore_properties_khr)CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR,
     CL_SEMAPHORE_EXPORT_HANDLE_TYPES_LIST_END_KHR,
     (cl_semaphore_properties_khr)CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR,
--- a/ext/cl_khr_external_semaphore.asciidoc
+++ b/ext/cl_khr_external_semaphore.asciidoc
@@ -104,6 +104,14 @@ cl_int clGetSemaphoreHandleForTypeKHR(
     size_t *handle_size_ret);
 ----
 
+API function added by `cl_khr_external_semaphore_sync_fd` 
+
+----    
+cl_int clImportSemaphoreSyncFdKHR(
+    cl_semaphore_khr sema_object,
+    int fd);
+----
+
 === New API Enums
 
 Accepted value for the _param_name_ parameter to {clGetPlatformInfo} to query external semaphore handle types that may be imported or exported by all devices in an OpenCL platform:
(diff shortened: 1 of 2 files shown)


bcalidas:
That makes sense. We will rename to clReimportSemaphoreSyncFdKHR.

bcalidas:
Regarding the discussion around whether the usage clImportSemaphore API call should be required for importing an external semaphore payload  - we prefer the current approach of importing the external semaphore payload when the semaphore is created using clCreateSemaphoreWithProperties. This allows us to have a clean description of  external semaphore state soon as it is created. Our recommendation is that usage of   clImportSemapore be considered as a convenience as opposed to a requirement.

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
 Please refer to handle specific specifications for more details on transference and
 permanence requirements specific to handle type.
 
+To import a handle of type {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR} into an existing semaphore, call the function
```
bashbaug, line 330: Discussed in the August 29th teleconference.  I think either:

1. This sentence is stale and should be updated to include all handle types (it sounded like this was the intent).
2. Or, we need to define an error code when this API is used with a non-SYNC_FD handle type.

bashbaug (commented):
A few other follow-up questions about this feature:

1. What happens if there are outstanding signal or wait operations on the semaphore that is being re-imported?  Should these signal or wait operations operate on the previous semphore handle (the one that was imported when the enqueue call was made), or the new semaphore handle (the one that was imported when the command executes), or should this case be implementation-defined or undefined behavior?
2. Pending resolution of the discussion on #945, what should happen to any exported handles from a semaphore that has been re-imported?  Are they invalid?  If they are valid, do they refer to the "old" semaphore, or to the "new" semaphore?
3. I'm not sure if this needs to be documented explicitly in the spec or not, but just to confirm: calling clImportSemaphoreKHR does not modify the reference counts on the semaphore object, correct?

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
 Please refer to handle specific specifications for more details on transference and
 permanence requirements specific to handle type.
 
+To import a handle of type _handle_type_ into an existing semaphore, call the function
+
+include::{generated}/api/protos/clImportSemaphoreKHR.txt[]
+
+_sema_object_ specifies a valid semaphore object with importable properties.
+
+_import_props_  Must be `NULL`.  Reserved for future use.
```
nikhiljnv, line 336: Do we have any insights into what properties we may want to pass here?
May be we can fold handle_type into import_props and keep it consistent with clCreateSemaphoreWithPropertiesKHR?

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
...
+
+_handle_ptr_ is a pointer to the imported external handle.
+If _handle_ptr_ is `NULL`, _handle_size_ret_ must not be `NULL`.
+
+_handle_size_ret_ returns the actual size in bytes for the external handle.
+If _handle_size_ret_ is `NULL`, it is ignored.
+
+Calling {clImportSemaphoreKHR} is equivalent to destroying _sema_object_ and re-creating it with the original _sema_props_
+from {clCreateSemaphoreWithPropertiesKHR}, except _handle_ will be imported.
+The semaphore _sema_object_ must have originally imported a handle of type _handle_type_.
```
nikhiljnv, line 351: Do we need handle_type and handle_size in that case? The handle_type and size associated with sema_object should be available from sema_object itself.

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
...
+
+_handle_size_ specifies the size of memory pointed by _handle_ptr_.
+
+_handle_ptr_ is a pointer to the imported external handle.
+If _handle_ptr_ is `NULL`, _handle_size_ret_ must not be `NULL`.
+
+_handle_size_ret_ returns the actual size in bytes for the external handle.
+If _handle_size_ret_ is `NULL`, it is ignored.
+
+Calling {clImportSemaphoreKHR} is equivalent to destroying _sema_object_ and re-creating it with the original _sema_props_
```
nikhiljnv, line 349: Are implementations strictly required to destroy sema_object and recreate it by associating with new handle? Or just closing earlier handle reference and associating new handle reference with sema_object good enough?

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
 Please refer to handle specific specifications for more details on transference and
 permanence requirements specific to handle type.
 
+To import a handle of type _handle_type_ into an existing semaphore, call the function
+
+include::{generated}/api/protos/clImportSemaphoreKHR.txt[]
+
+_sema_object_ specifies a valid semaphore object with importable properties.
```
nikhiljnv, line 334: Do we strictly need to pass handle value at the time of import if we can associate it with this new API at a later point?
We don't need to disallow, but would it make sense to make it optional or allow the handle value to be NULL?

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -317,6 +327,36 @@ imported payload, the semaphore's prior permanent payload will be restored.
 Please refer to handle specific specifications for more details on transference and
 permanence requirements specific to handle type.
 
+To import a handle of type _handle_type_ into an existing semaphore, call the function
+
+include::{generated}/api/protos/clImportSemaphoreKHR.txt[]
```
nikhiljnv, line 332: If the intent is to change some of the semaphore properties passed at the time of clCreateSemaphoreKHR API (associated handle_value in this case), would it make sense to rename this API to something like clSetSemaphorePropertiesKHR? May be it is easier to reuse and extend it to other cases if needed in the future?

nikhiljnv (commented):
Have left some comments based on initial review.

nikhiljnv:
> What happens if there are outstanding signal or wait operations on the semaphore that is being re-imported? Should these signal or wait operations operate on the previous semphore handle (the one that was imported when the enqueue call was made), or the new semaphore handle (the one that was imported when the command executes), or should this case be implementation-defined or undefined behavior?

+1 to Ben's comments.

ext/cl_khr_external_semaphore.asciidoc
nikhiljnv: Just noting down the summary of discussion last week -
This is not a preferred path in general for OpenCL objects.
We do not want OpenCL objects to be modifiable via direct Set APIs for good reason

bcalidas:
Regarding 
1) What happens if there are outstanding signal or wait operations on the semaphore that is being re-imported? Should these signal or wait operations operate on the previous semphore handle (the one that was imported when the enqueue call was made), or the new semaphore handle (the one that was imported when the command executes), or should this case be implementation-defined or undefined behavior?

>>> If a semaphore has a pending signal and clReleaseSemaphore is called, then the effect is likely implementation defined. We should put this clarification in the base spec.  With the latest version of the spec, the semaphore is safe to resignal/reimport as soon as WaitSemaphore returns. Therefore we only need to worry about the situation where the semaphore is released/re-imported  without WaitSemaphore ever having been called.

bcalidas:
Regarding
2) Pending resolution of the discussion on https://github.com/KhronosGroup/OpenCL-Docs/issues/945, what should happen to any exported handles from a semaphore that has been re-imported? Are they invalid? If they are valid, do they refer to the "old" semaphore, or to the "new" semaphore?
In this situation, the semaphore logically would need to be-exported from OpenCL. Once OpenCL re-imports the semaphore handle , the externally exported handle would continue pointing to the old semaphore. Hence the need for re-export. This is also consistent with what you would expect if you released the semaphore and created a new one via import.

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -103,6 +104,13 @@ cl_int clGetSemaphoreHandleForTypeKHR(
     size_t *handle_size_ret);
 ----
 
+----    
+cl_int clImportSemaphoreKHR(
+    cl_semaphore_khr sema_object,
+    cl_semaphore_import_properties_khr *import_props,
+    void *handle_ptr);
```
nikhiljnv, line 111: Any specific reason to pass handle_ptr as a separate parameter and not make it one of the props just like supported by clCreateSemaphoreKHR?

bcalidas:
Based on discussions around export of imported semaphores which we have decided is not permitted (  (https://github.com/KhronosGroup/OpenCL-Docs/issues/945)
and clarification of Release Semaphore Behavior ( https://github.com/KhronosGroup/OpenCL-Docs/pull/969 ) we believe that potential  behavior  issues relating to semaphore re-import have been resolved.

The next steps are to finalize the clImportSemaphoreKHR() function name and function signature.

nikhiljnv:
As discussed in memory subgroup call on November 14, 2023, agree to move the API to cl_khr_external_semaphore_sync_fd.

bcalidas:
Please review the latest changes. The semaphore re-import command has been moved into the sync fd extension

kpet (changes requested):
This change ought to bump the extension spec number

paulfradgley:
Looks good to me too.

joshqti:
I pushed a commit to advanced the extension number.

nikhiljnv:
Is the change ready to go? Or are there any reviews pending? Also, we need to resolve the conflicts on TOT.

ext/cl_khr_external_semaphore.asciidoc
```diff
@@ -90,6 +91,13 @@ Vivek Kini, NVIDIA +
 typedef cl_uint cl_external_semaphore_handle_type_khr;
 ----
 
+The `cl_khr_external_semaphore_sync_fd` extension adds:
+
+[source]
+----
+typedef cl_properties cl_semaphore_import_properties_khr;
```
nikhiljnv, line 98: Should this be renamed to cl_semaphore_reimport_properties_khr?
Are these going to be distinct from cl_semaphore_properties_khr that can be passed at the time of clCreateSeamphoreKHR possibly supporting initial import?

nikhiljnv: Can be merged after rename to cl_semaphore_reimport_properties_khr, version update and conflict resolution.

bashbaug (commented):
LGTM, once the properties are renamed (to reimport) and the version conflict is resolved (the new version should be 0.9.2 now).  Ping me once these are fixed and I'll do a final approval and merge.

nikhiljnv (approved):
LGTM.

Issue #861, opened 2022-11-22
Clarify image tiling requirements for cl_khr_external_memory


nikhiljnv:
cl_khr_external_memory spec currently doesn't specify if tiled images can be imported from APIs like Vulkan and
more specifically what minimal tiling modes needs to be supported.
This came up during review of https://github.com/KhronosGroup/OpenCL-CTS/pull/1559
This issue tracks required changes to the spec.

bcalidas:
We have done some thinking about how the  tiling considerations in  external memory and  image from buffer (https://github.com/KhronosGroup/OpenCL-Docs/pull/710) intersect with ext_image_tiling_control and exisiting spec behavior. We worked through some principles using which these considerations could be reconciled and we would be able to move forward with the external sharing extension.

The key goals of this exercise were.

1) To reason about the current spec implicitly defines image tiling behavior. To clarify the meaning of different tiling modes while not requiring any changes to existing implementations or applications.
2) To provide a reliable mechanism for creating linear and optimally tiled images from external memory without introducing a dependence on the ext_image_tiling_control extension. Allow the external memory spec to move forward without need to depend on ext_image_tiling_control.
3) To propose the key features of  ext_image_tiling_control  that would help bring clarity and flexibility to images created using different approaches.

There's a lot of text in our next comment, but hopefully this summary will help with reviewing our proposal.

Thanks,
Balaji Calidas,
Qualcomm

bcalidas:
Here are the broad takeaways from the most recent discussion.

1) There is general agreement that optimal tiling does not exclude a linear layout.
2) When images are created from external memory, it is important to be explicit about the tiling of the image. This may mean that external memory is now dependent on image_tiling_control. It is not clear if a robust mechanism for the import of tiled images can be defined based on the current core and extension specs.
3) There were concerns  about the idea of a uuid based image layout inference mechanism. Instead it was proposed that a drm format modifier type mechanism be used to  transfer the image layout information. It was proposed that the drm format modifier extension could be made cross platform or an alternate mechanism could be easily designed.

Thanks,
Balaji Calidas
Qualcomm

bcalidas:
An additional suggestion for implementing robust import of optimally tiled images from an API like Vulkan to OpenCL is to define a vendor extension. For example vendorXYZ_cl_vk_mem_import. The presence of the extension lets the application know that there is a robust mechanism in place for transferring the image tiling information even in the absence of drm format modifiers.

[...]

PR #940, opened 2023-06-13
External memory import: Add assume linear query (#861)

lakshmih:
Add query to clGetDeviceInfo for external memory handle types that are assumed to be linear.

### DIFF
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -108,7 +108,7 @@ Accepted value for the _param_name_ parameter to {clGetDeviceInfo} to query exte
 [source]
 ----
 CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR                      0x204F
-CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR        0x????
+CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR        0x2052
 ----
 
 New properties accepted as _properties_ to {clCreateBufferWithProperties} and {clCreateImageWithProperties}:
@@ -185,12 +185,14 @@ Following new enums are added to the list of supported _param_names_ by {clGetDe
       | Returns the list of importable external memory handle types supported by _device_.
 | {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR}
   | {cl_external_memory_handle_type_khr_TYPE}[]
-      | Returns the list of importable external memory handle types supported by _device_, that are assumed to have linear layout when no other layout information is provided.
+      | Returns the list of importable external memory handle types supported by _device_, that are assumed to apply linear layout to imported images when no other tiling information is provided.
 |====
 
 {clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR} must return a non-empty list of external memory handle types for at least one of the devices in the platform.
 
-{clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} returns a list of external memory handle types that are assumed to have a linear memory layout.  This list contains a subset of {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}.  The returned list may be empty.  External memory handle types not in {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} may have any memory layout.
+{clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} returns a list of external memory handle types that are assumed to have a linear memory layout when no other tiling information is provided.  This list contains a subset of {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}.  The returned list may be empty.
+
+External memory handle types not in {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} may have any memory layout. The layout interpretation of images imported with these handle types is implementation defined.
 
 Following new properties are added to the list of supported properties by {clCreateBufferWithProperties} and {clCreateImageWithProperties}.
 
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -30,6 +30,7 @@ Other related extensions define specific external memory types that may be impor
 [cols="1,1,3",options="header",]
 |====
 | *Date*     | *Version* | *Description*
+| 2023-08-29 | 0.9.3     | Added query for {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} (provisional).
 | 2023-08-01 | 0.9.2     | Changed device handle list enum to the memory-specific {CL_MEM_DEVICE_HANDLE_LIST_KHR} (provisional).
 | 2023-05-04 | 0.9.1     | Clarified device handle list enum cannot be specified without an external memory handle (provisional).
 | 2021-09-10 | 0.9.0     | Initial version (provisional).
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -107,8 +107,8 @@ Accepted value for the _param_name_ parameter to {clGetDeviceInfo} to query exte
 
 [source]
 ----
-CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR                      0x204F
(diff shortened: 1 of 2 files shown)


xml/cl.xml
```diff
@@ -1835,6 +1835,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x204D"        name="CL_DEVICE_SEMAPHORE_IMPORT_HANDLE_TYPES_KHR"/>
         <enum value="0x204E"        name="CL_DEVICE_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR"/>
         <enum value="0x204F"        name="CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR"/>
+        <enum value="0x????"        name="CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR"/>
```
bashbaug, line 1838: We should assign a value to this enum as part of this PR, probably something from the `0x2052` to `0x2054` range that is currently unused, below.

xml/cl.xml
```diff
@@ -6987,6 +6988,9 @@ server's OpenCL/api-docs repository.
             <require comment="cl_device_info">
                 <enum name="CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR"/>
             </require>
+            <require comment="cl_device_info">
+                <enum name="CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR"/>
+            </require>
```
bashbaug, line 6993: Nitpick: the header generation will look a little nicer if this enum is included in the same `<require>` block as `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR`.

bashbaug (commented):
Mostly LGTM, although I think we should assign a value to the enum before merging becuse the current `0x????` might break some tooling.

For the record also, it'd be really great if we could come up with a shorter name than `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR`, but I'd rather be clear and verbose than terse and ambiguous and I can't think of any obvious ways to shorten it...

xml/cl.xml
lakshmih: Made the requested changes.

xml/cl.xml
lakshmih: Made the requested changes.

xml/cl.xml
bashbaug: Thanks!  Could you update the spec text also?  It still says `0x????`.

ext/cl_khr_external_memory.asciidoc
```diff
@@ -181,10 +182,15 @@ Following new enums are added to the list of supported _param_names_ by {clGetDe
...
       | Returns the list of importable external memory handle types supported by _device_.
+| {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR}
+  | {cl_external_memory_handle_type_khr_TYPE}[]
+      | Returns the list of importable external memory handle types supported by _device_, that are assumed to have linear layout when no other layout information is provided.
 |====
 
 {clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR} must return a non-empty list of external memory handle types for at least one of the devices in the platform.
 
+{clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} returns a list of external memory handle types that are assumed to have a linear memory layout.  This list contains a subset of {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}.  The returned list may be empty.  External memory handle types not in {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} may have any memory layout.
+
```
kpet, line 193: I suggest we move all of this to the table and clearly state that the layout assumed or inferred by the implementation when it cannot assume linear is implementation-defined.

We might also want to clearly state that the only scenarios that are 100% guaranteed to work are those using linearly-tiled images.

ext/cl_khr_external_memory.asciidoc
```diff
@@ -181,10 +182,15 @@ Following new enums are added to the list of supported _param_names_ by {clGetDe
 | {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}
   | {cl_external_memory_handle_type_khr_TYPE}[]
       | Returns the list of importable external memory handle types supported by _device_.
+| {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR}
```
kpet, line 185: Should the name explicitly mention images and/or tiling? Something like `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEARLY_TILED_IMAGES_HANDLE_TYPES_KHR` would capture it all but gets a bit long-winded. Any better proposals?

ext/cl_khr_external_memory.asciidoc
```diff
@@ -181,10 +182,15 @@ Following new enums are added to the list of supported _param_names_ by {clGetDe
...
+  | {cl_external_memory_handle_type_khr_TYPE}[]
+      | Returns the list of importable external memory handle types supported by _device_, that are assumed to have linear layout when no other layout information is provided.
 |====
 
 {clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR} must return a non-empty list of external memory handle types for at least one of the devices in the platform.
 
+{clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} returns a list of external memory handle types that are assumed to have a linear memory layout.  This list contains a subset of {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}.  The returned list may be empty.  External memory handle types not in {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} may have any memory layout.
```
bashbaug, line 192: One more small suggested change, since you'll need to update the spec text to remove `0x????` anyhow: I recommend adding to the first sentence in this paragraph so it will not need to be updated if/when we add other explicit tiling controls.  Something like:

```suggestion
{clGetDeviceInfo} when called with param_name {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} returns a list of external memory handle types that are assumed to have a linear memory layout when no other tiling controls are provided.  This list contains a subset of {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_HANDLE_TYPES_KHR}.  The returned list may be empty.  External memory handle types not in {CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR} may have any memory layout.
```

As a small editorial note, it makes for easier editing and smaller diffs to avoid long lines like this one.  There are several recommended best practices to avoid long lines: asciidoctor [recommends](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) writing one sentence per line, or many editors can wrap lines longer than a certain length.

bashbaug (commented):
One more minor comment that would be nice to fix.

Also needs the CLA to be accepted before merging.  Thanks!

kpet (changes requested):
Thanks for this proposal! Fine with the general direction as a way of getting forward progress. We should bump the minor version of the extension as part of this PR as we're clearly adding additional functionality in a backwards-compatible fashion.

lakshmih (commented):
Addressed review comments.

ext/cl_khr_external_memory.asciidoc
bashbaug: Compromise?  `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_IMAGE_HANDLE_TYPES_KHR`

bashbaug (approved):
LGTM.

Did we decide whether we wanted "images" in the name of this query?  Specifically, do we want:

* `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR` (current name)
* `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_IMAGES_HANDLE_TYPES_KHR` (suggested alternative)

bcalidas:
> 
> Did we decide whether we wanted "images" in the name of this query? Specifically, do we want:
> 
> * `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_HANDLE_TYPES_KHR` (current name)
> * `CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_IMAGES_HANDLE_TYPES_KHR` (suggested alternative)

We have a preference for the current name.

bcalidas:
As agreed, we pushed a new commit with the CL_DEVICE_EXTERNAL_MEMORY_IMPORT_ASSUME_LINEAR_IMAGES_HANDLE_TYPES_KHR name.

lakshmih:
Fixed the merge conflict and bumped the minor version, should be good to merge.

bashbaug (approved):
LGTM!

nikhiljnv (approved):
LGTM.

bashbaug:
Merging as discussed in the September 5th teleconference - @kpet's comments have all been addressed.

PR #947, opened 2023-07-01
Function pointer naming

SunSerega:
Currently, function pointers definitions are placed inside `<command>` and require full-on parsing to properly interpret, because there are no tags to separate parameters and return type.

Also, the [cl_arm_printf](https://registry.khronos.org/OpenCL/extensions/arm/cl_arm_printf.html) extension requires its own function pointer type, but there is nowhere to define it in XML.

---

So, I've added `<type category="function">` tags. I've also merged function pointer types with the same parameters under a common name. Including cases where original spacing differs:

`clEnqueueSVMFreeARM`:
```
void * svm_pointers[], void *user_data
```
`clEnqueueSVMFree`:
```
void* svm_pointers[], void* user_data
```

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -372,6 +372,61 @@ server's OpenCL/api-docs repository.
             <member><type>cl_uint</type>                                 <name>num_mutable_dispatch</name></member>
             <member>const <type>cl_mutable_dispatch_config_khr</type>*   <name>mutable_dispatch_list</name></member>
         </type>
+        
+        <type category="function" name="cl_enqueue_svm_free_callback">
+            <proto><type>void</type></proto>
+            <param><type>cl_command_queue</type>            <name>queue</name></param>
+            <param><type>cl_uint</type>                     <name>num_svm_pointers</name></param>
+            <param><type>void</type>*                       <name>svm_pointers</name>[]</param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_create_context_callback">
+            <proto><type>void</type></proto>
+            <param>const <type>char</type>*                 <name>errinfo</name></param>
+            <param>const <type>void</type>*                 <name>private_info</name></param>
+            <param><type>size_t</type>                      <name>cb</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_context_destructor_callback">
+            <proto><type>void</type></proto>
+            <param><type>cl_context</type>                  <name>context</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_mem_object_destructor_callback">
+            <proto><type>void</type></proto>
+            <param><type>cl_mem</type>                      <name>memobj</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_program_callback">
+            <proto><type>void</type></proto>
+            <param><type>cl_program</type>                  <name>program</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_event_callback">
+            <proto><type>void</type></proto>
+            <param><type>cl_event</type>                    <name>event</name></param>
+            <param><type>cl_int</type>                      <name>event_command_status</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_enqueue_native_kernel_callback">
+            <proto><type>void</type></proto>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
+        <type category="function" name="cl_printf_callback">
+            <proto><type>void</type></proto>
+            <param>const <type>char</type>*                 <name>buffer</name></param>
+            <param><type>size_t</type>                      <name>len</name></param>
+            <param><type>size_t</type>                      <name>complete</name></param>
+            <param><type>void</type>*                       <name>user_data</name></param>
+        </type>
+        
     </types>
 
     <!-- SECTION: OpenCL enumerant (token) definitions. -->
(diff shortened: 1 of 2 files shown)


RocketRide9:
Nice PR, it makes writing a binding generator easier. Maybe functions should be added as OpenCL version/extension `require`'s too?

SunSerega:
@bashbaug, what do you think about require tags? I can add that in this PR, or a separate one, I'm not sure if that would be extra complexity here...

RocketRide9:
> How confident are we that introducing the function pointer types isn't going to cause a compatibility issue?

Can't C header generator emit the same header after these changes? It can translate functions as typedefs and then use them in command declarations, but it can also write out function prototype for every argument, like it's done currently. The only difference is going to be in spaces, like pointed out in the PR description.

Kerilk:
> Added function types to `<require>` tags.
> 
> Usage of function types per API function:
> 
> ```
> cl_enqueue_svm_free_callback			clEnqueueSVMFreeARM
> cl_enqueue_svm_free_callback			clEnqueueSVMFree
> 
> cl_create_context_callback				clCreateContext
> cl_create_context_callback				clCreateContextFromType
> 
> cl_context_destructor_callback			clSetContextDestructorCallback
> 
> cl_mem_object_destructor_callback		clSetMemObjectDestructorCallback
> cl_mem_object_destructor_callback		clSetMemObjectDestructorAPPLE
> 
> cl_program_callback						clBuildProgram
> cl_program_callback						clCompileProgram
> cl_program_callback						clLinkProgram
> cl_program_callback						clSetProgramReleaseCallback
> 
> cl_event_callback						clSetEventCallback
> 
> cl_enqueue_native_kernel_callback		clEnqueueNativeKernel
> ```
> 
> It seems this last commit broke generators, and I might as well try to do as @RocketRide9 described.

I like the fact that you tried to factor the callbacks. Though I think some type safety might be useful when the intent of the callback is different. If we factor the callback, then I think the callback function should be clear in the name, I am very bad at naming things, but here are a few counter proposal to get the ball rolling:

- cl_free_svm_callback: clEnqueueSVMFreeARM, clEnqueueSVMFree
- cl_context_error_callback: clCreateContext, clCreateContextFromType
- cl_context_destruction_callback: clSetContextDestructorCallback
- cl_mem_object_destruction_callback: clSetMemObjectDestructorCallback, clSetMemObjectDestructorAPPLE
- cl_program_build_completion_callback: clBuildProgram
- cl_program_compile_completion_callback: clCompileProgram
- cl_program_link_completion_callback: clLinkProgram
- cl_program_destruction_callback: clSetProgramReleaseCallback
- cl_event_status_change_callback: clSetEventCallback
- cl_native_kernel: clEnqueueNativeKernel

My reasoning for not factoring the different program callbacks is that you would most probably not want to use these interchangeably, since they have different intent.

SunSerega:
Agree with almost everything (and assume you meant `_callback` at the end of `cl_native_kernel`), and I still trust your naming sense more than mine))

But for now, I'm not convinced that we need to separate function types with the same parameters.
Especially between `clBuildProgram`, `clCompileProgram`, and `clLinkProgram`. These functions are very similar in their use in practice, and I can imagine a helper function for all of them - why make that harder?
More generally, as I see it now, the main reason for strongly typing enums is the IDE suggestions, with preventing the passing of incorrect values being secondary. And it's hard to imagine either reason being relevant for function types.
(I'm very open to discussion about this)

P.S. There is also `cl_printf_callback`, which I somehow missed in my list above, in case you also want to suggest its rename.

bashbaug:
In the cases when a callback is used by a core API and an extension API we have an additional wrinkle to consider: if an extension was added to "backport" a feature to a prior version of OpenCL, it would be undesirable for the extension to rely on a function pointer type that is only defined for a newer version of OpenCL.  Therefore, we would want the core function pointer type to be defined for all versions of OpenCL, not just when the core API is added.  Not the end of the world, but this feels a little weird.

We should also prepare for the eventuality where an extension adds a callback and later gets prompted to a core feature.

Given these issues, would it be easier to have a specific type for a core API callback vs. an extension API callback?

We could also do something like [this](https://github.com/KhronosGroup/OpenCL-Headers/blob/6137cfbbc7938cd43069d45c622022572fb87113/CL/cl_ext.h#L1618), which we've done a few times in the past, but it's always felt a little ugly:

```c++
#if !defined(CL_VERSION_2_1)
/* defined in CL.h for OpenCL 2.1 and newer */
typedef cl_uint             cl_kernel_sub_group_info;
#endif /* !defined(CL_VERSION_2_1) */
```

xml/cl.xml
```diff
@@ -5863,6 +5924,7 @@ server's OpenCL/api-docs repository.
         <extension name="cl_APPLE_SetMemObjectDestructor" revision="0.0.0" comment="not registered" supported="opencl">
             <require>
                 <type name="CL/cl.h"/>
+                <type name="cl_mem_object_destruction_callback"/>
```
Kerilk, line 5938: This would need guards so it is only defined when cl.h does not provide it. Redefinition of typedefs is illegal in c99.

xml/cl.xml
```diff
@@ -6338,6 +6400,7 @@ server's OpenCL/api-docs repository.
                 <type name="cl_svm_mem_flags_arm"/>
                 <type name="cl_kernel_exec_info_arm"/>
                 <type name="cl_device_svm_capabilities_arm"/>
+                <type name="cl_free_svm_callback"/>
```
Kerilk, line 6419: Same here, this would need guards so it is only defined when cl.h does not provide it. Redefinition of typedefs is illegal in c99.

xml/cl.xml
Kerilk: This is related to the comment from @bashbaug above.

SunSerega:
Added callback for `clSetProgramReleaseCallback` and the guards for callbacks defined in both core and ext.

@bashbaug, as I understand, we are now waiting on a decision: Do we want to have generators output inline callbacks as before, or to change the API to a (mostly) compatible one with separately defined callbacks, as in the sister PR https://github.com/KhronosGroup/OpenCL-Headers/pull/294

Issue #948, opened 2023-07-05
HTML errors in OpenCL specification


KhronosWebservices:
There are 11 issues in the HTML of the OpenCL spec [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html).  The issues are listed here:

https://validator.w3.org/nu/?doc=https%3A%2F%2Fregistry.khronos.org%2FOpenCL%2Fspecs%2F3.0-unified%2Fhtml%2FOpenCL_API.html

Fixing these issues will help us ensure the the major search engines can reliably pickup on these pages.

bashbaug:
I created a PR that fixes all of these but one.  The remaining issue is in our `khronos.css` style sheet:

```
#footer { max-width: 100%; background-color: none; padding: 1.25em; }
```

This produces the error:

> Error: CSS: background-color: none is not a background-color value.

@oddhack, any preference how to solve this?  It looks like other Khronos repos that share this style sheet have the same issue (checked [Vulkan-Docs](https://github.com/KhronosGroup/Vulkan-Docs/blob/2b7d9c9f65cde43580573a677490c3c7d099c09f/config/khronos.css#L366)).

outofcontrol:
```
#footer { max-width: 100%; background: none; padding: 1.25em; }
```

would be appropriate.

bashbaug:
Works for me!

bashbaug:
Fixed by #952 - closing.

oddhack:
> ```
> #footer { max-width: 100%; background: none; padding: 1.25em; }
> ```
> 
> would be appropriate.

@outofcontrol @bashbaug I'm wondering where you saw this error? I looked at the JS console in Chromium while loading the Vulkan single-page HTML spec, which contains the background-color: none CSS, and saw no errors reported (not arguing that it's valid, just trying to understand where it was diagnosed).

bashbaug:
It came from the validator in the initial issue description - we haven't published a new spec (yet) so you can still see the errors:

> https://validator.w3.org/nu/?doc=https%3A%2F%2Fregistry.khronos.org%2FOpenCL%2Fspecs%2F3.0-unified%2Fhtml%2FOpenCL_API.html

outofcontrol:
This month, we are finally seeing OpenCL in the registry appear in Google when searching for 'opencl spec'.

PR #952, opened 2023-07-18
fix html errors

bashbaug:
See #948 

This PR fixes all but one of the identified issues.  Fixes were:

* There were several places that defined duplicate anchor attributes.  I fixed these by de-duplicating the anchors.  If I have time I'll try to update some of the checker scripts to catch these issues, also.
* There was one place with incorrect asciidoctor formatting.  I fixed this by removing an underscore.
* The embedded PNG cannot have line breaks.  Fixed by removing line breaks - yes, this does produce a very long line.


bashbaug:
OK, I think I've fixed all of these now.  I also fixed a few other errors in the OpenCL C and OpenCL Extensions specs.  I didn't need to make any specific changes for the OpenCL SPIR-V Environment spec.

config/docinfo-header.html
```diff
@@ -1,286 +1,4 @@
 <div style="text-align: left;">
-    <span class="image"><img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAiYAAAD6CAYAAACVrYw7AAAMSHpUWHRSYXcgcHJvZmlsZSB0eXBl
```
kpet, line 2: Probably not for this PR but a separate PNG file might be easier to maintain.

kpet (approved):

bashbaug:
Merging as discussed in the September 12th teleconference.

PR #953, opened 2023-07-18
fix nextafter description

### LINKED ISSUE #951: "nextafter" function mis-defined for double precision
The `nextafter` function is defined in [section 6.15.2](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#math-functions) "Math Functions" of the OpenCL C specification as:

> ```
> gentype nextafter(gentype x, gentype y)
> ```
>
> Computes the next representable **single-precision** floating-point value following x in the direction of y. Thus, if y is less than x, nextafter() returns the largest representable floating-point number less than x.

(Emphasis mine.)

This function is available for both `float` and `double` types, so I think it is incorrect for the description to specifically say that it returns the next **single-precision** floating-point value.  If the input type is `double`, I presume it should return the next double-precision floating-point value.  It seems like the wording could simply drop these words:

> Computes the next representable floating-point value following x in the direction of y. Thus, if y is less than x, nextafter() returns the largest representable floating-point number less than x.

FWIW, I'm on the Khronos SYCL WG, and I'm reporting this because we have similar wording in the SYCL specification.

bashbaug:
Fixes #951 

The description of nextafter should not specifically describe single-precision floating-point numbers because it also works with doubles (and halfs, when cl_khr_fp16 is supported).

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -4811,7 +4811,7 @@ all arguments and the return type, unless otherwise specified.
     | Returns a quiet NaN.
       The _nancode_ may be placed in the significand of the resulting NaN.
 | gentype *nextafter*(gentype _x_, gentype _y_)
-    | Computes the next representable single-precision floating-point value
+    | Computes the next representable floating-point value
       following _x_ in the direction of _y_.
       Thus, if _y_ is less than _x_, *nextafter*() returns the largest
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM.

PR #955, opened 2023-07-24
fix a few places where the fp16 and fp64 frexp referred to a float

### LINKED ISSUE #954: "frexp" function typo for "half" inputs
The OpenCL cl_khr_fp16  extension defines the `frexp` function like this:

> Extract mantissa and exponent from x. For each component the mantissa returned is a **float** with magnitude in the interval [1/2, 1) or 0. Each component of x equals mantissa returned * 2<sup>exp</sup>.

(Emphasis mine.)

The word "float" should be "half" for this definition.  Note that the core OpenCL C spec has separate definitions of `frexp` for `float` and `double` and this word correctly matches the return type in those definitions.

FWIW, the SYCL spec had the same mistake.  Since we have a single description for all variants of `frexp`, I'm changing the description to say "... the mantissa returned is a floating point number with magnitude ...".  The exact return type is already defined by the function synopsis, so there is no need to repeat it in the description.

bashbaug:
fixes #954 

Fixes two issues in the extensions spec where the cl_khr_fp16 and cl_khr_fp64 descriptions of `frexp` still referred to a `float` rather than as a `half` or a `double`.

### DIFF
--- a/ext/cl_khr_fp16.asciidoc
+++ b/ext/cl_khr_fp16.asciidoc
@@ -222,8 +222,8 @@ all arguments and the return type.
   half__n__ **frexp** (half__n__ _x_, int__n__ *exp) +
   half **frexp** (half _x_, int *exp)
 | Extract mantissa and exponent from _x_.
-  For each component the mantissa returned is a float with magnitude in the
-  interval [1/2, 1) or 0.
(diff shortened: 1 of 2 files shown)


alycm (approved):
LGTM.

bashbaug:

PR #958, opened 2023-08-08
Update cl_arm_scheduling_controls to v0.6.0

kpet:
Change-Id: I018cd201eaa872c19ca58af3cdd5f14313593c13

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -2184,7 +2184,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x41EF"        name="CL_IMPORT_ANDROID_HARDWARE_BUFFER_PLANE_INDEX_ARM"/>
         <enum value="0x41F0"        name="CL_IMPORT_ANDROID_HARDWARE_BUFFER_LAYER_INDEX_ARM"/>
         <enum value="0x41F1"        name="CL_KERNEL_EXEC_INFO_COMPUTE_UNIT_MAX_QUEUED_BATCHES_ARM"/>
-        <enum value="0x41F2"        name="Reserved for cl_arm_core_builtins"/>
+        <unused start="0x41F2" end="0x41F2" comment="Reserved for cl_arm_core_builtins"/>
         <enum value="0x41F3"        name="CL_QUEUE_COMPUTE_UNIT_LIMIT_ARM"/>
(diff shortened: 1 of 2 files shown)


xml/cl.xml
```diff
@@ -2183,7 +2184,9 @@ server's OpenCL/api-docs repository.
         <enum value="0x41EF"        name="CL_IMPORT_ANDROID_HARDWARE_BUFFER_PLANE_INDEX_ARM"/>
         <enum value="0x41F0"        name="CL_IMPORT_ANDROID_HARDWARE_BUFFER_LAYER_INDEX_ARM"/>
         <enum value="0x41F1"        name="CL_KERNEL_EXEC_INFO_COMPUTE_UNIT_MAX_QUEUED_BATCHES_ARM"/>
-            <unused start="0x41F2" end="0x41FF"/>
+        <enum value="0x41F2"        name="Reserved for cl_arm_core_builtins"/>
```
SunSerega, line 2187:
```suggestion
        <unused start="0x41F2" end="0x41F2" comment="Reserved for cl_arm_core_builtins"/>
```

kpet: Fine with the suggestion but, before I apply it: what was breaking?

SunSerega: Well, it would probably break my bindings.

Wouldn't be hard to fix, but more importantly, even if it didn't break anything:
- Everywhere else in XML `Reserved for ...` enum ranges are defined using `<unused>` tags.
- Meanwhile `name` attribute of `<enum>` tags only ever refers to a valid OpenCL constant name - which can't have spaces.

These are very convenient to rely upon. Reading XML would be more hacky otherwise.

kpet: I completely agree and I wasn't challenging your request to change this. Your proposed change is absolutely right. I was (pleasantly) surprised to see that kind of feedback from someone outside of the OpenCL working group and assumed that you had something using the XML that this PR would have broken, hence my question.

We really ought to have better automated checks for the XML. I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/959. Feel free to suggest checks you think we should have on that issue. Thanks!

bashbaug (approved):
LGTM!

Issue #731, opened 2021-11-30
CL_M_* constants unspecified


kpet:
The XML registry defines a number of math constants named `CL_M_*`. These are provided by `cl_platform.h` but are not described in any of the specifications.

PR #960, opened 2023-08-09
Add a few missing uses of types and enums to XML

kpet:
- OpenCL 1.0 requires cl_char, etc types
- OpenCL 1.2 and cl_khr+_fp64 require cl_double
- cl_khr_fp16 requires CL_HALF_* constants
- cl_khr_icd requires cl_icd_dispatch
- OpenCL 1.0 requires all the CL_M_* constants. The specification does not state which version defines which constant (see #731)
- cl_khr_gl_sharing requires all cl_GL* types

Change-Id: I8eb34ab1eccf727700662ff5f61823d0e8c48ea1

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -4299,6 +4299,18 @@ server's OpenCL/api-docs repository.
             <type name="cl_image_format"/>
             <type name="cl_buffer_region"/>
         </require>
+        <require comment="API data types">
+            <type name="cl_char"/>
+            <type name="cl_uchar"/>
+            <type name="cl_short"/>
+            <type name="cl_ushort"/>
+            <type name="cl_int"/>
+            <type name="cl_uint"/>
+            <type name="cl_long"/>
+            <type name="cl_ulong"/>
+            <type name="cl_float"/>
+            <type name="cl_half"/>
+        </require>
         <require comment="Constants">
             <enum name="CL_CHAR_BIT"/>
             <enum name="CL_CHAR_MAX"/>
@@ -4340,6 +4352,32 @@ server's OpenCL/api-docs repository.
             <enum name="CL_HUGE_VAL"/>
             <enum name="CL_MAXFLOAT"/>
             <enum name="CL_INFINITY"/>
+            <enum name="CL_M_E"/>
+            <enum name="CL_M_LOG2E"/>
+            <enum name="CL_M_LOG10E"/>
+            <enum name="CL_M_LN2"/>
+            <enum name="CL_M_LN10"/>
+            <enum name="CL_M_PI"/>
+            <enum name="CL_M_PI_2"/>
+            <enum name="CL_M_PI_4"/>
+            <enum name="CL_M_1_PI"/>
+            <enum name="CL_M_2_PI"/>
+            <enum name="CL_M_2_SQRTPI"/>
+            <enum name="CL_M_SQRT2"/>
+            <enum name="CL_M_SQRT1_2"/>
+            <enum name="CL_M_E_F"/>
+            <enum name="CL_M_LOG2E_F"/>
+            <enum name="CL_M_LOG10E_F"/>
+            <enum name="CL_M_LN2_F"/>
+            <enum name="CL_M_LN10_F"/>
+            <enum name="CL_M_PI_F"/>
+            <enum name="CL_M_PI_2_F"/>
+            <enum name="CL_M_PI_4_F"/>
+            <enum name="CL_M_1_PI_F"/>
+            <enum name="CL_M_2_PI_F"/>
+            <enum name="CL_M_2_SQRTPI_F"/>
+            <enum name="CL_M_SQRT2_F"/>
+            <enum name="CL_M_SQRT1_2_F"/>
         </require>
         <require comment="Error codes">
             <enum name="CL_SUCCESS"/>
@@ -4864,6 +4902,7 @@ server's OpenCL/api-docs repository.
             <type name="cl_kernel_arg_access_qualifier"/>
             <type name="cl_kernel_arg_type_qualifier"/>
             <type name="cl_image_desc"/>
+            <type name="cl_double"/>
         </require>
         <require comment="Constants">
             <enum name="CL_DBL_DIG"/>
@@ -5039,6 +5078,7 @@ server's OpenCL/api-docs repository.
         <require>
             <type name="cl_device_svm_capabilities"/>
             <type name="cl_queue_properties"/>
+            <type name="cl_properties"/>
             <type name="cl_svm_mem_flags"/>
             <type name="cl_pipe_properties"/>
             <type name="cl_pipe_info"/>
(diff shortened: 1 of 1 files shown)


kpet:
With the change the only missing uses reported by https://github.com/KhronosGroup/OpenCL-Docs/pull/730 are vendor definitions and vendor-reserved enums.

xml/cl.xml
```diff
@@ -5484,6 +5524,7 @@ server's OpenCL/api-docs repository.
         <extension name="cl_khr_fp64" supported="opencl">
             <require>
                 <type name="CL/cl.h"/>
+                <type name="cl_double"/>
```
bashbaug, line 5527: This is debatably the right thing to do, but `cl_double` is currently defined unconditionally in `cl_platform.h`.  This means that the extension headers should never emit anything for `cl_double`, even if it matches what is currently in `cl_platform.h`, at least for C99:

https://stackoverflow.com/questions/26240370/why-are-typedef-identifiers-allowed-to-be-declared-multiple-times

So, what should we do about this?  Some options:

1. Consider `cl_double` to effectively be part of the API, unconditionally, meaning it's part of "OpenCL 1.0".  This feels a little dirty, but I can't find any headers without `cl_double` so I think it's encoding reality.
2. Consider `cl_double` to be an extension type only pre-OpenCL 1.2, and a core type for OpenCL 1.2 and newer.  This is probably the "most correct", though I'm worried what this will break.
3. Keep `cl_double` defined as-is in the headers and more-or-less as-defined here in the XML file, but add some sort of conditions or special-case code in the extension header generation scripts to avoid emitting anything for it.

(1) is the easiest thing to do.  I wouldn't recommend (2).  I think we could make (3) work but we'd need to figure out exactly what we want to do.

xml/cl.xml
```diff
@@ -5496,6 +5537,18 @@ server's OpenCL/api-docs repository.
...
+                <enum name="CL_HALF_DIG"/>
+                <enum name="CL_HALF_MANT_DIG"/>
+                <enum name="CL_HALF_MAX_10_EXP"/>
+                <enum name="CL_HALF_MAX_EXP"/>
+                <enum name="CL_HALF_MIN_10_EXP"/>
+                <enum name="CL_HALF_MIN_EXP"/>
+                <enum name="CL_HALF_RADIX"/>
+                <enum name="CL_HALF_MAX"/>
+                <enum name="CL_HALF_MIN"/>
+                <enum name="CL_HALF_EPSILON"/>
```
bashbaug, line 5550: These half constants are also defined in `cl_platform.h`.  Since these are `#defines` and not `typedefs` we could define them twice without a diagnostic (assuming they're defined to the same values, which hopefully they are!), though it might be better not to do this.

Do we want to:

1. Consider these constants to be parts of the API, unconditionally, also, by including them as part of "OpenCL 1.0" and not part of the `cl_khr_fp16` extension?
2. Take them out of `cl_platform.h` and define them in `cl_ext.h` instead, as part of the `cl_khr_fp16` extension?
3. Some other solution that involves special-casing these constants in the extension header generation scripts.

Note, I originally thought it was weird that `cl_half` was in "OpenCL 1.0", especially because `cl_double` is not, but perhaps this is intended and correct due to the `vload_half` and `vstore_half` built-in functions?

xml/cl.xml
```diff
@@ -5518,6 +5571,7 @@ server's OpenCL/api-docs repository.
         <extension name="cl_khr_icd" supported="opencl">
             <require>
                 <type name="CL/cl.h"/>
+                <type name="cl_icd_dispatch"/>
```
bashbaug, line 5574: I'm not sure what to do with this one.

It doesn't seem right to include it with `cl_khr_icd`, since it doesn't define any APIs that take a `cl_icd_dispatch` or a pointer to a `cl_icd_dispatch`, and the [extension text itself](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_icd-inferring-vendors-from-function-call-arguments) only describes `struct _cl_icd_dispatch`.

Maybe it should be part of [cl_loader_layers](https://github.com/KhronosGroup/OpenCL-Docs/blob/main/ext/cl_loader_layers.asciidoc), since it at least defines a few APIs that take a pointer to a `cl_icd_dispatch`?

If we consider this type to be a part of `cl_loader_layers`, though, should we [typedef](https://github.com/KhronosGroup/OpenCL-Headers/blob/e049b16b5f157e2f28e7b5c301e71e1ccb3fe288/CL/cl_icd.h#L38) off of `struct _cl_icd_dispatch` though?  Based on the current headers the type doesn't appear to be associated with any extension...

xml/cl.xml
```diff
@@ -6444,6 +6498,9 @@ server's OpenCL/api-docs repository.
         <extension name="cl_khr_gl_depth_images" requires="cl_khr_gl_sharing" comment="no API - reuses tokens from core API" supported="opencl">
             <require>
                 <type name="CL/cl.h"/>
+                <type name="cl_GLint"/>
+                <type name="cl_GLenum"/>
+                <type name="cl_GLuint"/>
```
bashbaug, line 6503: If we're going to attach these to an extension then I think it makes the most sense to attach them to `cl_khr_gl_sharing` and not to `cl_khr_gl_depth_images`.

Note that these types are also defined in `cl_platform.h`, though that's probably a bug, and I think it'd be pretty low-risk to define them in an extension header (specifically `cl_gl.h`) instead.

bashbaug (commented):
Thanks for doing this!  I have a few comments and suggestions, mostly from a header organization (and generation) point of view.

xml/cl.xml
kpet: Agree, which, looking at my commit, message seemed to have been the intent.

Also agree, I've created https://github.com/KhronosGroup/OpenCL-Headers/pull/257

xml/cl.xml
kpet: I suggest we move the require tag to cl_loader_layers, at least as a first step. It's obviously the case and results at least one user for this which is the goal of this change.

xml/cl.xml
kpet: The other PRs have gone in so I'll resolve this thread. I've rebased this change again, over to you for when you get a chance to look at the other threads on this PR.

xml/cl.xml
bashbaug: Noting for completeness: I think the XML change to move `cl_icd_dispatch` to `cl_loader_layers` is probably the best we can do, but it's still a little bit of a mess because it's used by both layers and by the OpenCL ICD loader.  If we don't do anything, we will end up with a typedef for `cl_icd_dispatch` in both `cl_icd.h` and `cl_layer.h`, which seems like a recipe for trouble for any library or application that choses to include both files.

Some options to "fix" this (not in this PR!) are:

1. Update the header generation scripts so `cl_icd_dispatch` is on a "denylist" and doesn't get generated like the other typedefs.  Then, the headers will be generated more-or-less as-is, and neither the ICD loader nor any layers need to change.
2. Remove the typedef from `cl_icd.h` and generate it in `cl_layer.h`.  This may require minor changes to the OpenCL ICD loader, to include `cl_layer.h` instead of `cl_icd.h`.  Note that `cl_layer.h` currently includes `cl_icd.h`, so no changes should be required for layers (in theory, at least).

xml/cl.xml
bashbaug: Confirming: we decided to go with (2)?  If so, this means we need to add a "denylist" to the header generation script for these enums.  Not a big deal, but it's not something I'll do unless we need to do it.

PR #967, opened 2023-09-12
Restrict execution scope to subgroup in environment for cl_khr_subgroup_rotate

### LINKED ISSUE #965: SPIR-V cl_khr_subgroup_rotate should require Subgroup scope
The SPIR-V environment spec for cl_khr_subgroup_rotate currently does not say anything about valid Scopes for Execution.

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_cl_khr_subgroup_rotate

Because the SPIR-V instruction supports both Subgroup and Workgroup Scope...

http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_subgroup_rotate.html

... the SPIR-V environment spec should only allow Subgroup as a valid Scopes for Execution.  This would be consistent with other subgroup extensions.

kpet:
Fixes #965

Change-Id: I74c046ed5dd3dbead20f5ba33d8f4f1b94e3da9b

### DIFF
--- a/env/extensions.asciidoc
+++ b/env/extensions.asciidoc
@@ -355,6 +355,10 @@ declare the following SPIR-V capabilities:
 
 * *GroupNonUniformRotateKHR*
 
+For instructions requiring these capabilities, _Scope_ for _Execution_ may be:
+
+* *Subgroup*
+
 ==== `cl_khr_work_group_uniform_arithmetic`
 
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM!

PR #969, opened 2023-09-15
Clarify clReleaseSemaphoreKHR

joshqti:
If a semaphore is deleted while there is a pending signal, specify the behavior.

### DIFF
--- a/ext/cl_khr_semaphore.asciidoc
+++ b/ext/cl_khr_semaphore.asciidoc
@@ -465,7 +465,8 @@ Otherwise, it returns one of the following errors:
 
 After the reference count becomes zero and commands queued for execution on a
 command-queue(s) that use _sema_object_ have finished, the semaphore object is
-deleted.
+deleted. Undefined behavior may occur if _sema_object_ has a pending signal and
+{clEnqueueWaitSemaphoresKHR} has not been called for _sema_object_.
 Using this function to release a reference that was not obtained by creating the
 object via {clCreateSemaphoreWithPropertiesKHR} or by calling
 {clRetainSemaphoreKHR} causes undefined behavior.


ext/cl_khr_semaphore.asciidoc
```diff
@@ -465,7 +465,8 @@ Otherwise, it returns one of the following errors:
 
 After the reference count becomes zero and commands queued for execution on a
 command-queue(s) that use _sema_object_ have finished, the semaphore object is
-deleted.
+deleted. Undefined behavior may occur if _sema_object_ has a pending signal and
+{clEnqueueWaitSemaphoresKHR} has not been called for _sema_object_.
```
nikhiljnv, line 469: Just trying to understand what may cause undefined behavior and how having called clEnqueueWaitSemaphoresKHR or not would make a difference here at least in case of core semaphore to be used within OpenCL if we try to release a semaphore with pending signal. Is this because the pending signal  operation may not complete and we end up releasing semaphore earlier? I guess that would be more problematic if we had called clEnqueueWaitSemaphoresKHR. Also, if there is a pending signal on the semaphore that we are trying to release, shouldn't the reference count be non-zero and shouldn't we wait for it to complete before releasing it? Even if we don't do any of these and end up releasing the semaphore with pending signal, there is no way to access the semaphore post the release operation and any use of cl_semaphore object would anyway be invalid/illegal.

bcalidas: At the WG discussion,  there was a suggestion that semaphores could hold an additional internal reference count when in pending signal mode and release that internal reference count when the semaphore had signaled. This would mean that an application calling clReleaseSemaphoreKHR on a semaphore that had a pending signal would become disconneted from the semaphore, but the semaphore object itself would be valid until the signal happened. This in turn would mean that it was safe for the application to call clReleaseSemaphoreKHR on a sempahore with a pending signal.

Our analysis has shown that it is not possible for an implementation to know in all cases if a semaphore can a pending signal. For example, an application could import a semaphore with a pending signal using the opaque fd handle type. In this case, the application could not reliably infer the state of the semaphore at the time of import. If the application were to then immediately call, clReleaseSemaphoreKHR, the implementation would have to release the semaphore  object.

Our recommendation is to add language to the specification that in such cases the behavior of the released semaphore  is implementation defined, but that there is no effect on other semaphores,  semaphores that will be created later  or the OpenCL context.

nikhiljnv: As discussed on memory subgroup call on October 3, need more thinking to analyze different cases for external semaphores.

bcalidas: On further analysis from our side , it seems that there are no obvious bad effects from releasing a semaphore even if has a pending signal.  At the same time, the behavior that the application can expect if it calls clReleaseSemaphoreKHR on a semaphore  is fairly intuitive across the different semaphore states. we may not need to explicitly document the behavior of clReleaseSeamporeKHR for corner cases.

nikhiljnv:
As discussed in Memory subgroup call on October 10, 2023, decided not to make any spec edits around this.
Closing the PR.

PR #970, opened 2023-09-15
External memory import of dedicated/non-dedicated memory

joshqti:
Specifies when OpenCL becomes aware of imported image memory layout metadata.  This is meant to clarify how OpenCL interacts with dedicated memory.

### DIFF
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -211,7 +211,7 @@ If {CL_MEM_DEVICE_HANDLE_LIST_KHR} is not specified as part of _properties_, the
 
 The properties used to create a buffer or image from an external memory handle are described by related extensions.
 When a buffer or image is created from an external memory handle, the _flags_ used to specify usage information for the buffer or image must not include {CL_MEM_USE_HOST_PTR}, {CL_MEM_ALLOC_HOST_PTR}, or {CL_MEM_COPY_HOST_PTR}, and the _host_ptr_ argument must be `NULL`.
-For images created from an external memory handle, implementations will acquire information about image attributes such as format and layout at the time of creation and continue to apply those attributes for the lifetime of the image object.
+For images created from an external memory handle, implementations may acquire information about image attributes such as format and layout at the time of creation and continue to apply those attributes for the lifetime of the image object.
 
 Add to the list of error conditions for {clCreateBufferWithProperties} and {clCreateImageWithProperties}:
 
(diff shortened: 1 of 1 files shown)


ext/cl_khr_external_memory.asciidoc
```diff
@@ -211,6 +211,7 @@ If {CL_MEM_DEVICE_HANDLE_LIST_KHR} is not specified as part of _properties_, the
 
 The properties used to create a buffer or image from an external memory handle are described by related extensions.
 When a buffer or image is created from an external memory handle, the _flags_ used to specify usage information for the buffer or image must not include {CL_MEM_USE_HOST_PTR}, {CL_MEM_ALLOC_HOST_PTR}, or {CL_MEM_COPY_HOST_PTR}, and the _host_ptr_ argument must be `NULL`.
+For images created from an external memory handle, implementations will acquire information about image attributes such as format and layout at the time of creation and continue to apply those attributes for the lifetime of the image object.
```
nikhiljnv, line 214: This issue may affect buffers created from external memory handle as well. Also, the statement in a way is assuming kind of inference at the time of creation which may not be necessary/applicable for all implementations. 

How about something like this -
"For cl_mem created from an external memory handle, the memory attributes like layout, format etc will be borrowed from external memory handle properties and the information provided by the application at the time of creating such a cl_mem. The same shall continue to apply throughout the lifetime of the cl_mem. Any changes to external memory properties happened outside the scope of defined OpenCL APIs will not be reflected in OpenCL's view of the cl_mem. A new cl_mem must be created by re-importing the external memory handle with appropriate properties in order to see these changes reflected in OpenCL."

nikhiljnv (changes requested):
Left some thoughts/comments.

ext/cl_khr_external_memory.asciidoc
kpet, line 214: Suggest s/will/may/ as acquiring the information is not mandatory. The second part of the sentence could be tweaked accordingly, maybe:

>When images are created from an external memory handle, implementations may acquire information about image attributes such as format and layout at the time of creation. When such information is acquired at image creation time, it is used as  for the lifetime of the image object.

kpet (commented):
Looks generally good, I've suggested a few language tweaks.

ext/cl_khr_external_memory.asciidoc
joshqti: I agree, and have made the requested change.

Thanks,
Joshua Kelly, Qualcomm

kpet (approved):
Thanks!

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as discussed on Memory subgroup call on October 3.

PR #974, opened 2023-09-29
document USM error conditions for clSetKernelExecInfo

bashbaug:
Adds a few error conditions to the `cl_intel_unified_shared_memory` extension.

Specifically, describes expected error behavior for clSetKernelExecInfo when devices in a context do not support different types of USM, or do not support USM at all.


alycm (approved):
LGTM, merging.

PR #976, opened 2023-10-03
Remove text relaxing core semaphore requirement for external semaphores

### LINKED ISSUE #944: Clarify dependencies between cl_khr_semaphore and cl_khr_external_semaphore
cl_khr_semaphore provides basic support for semaphore creation and wait/signal operations on them.
cl_khr_external_semaphore provides mechanism to import semaphores into OpenCL and export semaphores from OpenCL.
While being able to export semaphores from OpenCL requires OpenCL to be able to create semaphores and for some semaphore handle types like SYNC_FD, semaphores inherently require both import and export functionalities to work properly, for some other semaphore handle types (like OPAQUE_FD and WIN32) semaphores can be independently imported and/or exported. 
So, for implementations just willing to support importing semaphores in OpenCL, semaphore creation is strictly not required.

cl_khr_external_semaphore depends on cl_khr_semaphore, but [cl_khr_external_semaphore spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_external_semaphore) says (largely to facilitate import-only scenario described above)

_clCreateSemaphoreWithPropertiesKHR may return a NULL value on some implementations if sema_props does not contain an external semaphore handle type to import. Such implementations are required to return a valid semaphore when a supported external memory handle type and valid external semaphore handle is specified._

This makes cl_khr_semaphore base functional guarantees ambiguous as cl_khr_external_semaphore leads to subtracting functionality provided by required cl_khr_semaphore.

This issue is to track any spec / CTS changes needed to remove inconsistencies in the descriptions.

nikhiljnv:
cl_khr_external_semaphore requires cl_khr_semaphore and cl_khr_semaphore requires support for base semaphores. However, cl_khr_external_semaphore allows relaxing core cl_khr_semaphore requirement to support base semaphores if importing external semaphores is supported.

Clean up the text related to above and make it mandatory to support core semaphores to honor dependency of
cl_khr_semaphore.

Fixes #944

### DIFF
--- a/ext/cl_khr_external_semaphore.asciidoc
+++ b/ext/cl_khr_external_semaphore.asciidoc
@@ -219,9 +219,6 @@ Add to the list of error conditions for {clCreateSemaphoreWithPropertiesKHR}:
 
 * {CL_INVALID_DEVICE} if one or more devices identified by properties {CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR} can not import the requested external semaphore handle type.
 
-{clCreateSemaphoreWithPropertiesKHR} may return a NULL value on some implementations if _sema_props_ does not contain an external semaphore handle type to import.
-Such implementations are required to return a valid semaphore when a supported external memory handle type and valid external semaphore handle is specified.
-
 Add to the list of supported _param_names_ by {clGetSemaphoreInfoKHR}:
 
 .List of supported param_names by {clGetSemaphoreInfoKHR}


bashbaug (approved):
LGTM, thanks!

bashbaug:
Merging as discussed in the October 3rd memory subgroup.

PR #979, opened 2023-10-16
add an error condition when unmapping a SVM region that was not mapped

### LINKED ISSUE #890: clEnqueueSVMUnmap
about the function clEnqueueSVMUnmap , what happens if we call this function without calling clEnqueueSVMMap first? this is not well-defined in openCL 3.0 Reference Page (spec)

bashbaug:
fixes #890 

Adds an error condition (`CL_INVALID_VALUE`) when attempting to unmap a SVM region that was not mapped.


alycm (approved):
LGTM.

PR #980, opened 2023-10-17
allow clSetCommandQueueProperty to return an error for non-OpenCL 1.0 devices

### LINKED ISSUE #621: Status of clSetCommandQueueProperty
See: https://github.com/KhronosGroup/OpenCL-CTS/pull/1182

The current OpenCL spec says that `clSetCommandQueueProperty` is _deprecated_, where the glossary definition of deprecated is:

> Existing features are marked as deprecated if their usage is not recommended as that feature is being de-emphasized, superseded and may be removed from a future version of the specification.

This was an addition to the OpenCL 3.0 spec.  Prior to OpenCL 3.0, the last version of the OpenCL spec to include `clSetCommandQueueProperty` was the OpenCL 1.0 spec, and the API was removed from the OpenCL 1.1 spec with the comment:

> The following features in OpenCL 1.0 are deprecated:
> * The clSetCommandQueueProperty API is no longer supported in OpenCL 1.1.

The current OpenCL headers have the comment:

https://github.com/KhronosGroup/OpenCL-Headers/blob/v2021.04.29/CL/cl.h#L1845

```c
#ifdef CL_USE_DEPRECATED_OPENCL_1_0_APIS
    /*
     *  WARNING:
     *     This API introduces mutable state into the OpenCL implementation. It has been REMOVED
     *  to better facilitate thread safety.  The 1.0 API is not thread safe. It is not tested by the
     *  OpenCL 1.1 conformance test, and consequently may not work or may not work dependably.
     *  It is likely to be non-performant. Use of this API is not advised. Use at your own risk.
     *
     *  Software developers previously relying on this API are instructed to set the command queue
     *  properties when creating the queue, instead.
     */
    extern CL_API_ENTRY cl_int CL_API_CALL
    clSetCommandQueueProperty(cl_command_queue              command_queue,
                              cl_command_queue_properties   properties,
                              cl_bool                       enable,
                              cl_command_queue_properties * old_properties) CL_API_SUFFIX__VERSION_1_0_DEPRECATED;
#endif /* CL_USE_DEPRECATED_OPENCL_1_0_APIS */
```

What is the status of `clSetCommandQueueProperty`?

* Is it deprecated using the definition described in the spec glossary, in which case it should be tested and functional even if its use is not recommended?
* Or, is it removed or otherwise unsupported, in which case it may not be functional?
    * If this is the case, what level of CTS testing is appropriate for this API?

bashbaug:
fixes #621

This follows the same pattern and text used by clSetProgramReleaseCallback, which may return an error for non-OpenCL 2.2 devices.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -428,25 +428,28 @@ include::{generated}/api/version-notes/clSetCommandQueueProperty.asciidoc[]
     changed by {clSetCommandQueueProperty}.  If _old_properties_ is `NULL`, it
     is ignored.
 
-[NOTE]
-====
-Changing the {CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE} command-queue property
-will cause the OpenCL implementation to block until all previously queued
-commands in _command_queue_ have completed.  This can be an expensive operation
-and therefore changes to this property should only be done when absolutely
-necessary.
-====
-
 // refError
+
+{clSetCommandQueueProperty} may unconditionally return an error if no
+devices in the context associated with _command_queue_ support modifying
+the properties of a command-queue.
+Support for modifying the properties of a command-queue is required only for
+OpenCL 1.0 devices.
 
 {clSetCommandQueueProperty} returns {CL_SUCCESS} if the function is executed
 successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_COMMAND_QUEUE} if _command_queue_ is not a valid command-queue.
+  * {CL_INVALID_OPERATION} if no devices in the context associated with
+    _command_queue_ support modifying the properties of a command-queue.
   * {CL_INVALID_VALUE} if values specified in _properties_ are not valid.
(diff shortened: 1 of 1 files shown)


alycm (approved):
Returning an error is consistent with `clSetProgramReleaseCallback`.

I had a look at a few implementations and they generally return an error, or have some level of implementation that can return `CL_SUCCESS`.

The auto-generated version note still says "[clSetCommandQueueProperty](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetCommandQueueProperty) is [deprecated by](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#unified-spec) version 1.1." but I'm not immediately sure if `cl.xml` or the scripts around it have a concept of removal.  I'd have to look at the XML for some other specifications.

I don't think the the previous paragraph is a blocker to merging this PR.

bashbaug:
@kpet did you want to review this before merging?

kpet (approved):
LGTM

PR #982, opened 2023-10-17
fix version note links in the reference pages

### LINKED ISSUE #726: Reference page "missing before" style notes link to nowhere
The "missing before" and "deprecated by" version notes in the reference pages are links, but currently point to an anchor that does not exist on the reference page.

Possible options:
* Rewrite the link to the [relevant section](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#unified-spec) of the API specification (where the link would go in the specification itself).
* Have the text without it being a link.

The steps to take for both options are probably identical, the scripts generating the pages need to adjust the link.

bashbaug:
fixes #726 

This feels like a bit of a hack, but it's working, and the same possible solution was proposed on the issue also so maybe it's not _that_ bad?

This fixes the version note links by adding a config file defining asciidoctor attributes for "missing before" and "deprecated by" with two variants:

1. The first variant has local links that work within the same document and is included by the main OpenCL API spec.
2. The second variant has full links that will work anywhere and is included by the reference pages.

### DIFF
--- a/config/version-full-links.asciidoc
+++ b/config/version-full-links.asciidoc
@@ -2,7 +2,7 @@
 // Creative Commons Attribution 4.0 International License; see
 // http://creativecommons.org/licenses/by/4.0/
 
-// Attributes for version notes, with local links.
+// Attributes for version notes, with full links.
 
 // "missing before"
 :missing_before_label: pass:q[missing before]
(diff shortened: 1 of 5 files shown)


config/version-full-links.asciidoc
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023 The Khronos Group. This work is licensed under a
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+// Attributes for version notes, with local links.
```
alycm, line 5:
```suggestion
// Attributes for version notes, with full links.
```

alycm (approved):
LGTM, I honestly can't think of a tidier way to do this than the double config files.

PR #983, opened 2023-10-18
Fix broken refpage links for CL_VERSION_X_Y macros

oddhack:
Aliases them all to the preprocessorDirectives page.

This will need to be tried out on a running Apache server since it amounts to modifying .htaccess to do rewrites of the aliases.


bashbaug (approved):
Thanks, I'm not sure how I missed this!

I still think we should consider #984, since these CL_VERSION macros do not appear in the spec (they're only in the headers), but at least this way they'll link to something if we do decide to keep them.

PR #984, opened 2023-10-18
filter CL_VERSION_X_Y macros from reference page see also

### LINKED ISSUE #981: broken refpage links for CL_VERSION_X_Y macros
The refpage links for the `CL_VERSION_X_Y` macros in the "see also" sections all appear to be broken.

Example for [clCreateBuffer](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clCreateBuffer.html#_see_also) and [CL_VERSION_1_0](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/CL_VERSION_1_0.html) (the first link works, the second doesn't).

I suspect this should redirect to [preprocessorDirectives](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/preprocessorDirectives.html) but I'm not quite sure what magic is needed to make this happen.

@oddhack might be able to help...

bashbaug:
fixes #981 

Instead of trying to make the CL_VERSION_X_Y macros link to the right page, just filter them from the "see also" list, because they aren't adding much value.

I tried to make this somewhat general if we decide to filter other references later, also.  I'm open to suggestions how to do this better though, if desired.

### DIFF
--- a/scripts/genRef.py
+++ b/scripts/genRef.py
@@ -103,13 +103,6 @@ def macroPrefix(name):
     return 'reflink:' + name
 
 
-def filterRef(name):
-    # Do not include CL_VERSION_X_Y macro references
-    if name.startswith('CL_VERSION_'):
-        return True
-    return False
-
-
 def seeAlsoList(apiName, explicitRefs=None, apiAliases=[]):
     """Return an asciidoc string with a list of 'See Also' references for the
     API entity 'apiName', based on the relationship mapping in the api module.
@@ -146,8 +139,8 @@ def seeAlsoList(apiName, explicitRefs=None, apiAliases=[]):
                 if dependency is not None:
                     refs.add(dependency)
 
-    # Remove CL_VERSION_X_Y macro references
-    refs = set([ref for ref in refs if not filterRef(ref)])
+    # Filter references based on API conventions
+    refs = set([ref for ref in refs if not conventions.filter_ref(ref)])
 
     if len(refs) == 0:
         return None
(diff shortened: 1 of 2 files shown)


oddhack:
Did you look at #983 as an alternative?

If doing this I suggest punting the actual exclusion list into a method of clconventions.py

bashbaug:
> If doing this I suggest punting the actual exclusion list into a method of clconventions.py

I moved the exclusion criteria to clconventions.py, if we do end up deciding to do this.

scripts/genRef.py
```diff
@@ -139,6 +139,9 @@ def seeAlsoList(apiName, explicitRefs=None, apiAliases=[]):
                 if dependency is not None:
                     refs.add(dependency)
 
+    # Filter references based on API conventions
+    refs = set([ref for ref in refs if not conventions.filter_ref(ref)])
+
```
oddhack, line 144: I think you could also, slightly more pythonically, do

> refs = set(filter(lambda ref: conventions.filter_ref(ref), refs))

(which I just learned about on asking "hm, seems like there should be a builtin for this" :-).

oddhack (commented):
I don't have any issues with this codewise, but think the issue has been dealt with by the other PR so am slightly against including this.

bashbaug:
I'll convert this to a draft for now.  We can resurrect it after the next release if the linking still seems weird, or we can close it if we're happy with the updated solution.  Thanks!

bashbaug:
I don't think this is needed anymore - closing.

PR #985, opened 2023-10-18
add old command queue APIs to the reference page table of contents

bashbaug:
Found while working on #980.  The older clCreateCommandQueue and clSetCommandQueueProperties APIs have generated HTML pages but they weren't appearing in the TOC.  This changes adds them to the TOC so they are easier to find (even if these APIs are deprecated and their use is discouraged).


alycm (approved):
LGTM, merging as obvious.

PR #986, opened 2023-10-19
add rewrite rule for cl_mem_properties

bashbaug:
Adds a redirect from the nonexistent `cl_mem_properties.html` to the `clCreateBuffer.html` page, which also includes `clCreateBufferWithProperties` and hence is as good of a page for `cl_mem_properties` as any.

With #984 I believe this fixes all of the broken links in the OpenCL reference pages.


oddhack (approved):
LGTM

PR #987, opened 2023-10-19
fix the link appearance in the online reference pages

bashbaug:
Most of the links in the online reference pages are formed by parsing the asciidoctor source to find `<<anchor,link text>>` cross references and then converting them to HTML `<a href="spec_url#anchor">link text</a>` links instead.  This works great if there is link text in the cross references, but it doesn't work well when not using [explicit link text](https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#explicit-link-text).

An example showing the problem can be found here:
https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetPlatformInfo.html

The link in table 1 is not broken, but it is very long and shows the explicit URL.

This PR either switches the asciidoctor cross references to use explicit link text, or uses the asciidoctor dictionaries for APIs and enums instead.  This produces nice, short links instead.


oddhack:
LGTM. The refpage extractor goes to some lengths to rewrite xrefs into links into the specification. In the Vulkan spec all the API names are tagged as asciidoctor macros that are resolved at runtime, which is equivalent to what you're doing here except you're resolving them in a preprocessing step to set the attributes.

N.b. the refpage extractor is not currently clever enough to realize when an xref is local to the refpage, e.g. the "Platform Queries" links from the example all go to the main spec even though the table is imbedded in the clGetPlatformInfo refpage. It doesn't affect more than handful of xrefs in the Vulkan spec, but OpenCL is more liberal with imbedding tables so it comes up more often. I should take another look at this.

alycm (approved):
LGTM.

bashbaug:
Merging as discussed in the December 5th teleconference.

PR #991, opened 2023-11-06
Use hexapdf instead of ghostscript for PDF optimization

oddhack:
Resulting PDFs tend to be considerably smaller, and also runs about 15% faster when doing a full PDF build (2:39 vs. 3:06 on my machine).

The hexapdf tool does need to be installed in the build environment - it is in the khronosgroup/docker-images:asciidoctor-spec Docker image.


bashbaug (approved):
Nice, the generated PDFs using hexapdf are considerably smaller and still look great.  Thanks!

PR #992, opened 2023-11-06
Update cl_khr_command_buffer_mutable_dispatch.asciidoc

aharon-abramson:
Add type cl_mutable_dispatch_promises_khr and its possible values


ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -233,6 +236,7 @@ CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR        0x12B0
 
 // Property to cl_ndrange_kernel_command_properties_khr
 CL_MUTABLE_DISPATCH_UPDATABLE_FIELDS_KHR           0x12B1
+CL_MUTABLE_DISPATCH_PROMISES_KHR                   0x12B2
```
EwanC, line 239: The ``0x12B2` value is already taken https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml#L1777 , `0x12B7` looks like the next available value to use

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
EwanC, line 1: We should bump the spec version number to `0.9.1` for this change

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -101,6 +102,9 @@ typedef cl_uint cl_mutable_command_info_khr;
 
 // Identifies the type of a structure to allow structure pointer chains
 typedef cl_uint cl_command_buffer_structure_type_khr;
+
+// Bitfield covering certain promises by the user to the implementation, enabling possible optimizations 
+typedef cl_bitfield cl_mutable_dispatch_promises_khr;
```
EwanC, line 107: We need to define this in the XML, for example https://github.com/KhronosGroup/OpenCL-Docs/blob/c37e00f1423379a882c173c12ed1aa5ca9d8e51d/xml/cl.xml#L254

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -20,6 +20,7 @@ commands between command-buffer enqueues.
 |====
 | *Date*     | *Version* | *Description*
 | 2022-08-31 | 0.9.0     | First assigned version (provisional).
+| 2023-11-07 | 0.9.1     | Add type cl_mutable_dispatch_promises_khr and its possible values
```
EwanC, line 23:
```suggestion
| 2023-11-07 | 0.9.1     | Add type cl_mutable_dispatch_promises_khr and its possible values (provisional).
```

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -233,6 +237,7 @@ CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR        0x12B0
 
 // Property to cl_ndrange_kernel_command_properties_khr
 CL_MUTABLE_DISPATCH_UPDATABLE_FIELDS_KHR           0x12B1
+CL_MUTABLE_DISPATCH_PROMISES_KHR                   0x12B7
```
EwanC, line 240: This needs added to the XML file too, https://github.com/KhronosGroup/OpenCL-Docs/blob/c37e00f1423379a882c173c12ed1aa5ca9d8e51d/xml/cl.xml#L1782

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -254,6 +259,9 @@ CL_MUTABLE_COMMAND_COMMAND_TYPE_KHR                0x12AD
 
 // Bits for cl_command_buffer_flags_khr
 CL_COMMAND_BUFFER_MUTABLE_KHR                    (0x1 << 1)
+
+// Bits for cl_mutable_dispatch_promises_khr bitfield
+CL_MUTABLE_DISPATCH_PROMISE_NO_ADDITIONAL_WORK_GROUPS_KHR	(0x1 << 0)
```
EwanC, line 264: Should also be added to XML file
```
 <enums name="cl_mutable_dispatch_promises_khr" vendor="Khronos" type="bitmask">
        <enum bitpos="0"            name="CL_MUTABLE_DISPATCH_PROMISE_NO_ADDITIONAL_WORK_GROUPS_KHR"/>
         <unused start="1" end="31"/>
 </enums>
```

The XML definition of the extension should also be updated to reflect these new XML additions https://github.com/KhronosGroup/OpenCL-Docs/blob/c37e00f1423379a882c173c12ed1aa5ca9d8e51d/xml/cl.xml#L7268

EwanC (commented):
Left some comments about reflecting these changes in cl.xml, you can check these have been made correctly by running the OpenCL-Headers generation script https://github.com/KhronosGroup/OpenCL-Headers/tree/main/scripts#generating-headers 

This is also the best way to create the patch for an OpenCL-Headers PR to match this change.

xml/cl.xml
```diff
@@ -1779,7 +1783,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x12B4"        name="CL_IMAGE_REQUIREMENTS_MAX_HEIGHT_EXT"/>
         <enum value="0x12B5"        name="CL_IMAGE_REQUIREMENTS_MAX_DEPTH_EXT"/>
         <enum value="0x12B6"        name="CL_IMAGE_REQUIREMENTS_MAX_ARRAY_SIZE_EXT"/>
-            <unused start="0x12B7" end="0x1FFF" comment="Reserved for core API tokens"/>
+        <enum value="0x12B7"		name="CL_MUTABLE_DISPATCH_PROMISES_KHR"/>
```
SunSerega, line 1786: Tabs got placed instead of spaces:
``` suggestion
        <enum value="0x12B7"        name="CL_MUTABLE_DISPATCH_PROMISES_KHR"/>
```

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: It is defined - in line 255.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: line 1786

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: CL_MUTABLE_DISPATCH_PROMISE_NO_ADDITIONAL_WORK_GROUPS_KHR is defined in line 1348.

EwanC (commented):
This spec language is using the terminology "asserts", but all the XML changes uses the terminology "promises", we should use one or the other.

xml/cl.xml
```diff
@@ -7313,6 +7319,9 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR"/>
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR"/>
             </require>
```
EwanC, line 7321: We need to pull in the definition of `CL_MUTABLE_DISPATCH_ASSERTS_KHR` too
```suggestion
            </require>
            <require comment="cl_mutable_dispatch_asserts_khr">
                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
```

EwanC:
OpenCL-Headers PR with changes generated for cl.xml here - https://github.com/KhronosGroup/OpenCL-Headers/pull/243

aharon-abramson:
@EwanC, do you approve of this PR?

xml/cl.xml
```diff
@@ -1343,6 +1344,9 @@ server's OpenCL/api-docs repository.
         <enum bitpos="4"            name="CL_MUTABLE_DISPATCH_EXEC_INFO_KHR"/>
             <unused start="5" end="31"/>
     </enums>
+    <enums name="cl_mutable_dispatch_asserts_khr" vendor="Khronos" type="bitmask">
+    	<enum bitpos="0"            name="CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR"/>
```
SunSerega, line 1348:
``` suggestion
        <enum bitpos="0"            name="CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR"/>
```
And another tab in the `<require>` section.

xml/cl.xml
```diff
@@ -7313,6 +7319,12 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR"/>
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR"/>
             </require>
+            <require comment="cl_mutable_dispatch_asserts_khr">
+                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
+            </require>
+            <require comment="cl_mutable_dispatch_asserts_khr - bitfield">
+            	<enum name="CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR"/>
```
SunSerega, line 7326:
``` suggestion
                <enum name="CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR"/>
```

xml/checklinks.py
EwanC, line 1: There's a lot of files under the `xml` directory being touched here, I think only the modifications to `xml/cl.xml` should be committed.

aharon-abramson: You are right. I'll remove them.

xml/cl.xml
```diff
@@ -1779,7 +1783,8 @@ server's OpenCL/api-docs repository.
         <enum value="0x12B4"        name="CL_IMAGE_REQUIREMENTS_MAX_HEIGHT_EXT"/>
         <enum value="0x12B5"        name="CL_IMAGE_REQUIREMENTS_MAX_DEPTH_EXT"/>
         <enum value="0x12B6"        name="CL_IMAGE_REQUIREMENTS_MAX_ARRAY_SIZE_EXT"/>
-            <unused start="0x12B7" end="0x1FFF" comment="Reserved for core API tokens"/>
+        <enum value="0x12B7"        name="CL_MUTABLE_DISPATCH_PROMISES_KHR"/>
```
EwanC, line 1786: The spec says `CL_MUTABLE_DISPATCH_ASSERTS_KHR` not `CL_MUTABLE_DISPATCH_PROMISES_KHR`

xml/cl.xml
```diff
@@ -252,6 +252,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_fp_atomic_capabilities_ext</name>;</type>
         <type category="define">typedef <type>cl_uint</type>          <name>cl_image_requirements_info_ext</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_platform_command_buffer_capabilities_khr</name>;</type>
+        <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mutable_dispatch_promises_khr</name></type>
```
EwanC, line 255: spec says `cl_mutable_dispatch_asserts_khr` not `cl_mutable_dispatch_promises_khr`

EwanC (approved):
LGTM

oddhack:
@aharon-abramson @bashbaug @ntrevett this is one of the last couple of extensions needing integration into the core spec markup in my working draft #950, and there are quite a few changes happening here. Is it likely this will be signed off soon? It's unfortunate to be integrating extensions undergoing changes, as any changes that are done in the separate extension doc will have to be reflected into the core spec additions as well.

aharon-abramson:
Balaji requested another week to review it. Hopefully, we'll be able to merge it next week.

oddhack:
> Balaji requested another week to review it. Hopefully, we'll be able to merge it next week.

Thanks. A clarification: the extension language modifies the description of mutable_handle and the CL_INVALID_OPERATION error, but I'm not clear as to whether these modifications only apply to clCommandNDRangeKernelKHR or more broadly to other commands taking mutable_handle parameters? A literal reading of the extension suggests the former. But I just went through integrating the cl_khr_command_buffer_multiple_device extension language, which makes similar edits to a parameter & errors of (all? most of?) the command recording commands, so wanted to check.

oddhack:
One of the edits here is to replace the error

> * {CL_INVALID_OPERATION} if _mutable_handle_ is not `NULL`.

However, AFAICT the corresponding error in the cl_khr_command_buffer extension is CL_INVALID_VALUE, not CL_INVALID_OPERATION, so this may be a typo.

Also ISTM this isn't "replacing" this error (with something else related to _mutable_handle_) so much as removing it and adding a new, unrelated error, so I'd express it that way.

aharon-abramson:
> > Balaji requested another week to review it. Hopefully, we'll be able to merge it next week.
> 
> Thanks. A clarification: the extension language modifies the description of mutable_handle and the CL_INVALID_OPERATION error, but I'm not clear as to whether these modifications only apply to clCommandNDRangeKernelKHR or more broadly to other commands taking mutable_handle parameters? A literal reading of the extension suggests the former. But I just went through integrating the cl_khr_command_buffer_multiple_device extension language, which makes similar edits to a parameter & errors of (all? most of?) the command recording commands, so wanted to check.

Yes, it relates to the former.

aharon-abramson:
> One of the edits here is to replace the error
> 
> > * {CL_INVALID_OPERATION} if _mutable_handle_ is not `NULL`.
> 
> However, AFAICT the corresponding error in the cl_khr_command_buffer extension is CL_INVALID_VALUE, not CL_INVALID_OPERATION, so this may be a typo.
> 
> Also ISTM this isn't "replacing" this error (with something else related to _mutable_handle_) so much as removing it and adding a new, unrelated error, so I'd express it that way.

This isn't part of my change. Where do you see it in the diff?

oddhack:
> > One of the edits here is to replace the error
> > > * {CL_INVALID_OPERATION} if _mutable_handle_ is not `NULL`.
> > 
> > 
> > However, AFAICT the corresponding error in the cl_khr_command_buffer extension is CL_INVALID_VALUE, not CL_INVALID_OPERATION, so this may be a typo.
> > Also ISTM this isn't "replacing" this error (with something else related to _mutable_handle_) so much as removing it and adding a new, unrelated error, so I'd express it that way.
> 
> This isn't part of my change. Where do you see it in the diff?

Sorry, I was unclear - it is in the main branch version of the extension. I am just asking questions here because it seems like an opportune place to clear up minor issues vs. opening a new issue, but if you don't want to engage on it here that's fine and I'll open a new issue.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
EwanC, line 1: At the end of this document in the Issues section we have an entry for.
```asciidoc
Introduce a `CL_MUTABLE_DISPATCH_ADDITIONAL_WORK_GROUPS_KHR` capability to
allow the number of work-groups in kernel execution to be increased during
update.
+
--
*Resolved*: Can be included in the final release of the extension if there is
implementation coverage.
--
```

I feel like we've progressed this in this PR, and the entry can be removed

aharon-abramson: done

bcalidas:
After careful review, we feel that the no_additional workgroups assert feels more like a vendor specific feature. We would not be able to make use of it, for example. In addition checking for assert violation would be expensive and reduces a major advantage of using command buffers.

The typical sequence that a developer would follow on our HW is -

Record Command Buffer
UpdateCommandBuffer
EnqueueCommandBuffer
UpdateCommandBuffer
EnqueueCommandBuffer
UpdateCommandBuffer
EnqueueCommandBuffer....

Our suggestion is to add a DeviceInfo query for which mutable command buffer asserts are supported. If an implementation returns 0 it means that it does not support any asserts.  For example CL_DEVICE_MUTABLE_COMMAND_BUFFERS_SUPPROTED_ASSERTS.

aharon-abramson:
@bcalidas, if instead of returning an error, we specify that the behavior is undefined, will it be good enough?

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -233,6 +237,7 @@ CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR        0x12B0
 
 // Property to cl_ndrange_kernel_command_properties_khr
 CL_MUTABLE_DISPATCH_UPDATABLE_FIELDS_KHR           0x12B1
+CL_MUTABLE_DISPATCH_ASSERTS_KHR                    0x12B7
```
bashbaug, line 240: This is actually used for two purposes right now:

1. It can be passed as part of the `cl_command_buffer_properties_khr` during command buffer creation to specify properties for the entire command buffer.
2. It can be passed as part of the `cl_ndrange_kernel_command_properties_khr` to specify properties for the specific ND-range kernel recorded into the command buffer.

Would it be better to define two different properties for these two different usages?

Note that the current naming is consistent with (2).  To be consistent with the naming for (1) I think the property should begin with `CL_COMMAND_BUFFER...`, like `CL_COMMAND_BUFFER_FLAGS_KHR`.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -533,9 +561,9 @@ the array violates the defined conditions:
 * {CL_INVALID_VALUE} if _type_ is not
   {CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR}.
 
-* {CL_INVALID_OPERATION} if values of _local_work_size_ and/or
-  _global_work_size_ result in an increase to the number of work-groups in the
-  ND-range.
+* {CL_INVALID_OPERATION} if {CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR} is specified, 
+  and values of _local_work_size_ and/or _global_work_size_ result in an increase to the number of work-
+  groups in the ND-range over the number specified when the ND-range kernel was recorded.
```
bashbaug, line 566: What is the expected behavior if a command buffer or command specifies `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR`, the local work size is initially non-NULL, and then the updated local work size **is** NULL?

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: I agree.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: Hmm, I haven't thought of that. I think that, in this case, we should allow this, and the implementation should be able to handle it.

bashbaug: Hmm, perhaps there are two scenarios to consider:

1. The global work size is not changing, but the local work size is changing to NULL.  I this case I suppose that the implementation could handle it, although it may need to choose a different local work size than it would otherwise, subject to the "no additional work-groups" constraint.

2. The global work size is changing, and the local work size is also changing to NULL.  I think it would be very difficult to handle this in a general case, though, and there may be some global work sizes that satisfy the "no additional work-groups" constraint.
    * Note that this is true even if the global work size gets smaller, especially if the work-group size must be uniform; for example, consider changing from a very large power-of-two (with lots of prime factors and options for work-group sizes) to a slightly smaller prime number (with few work-group size options).

I see that the latest updates added the line:

```
If the user updates _local_work_size_ to be _NULL_, the assertion is ignored.
```

We may want to go a bit stronger though, and either make this an error (assuming we're checking errors), or undefined behavior (if we aren't defining errors).

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
bashbaug: > I agree, but is there a way to reuse the new definitions for both usages?

Sure, I don't see why not: The two properties would be different so we would have one property for each purpose, but the value of each property could be the same since they are both defining mutable dispatch assertions.

bashbaug: More detail: I propose that we split `CL_MUTABLE_DISPATCH_ASSERTS_KHR` into two properties.  One of these properties can re-use the value `0x12B7` currently assigned to `CL_MUTABLE_DISPATCH_ASSERTS_KHR` - I don't particularly care which one:

```c++
// This is a property that may be specified when the command buffer is created.
// It specifies mutable dispatch assertions for the entire command buffer.

/* cl_command_buffer_properties_khr */
#define CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR   0xXXXX

// This is a property that may be specified for a single ND-range kernel command.
// It specifies mutable dispatch assertions for that ND-range kernel command only.

/* cl_ndrange_kernel_command_properties_khr */
#define CL_MUTABLE_DISPATCH_ASSERTS_KHR                  0xYYYY

// The value for both properties is the same and is a bitfield of mutable dispatch
// assertions.  We will define one bit in the bitfield currently, to assert that no
// additional work-groups will be required.

typedef cl_bitfield         cl_mutable_dispatch_asserts_khr;

/* cl_mutable_dispatch_asserts_khr - bitfield */
#define CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR (1 << 0)
```

xml/cl.xml
```diff
@@ -1779,7 +1783,9 @@ server's OpenCL/api-docs repository.
         <enum value="0x12B4"        name="CL_IMAGE_REQUIREMENTS_MAX_HEIGHT_EXT"/>
         <enum value="0x12B5"        name="CL_IMAGE_REQUIREMENTS_MAX_DEPTH_EXT"/>
         <enum value="0x12B6"        name="CL_IMAGE_REQUIREMENTS_MAX_ARRAY_SIZE_EXT"/>
-            <unused start="0x12B7" end="0x1FFF" comment="Reserved for core API tokens"/>
+        <enum value="0x12B7"        name="CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR"/>
```
bashbaug, line 1786: This new enum needs to be added to a `<require>` block below, also, so it will be included in the generated headers.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
bashbaug: Sorry about the false alarm, this may not actually be an issue.

In the current mutable dispatch extension we say (**emphasis** mine):

> local_work_size Points to an array of work_dim unsigned values that describe the number of work-items that make up a work-group that will execute the kernel. **If local_work_size is NULL then the number of local work-items in the dispatch is not changed.** See clEnqueueNDRangeKernel for valid usage.

This means that it is not possible to change from a non-NULL local work size to a NULL local work size, because the NULL local work size will be interpreted as a request to keep the local work size unchanged.  Therefore, the case where the "user updates _local_work_size_ to be _NULL_" will never happen, and we don't need to include this case in the spec, although I suppose we could include it in a footnote or an informative note, if we so choose.

The situations we need to consider are for the assertion are:

1. Non-NULL initial local work size, non-NULL updated local work size.  In this case the updated work-group count is computed from the updated local work size and must be less than or equal to the initial work-group count or behavior is undefined.
2. Non-NULL initial local work size and "NULL" updated local work size.  In this case, the work-group count is computed from the initial local work size (since the "NULL" local work size indicates that it is unchanged), and the updated work-group count must be less than or equal to the initial work-group count or behavior is undefined.
3. NULL initial local work size.  For this case, the PR currently says "In case _local_work_size_ is NULL, the effect of this flag is undefined", so I suppose behavior is undefined in all of these scenarios.  Can we do any better?  Some possibilities:
    * The assertion is ignored when the local work size is NULL.
    * The assertion is invalid when the local work size is NULL.  Note, this condition would be checked when the command buffer is initially recorded, not whenever it is modified, so error checking should be possible.  (I prefer this option.)
    * The initial work-group count is computed assuming a worst-case work-group size of one.  (This seems a little too clever.).

Does this sound correct?

aharon-abramson: You are right; I also forgot about this part of the spec.
I, too, think that the 2nd option is preferable. I'll update the PR accordingly.

xml/cl.xml
```diff
@@ -7313,6 +7321,12 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR"/>
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR"/>
             </require>
+            <require comment="cl_mutable_dispatch_asserts_khr">
+                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
+            </require>
```
SunSerega, line 7326:
```suggestion
            <require comment="cl_ndrange_kernel_command_properties_khr">
                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
```
`comment` denotes the type, which accepts this enum. I guess this is the parameter:
https://github.com/KhronosGroup/OpenCL-Docs/blob/00bbbcf924d386d62dcc202573abc606659f5653/xml/cl.xml#L3211

xml/cl.xml
```diff
@@ -1779,7 +1783,9 @@ server's OpenCL/api-docs repository.
         <enum value="0x12B4"        name="CL_IMAGE_REQUIREMENTS_MAX_HEIGHT_EXT"/>
         <enum value="0x12B5"        name="CL_IMAGE_REQUIREMENTS_MAX_DEPTH_EXT"/>
         <enum value="0x12B6"        name="CL_IMAGE_REQUIREMENTS_MAX_ARRAY_SIZE_EXT"/>
-            <unused start="0x12B7" end="0x1FFF" comment="Reserved for core API tokens"/>
+        <enum value="0x12B7"        name="CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR"/>
+        <enum value="0x12B8"		name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
```
SunSerega, line 1787:
```suggestion
        <enum value="0x12B8"            name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
```
And those tabs I reported earlier... It seems that the old comment got hidden after this enum was renamed.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -392,6 +431,17 @@ in the table below.
...
+| {CL_MUTABLE_DISPATCH_ASSERTS_KHR}
+| {cl_mutable_dispatch_asserts_khr_TYPE}
+| This is a bitfield and can be set to a combination of the following values:
+
+  {CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR}
+  An assertion by the user that the number of work-groups of this ND-range kernel will not be updated beyond
+  the number defined when the ND-range kernel was recorded. The number of work-groups is
+  defined as the product for each _i_ from _0_ to _work_dim - 1_ of
+  _ceil((global_work_size[i] - global_work_offset[i])/local_work_size[i])_ (if
+  _global_work_offset_ is NULL, _global_work_offset[i]_ should be replaced with _0_).
```
bashbaug, line 444: I don't think this is correct - the number of work-groups is only dependent on the global work size and the local work size, and is not dependent on the global work offset.

See: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_mapping_work_items_onto_an_nd_range

> The number of work-groups can be computed as:
> 
> (Wx, Wy) = (ceil(Gx / Sx), ceil(Gy / Sy))

xml/cl.xml
aharon-abramson: This enum is reused for both cl_mutable_dispatch_asserts_khr in clCreateCommandBufferKHR and clCommandNDRangeKernelKHR. So, I guess I need both `require`s.

ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
aharon-abramson: You are right, my mistake.

xml/cl.xml
SunSerega: Ok, fixed the suggestion.

bashbaug: Just FYI, the header generation scripts will generate a block of enums for each "require" section, so if the enum is listed in each block it will be generated twice.  The header generation scripts will handle this (they need to, since sometimes an enum is required by two different extensions) but it does look a little odd if you don't know what's happening:

```c++
/* cl_command_buffer_properties_khr */
#define CL_MUTABLE_DISPATCH_ASSERTS_KHR                     0x12B8

/* cl_ndrange_kernel_command_properties_khr */
/* enum CL_MUTABLE_DISPATCH_ASSERTS_KHR */
```

Note that the second enum is commented out because it is a duplicate.

If you're OK with this, no action required, just thought I'd mention it.  Thanks!

xml/cl.xml
```diff
@@ -7313,6 +7321,15 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR"/>
                 <enum name="CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR"/>
             </require>
+            <require comment="cl_command_buffer_properties_khr">
+                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
+            </require>
+            <require comment="cl_ndrange_kernel_command_properties_khr">
+                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
```
bcalidas, line 7328: would CL_COMMAND_MUTABLE_DISPATCH_ASSERTS be a better name?

aharon-abramson: But all other names begin with "CL_MUTABLE_DISPATCH_".

bashbaug:
Merging as discussed in the February 13th teleconference.

SunSerega:
Sorry to ask after the merge, but how is `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` used?
Is it possibly an editing leftover, an old name of `CL_MUTABLE_DISPATCH_ASSERTS_KHR`?

aharon-abramson:
> Sorry to ask after the merge, but how is `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` used? Is it possibly an editing leftover, an old name of `CL_MUTABLE_DISPATCH_ASSERTS_KHR`?

No, it isn't a leftover. `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` is used in `clCreateCommandBufferKHR`, and `CL_MUTABLE_DISPATCH_ASSERTS_KHR` - in `clCommandNDRangeKernelKHR`.

SunSerega:
Ah, in that case this part:
```
            <require comment="cl_command_buffer_properties_khr">
                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
            <require comment="cl_ndrange_kernel_command_properties_khr">
                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
```
Isn't reusing this enum after all and should instead be:
```
            <require comment="cl_command_buffer_properties_khr">
                <enum name="CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
            <require comment="cl_ndrange_kernel_command_properties_khr">
                <enum name="CL_MUTABLE_DISPATCH_ASSERTS_KHR"/>
            </require>
```
Right?

aharon-abramson:
Hmm, it seems you're right. @bashbaug, did we miss this?

SunSerega:
Or, wait, this part is in `cl_khr_command_buffer_mutable_dispatch` extension, while `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` is already added to `<require comment="cl_command_buffer_properties_khr">` block of `cl_command_buffer_properties_khr` extension. So the `<require>` in `cl_khr_command_buffer_mutable_dispatch` was probably unnecessary.

More importantly, there is no mention of the `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` enum in [spec](https://github.com/aharon-abramson/OpenCL-Docs/blob/e89bf7f8410948b05977712f375927dd4aef1621/ext/cl_khr_command_buffer_mutable_dispatch.asciidoc). Instead, there is this `CL_COMMAND_BUFFER_ASSERTS_KHR`, which isn't defined in XML.
That's what started my confusion.

SunSerega:
Or, another possibility - this PR shouldn't have added the `<require>` block in `cl_khr_command_buffer`. Because this PR is about `cl_khr_command_buffer_mutable_dispatch` extension, and the `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` enum is not added by `cl_khr_command_buffer`.

PR #999, opened 2023-11-15
move the layers spec to the extensions directory

bashbaug:
Since the layers spec is not published in the OpenCL extension spec and is instead published on the OpenCL registry similar to EXT and vendor extensions, it makes more sense to put it in the extensions directory.


Kerilk (approved):
LGTM!

PR #1000, opened 2023-11-16
add description metadata

### LINKED ISSUE #994: add description metadata to HTML specs
One of the issues identified by https://pagespeed.web.dev/ is that our HTML specs do not currently have description metadata.  It looks like we can add this using Asciidoctor attributes:

https://docs.asciidoctor.org/asciidoc/latest/document/metadata/

We might want to add appropriate keywords also.

This may help search engines to prioritize our specifications higher.

bashbaug:
fixes #994 

This PR adds an asciidoctor "description" attribute that gets assigned to the HTML `<meta>` description element.

Best practices seem to be to keep the description approximately 150 characters long, so I've tried to adhere to that.


alycm (approved):
LGTM, merging as obvious.

Issue #1003, opened 2023-11-20
cl_khr_icd 2.0.0 Discussion


Kerilk:
This is a discussion thread, the intent is to gather feedback on the new dispatching strategy proposed at the Montreal F2F.

For Khronos members, slides are there:
https://members.khronos.org/wg/OpenCL/document/31662

The gist of the idea is:

## Fix Dispatching Issues

the current dispatch strategy leaves us open to  segfaults, as the dispatch tables are not managed by the loader, and thus the loader has minimal information about them. The idea is to switch to loader managed dispatch as Vulkan does.

### Change OpenCL objects layout

From:
```C
struct _cl_platform_id { cl_icd_dispatch *dispatch; };
```
to:
```C
struct _cl_platform_id {
  cl_icd_dispatch *dispatch;
  void            *disp_data;
};
```

The loader will be able to distinguish objects supporting the new dispatching strategy by using the first entry in the dispatch table, `clGetPlatformIDs`, that is not currently used by the loader since it uses `clIcdGetPlatformIDsKHR` to query platforms:
```C
typedef struct _cl_icd_dispatch {
  /* OpenCL 1.0 */
  clGetPlatformIDs_t *clGetPlatformIDs;
  clGetPlatformInfo_t *clGetPlatformInfo;
  clGetDeviceIDs_t *clGetDeviceIDs;
  /* ... */
}
```

### Dispatch through `disp_data`

The check and dispatch would look like this
```C
  /* Dispatch test on 64 bit architecture */
  if ((uintptr_t)platform->dispatch->clGetPlatformIDs == 0x4F50454E434C3331) { // (ASCII “OPENCL31”)
    /* New dispatch through `disp_data` */
  } else {
    /* Old dispatch through `dispatch` */
  }
```

The `disp_data` field would be set by the loader on platforms returned by `clGetPlatformIDsKHR`, and populated by the new `clGetFunctionAddressForPlatformKHR` extension function. This new function allows querying regular entry points for a platform.

### Vendor Responsible to Propagate `disp_data`

This `disp_data` has to be propagated to child object, and this is left to the implementation to do. For instance, when calling:
```C
cl_int clGetDeviceIDs(
  cl_platform_id platform,
  cl_device_type device_type,
  cl_uint num_entries,
  cl_device_id* devices,
  cl_uint* num_devices);
```
devices returned by the driver in `devices` should have their `disp_data` pointer set to the same value as the one `platform` currently has.

[...]

PR #1005, opened 2023-11-22
Extension of cl_khr_icd to 2.0.0 with loader managed dispatch.

Kerilk:
This is a formal description of the proposal discussed here:
https://github.com/KhronosGroup/OpenCL-Docs/issues/1003

### DIFF
--- a/api/cl_khr_icd.asciidoc
+++ b/api/cl_khr_icd.asciidoc
@@ -9,7 +9,7 @@ include::{generated}/meta/{refprefix}cl_khr_icd.txt[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2020-04-21
+    2024-09-04
 *IP Status*::
     No known IP claims.
 
@@ -48,6 +48,8 @@ An object is said to be ICD compatible if it is of the following structure:
 struct _cl_<object>
 {
     struct _cl_icd_dispatch *dispatch;
+    // Since version 2.0.0
+    void *dispatch_data;
     // ... remainder of internal data
 };
 ----
@@ -67,6 +69,13 @@ can be appended.
 Functions which do not have an argument from which the vendor implementation
 may be inferred have been deprecated and may be ignored.
 
+New in version 2.0.0, object that are ICD 2 compatible should also contain
+a `dispatch_data` field of type `void *` that the loader is free to modify
+for a platform through the {clIcdSetPlatformDispatchDataKHR} new API entry
+point. In order to distinguish ICD 1 and ICD 2 objects, the pointers to
+`clGetPlatformIDs` and `clUnloadCompiler` in the dispatch structure must be
+set to `CL_ICD2_TAG_KHR`. The vendor driver is responsible for propagating
+the `dispatch_data` value when returning objects to users.
 
 === ICD Data
 
@@ -230,9 +239,30 @@ value {CL_PLATFORM_ICD_SUFFIX_KHR}.
 If any of these steps fail, the ICD Loader will ignore the Vendor ICD and
 continue on to the next.
 
+New in version 2.0.0, the ICD Loader will also query the following functions
+from the library: {clIcdGetFunctionAddressForPlatformKHR} and
+{clIcdSetPlatformDispatchDataKHR}. If these two functions are present and the
+pointers to `clGetPlatformIDs` and `clUnloadCompiler` in the dispatch
+structure of a platform is set to `CL_ICD2_TAG_KHR` the platform will be
+deemed ICD 2 compatible and dispatch will be managed by the ICD Loader. If the
+`CL_ICD2_TAG_KHR` tag is present but one of the two functions above is
+missing or it the tag is present in only one of the pointers, the ICD Loader
+will ignore the Vendor ICD and continue on to the next.
+
+During initialization, after calling {clIcdGetPlatformIDsKHR} to query the
+available platforms, the ICD Loader will, for each ICD 2 compatible platform,
+query dispatchable entry points using {clIcdGetFunctionAddressForPlatformKHR}
+and then set the platform dispatch data using
+{clIcdSetPlatformDispatchDataKHR}.
+
 === New Commands
 
   * {clIcdGetPlatformIDsKHR}
+
+New in version 2.0.0:
+
+  * {clIcdGetFunctionAddressForPlatformKHR}
+  * {clIcdSetPlatformDispatchDataKHR}
 
 === New Enums
 
@@ -243,6 +273,11 @@ Accepted as _param_name_ to the function {clGetPlatformInfo}:
 Returned by {clGetPlatformIDs} when no platforms are found:
 
   * {CL_PLATFORM_NOT_FOUND_KHR}
+
+New in version 2.0.0, used as a value in the pointers to `clGetPlatformIDs`
+and `clUnloadCompiler` in the dispatch structure:
+
+  * `CL_ICD2_TAG_KHR`
 
 === Issues
 
@@ -288,7 +323,29 @@ error values (e.g. {CL_INVALID_DEVICE} corresponding to the object in
 question.
 --
 
+  . How will the ICD loader avoid calling entry points outside of the dispatch
+    table of old implementations, potentially causing segfaults?
++
+--
+*RESOLVED*: From versions 2.0.0 and above, the loader will be responsible for
+managing dispatch data information per platform. The content of the dispatch
+table is obtained through a dedicated entry point, and missing entry points are
+replaced by stubs that return {CL_INVALID_OPERATION}.
+--
+
+  . How will the ICD loader set the loader managed dispatch information in new
+    objects? Some of these objects may be created by extension functions that
+    the loader is unaware of.
++
+--
+*RESOLVED*: During initialization, the loader will set the platform dispatch
+information through a new entry point. The implementation will be responsible
+for propagating this information to new objects.
+--
+
 === Version History
 
   * Revision 1.0.0, 2020-04-21
   ** First assigned version.
+  * Revision 2.0.0, 2024-09-04
+  ** Loader managed dispatch.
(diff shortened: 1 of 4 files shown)


ext/cl_khr_icd.asciidoc
```diff
@@ -28,6 +28,7 @@ implementation, the string *cl_khr_icd* will be present in the
 |====
 | *Date*     | *Version* | *Description*
 | 2020-04-21 | 1.0.0     | First assigned version.
+| XXXX-XX-XX | 2.0.0     | Loader managed dispatch.
```
pjaaskel, line 31: It might be informative to add a paragraph to the intro about the problems of 1.0.0 that 2.0.0 solves. A simple example exposing the problems (segfaults) would help understanding. Is it only related to unloading of the drivers/layers? It's not obvious to me who haven't been able to follow the ICD2 discussion closely, thus it likely won't be obvious to others new to the extension text.

ext/cl_khr_icd.asciidoc
```diff
@@ -42,6 +43,8 @@ An object is said to be ICD compatible if it is of the following structure:
 struct _cl_<object>
 {
     struct _cl_icd_dispatch *dispatch;
+    // Since version 2.0.0
+    void *disp_data;
```
pjaaskel, line 47:
```suggestion
    void *dispatch_data;
```
IMHO better be readable than short :)

ext/cl_khr_icd.asciidoc
```diff
@@ -60,6 +63,14 @@ members can be appended.
 Functions which do not have an argument from which the vendor implementation
 may be inferred have been deprecated and may be ignored.
 
+*New in version 2.0.0*, object that are ICD 2 compatible should also contain
+a `disp_data` field of type `void *` that the loader is free to modify for a
```
pjaaskel, line 67: Does it mean that the loader is supposed to set disp_data of all objects involved in the call to loader-specific dispatch data to transfer the state, but it doesn't have to?

ext/cl_khr_icd.asciidoc
```diff
@@ -60,6 +63,14 @@ members can be appended.
...
+platform through the `clSetPlatformDispatchDataKHR` new API entry point
+In order to distinguish ICD 1 and ICD 2 objects, the pointer to
+`clGetPlatformIDs` in the dispatch structure must be set to
+`CL_ICD2_TAG_KHR`. The vendor driver is responsible for propagating the
```
pjaaskel, line 71: OK, so loader sets it to some objects, and then when creating new objects that are related to another object (cl_context), the driver is assumed to copy it by default? Could we make it more consistent by forcing the loader to manage this data always? It's supposed to be loader-specific data, not driver specific, right? And it's always in a well-specified position in the beginning of the CL objects?

pjaaskel (changes requested):

ext/cl_khr_icd.asciidoc
pjaaskel: OK, found the explanation in this discussion: https://github.com/KhronosGroup/OpenCL-Docs/issues/1003 (the extensions of which obj args the loader doesn't know about). Could you perhaps move some of these justifications and examples to the proposal text to make it more self-contained?

ext/cl_khr_icd.asciidoc
Kerilk: Will do, thanks for the suggestion. I will add a rationale section.

ext/cl_khr_icd.asciidoc
Kerilk: I agree, will do!

ext/cl_khr_icd.asciidoc
Kerilk: Only of the platforms, and only at initialization (and when I am done with the instance support, just after an instance platform is created).

ext/cl_khr_icd.asciidoc
Kerilk: You are correct! I will add this to the rationale.

ext/cl_khr_icd.asciidoc
pjaaskel: OK. It's not obvious from the text why this special casing for loader setting it for the platform only.

ext/cl_khr_icd.asciidoc
Kerilk: Added in the issues section.

ext/cl_khr_icd.asciidoc
Kerilk: Added in the issues section.

xml/cl.xml
```diff
@@ -382,6 +384,15 @@ server's OpenCL/api-docs repository.
...
             <member>const <type>size_t</type>*                             <name>local_work_size</name></member>
         </type>
+
+        <type category="define" requires="stdint" name="CL_ICD2_TAG_KHR" comment="ICD 2 tag, in cl_ext.h">
+#if INTPTR_MAX == INT32_MAX
+#define CL_ICD2_TAG_KHR ((intptr_t)0x434C3331)
+#else
+#define CL_ICD2_TAG_KHR ((intptr_t)0x4F50454E434C3331)
+#endif
+        </type>
```
SunSerega, line 394: This is different from how other `#define` constants are defined, making it very hard to read this definition automatically.

How about we instead specify what happens to upper bits in a comment?
```suggestion
        <type category="define" requires="stdint" comment="ICD 2 tag, in cl_ext.h; upper bits are cut off on 32-bit systems">#define <name>CL_ICD2_TAG_KHR</name> ((intptr_t)0x4F50454E434C3331)</type>
```

Kerilk: That was my original idea. Unfortunately, since this is downcasting an integer it generates warning on some compilers. I checked, and Vulkan also use these king of logic:

```xml
        <type category="define" name="VK_USE_64_BIT_PTR_DEFINES">
#ifndef VK_USE_64_BIT_PTR_DEFINES
    #if defined(__LP64__) || defined(_WIN64) || (defined(__x86_64__) &amp;&amp; !defined(__ILP32__) ) || defined(_M_X64) || defined(__ia64) || defined (_M_IA64) || defined(__aarch64__) || defined(__powerpc64__) || (defined(__riscv) &amp;&amp; __riscv_xlen == 64)
        #define VK_USE_64_BIT_PTR_DEFINES 1
    #else
        #define VK_USE_64_BIT_PTR_DEFINES 0
    #endif
#endif</type>
```

SunSerega: Hmm... What about `(intptr_t)(int64_t)0x4F50454E434C3331`? The idea is that it shouldn't generate a warning if the downcast is more clearly intended...

Kerilk: That's what I though, but the warning was still there. And someone warned me it would happen beforehand, like you I was in disbelief.

SunSerega: Alright, then, I agree with this solution; that's better than what I can suggest...

Kerilk:
@keryell was concerned clPlatformID pointer for a non ICD 2.0.0 compatible could have the same value as the tag.

One way to avoid this problem would be to have a second entry in the dispatch table set to the tag value (clUnloadCompiler comes to mind) and check that both of these have the tag while loading the library. If the tag were not present in both, the implementation would not be loaded.

keryell:
> @keryell was concerned clPlatformID pointer for a non ICD 2.0.0 compatible could have the same value as the tag.
> 
> One way to avoid this problem would be to have a second entry in the dispatch table set to the tag value (clUnloadCompiler comes to mind) and check that both of these have the tag while loading the library. If the tag were not present in both, the implementation would not be loaded.

Thanks. This is quite simpler and more portable than the solution I was thinking to.

api/opencl_platform_layer.asciidoc
```diff
@@ -84,6 +84,43 @@ Otherwise, it returns one of the following errors:
...
+include::{generated}/api/protos/clIcdSetPlatformDispatchDataKHR.txt[]
+include::{generated}/api/version-notes/clIcdSetPlatformDispatchDataKHR.asciidoc[]
+
+  * _platform_ refers to the platform ID returned by {clIcdGetPlatformIDsKHR}
+  * _dispatch_data_ is the value to set the `dispatch_data` field of the
+    structure to
+
+{clIcdSetPlatformDispatchDataKHR} returns {CL_SUCCESS} if the function is
+executed successfully.
+It returns {CL_INVALID_PLATFORM} if _platform_ is not a valid platform.
```
bashbaug, line 121: What is the behavior if _platform_ is an ICD 1 platform?

Kerilk: It should return `CL_INVALID_PLATFORM` but the loader should never call this for an ICD 1 platform, either because the platform did not provide `clIcdSetPlatformDispatchDataKHR`or because the loader didn't find the correct tags in the platform dispatch table. If you think this needs additional we can add `ICD 2`to the mention above.

api/cl_khr_icd.asciidoc
```diff
@@ -9,7 +9,7 @@ include::{generated}/meta/{refprefix}cl_khr_icd.txt[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2020-04-21
+    2024-09-04
```
bashbaug, line 12:
```suggestion
    2025-04-01
```

api/cl_khr_icd.asciidoc
```diff
@@ -286,9 +321,31 @@ obtaining the ICD dispatch table.
...
+*RESOLVED*: During initialization, the loader will set the platform dispatch
+information through a new entry point. The implementation will be responsible
+for propagating this information to new objects.
 --
 
 === Version History
 
   * Revision 1.0.0, 2020-04-21
   ** First assigned version.
+  * Revision 2.0.0, 2024-09-04
```
bashbaug, line 350:
```suggestion
  * Revision 2.0.0, 2025-04-01
```

bashbaug (approved):

PR #1006, opened 2023-11-23
[RFC] cl_exp_tensor

linehill:
This draft introduces an extension for tensor data type (`cl_tensor`) for storing N-dimensional data in implementation-defined memory layout. This extension is designed for the defined [built-in kernel extension](TODO) for starters. In the future we may extend the tensor data type support in OpenCL C / SPIR-V.

Any feedback is appreciated.

HTML is rendered [here](https://htmlpreview.github.io/?https://github.com/linehill/OpenCL-Docs/blob/cl_exp_tensor/ext/cl_exp_tensor.html).


FreddieWitherden:
The idea of an 'implementation defined memory layout' is a fallacy.  The optimal layout is *always* contextual.

As an example [here](https://github.com/intel/intel-graphics-compiler/issues/254#issuecomment-1276861768) is a high-performance OpenCL SGEMM kernel for Intel GPUs which can get 75% of peak (and a good candidate for a 'built in').  However, to perform `C = A*B` it wants `C` and `B` to be row-major and `A` to be an obscure tiled format with column-major tiles of size 8 by 4 packed in a row-major fashion.  This is no problem in practice as you run a precursor kernel to take A from row-major and copy it into a temporary tiled array).  For large matrices which are FLOP bound the overhead here is a few percent.

The key thing, however, is the optimal layout for a parameter depends on the parameter itself.  Unless you know exactly what you want to do with a tensor/matrix you can not put it into an optimal format ahead of time.  Thus, this extra complexity solves nobodies problem (a GEMM will still likely need to rearrange some of the data before operating on it) and comes at an **extreme** cost (none of my other kernels can operate on these tensors without going through import/export operations).  Plus, even if the language were extended to enable kernels to operate on these tensors it would not help much: since they are a black box how can I construct my kernel to ensure things like coalesced access?

linehill:
Thanks for the feedback. I think the extension needs to be improved by illustrating further on how drivers may gather context and use it to optimize memory layouts.

A driver should be able to gather context from kernels and the tensor arguments given to them and use it to optimize memory layouts of the tensors. For example, in the command queue code sample at the following point:

```
cl_mem in0_mem = clCreateBufferWithProperties(
  ctx, {CL_MEM_BIND_TO_TENSOR, in0, 0}, CL_MEM_READ_ONLY,
  0 /* must be zero for CL_MEM_BIND_TO_TENSOR. */, nullptr, &err);
```

The driver would at least know the `in0` tensor’s shape (64, 100, 200) and know it may be passed to the `matmul` kernel. The driver may discover what memory layout gives the `matmul` kernel the best performance (especially if it is a built-in kernel) and use the best layout for the `in0` tensor. Drivers may also defer the layout selection up until kernel execution which may provide more context for the layout optimization.

> Unless you know exactly what you want to do with a tensor/matrix you can not put it into an optimal format ahead of time.

This could be true with command queues. With command buffers, however, we should already be able to tell drivers in precision what we want to do with the tensors and that should give the drivers an opportunity to optimize the layouts ahead of time.

>  Thus, this extra complexity solves nobodies problem (a GEMM will still likely need to rearrange some of the data before operating on it) and comes at an extreme cost (none of my other kernels can operate on these tensors without going through import/export operations).

The cost of using import/export tensor commands shouldn't be extreme - at least I’m unaware of the reasoning that led to the “extreme cost” conclusion. Could you elaborate?

The worst case I can see is that a kernel compiler is unable to analyze kernel code for concluding optimal layout (respect to the context it can muster). In that case the driver may fallback to linear memory layout and `clEnqueueExportToTensor` and `clEnqueueImportFromTensor` would then behave similarly as they were `clEnqueueWriteBufferRect` and `clEnqueueReadBufferRect`, respectively. 

> Plus, even if the language were extended to enable kernels to operate on these tensors it would not help much: since they are a black box how can I construct my kernel to ensure things like coalesced access?

I think I’m missing something here. AFAIK, OpenCL does not give guarantees for coalesced accesses and they are very hardware dependent. Are you hoping that tensor extension would bring coalesced access guarantees for software defined (OpenCL C / SPIR-V) kernels?

I’m not aware of doors being closed for the possible language extension regarding improving coalescing opportunities. Can you provide insight on this matter?

FreddieWitherden:
> The driver would at least know the `in0` tensor’s shape (64, 100, 200) and know it may be passed to the `matmul` kernel.

Or any other built-in kernel.  Or, in the future, any user defined kernel (assuming the OpenCL kernel language is extended to support tensor arguments).  It provides no useful information. Moreover, it is highly likely that a single tensor will partake in multiple matrix multiplications throughout its life of varying sizes, with each multiplication wanting a different layout.

> > Unless you know exactly what you want to do with a tensor/matrix you can not put it into an optimal format ahead of time.
> 
> This could be true with command queues. With command buffers, however, we should already be able to tell drivers in precision what we want to do with the tensors and that should give the drivers an opportunity to optimize the layouts ahead of time.

Giving opportunities in theory is very different to being useful in practice.  This would add a large amount of additional complexity to the driver that goes well beyond what any of them currently do.  Moreover, compare it to the alternative: pick a fixed layout for the tensors (or a couple of layouts) and have a built-in kernels for them.  This would not only make it easier for vendors (just a case of wiring up their existing BLAS library kernels) but also for users (no need to wait for their implementation to support command buffers and for their application to be ported over them to get optimal performance).  Fixed layouts are exactly what NVIDIA and AMD do in CUDA and HIP, respectively.  It works well.

> > Thus, this extra complexity solves nobodies problem (a GEMM will still likely need to rearrange some of the data before operating on it) and comes at an extreme cost (none of my other kernels can operate on these tensors without going through import/export operations).
> 
> The cost of using import/export tensor commands shouldn't be extreme - at least I’m unaware of the reasoning that led to the “extreme cost” conclusion. Could you elaborate?

Superfluous memory copies to/from device memory.  In contrast a fixed layout allows the decision about if to make a copy or not to be deferred until the last minute (and made by the kernel itself which has the most context).  In contrast, having two operations requires driver magic.  Further, if the stars align it may be possible for the matmul to use a so-called direct kernel which makes no temporary copies at all.

> > Plus, even if the language were extended to enable kernels to operate on these tensors it would not help much: since they are a black box how can I construct my kernel to ensure things like coalesced access?
> 
> I think I’m missing something here. AFAIK, OpenCL does not give guarantees for coalesced accesses and they are very hardware dependent. Are you hoping that tensor extension would bring coalesced access guarantees for software defined (OpenCL C / SPIR-V) kernels?

The data for matrix multiplications has to come from somewhere.  Most of the time that is an OpenCL kernel.  Moreover, if you don't want superfluous copies then kernels need support for tensor arguments.  But, if the layout is a black box then how do you write the kernel efficiently?  As a simple example consider a matrix and storing it in row-major vs. column-major.  If you tell me (the kernel author) what you want, I can ensure that my kernel is mostly doing stride-1 reads and writes.  If the layout is abstracted then a lot of my accesses could end up as gather/scatter operations.

This is why fixed layouts make so much more sense and are the norm in BLAS libraries; `cublasSgemmBatched` and `cublasSgemmStridedBatched` as a nice example of how these APIs can work.  Notice how two APIs are needed (the first is useful for saving memory and improving cache hit rates if you have a batch of B matrices and want to multiply them all by the same A matrix since you only need to allocate one A matrix).

linehill:
Thanks for the input.  Fair enough - this extension in conjunction with the defined built-in kernel extension was designed to be used with command buffers. To get most benefit out of the "opaque layouts" the compiler/runtime needs to know how the kernel connects to others which may only be accomplished reasonably with command buffers.

For usability without command buffers we could add a feature that lets OpenCL programmers set the tensor layouts and tune their kernels with respect to them, and have a way to "manually" eliminate the need of import/export tensor commands.

We’ll draft new version which should also cover the scenario you described.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+== Dependencies
+
+This extension is written against the OpenCL Specification version 3.0.14.
+
+This extension requires OpenCL 1.2 or later.
+
+== Overview
+
+The extension provides new tensor object abstraction. Tensor objects
+are similar to image types in regard they represents N-dimensional
```
pjaaskel, line 50:
```suggestion
are similar to image types in regard that they represent N-dimensional
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+data of some application chosen data type and they may be mapped to
```
pjaaskel, line 51:
```suggestion
data of an application chosen data type and they may be mapped to
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+dedicated hardware except that
```
pjaaskel, line 52:
```suggestion
dedicated hardware, with the following key differences:
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+* higher than 3-dimensional data can be supported (limited by
```
pjaaskel, line 54:
```suggestion
* Higher than 3-dimensional data can be supported (limited by
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  devices' capabilities).
+
+* applications may choose how the data elements of the tensors are
```
pjaaskel, line 57:
```suggestion
* Applications may choose how the data elements of the tensors are
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  laid out in the buffers using the tensor layout descriptions
+  provided in this extension.
+
+Applications may also choose the memory layouts of the tensors be
```
pjaaskel, line 61:
```suggestion
Applications may also choose the memory layouts of the tensors to be
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+implementation-specified, letting the driver to optimize the tensor
+data layout for better performance or to lay out the data as required by
+hardware functions (e.g. exposed via builtin kernels).
```
pjaaskel, line 64:
```suggestion
hardware accelerated functions (e.g. exposed via builtin kernels).
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+The scope of this extension to provide host APIs for creating tensor
```
pjaaskel, line 66:
```suggestion
The scope of this extension is to provide host APIs for creating tensor
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+objects and transfer data between tensors, host and other memory
+objects.
+
+A separate extension implemented on top of this extension,
+cl_exp_defined_builtin_kernels which provides "defined built-in
```
pjaaskel, line 71:
```suggestion
cl_exp_defined_builtin_kernels provides "defined built-in
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+kernels" (DKBs) which operates on tensors. It also provides mechanism
```
pjaaskel, line 72:
```suggestion
kernels" (DKBs) which can operate on tensors. It also provides mechanism
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+} cl_tensor_layout_blas_pitched__exp;
+
+typedef struct cl_tensor_layout_ml_exp {
+  cl_tensor_layout_ml_type_exp ml_type;
+} cl_tensor_layout_ml_exp;
+----
+
+== New API Enums
+
+Accepted value for _properties_ parameter to
```
pjaaskel, line 187:
```suggestion
Accepted value for the _properties_ parameter to
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+The scope of this extension to provide host APIs for creating tensor
+objects and transfer data between tensors, host and other memory
+objects.
+
+A separate extension implemented on top of this extension,
+cl_exp_defined_builtin_kernels which provides "defined built-in
+kernels" (DKBs) which operates on tensors. It also provides mechanism
+for drivers to create DBKs that are optimized for the tensor arguments
+they operate on.
+
```
pjaaskel, line 75: Perhaps mention that it's designed for use with command buffers for whole task-graph optimizations. And maybe also about Vulkan layering (if those referred APIs are public).

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+--
+// End (Add the following to Section 5.2.2, *Reading, Writing and Copying Buffer Objects*)
+
+
+(Add the following to Section 5.17.5, *Recording Commands to a Command-Buffer*) ::
++
+--
+
+If *cl_khr_command_buffer* is supported, then the following command
```
pjaaskel, line 656: Refer to this extension in the intro spec, state that optionally cl_khr_command_buffer is extended.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+| CL_TENSOR_DTYPE_INT4_EXP        | 4-bit signed integer.            | cl_char.
+| CL_TENSOR_DTYPE_INT8_EXP        | 8-bit signed integer.            | cl_char.
+| CL_TENSOR_DTYPE_INT16_EXP       | 16-bit signed integer.           | cl_short.
+| CL_TENSOR_DTYPE_INT32_EXP       | 32-bit signed integer.           | cl_int.
+| CL_TENSOR_DTYPE_INT64_EXP       | 64-bit signed integer.           | cl_long.
+| CL_TENSOR_DTYPE_UINT8_EXP       | 8-bit unsigned integer.          | cl_uchar.
+| CL_TENSOR_DTYPE_UINT16_EXP      | 16-bit unsigned integer.         | cl_ushort.
+| CL_TENSOR_DTYPE_UINT32_EXP      | 32-bit unsigned integer.         | cl_uint.
+| CL_TENSOR_DTYPE_UINT64_EXP      | 64-bit unsigned integer.         | cl_ulong.
+| CL_TENSOR_DTYPE_FP8_EXP         | Half precision floating-point.   | cl_char.
```
pjaaskel, line 798: Refer to a specific FP8 format.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  64-bit real and imaginary part. | cl_double2
+|===
+
+[[layout-types-table]]
+.Optional tensor memory layout types.
+[cols="1,1,4",stripes=even]
+|===
+| *layout type* | *tensor layout type* | *Description*
+
+| CL_TENSOR_LAYOUT_OPAQUE_EXP | N/A | The tensor don't have application
```
pjaaskel, line 815:
```suggestion
| CL_TENSOR_LAYOUT_OPAQUE_EXP | N/A | The tensor doesn't have application
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  defined memory layout. Driver controls the tensors layout. To read
+  or write elements of the tensor
```
pjaaskel, line 817: Is this sentence unfinished?

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+| CL_TENSOR_LAYOUT_BLAS_EXP
+|<<cl-tensor-layout-blas,cl_tensor_layout_blas_exp>>
+| A type that describe packed memory layout similar ones used in BLAS APIs.
```
pjaaskel, line 821:
```suggestion
| A type that describes a packed memory layout similar ones used in BLAS APIs.
```

pjaaskel: Is this accurate enough description. Is the BLAS layout well-specified somewhere to which we can refer?

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+| CL_TENSOR_LAYOUT_BLAS_EXP
```
pjaaskel, line 823: Pitched? Unfinished copy-pasta?

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+|<<cl-tensor-layout-blas,cl_tensor_layout_blas_pitched_exp>>
+| A type that describe memory layout similar ones used in BLAS APIs.
+
+| CL_TENSOR_LAYOUT_ML_EXP       | <<cl-tensor-layout-ml,cl_tensor_layout_ml_exp>> |
+A convenience layout type over `CL_TENSOR_LAYOUT_BLAS_EXP`.
```
pjaaskel, line 828: What does it mean?

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+--
+// End (Add the following to new Section 5.X.Y, *Tensor Descriptions*)
+
+
+[[cl-tensor-layout-blas]]
+(Add the following to new Section 5.X.Y.1, *BLAS Tensor Layout*) ::
++
+--
+The following structures describes packed / pitched BLAS-like memory
```
pjaaskel, line 840:
```suggestion
The following structures describe packed / pitched BLAS-like memory
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+    cl_tensor_pitch      leading_pitches[CL_TENSOR_DESC_MAX_RANK_EXP];
+} cl_tensor_layout_blas_pitched_exp;
+
+typedef struct cl_tensor_layout_ml_exp {
+    cl_tensor_layout_ml_type_exp ml_type;
+} cl_tensor_layout_ml_exp;
+----
+
+* _leading_dims_ describes which elements along the tensor dimension
+  are laid out in the memory. `leading_dims[0]` point to dimension
```
pjaaskel, line 860:
```suggestion
  are laid out in the memory. `leading_dims[0]` points to the dimension
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  whose elements are laid out first, followed by elements along
+  dimension by `leading_dims[1]` and so on. The first N elements must
```
pjaaskel, line 862:
```suggestion
  the dimension by `leading_dims[1]` and so on. The first N elements must
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  be non-zero where N is tensor's rank and the values must be unique
```
pjaaskel, line 863:
```suggestion
  be non-zero where N is a tensor's rank and the values must be unique
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  and within range `[0, tensor_rank)`.
+
+* _leading_pitches_ describes distance between from an element to the
```
pjaaskel, line 866:
```suggestion
* _leading_pitches_ describes the distance between an element to the
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+   rank is greater than one and
+
+** `leading_pitches[i + 1] >= tensor_shape[leading_dims[i]] *
+  leading_pitches[i]` for `i` in `[0, tensor_rank - 1)` if the tensor
+  rank is greater than two.
+
+// ^ This condition is meant to ensure that the tensor elements at different
+// coordinates don't alias.
+
+* _ml_type_ defines memory layout via enumerators which corresponds to
```
pjaaskel, line 883:
```suggestion
* _ml_type_ defines the memory layout via enumerators which corresponds to
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+  predefined configurations of `cl_tensor_layout_blas_exp` structure
+  as listed in <<tensor-layout-ml-types,ML tensor layout type>> table.
+
+The memory layout descriptions map tensor coordinates to buffer's
+memory byte locations respect to buffer's base address as followed in
```
pjaaskel, line 888:
```suggestion
memory byte locations respect to buffer's base address as in the following pseudo code example.
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+[source,cpp]
+----
+constexpr size_t b = 64, m = 100, n = 200, k = 50;
+
+std::vector<float> in0_data = ...;
+std::vector<float> in1_data = ...;
+std::vector<float> out_data(b * m * n);
+
+// Create tensor with opaque layout.
```
pjaaskel, line 936:
```suggestion
// Create a tensor with an opaque layout.
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+in0_desc.shape[2] = k;
+in0_desc.layout = nullptr;
+in0_desc.layout_type = CL_TENSOR_LAYOUT_OPAQUE_EXP;
+
+cl_int err;
+cl_mem in0_tensor = clCreateBufferWithProperties(
+  ctx, {CL_MEM_TENSOR_EXP, in0_desc, 0},
+  CL_MEM_READ_ONLY, 0, nullptr, &err);
+
+// Create tensor from a host allocation using an application defined
```
pjaaskel, line 951:
```suggestion
// Create tensor from a host allocation using an application-defined
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+col_major.leading_dims[0] = 1,
+col_major.leading_dims[1] = 2,
+in1_desc.layout = &col_major;
+in1_desc.layout_type = CL_TENSOR_LAYOUT_BLAS_EXP;
+
+cl_mem in1_tensor = clCreateBufferWithProperties(
+  ctx, {CL_MEM_TENSOR_EXP, in1_desc, 0},
+  CL_MEM_USE_HOST_PTR | CL_MEM_READ_ONLY, 0, in1_data.data(), &err);
+
+// Create another tensor with application defined layout.
```
pjaaskel, line 970:
```suggestion
// Create another tensor with an application-defined layout.
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+// optimized for them using via yet realized API extension.
+cl_kernel batched_matmul_kernel = create_batched_matmul_kernel(
+  ctx, device_span, in1_desc, in2_desc, out_desc);
+
+clSetKernelArg(batched_matmul_kernel, 0, sizeof(cl_mem), &in0_tensor);
+clSetKernelArg(batched_matmul_kernel, 1, sizeof(cl_mem), &in1_tensor);
+clSetKernelArg(batched_matmul_kernel, 2, sizeof(cl_mem), &out_tensor);
+
+// Required command for transferring data to layout-opaque tensors and
+// from it to elsewhere.
```
pjaaskel, line 998:
```suggestion
// from it elsewhere.
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+clEnqueueNDRangeKernel(
+  cmd_q, batched_matmul_kernel, 3, matmul_grid, nullptr, nullptr, 0, nullptr, nullptr);
+
+clEnqueueMapBuffer(
+  cmd_q, out_tensor, CL_TRUE, CL_MAP_READ, 0, b * m * n, 0, nullptr, nullptr);
+----
+
+
+== Issues
```
pjaaskel, line 1011:
```suggestion
== Issues and Open Questions
```

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+
+. Should we have support for tensors with undefined shape and tensors
```
pjaaskel, line 1013:
```suggestion
. Should we support tensors with undefined shape and tensors
```

pjaaskel: Should we also look at Khronos NNEF or is it inactive now?

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
++
+--
+// https://onnx.ai/onnx/repo-docs/ShapeInference.html
+*UNRESOLVED*
+--
+
+. Should we define OpenCL C language features for accessing tensors?
++
+--
+*RESOLVED*: OpenCL C support for tensors can be introduced later in a
```
pjaaskel, line 1024: I'd just add those built-ins to this extension right away to avoid expanding the extension fragmentation.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+. What is the use case of `cl_tensor_layout_blas_pitch_exp`?
++
+--
+*UNRESOLVED*
+--
+
+. Should image types be extended instead of adding a separate tensor type?
++
+--
+*UNRESOLVED*
```
pjaaskel, line 1038: I think it logically might be the other way around: images a special case of tensors. Perhaps something to consider in the longer run if this one takes off.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -0,0 +1,1058 @@
...
+| Version | Date       | Author          | Changes
+| 0.1.0   | 2023-11-23 | Henry Linjamäki | *Initial revision*
+
+| 0.2.0   | 2024-8-14 | Henry Linjamäki |
+
+* Rework document structure match to the cl_khr_extension_template.
+
+* Added clEnqueueCopyTensor.
+
+* Added API for setting memory layout for tensors.
```
pjaaskel, line 1056:
```suggestion
* Added an API for setting the memory layout for tensors.
```

pjaaskel (changes requested):
Thanks! Mainly just grammar/typos found so far.

extensions/cl_exp_tensor.asciidoc
```diff
@@ -878,15 +898,15 @@ typedef struct cl_tensor_layout_ml_exp {
...
 
 * _ml_type_ defines the memory layout via enumerators which corresponds to
   predefined configurations of `cl_tensor_layout_blas_exp` structure
   as listed in <<tensor-layout-ml-types,ML tensor layout type>> table.
 
 The memory layout descriptions map tensor coordinates to buffer's
-memory byte locations respect to buffer's base address as followed in
-pseudo C:
+memory byte locations respect to buffer's base address as in the
+followed in pseudo C code example:
```
pjaaskel, line 909:
```suggestion
following pseudo C code example:
```

pjaaskel (changes requested):

PR #1010, opened 2023-11-27
consistently put newer versions at the bottom of version history tables

### LINKED ISSUE #1001: adopt a consistent version history table convention
We seem to have several different conventions for version history in the OpenCL Extension Specification:

[Some extensions](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_version_history_33) (a total of 3) have the older versions at the top of the table and the newer versions at the bottom.

[Other extensions](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_version_history_41) (also a total of 3) have the newer versions at the top of the table and the older versions at the bottom.

I think we should have one convention and use it consistently.  Which to we prefer?

bashbaug:
fixes #1001 

Consistently puts newer versions at the bottom of "version history" tables.

This only changes KHR extensions in the OpenCL extension spec.  I didn't modify any vendor or EXT extensions.

### DIFF
--- a/ext/cl_khr_external_memory.asciidoc
+++ b/ext/cl_khr_external_memory.asciidoc
@@ -30,10 +30,10 @@ Other related extensions define specific external memory types that may be impor
 [cols="1,1,3",options="header",]
 |====
 | *Date*     | *Version* | *Description*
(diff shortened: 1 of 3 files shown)


alycm (approved):
LGTM, merging as obvious.

PR #1011, opened 2023-11-27
use title case for more chapter headings

bashbaug:
This is a purely editorial change.

We tend to use title case (all major words capitalized) for most of our chapter headings, but in a few places we only capitalized the first letter.  Switched some of the more common chapter headings to title case:

* "General information" -> "General Information"
* "Version history" -> "Version History"


alycm (approved):
LGTM, merging as obvious.

PR #1012, opened 2023-11-27
remove TODO comment in semaphore spec

bashbaug:
related to #957

The default behavior when the device handle list is not specified is now properly described, so the TODO comment can be removed.

CC @joshqti


alycm (approved):
LGTM, merging.

PR #1015, opened 2023-11-28
clarify async copies and wait group events must be convergent

### LINKED ISSUE #535: Should wait_group_events and async_work_group_copy be "convergent"
In OpenCL C `wait_group_events` and friends have wording that says:

> This function must be encountered by all work-items in a work-group executing the kernel with the same num_events and event objects specified in event_list; otherwise the results are undefined.

`barrier` and `work_group_barrier` has wording that effectively says this, but also more:

> For these functions, if any work-item in a work-group encounters a barrier, the barrier must be encountered by all work-items in the work-group before any are allowed to continue execution beyond the barrier.
> 
> If the barrier is inside a conditional statement, then all work-items in the work-group must enter the conditional if any work-item in the work-group enters the conditional statement and executes the barrier.
> 
> If the barrier is inside a loop, then all work-items in the work-group must execute the barrier on each iteration of the loop if any work-item executes the barrier on that iteration.
>
> [...]
>
>  The values of flags and scope must be the same for all work-items in the work-group.

However on these additional control-flow point the SPIR-V `OpGroupWaitEvents` and `OpControlBarrier` definitions have essentially identical wording.  Respectively:

> Behavior is undefined unless all invocations within Execution execute the same dynamic instance of this instruction.

> When Execution is Workgroup or larger, behavior is undefined unless all invocations within Execution execute the same dynamic instance of this instruction.

The following kernel follows the specification rules for `wait_group_events`, but not the stricter rules that are placed on `barrier` or `OpGroupWaitEvents`.

```c
void my_kernel(__global int* src, __local int* dst) {
  size_t id = get_global_id(0);
  
  event_t my_event = async_work_group_copy(dst, src, 1, 0);

  // Below does not violate any wording in OpenCL C specification, but if naively
  // translated to SPIR-V it is bound to hit the undefined behaviour case.
  //
  // Also, any CPU implementation wanting to treat `wait_group_events` with similar
  // semantics to `barrier` may struggle with this case.
  if (id % 2) {
    wait_group_events(1, &my_event);
  } else {
    wait_group_events(1, &my_event);
  }
  
  // Do something with 'dst' here.
}
```

Using the stricter definition from `barrier` or `OpGroupWaitEvents` for `wait_group_events` allows implementations more freedom, especially CPU implementations that are likely executing work-items in serial (i.e. you could now use the same control-flow transformations that may be used around `barrier` etc).

An example of following the current OpenCL C wording precisely is that the Clang default OpenCL headers mark [`barrier` as convergent](https://github.com/llvm/llvm-project/blob/d8938c8bb5479b168d27d3e161cb3a53e8ff09f0/clang/lib/Headers/opencl-c.h#L12535-L12539), but [not `wait_group_events`](https://github.com/llvm/llvm-project/blob/d8938c8bb5479b168d27d3e161cb3a53e8ff09f0/clang/lib/Headers/opencl-c.h#L12938).

Should the `async_work_group_copy` and `wait_group_events` family of functions in OpenCL C have the additional control flow wording that `barrier` has?  I.e. should OpenCL C specification match the semantics of the equivalent SPIR-V instructions?  I'm not intending to propose a behaviour change here, obviously OpenCL C pre-dates SPIR-V, but thinking perhaps that this was the intent all along.  The current wording basically comes from OpenCL 1.0.

bashbaug:
fixes #535

Note, I factored out the descriptions for each of the different functions into the common section header because otherwise the table cells extended beyond one page and generated asciidoctor errors.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -6379,6 +6379,23 @@ The OpenCL C programming language implements the <<table-builtin-async-copy,
 following functions>> that provide asynchronous copies between `global` and
 local memory and a prefetch from `global` memory.
 
+The async copy and wait group events functions are performed by all work-items
+in a work-group and therefore must be encountered by all work-items in a
+work-group executing the kernel with the same argument values, otherwise the
+results are undefined.
+This rule applies to ND-ranges implemented with uniform and non-uniform
+work-groups.
+
+If an async copy or wait group events function is inside a conditional statement
+then all work-items in the work-group must enter the conditional if any
+work-item in the work-group enters the conditional statement and executes the
+async copy or wait group events function.
+
+If an async copy or wait group events function is inside a loop then all
+work-items in the work-group must execute the async copy or wait group events
+function on each iteration of the loop if any work-item executes the async copy
+or wait group events function on that iteration.
+
 We use the generic type name `gentype` to indicate the built-in data types `char`,
 `char__n__`, `uchar`, `uchar__n__`, `short`, `short__n__`,
 `ushort`, `ushort__n__`, `int`, `int__n__`, `uint`,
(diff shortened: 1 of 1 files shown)


alycm (approved):
Thanks for this!  LGTM.

PR #1016, opened 2023-11-28
clarify the free function for clEnqueueSVMFree must be thread-safe

### LINKED ISSUE #977: `clEnqueueSVMFree` doesn't state that the `pfn_free_func` parameter needs to point to a thread-safe callback
Is this intentional? All the other callback functions are required to be thread-safe by the specification, this seems to be the only exception.

Or is there a different part of the spec stating that all callback functions are thread-safe by default?

bashbaug:
Clarify that the free function pointer passed to clEnqueueSVMFree must be thread-safe.  See #977.

Also:

* Fixes a grammatical error: "applications reponsibility" -> "application's responsibility".
* Fixes a few formatting errors in the description of _user_data_.


alycm (approved):
Change makes sense to me, this just makes this callback match all the other callback.

LGTM.

PR #1018, opened 2023-11-28
include dictionaries before footnotes for reference pages

bashbaug:
The generated asciidoctor dictionaries need to be included before footnotes because some of the footnotes use the attributes defined in the dictionaries.  If the footnotes are included first then the attributes are not properly substituted.

To see the problem, notice how the current footnote for clGetDeviceIds is not properly formatted:

https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetDeviceIDs.html#_footnotedef_1

### DIFF
--- a/scripts/clconventions.py
+++ b/scripts/clconventions.py
@@ -201,10 +201,10 @@ class OpenCLConventions(ConventionsBase):
         """Return any extra text to add to refpage headers."""
         return 'include::{config}/attribs.txt[]\n' + \
             'include::{config}/opencl.asciidoc[]\n' + \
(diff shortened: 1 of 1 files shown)


oddhack:
Related question, the way the footnotes are currently marked up as, more or less 'footnote:[{some attribute containing text}]' resulting in a lot of copies of some footnotes. I believe if they were re-done as '{some attribute containing footnote:label[text]}', each footnote would only appear once, yet would still work when they appear in refpages. I could pick this up in the draft extension integration branch, if you like.

alycm (approved):
LGTM, merging as obvious.

PR #1019, opened 2023-11-28
WIP cl_ext_image_drm_format_modifier

kpet:


extensions/cl_ext_image_drm_format_modifier.asciidoc
```diff
@@ -0,0 +1,193 @@
...
+  * {CL_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT} requires image data be laid out according
+    to the DRM format modifier passed via the {CL_MEM_IMAGE_DRM_FORMAT_MODIFIER_EXT}
+    property. Applications are required to provide a value for
+    {CL_MEM_IMAGE_DRM_FORMAT_MODIFIER_EXT} when they set {CL_MEM_IMAGE_TILING_EXT} to
+    {CL_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT}.
+
+{CL_MEM_IMAGE_DRM_FORMAT_MODIFIER_EXT} can be passed as part of the _properties_
+parameter to *clCreateImageWithProperties* to provide the DRM format modifier
+that describes the layout of image data when an image is created with
+{CL_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT}.
```
bashbaug, line 141: At least for Vulkan, there are two ways to create an image with a DRM format modifier:

1. Using [VkImageDrmFormatModifierExplicitCreateInfoEXT](https://registry.khronos.org/vulkan/specs/latest/man/html/VkImageDrmFormatModifierExplicitCreateInfoEXT.html), which requires creating the image with the exact DRM format modifier specified.  This is analogous to the `CL_MEM_IMAGE_DRM_FORMAT_MODIFIER_EXT` property added by this extension.
2. Using [VkImageDrmFormatModifierListCreateInfoEXT](https://registry.khronos.org/vulkan/specs/latest/man/html/VkImageDrmFormatModifierListCreateInfoEXT.html), which allows creating the image from a set of specified DRM format modifiers.  This provides the implementation with some freedom to choose what it considers to be the best DRM format modifier.

Should we support (2) and add the ability to pass a list of viable DRM format modifiers as well?

Note, if we added this functionality, then the application should still be able to query `CL_IMAGE_DRM_FORMAT_MODIFIER_EXT` to determine which of the DRM format modifiers in the set was chosen.

kpet: Discussed in the 2025/04/15 memory TSG teleconference: another way of allowing implementations to express a preferred DRM modifier (which is what we'd really get with (2)) would be to state that the list of supported DRM modifiers reported by the implementation is ordered by preference. No strong preference for one or the other approach expressed on the call.

nyanmisaka:
Just wondering if there is a way in ARM libMali to import AFBC compressed images? I came across this draft by accident while searching.

In my use case the hardware video decoder is able to produce images with the `AFBC_FORMAT_MOD_BLOCK_SIZE_16x16` DRM format modifer and it would help performance (reduce memory bandwidth utilizing) if libMali could directly import that instead of requiring linear images.

kpet:
>Just wondering if there is a way in ARM libMali to import AFBC compressed images? I came across this draft by accident while searching.

@nyanmisaka Yes, this is supported in shipping drivers using this extension. I'm happy to help you get going with it. Feel free to reach out at kevin.petit@arm.com.

PR #1027, opened 2023-12-04
remove unnecessary rounding mode text for geometric and common functions

### LINKED ISSUE #1024: geometric functions and rounding modes
The section text for "geometric functions" currently contains the lines ([link](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#geometric-functions)):

> The built-in geometric functions are implemented using the round to nearest even rounding mode. The built-in geometric functions may be implemented using contractions such as mad or fma.

This text has been around since OpenCL 1.0.

Without knowing the exact operations that are performed though, and the order in which they are performed, it's not clear how this information is useful or actionable.  Can it be removed?

Note:
* We do describe accuracy requirements for these geometric functions.
* We do not have similar text in the OpenCL SPIR-V environment spec, though we do in the OpenCL extended instruction set ([link](https://registry.khronos.org/SPIR-V/specs/1.0/OpenCL.ExtendedInstructionSet.100.html#_a_id_geometric_a_geometric_instructions)).
* The section text for "common functions" contains similar text as well - can it also be removed ([link](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#common-functions))?

bashbaug:
fixes #1024

Without knowing how these functions are implemented any statements about rounding modes or contractions are unnecessary and confusing. We have defined error bounds for these functions, and as long as an implementation meets these error bounds it should be considered correct.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -5473,11 +5473,6 @@ We use the generic type name `gentyped` to indicate that the function can
 take `double`, `double2`, `double3`, `double4`, `double8` or `double16` as
 the type for the arguments.
 
-The built-in common functions are implemented using the round to nearest
-even rounding mode.
-The built-in common functions may be implemented using contractions such
-as *mad* or *fma*.
-
 [[table-builtin-common]]
 .Built-in Scalar and Vector Argument Common Functions
 [cols=",",]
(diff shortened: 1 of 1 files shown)


kpet (approved):
LGTM

PR #1028, opened 2023-12-04
publish cl_intel_subgroup_local_block_io

bashbaug:
This is the specification source for the cl_intel_subgroup_local_block_io extension.

We have been shipping this extension for a while now ([gpuinfo.org link](https://opencl.gpuinfo.org/listreports.php?extension=cl_intel_subgroup_local_block_io)), but it looks like the extension specification was never published.


alycm (approved):
Only reviewed editorially, but it all LGTM.  Merging.

PR #1034, opened 2023-12-09
Remove `cl_semaphore_import_properties_khr`

SunSerega:
I cannot find the spec description, and there is no related `<type category="define">` tag.
Probably editing leftover from #939? @lakshmih


bashbaug (approved):
I was initially wondering why the header generation scripts weren't complaining loudly about this, but a) it's only in the block "comment", and b) it's an empty block so the script is omitting it completely.  This is still a good cleanup to make though, thanks for the PR!

Mikolaj-Filar-Mobica (approved):

lakshmih (approved):

bashbaug:
Merging as discussed in the December 12th teleconference.

PR #1035, opened 2024-01-02
update the spec source copyright dates to 2024

bashbaug:
Updates the spec source copyright dates to 2024, using the script from #757.

As with #881 this PR also updates the spec source copyright dates for several vendor extensions.


alycm (approved):
LGTM, merging immediately as the changes are formulaic and obviously correct.

This also changes copyright dates on vendor extensions, which is consistent with what we did last year.

PR #1039, opened 2024-01-09
cl_khr_external_memory: Clarify offset is zero

joshqti:
Clarify that the offset is zero when importing memory, and the layout is not inferred, such as linear memory from a handle specified as "assume linear". This implies that when specifying an image slice_pitch, array layer 0 is assumed to start at offset 0.


nikhiljnv:
Closing without adding any clarification as discussed in memory subgroup call on January 16th, 2024

PR #1042, opened 2024-01-15
Add CL_CANCELLED_IMG error code for cl_img_cancel_command extension.

paulfradgley:
Added a new error code CL_CANCELLED_IMG, which will be used by future cl_img_cancel_commands extension.


bashbaug (approved):
LGTM.  This comes out of a block previously reserved for IMG, so no reason to wait to merge this, I think.

Eventually this error code should be attached to a specific extension so the header generation will pick it up, but that can happen later.

Issue #1041, opened 2024-01-10
Reconsider mutable command-buffer structure pointer chaining


EwanC:
The [cl_command_buffer_mutable_dispatch](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer_mutable_dispatch) extension introduced structure pointer chains used as part of the update API
```c
cl_int clUpdateMutableCommandsKHR(
    cl_command_buffer_khr command_buffer,
    const cl_mutable_base_config_khr* mutable_config);

```

The advantages of pointer chaining are extensibility without API breakages, and also the ability for layers to inject structs into the linked list. However, because the structure pointer chaining design is limited to one corner of the OpenCL standard, there are no layers using it, and the API may be more complicated than necessary as it also throws up the issues highlighted in https://github.com/KhronosGroup/OpenCL-Docs/issues/990 .

An alternative approach that would keep the benefit of extensibility without API breakages, but drop the extra complexity, is an API that uses an array approach rather than a linked list. Credit to @bashbaug for this idea:
```c
cl_int clUpdateMutableCommandsKHR(
    cl_command_buffer_khr command_buffer,
    cl_uint num_configs,                    /* the number of configs in the arrays */
    const cl_config_type_khr* config_types, /* the type of each config */
    const void** configs) ;                 /* the config data, interpreted using the type */
```

This would be a breaking change to existing `cl_khr_command_buffer_mutable_dispatch` implementations and code, but this is allowed due to the provisional nature of the extension. I think this is a better final design, as in future layered extensions it's not unlikely that we might want to have more than one instance of a struct, which is forbidden in Vulkan pointer chaining semantics, and defining our own pointer chaining semantics separate from Vulkan seems like more effort than it's worth when pointer chaining isn't giving us tons of value.

In summary, I'd like to propose removing the pointer chaining update API and replace it with the proposed update API entry-point.

## PRs containing proposed changes

* OpenCL-Docs https://github.com/KhronosGroup/OpenCL-Docs/pull/1045
* OpenCL-Headers https://github.com/KhronosGroup/OpenCL-Headers/pull/245
* CTS - https://github.com/KhronosGroup/OpenCL-CTS/pull/1984
* Poc Implementation change to OCK - https://github.com/EwanC/oneapi-construction-kit/pull/1

EwanC:
Drafted into docs & headers changes:
[...]

PR #1045, opened 2024-01-16
Use array for clUpdateMutableCommandsKHR.

EwanC:
Proposal to pass the update configs to `clUpdateMutableCommandsKHR` as an array, rather than pointer chained linked list.

See https://github.com/KhronosGroup/OpenCL-Docs/issues/1041 for motivation.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -15620,7 +15620,7 @@ endif::cl_khr_command_buffer_multi_device[]
 
 ifdef::cl_khr_command_buffer_mutable_dispatch[]
 [[mutable-commands]]
-=== Mutable Commands:
+=== Mutable Commands
 
 A generic {cl_mutable_command_khr_TYPE} handle is called a _mutable-command_
 object as it can be returned from any command recording entry-point in the
@@ -15631,11 +15631,10 @@ modified through the fields of {cl_mutable_dispatch_config_khr_TYPE}.
 
 Mutable-command handles are updated between enqueues using entry-point
 {clUpdateMutableCommandsKHR}.
-To enable performant usage, all aspects of mutation are encapsulated inside
-a single {cl_mutable_base_config_khr_TYPE} parameter.
-This means that the runtime has access to all the information about how the
-command-buffer will change, allowing the command-buffer to be rebuilt as
-efficiently as possible.
+To enable performant usage, all aspects of mutation can be passed in a single
+call using an array. This means that the runtime has access to all the
+information about how the command-buffer will change, allowing the
+command-buffer to be rebuilt as efficiently as possible.
 Any modifications to the arguments or execution info of a mutable-dispatch
 handle using {cl_mutable_dispatch_arg_khr_TYPE} or
 {cl_mutable_dispatch_exec_info_khr_TYPE} have no affect on the original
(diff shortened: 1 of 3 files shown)


ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
```diff
@@ -501,40 +463,29 @@ mutable-command objects are preserved and one of the errors below is returned:
...
+If _configs_ is non-`NULL`, then for any {cl_mutable_dispatch_config_khr_TYPE}
+element of the array the errors defined by {clEnqueueNDRangeKernel},
+{clSetKernelExecInfo}, {clSetKernelArg}, and {clSetKernelArgSVMPointer} are
+returned by {clUpdateMutableCommandsKHR} if any of the struct elements are set
+to an invalid value. Additionally, the following errors are returned if any
+{cl_mutable_dispatch_config_khr_TYPE} element of the array violates the defined
+conditions:
 
 * {CL_INVALID_MUTABLE_COMMAND_KHR} if _command_ is not a valid mutable
   command object, or created from _command_buffer_.
```
EwanC, line 487:
```suggestion
  command object returned from {clCommandNDRangeKernelKHR}, or created from _command_buffer_.
```

EwanC:
I've rebased this change after #950 merged, however I'm still getting familiar with how the spec looks now. So changing this PR back to draft until I double check I'm happy with how everything looks.

aharon-abramson (approved):

bashbaug:
Discussed in the July 9th teleconference.  We set a deadline for two weeks to merge (July 23rd).

api/opencl_runtime_layer.asciidoc
```diff
@@ -15726,19 +15727,17 @@ parameters are updated so that the new number of work-groups exceeds the
...
-{clUpdateMutableCommandsKHR} if any of the array elements are set to an
-invalid value.
-Additionally, the following errors are returned if any
-{cl_mutable_dispatch_config_khr_TYPE} element of the array violates the
-defined conditions:
+If _configs_ is non-`NULL`, then for any {cl_mutable_dispatch_config_khr_TYPE}
+element of the array the errors defined by {clEnqueueNDRangeKernel},
+{clSetKernelExecInfo}, {clSetKernelArg}, and {clSetKernelArgSVMPointer} are
+returned by {clUpdateMutableCommandsKHR} if any of the struct elements are set
+to an invalid value. Additionally, the following errors are returned if any
```
kpet, line 15734: I think we should try to improve how this is described so we make the list of errors that _may_ be returned a bit clearer. This is not something that this PR changes so I'm prefectly fine with capturing an issue that we'd look at separately.

kpet (approved):
Thanks for proposing this change! Happy with the direction. I have not turned every stone and so might have missed points of detail but I suggest we prioritise aligning all components to this proposal to enable forward progress and fix any minor issues on top.

api/opencl_runtime_layer.asciidoc
EwanC: Creating an issue makes sense, is there any particular way you imagine the improvement looking? e.g. Being more explicit about the errors rather than referencing core API entry-points, or just making the layout of the existing information more readable.

kpet: The first idea that came to mind was to categorise errors in the description of those commands such that we could refer to the categories here but I have not though enough about this you really have good input or opinions.

bashbaug:
Merging as discussed in the July 16th teleconference.

PR #1046, opened 2024-01-16
Upload cl_img_cancel_command asciidoc specification

paulfradgley:
This extension adds the functionality to instruct the OpenCL implementation that an incomplete OpenCL command no longer needs to be executed.


bashbaug (approved):
Interesting extension!

I see no reason not to merge this PR now, but you will probably want to update the XML file with information about this extension as well so it will be included in e.g. the generated extension headers.

PR #1048, opened 2024-01-22
fixes derived formula for atanh

bashbaug:
Asciidoctor treats words surrounded by double parentheses as index terms, so we need to escape the first parenthesis so the derived formula for atanh is properly preserved.

See: #142 

I did a search through the spec and I believe all of the other uses of double parentheses (typically for OpenCL C attributes, like the reqd_work_group_size) are properly handled.

It would be nice to have a way to disable this expansion since we don't use it, or at least to detect when it occurs.


kpet (approved):

bashbaug:
Merging as discussed in the January 30th teleconference.

PR #1049, opened 2024-01-26
clarify that 16-bit and 64-bit floats can be passed as kernel arguments

### LINKED ISSUE #1037: SPV environment spec does not allow OpTypeFloat kernel arguments of width 16 or 64
According to [this section](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_kernel_arguments) of the OpenCL SPIR-V Environment spec, kernel arguments of type **OpTypeFloat** can only have width of 32.  This seems like an oversight.  Shouldn't widths of 16 and 64 be supported if the device has support for 16- or 64-bit floating point data types?

Note that the SPIR-V environment spec *does* say that basic scalar and vector types are allowed to be **OpTypeFloat** with any of the widths 16, 32, or 64 (see [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_basic_scalar_and_vector_types)).  It's only the section on kernel arguments that is missing the widths of 16 and 64.

bashbaug:
fixes #1037

### DIFF
--- a/env/common_properties.asciidoc
+++ b/env/common_properties.asciidoc
@@ -370,7 +370,7 @@ Allowed types for OpenCL kernel arguments are:
 For *OpTypeInt* parameters, supported _Widths_ are 8, 16, 32, and 64, and
 must have no signedness semantics.
 
-For *OpTypeFloat* parameters, _Width_ must be 32.
+For *OpTypeFloat* parameters, supported _Width_ are 16, 32, and 64.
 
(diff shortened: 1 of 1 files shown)


kpet (approved):

bashbaug:
Merging as discussed in the January 30th teleconference.

PR #1050, opened 2024-01-26
add the numerical value of the image channel order and data type

bashbaug:
This may make it easier to extend these tables or enable additional simplifications in the future.

see internal SPIR-V issue 773


alycm (approved):
LGTM, merging immediately.

PR #1051, opened 2024-01-29
Fix unterminated table in IMG extension

kpet:
Change-Id: Ib3e14b09c052930da79c74b2e09654c89df93d45


paulfradgley (approved):
Good catch, thanks.

PR #1054, opened 2024-02-04
Fix markups of parameter names ``obj`` and ``value``

F3eQnxN3RriK:
In other part of the document, parameter names are enclosed in ``_``, so these parameters also should be marked up the same way.


bashbaug (approved):
Thanks for fixing these!

I don't think we're as consistent using this convention as I'd like, but it's definitely better to use some special text treatment than none at all, and using _italics_ seems to be the most common convention (though a few other places use `monospace`).  At some point it'd be a good idea to take a pass through the spec and fix a few more of these.

PR #1056, opened 2024-02-08
Add `cl_img_cancel_command` extension to XML

SunSerega:
`CL_CANCELLED_IMG` being defined but not used was bothering me when pulling upstream changes...

@paulfradgley is this all right?

For context:
1. `CL_CANCELLED_IMG` was added to XML separately in #1042
2. Specification was added in #1046, but it didn't add the rest of the definitions to XML


paulfradgley (approved):
Looks good to me.

bashbaug:
Note, we should create an OpenCL-Headers PR to regenerate headers with these changes, also.

PR #1058, opened 2024-02-12
add USM host mem alloc size issue

bashbaug:
Adds an issue to the Intel USM spec regarding the maximum size of a host USM allocation.


alycm (approved):
LGTM, only updating an unresolved issue on a vendor extension, so just merging immediately.

PR #1059, opened 2024-02-14
Fix `cl_khr_command_buffer_mutable_dispatch` extension

SunSerega:
Found at the end of https://github.com/KhronosGroup/OpenCL-Docs/pull/992

`CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR` enum has the wrong name in the spec and was added to the wrong extension block in XML. It was also added to the right extension block, but under the wrong name: `CL_MUTABLE_DISPATCH_ASSERTS_KHR`.

### DIFF
--- a/ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
+++ b/ext/cl_khr_command_buffer_mutable_dispatch.asciidoc
@@ -337,7 +337,7 @@ description of property values.
   
 |====
 
-Add a {CL_COMMAND_BUFFER_ASSERTS_KHR} property to the
+Add a {CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR} property to the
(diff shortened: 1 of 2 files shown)


SunSerega:
@aharon-abramson

bashbaug (commented):
Thank you for fixing these, mostly LGTM.

I think if you apply the suggestion below the automated builds will start working again.

If they are still failing I think it is because there is an "unterminated open block" at the end of this file - the closing `--` is missing from the issue description.  If this doesn't cause a problem with the automated builds though, don't worry about it, and I'll make a PR to fix it and a few other things I've noticed.

Thanks again!

aharon-abramson:
I'm OK with this.

SunSerega:
> If they are still failing I think it is because there is an "unterminated open block"

It doesn't fail the checks. I'll leave it to you then.

bashbaug (approved):

PR #1062, opened 2024-02-16
fix a few minor issues in the extensions spec

bashbaug:
* Fixes "unterminated open block" at the end of the mutable dispatch extension.
* Fixes a typo in the assume linear images enum.
* Uses the asciidoctor attribute for the mutable dispatch asserts type.
* Minor wordsmithing and ordering improvements in the mutable dispatch new enums section.


alycm (approved):
LGTM, all editorial so will mege immediately.

Issue #1061, opened 2024-02-16
clCreateContextFromType with properties != NULL and unspecified or invalid platform


s-barannikov:
1. The documentation does not specify what error code should be returned by `clCreateContextFromType` if all of the following conditions are met:
* `properties != NULL`;
* `properties` didn't specify a platform;
* no platform could be selected.

I guess it should specify that `CL_INVALID_PLATFORM` is returned in this case (the same as if `properties == NULL`).

2. The documentation is ambigous on which error code should be returned by `clCreateContextFromType` if `properties` specify an invalid platform:
* CL_INVALID_PLATFORM ... if platform value specified in properties is not a valid platform.
* CL_INVALID_PROPERTY ..., if the value specified for a supported property name is not valid ...

bashbaug:
For (1) we could modify the error description to cover this scenario if you'd like.  I agree `CL_INVALID_PLATFORM` makes the most sense.  Maybe something like:

> [CL_INVALID_PLATFORM](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_PLATFORM) if ~~properties is NULL~~ **no platform value is specified** and no platform could be selected or if the platform value specified in properties is not a valid platform.

For (2), I think either error code would be valid, and I don't believe we currently test for this case.  If it's helpful though, I recommend chosing the more descriptive error, so `CL_INVALID_PLATFORM`.

Note, if you are implementing an ICD then this checking will happen in the OpenCL ICD loader: [link](https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/861b68b290e76d08e7241608479c16431f529945/loader/icd_dispatch_generated.c#L249).

s-barannikov:
@bashbaug 
Thank you for your reply.

> For (1) we could modify the error description to cover this scenario if you'd like. I agree `CL_INVALID_PLATFORM` makes the most sense. Maybe something like:

This is exactly what I had in mind.

> I recommend chosing the more descriptive error, so `CL_INVALID_PLATFORM`.

Thanks, I'll follow your suggestion. This would also be closer to how other interfaces handle invalid platforms.

> Note, if you are implementing an ICD then this checking will happen in the OpenCL ICD loader:

[...]

PR #1063, opened 2024-02-16
clarify conditions for CL_INVALID_PLATFORM

bashbaug:
See discussion: https://github.com/KhronosGroup/OpenCL-Docs/issues/1061#issuecomment-1949130684

Clarifies the conditions when clCreateContext and clCreateContextFromType should return CL_INVALID_PLATFORM.

CC @s-barannikov.


alycm (approved):
LGTM, an intuitive change and matches outcome in issue discussion.

Issue #224, opened 2020-02-20
Clarify value returned for CL_DEVICE_TYPE query


kpet:
The specification currently states:

>The OpenCL device type. Currently supported
values are:
CL_DEVICE_TYPE_CPU, CL_DEVICE_TYPE_GPU, CL_
DEVICE_TYPE_ACCELERATOR, CL_DEVICE_TYPE_DEFAULT,
a combination of the above types, or CL_DEVICE_
TYPE_CUSTOM.

This means that an implementation could return `CL_DEVICE_TYPE_CPU|CL_DEVICE_TYPE_GPU, CL_DEVICE_TYPE_ACCELERATOR|CL_DEVICE_TYPE_DEFAULT`. What is the rationale for allowing CPU|GPU for example?

Note that the spec currently doesn't describe anything CPU- or GPU-specific.

Oblomov:
The possibility to combine flag values was mostly used by platforms to specify which device was the default for that platform (e.g. returning `CL_DEVICE_TYPE_CPU|CL_DEVICE_TYPE_DEFAULT`), but even that is being phased out, since  a lot of client code erroneously checks for device types by equality (e.g. it tries to skip the CPU device during enumeration by checking if the returned value is `CL_DEVICE_TYPE_CPU`, rather than checking only the corresponding bit).

As for a mixed-type device, it wouldn't be inconceivable to have a single piece of hardware acting as both the CPU and GPU of the host, even though there aren't devices that do that AFAIK.

bashbaug:
> that is being phased out, since a lot of client code erroneously checks for device types by equality

Interesting.  We are considering adding new device types to distinguish between different types of `ACCELERATOR` devices, or integrated / discrete `GPU` devices, for example.  If we did this, we would recommend that these devices continue to return the "old" device type for backwards compatibility in addition to the newer more detailed device type, e.g. `CL_DEVICE_TYPE_FANCY_NEW_ACCELERATOR | CL_DEVICE_TYPE_ACCELERATOR`.  This obviously won't work, though, if client code is checking the device types by equality.  Should we consider adding the additional device types via a separate query instead?

Oblomov:
[...]

PR #1069, opened 2024-02-26
strengthen requirements for CL_DEVICE_TYPE query

bashbaug:
Proposed fix for #224.

This PR:

1. Moves the requirement that `CL_DEVICE_TYPE_CUSTOM` must not be combined with any other device types to the description of `CL_DEVICE_TYPE`.  (see (4) below)
2. Adds the requirement that `CL_DEVICE_TYPE_DEFAULT` must only be used to get device IDs using `clGetDeviceIDs` and must NOT be returned as the `CL_DEVICE_TYPE` for any device.
3. Clarifies that `CL_DEVICE_TYPE_ALL` also must only be used to get device IDs using `clGetDeviceIDs`.
4. ~~Adds an informative note that most implementations should only return a single device type for `CL_DEVICE_TYPE`, and should not return a combination of device types.~~  Updated: Adds the requirement that the device must return only a single device type for `CL_DEVICE_TYPE`.


api/opencl_platform_layer.asciidoc
```diff
@@ -239,34 +239,40 @@ include::{generated}/api/version-notes/CL_DEVICE_TYPE_ACCELERATOR.asciidoc[]
...
-For example, a {CL_DEVICE_TYPE_CPU} device may also be a
-{CL_DEVICE_TYPE_GPU} device, or a {CL_DEVICE_TYPE_ACCELERATOR} device
-may also be some other, more descriptive device type.
-{CL_DEVICE_TYPE_CUSTOM} devices must not be combined with any other
+[NOTE]
+====
+Although the standard allows an OpenCL device to report more than one device
+type in its {CL_DEVICE_TYPE}, many applications assume that devices will only
+report a single device type. Implementations are strongly encouraged to report a
+single device type for {CL_DEVICE_TYPE} and not to report a combination of
```
alycm, line 273: Text above says `{CL_DEVICE_TYPE_DEFAULT} **must not** be reported in {CL_DEVICE_TYPE}`, but the wording of this note is just **strongly encouraged**.

I'm not sure if any devices have reported more than one non-default device here, e.g. `GPU | ACCELERATOR`, which thing wording discourages but doesn't outright forbid.

alycm (approved):
Textually LGTM.  Group needs to approve overall change.

api/opencl_platform_layer.asciidoc
bashbaug: I strengthened the requirement further and now the device must report only a single device type.

alycm (approved):
New wording still approved, it's very clear on returning a single value.

bashbaug:
Had a thought: Would we want `CL_DEVICE_TYPE_DEFAULT` and `CL_DEVICE_TYPE_ALL` to be valid values to pass to `clCreateContextFromType`?  I think we would.

If so, I should adjust the descriptions of these devices types so it is clearer that this is allowed.  In other words, `CL_DEVICE_TYPE_ALL` could be passed to `clGetDeviceIDs` or to `clCreateContextFromType`, though it will never be returned as a device type for `CL_DEVICE_TYPE`.

bashbaug:
I've updated this PR so it describes `clCreateContextFromType` also.  I think it's ready to go.

kpet (approved):

PR #1070, opened 2024-02-26
fix a few notes in the spec for consistency

bashbaug:
Fixes a few minor issues with the "note" markup in the spec.

The first change doesn't seem to be a functional change, but I'm not quite sure how it is working right now, and we may as well make them all consistent.

The second change removes an extraneous ".Note" that we do not have for any of the other "notes".


alycm (approved):
LGTM, minor formatting clean-up, merging.

PR #1071, opened 2024-02-27
clarify clSetEventCallback behavior for command errors

### LINKED ISSUE #1068: clSetEventCallback callbacks being called on errors shouldn't be a footnote.
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetEventCallback

There is this footnote:

> The callback function registered for a command_exec_callback_type value of [CL_COMPLETE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_COMPLETE) will be called when the command has completed successfully or is abnormally terminated.

the "or is abnormally terminated" part is pretty important to know about, so I argue that it shouldn't be mentioned in a footnote like this, but in the actual API description directly. The current wording is also confusing, because it only states (ignoring the footnote):

> The registered callback function will be called when the execution status of command associated with event changes to an execution status equal to or past the status specified by command_exec_status.

bashbaug:
Fixes #1068.

The callback registered for CL_COMPLETE will be called when the command completes successfully or when the command is abnormally terminated.  This behavior used to be documented in a footnote but it is belongs in the main spec text instead.

### DIFF
--- a/api/footnotes.asciidoc
+++ b/api/footnotes.asciidoc
@@ -43,10 +43,6 @@ Either of these choices would mean that no big swap would need to occur in hardw
 The OpenCL specification does not describe the order of precedence for error codes returned by API calls. \
 ]
 
-:fn-event-callback-complete: pass:n[ \
-The callback function registered for a _command_exec_callback_type_ value of {CL_COMPLETE} will be called when the command has completed successfully or is abnormally terminated. \
-]
-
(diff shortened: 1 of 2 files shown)


alycm (approved):
LGTM, I do not believe this changes any meanings, it just moves wording around and out of a footnote.

PR #1076, opened 2024-03-03
Building individual extensions depends on generated files

alycm:
Fix an occasional parallel build failure.  I don't think this is a new issue, just it rarely occurs and only on a clean build.


bashbaug (approved):
Nice catch!

PR #1077, opened 2024-03-03
Fix building specs on macOS

alycm:
The shell code that automatically builds up the Makefile dependencies from the include lines in Asciidoc was not working on macOS.  Carriage returns were present in lines, and this meant that xargs did not behave as intended.  Just strip out the carriage returns.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ src:
 # Top-level spec source file
 APISPEC = OpenCL_API
 APISPECSRC = $(APISPEC).txt $(GENDEPENDS) \
-    $(shell grep ^include:: $(APISPEC).txt | sed -e 's/^include:://' -e 's/\[\]/ /' | xargs echo)
+    $(shell grep ^include:: $(APISPEC).txt | sed -e 's/^include:://' -e 's/\[\]/ /' -e 's/\r//g' | xargs echo)
 
 apihtml: $(HTMLDIR)/$(APISPEC).html $(APISPECSRC)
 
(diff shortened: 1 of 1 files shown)


alycm:
There is almost certainly a cleaner way to do this, but I haven't seen this before so I just did the most direct thing to remove the problem.

I haven't confirmed that this doesn't break the build on other OS, but currently relying on CI to check that.

bashbaug (approved):
Could it be that there is some weird git line ending thing going on?

I checked and this worked fine for my Linux build also, so if this fixes your problem it doesn't seem to have caused any problems here.

CC @oddhack, in case he's seen anything like this before, or perhaps wants to make a similar change to the Vulkan build scripts.

(Note, there is a merge conflict now, probably simple to resolve.)

oddhack:
Do you have core.autocrlf=input set in your .gitconfig (or other git preferences)? I don't have access to a MacOS system but this seems to handle things OK for me in past encounters with EOL conventions. In Vulkan we have a .gitattributes set in the repo root as well, but only to force Bash scripts to use Unix eol=lf conventions everywhere, motivated by some issues in WSL on Windows IIRC.

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=mac

alycm:
> Do you have core.autocrlf=input set in your .gitconfig (or other git preferences)?

I did not, and I now do.  But that won't affect this issue as it only affects writes into the git database, not reads out from it.  I don't think there is a setting that rewrites CRLF to LF on output (`true` rewrites to CRLF on output, `false` does nothing, `input` rewrites to LF on input).

This, however, did cause me to realize that the root issue is that  small number of files in the repository are stored with CRLF line endings.  So #1079 is a more direct change.

alycm:
Closing in favour of #1079

PR #1079, opened 2024-03-05
Remove CRLF line endings from two core spec files

alycm:
All other text files in the repository use LF line endings.  The presence of the carriage returns in the files means that on macOS the shell code that automatically builds up the Makefile dependencies from the include lines in Asciidoc was not working as xargs did not behave as intended.

Note that various Intel vendor extensions also use CRLF line endings but this PR does not touch them, they do not contain includes so do not actually cause any problems.


alycm:
This is an alternative fix for #1077

bashbaug:
> Note that various Intel vendor extensions also use CRLF line endings but this PR does not touch them, they do not contain includes so do not actually cause any problems.

Out of curiosity, how did you find (and fix) the files with CRLF?

Doesn't need to be done with this PR obviously, but I may want to go back and clean up these Intel extensions eventually...

bashbaug (approved):

alycm:
```sh
for a in $(find . | grep -v '\.git') ; do file "$a" ; done | grep CRLF

# Above prints a list of file names and descriptions

dos2unix manually_typed_file_names_that_caused_actual_build_issues_here.txt
```

The result of the above file command on the branch for this PR is

```
./extensions/cl_intel_spirv_media_block_io.asciidoc: ASCII text, with very long lines (307), with CRLF line terminators
./extensions/cl_intel_subgroups.asciidoc: ASCII text, with very long lines (342), with CRLF line terminators
./extensions/cl_intel_subgroups_short.asciidoc: ASCII text, with very long lines (341), with CRLF line terminators
./extensions/cl_intel_subgroup_local_block_io.asciidoc: ASCII text, with very long lines (360), with CRLF line terminators
./extensions/cl_intel_spirv_device_side_avc_motion_estimation.asciidoc: ASCII text, with very long lines (320), with CRLF line terminators
./extensions/cl_intel_spirv_subgroups.asciidoc: ASCII text, with very long lines (554), with CRLF line terminators
./extensions/cl_intel_mem_alloc_buffer_location.asciidoc: ASCII text, with CRLF line terminators
./extensions/cl_intel_required_subgroup_size.asciidoc: ASCII text, with very long lines (308), with CRLF line terminators
```

But these aren't causing any problems, all text editors are completely capable of handling all line ending variations, it just seems to be xargs on macOS that's causing a problem.  Running dos2unix trashes future use of `git blame` on the touched files because you've technically changed every line in the file.

bashbaug:
OK cool, I agree there's no reason to fix the Intel extensions if they aren't causing a problem right now, and I'd rather preserve `git blame` (even though these extensions don't change frequently).

I'm fine clicking the button to merge this now if you'd like, or feel free to merge it yourself.

PR #1081, opened 2024-03-10
Removing separate OpenCL Extensions from the repository

oddhack:
Once we integrate all the khr extensions into the API and C specs, the separate documents under ext/ will serve no purpose and will have the risk of creating divergences between the sme content in the API / C specs and the separate extension spec.

This removes the separate khr extension documents from the repo, and strips down the OpenCL Extension spec to just point to the corresponding extension appendix in the API spec (and include a small amount of content about using extensions, which was separate from the extension specs themselves). It will also serve as a forcing function for all future khr extension development to be done against the API / C specs. Any outstanding PRs against these documents will have to be rebased and the same edits reapplied to those specs, whether changes to published extensions or in-flight, but as yet unpublished khr extensions.

Ideally this would merge into #950 before that is merged to main, if the WG can agree to that. It could also be retargeted to main instead, if this decision will take longer than the timeframe for merging #950.


bashbaug:
Merging as discussed in the March 12th teleconference.

bashbaug (approved):

PR #1084, opened 2024-03-19
Update XML to mark recently ratified extensions as non-provisional

oddhack:
Net effect is to rearrange the extension appendices so these extensions are no longer in a 'provisional' subsection, and to remove the generated comments about their being provisional from the extension refpages.

@neiltrevett @bashbaug once #950 is merged into main, doing this sort of update will be needed whenever an extension is moved from non-ratified to ratified, or provisional ratified to fully ratified. For Vulkan, we've incorporated this into the extension release / ratification process checklists (along with many other minor, easy to overlook tasks).

I encourage merging #950 (with this included) ASAP as this is the start of generating divergences and merge conflicts that will have to be synced up with #950 every time one of the old independent khr extension docs is modified.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -6950,7 +6950,7 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_4x8BIT_PACKED_KHR"/>
             </require>
         </extension>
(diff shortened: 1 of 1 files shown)


bashbaug (approved):

bashbaug:
CI seems stuck, but I'm going to go ahead and merge these changes.  We should still get one more run in when #950 gets merged.

PR #1087, opened 2024-03-22
Fix section titles in Appendix I - OpenCL extensions

kpet:
The section headers for "Name String", "Ratification Status", etc should be one level below the headers for individual extensions.

Another, possibly cleaner, way to fix this would be to add one level of nesting to headers in all the extension files under api/

Change-Id: I97bc12a845ccb39c421dfed16223b3cdee5203af

### DIFF
--- a/scripts/clconventions.py
+++ b/scripts/clconventions.py
@@ -199,7 +199,7 @@ class OpenCLConventions(ConventionsBase):
            file.
             - name - extension name"""
 
-        return 'include::{{apispec}}/{}[]'.format(
+        return 'include::{{apispec}}/{}[leveloffset=+1]'.format(
                 self.extension_file_path(name))
 
     @property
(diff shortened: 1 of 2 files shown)


bashbaug:
@oddhack do you have a preference how we do this?

I agree the spec TOC looks much nicer after this change.

oddhack:
I did something along these lines in the Vulkan scripts quite recently as well - let me take another pass at syncing up the scripts, as I'd like to keep those linked up between the various repos making use of them. It is a bit tricky because the refpage extraction script that assembles the extension refpages has to make some assumptions about the heading level of the extension appendix bits it extracts.

oddhack:
Please take a look at #1116 as an alternative. The nesting level adjustment falls out of synchronizing the scripts with the Vulkan scripts.

oddhack:
@kpet this should probably be closed, with #1116 merged.

kpet:
@oddhack Thanks! Closing this PR.

PR #1088, opened 2024-03-23
XML `<extension>`: Replace remaining `condition` with `depends`

SunSerega:
Asked at the end of #950. This in particular wasn't answered yet, but after getting a bit more comfortable with new XML - I'm pretty sure this does need to be fixed.

Note: There are also a bunch of `<require>` tags with `condition` instead of the `depends` attribute.
The `depends` attribute is being used for these too, but only in one place:
```
            <require depends="CL_VERSION_2_0" comment="From version 0.9.4 of the extension">
```
But half of the `condition` attributes here use negation:
```
            <require condition="!defined(CL_VERSION_1_2)" comment="cl_device_info - defined in CL.h for OpenCL 1.2 and newer">
```
And there is also one like this:
```
            <require condition="defined(_WIN32)">
```
Which is neither a version nor an extension name, so I guess this one cannot be converted to `depends`?

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -6071,7 +6071,7 @@ server's OpenCL/api-docs repository.
                 <command name="clSetKernelExecInfoARM"/>
             </require>
         </extension>
-        <extension name="cl_arm_get_core_id" condition="defined(CL_VERSION_1_2)" supported="opencl">
+        <extension name="cl_arm_get_core_id" depends="CL_VERSION_1_2" supported="opencl">
             <require>
                 <type name="CL/cl.h"/>
             </require>
@@ -7278,7 +7278,7 @@ server's OpenCL/api-docs repository.
                 <command name="clEnqueueWriteHostPipeINTEL"/>
             </require>
         </extension>
-        <extension name="cl_ext_image_requirements_info" condition="defined(CL_VERSION_3_0)" supported="opencl">
+        <extension name="cl_ext_image_requirements_info" depends="CL_VERSION_3_0" supported="opencl">
             <require comment="Types">
                 <type name="cl_image_requirements_info_ext"/>
             </require>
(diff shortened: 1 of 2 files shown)


bashbaug:
The "conditions" are currently used by the extension header generation scripts:

https://github.com/KhronosGroup/OpenCL-Headers/blob/59452533d2afa817bc2dc0da4f783097f4cdbcb0/scripts/cl_ext.h.mako#L307

I'm OK doing something like this PR to remove redundancy, but we'll need to update the header generation scripts as well if we make this change.

SunSerega:
Ah, I was somehow sure there were more `condition` attributes before => most of them were replaced by `depends` => in-house scripts already work with `depends` to do the same thing.

oddhack:
The 'condition' is something OpenCL-specific. The 'depends' was introduced when syncing with the Vulkan scripts and has the same restrictions, it only speaks to extension and core version dependencies.

SunSerega:
> and has the same restrictions

Well, it can represent more cases. For instance in this case:
```
        <extension name="cl_khr_subgroups" supported="opencl" promotedto="CL_VERSION_2_1" ratified="opencl">
            <require>
                <type name="CL/cl.h"/>
            </require>
            <require condition="!defined(CL_VERSION_2_1)" comment="defined in CL.h for OpenCL 2.1 and newer">
                <type name="cl_kernel_sub_group_info"/>
            </require>
            <require comment="cl_kernel_sub_group_info">
                <enum name="CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE_KHR"/>
                <enum name="CL_KERNEL_SUB_GROUP_COUNT_FOR_NDRANGE_KHR"/>
            </require>
            <require>
                <command name="clGetKernelSubGroupInfoKHR"/>
            </require>
        </extension>
```
Here, I don't know how to convert this `condition` to `depends`, because of the `NOT` operator.
But then again, maybe this one doesn't add any new info. Presumably, if an extension is `promotedto="CL_VERSION_2_1"`, then all types it adds should only be defined when `CL_VERSION_2_1` is not defined.

But what about all the extensions with `depends` but not `condition` - are there cases where generating a macro condition in the header based on `depends` would be wrong?

PR #1090, opened 2024-03-25
Reference layered extensions in command-buffer intro

EwanC:
The "Interaction With Other Extensions" introductory section of `cl_khr_command_buffer` references future layered extensions as part of the design rationale.

Two of these layered extensions currently exist, so we can update this section to reference them directly as opposed to the existing speculative language.


bashbaug (commented):
Merging as discussed in the March 26th teleconference.

bashbaug (approved):

PR #1091, opened 2024-03-25
fixes a few incorrect asciidoctor attributes

bashbaug:
With these changes I believe there are no remaining asciidoctor attribute that are not substituted correctly.


bashbaug:
Merging as discussed in the March 26th teleconference.

PR #1104, opened 2024-03-26
fix a markup issue regarding cl_khr_subgroup_extended_types

bashbaug:
Looks like asciidoctor was getting confused by the "``s" at the end of "gentype", so just rephrase this sentence to avoid the confusion.


bashbaug:
Formatting change - merging without a review.

PR #1107, opened 2024-03-26
use the feature macro dictionary attributes in more places

### LINKED ISSUE #1100: increase use of asciidoctor feature dictionary
If we start including the OpenCL C feature dictionary in the OpenCL API spec (like is proposed for https://github.com/KhronosGroup/OpenCL-Docs/pull/1091), we should start to use it for more OpenCL C features described in the OpenCL API spec also.  For example, there are a lot of features mentioned in Appendix H (the backwards compabitlity appendix) that are currently formatted "manually", that could make use of the feature dictionary instead.

bashbaug:
fixes #1100


bashbaug:
This text is unchanged from the internally reviewed and ratified text.  Merging immediately.

PR #1108, opened 2024-03-26
update generated version text for extension APIs and enums

### LINKED ISSUE #1055: Preferred phrasing for generated extension dependencies
Currently #950 emits version-notes for extensions which are grammatically and probably technically wrong as they look just like the core version dependencies:

> {apiname} is missing before version cl_khr_some_extension

@bashbaug what would you recommend to use when the dependency is on an extension? I'm thinking

> {apiname} requires support for <<extension_name>>

or

> {apiname} requires the <<extension_name>> extension

which are consistent with patterns in a small number of handcoded comments already present.

bashbaug:
Fixes #1055.

### DIFF
--- a/scripts/gen_version_notes.py
+++ b/scripts/gen_version_notes.py
@@ -33,29 +33,37 @@ def GetFooter():
     return """
 """
 
-def FullNote(name, added_in, deprecated_by):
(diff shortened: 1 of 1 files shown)


oddhack:
LGTM.

PR #1112, opened 2024-03-28
Fix markup for a few extension names.

### LINKED ISSUE #1101: missing asciidoctor attribute value for the name of an extension?
There is one place in the OpenCL C spec that uses an extension name as an asciidoctor attribute (`cl_khr_integer_dot_product`):

```
ifdef::cl_khr_integer_dot_product[]
| {opencl_c_integer_dot_product_input_4x8bit_packed} +
  (when the {cl_khr_integer_dot_product} extension macro is defined)
| The OpenCL C compiler supports built-in functions that perform dot
products on 4x8 bit packed integer vectors.
```

This attribute is defined to get into this code section, but it is not currently defined to be anything, so in the rendered specification it looks like "(when the extension macro is defined)".  I don't think this is the expected behavior.

Should we define a value for this asciidoctor attribute?  Or should we just write the name of the extension manually?

oddhack:
Closes #1101

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -218,13 +218,13 @@ operations across a work-group.
 
 ifdef::cl_khr_integer_dot_product[]
 | {opencl_c_integer_dot_product_input_4x8bit_packed} +
-  (when the {cl_khr_integer_dot_product} extension macro is defined)
+  (when the `<<cl_khr_integer_dot_product>>` extension macro is defined)
 
 | The OpenCL C compiler supports built-in functions that perform dot
 products on 4x8 bit packed integer vectors
 
 | {opencl_c_integer_dot_product_input_4x8bit} +
-  (when the {cl_khr_integer_dot_product} extension macro is defined)
+  (when the `<<cl_khr_integer_dot_product>>` extension macro is defined)
 | The OpenCL C compiler supports built-in functions that perform dot
 products on 4x8 bit integer vectors
 endif::cl_khr_integer_dot_product[]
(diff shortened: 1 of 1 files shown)


oddhack:
@outofcontrol I'm seeing 'license/cla Expected - Waiting for status to be reported' still, about 90 minutes after the PR was submitted. Maybe the CLA bot needs to be kicked again? It seems OK on my other PR #1113 I submitted a few minutes later.

outofcontrol:
@oddhack Does this go away once the review is done? We have been seen this message from cla-assistant for a couple of weeks now. Not entirely clear on where the issue lies. Wondering if perhaps this is a new message that shows when the PR has not met all the other requirements.

oddhack:
> @oddhack Does this go away once the review is done? We have been seen this message from cla-assistant for a couple of weeks now. Not entirely clear on where the issue lies. Wondering if perhaps this is a new message that shows when the PR has not met all the other requirements.

Compare to #1113 - it also has requested reviews that have not been completed, but has cleared the CLA step. So that particular relationship does not seem to be the problem. Reviews are a separate block from 'checks' AFAICT - you can pass all the checks and still be blocked from merging because nobody has reviewed.

bashbaug (approved):

bashbaug:
I'm going to go ahead and merge this even though the CLA bot hasn't come back, since it's obviously OK in this case, but we really should figure out what's going on with it.

oddhack:
> I'm going to go ahead and merge this even though the CLA bot hasn't come back, since it's obviously OK in this case, but we really should figure out what's going on with it.

@outofcontrol it feels like the CLAbot is forgetting about the PR somehow - did not clear the condition after 4 days. Unfortunately one of those difficult to reproduce situations.

PR #1113, opened 2024-03-28
Fix provisional extension notes in appendices

### LINKED ISSUE #1106: update links and provisional extension disclaimers
There are a few items we should tidy up regarding provisional extensions:

1. There is auto-generated text for each provisional extension that says:

    > This is a provisional extension and must be used with caution. See the description of provisional header files for enablement and stability details.

    But, the "description" link goes nowhere.  We should find something to link it to, or remove the link.

2. Our current "provisional extension disclaimer" is included before the version history, which is a litle odd.  We should find a better place for it, or perhaps the auto-generated text described above could replace it?

oddhack:
The autogenerated metadata for provisional extensions includes a sentence linking to what was previously a dead anchor. The anchor now points to a short subsection at the start of the extensions appendix capturing substantially the same language as the explicitly included provisional_notice.asciidoc in the separate extension documents - it has been rephrased slightly and reads slightly better now, I think.

The explicit includes of provisional_notice.asciidoc have been removed (and also from some extensions that were recently ratified as non-provisional).

N.b. some older extensions that were ratified initially as provisional later had the extension version number bumped and a version commented added to the effect that they were no longer provisional. I don't know if that is really needed when they are ratified as non-provisional without any API changes, but for consistency that might be something you wish to do. It is out of scope for this PR and would require changes to both XML and the extension appendix version notes.

Closes #1106

### DIFF
--- a/api/appendix_extensions.asciidoc
+++ b/api/appendix_extensions.asciidoc
@@ -25,6 +25,22 @@ alphabetically by author ID.
 Within each group, extensions are listed in alphabetical order by their
 names.
 
+
+[[boilerplate-provisional-header]]
+== Provisional Extensions
+
+_Provisional_ OpenCL extensions described in this appendix have been
+Ratified under the Khronos Intellectual Property Framework.
+They are being made publicly available as provisional extensions to enable
+review and feedback from the community.
+While an extension is provisional, features may be added, removed, or
+changed in non-backward compatible ways.
+
+If you have feedback on a provisional extension, please create an issue on
+the link:https://github.com/KhronosGroup/OpenCL-Docs/[OpenCL-Docs
+repository].
+
+
 == Extension Dependencies
 
 Extensions which have dependencies on specific core versions or on other
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -419,8 +419,6 @@ features:
 *UNRESOLVED*
 --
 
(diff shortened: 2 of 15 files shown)


api/appendix_extensions.asciidoc
```diff
@@ -25,6 +25,22 @@ alphabetically by author ID.
 Within each group, extensions are listed in alphabetical order by their
 names.
 
+
+[[boilerplate-provisional-header]]
```
oddhack, line 29: N.b. this anchor name is hardwired into the generator script and looks a bit odd in context since Vulkan has an entire 'boilerplate' chapter for stuff like this. It would be a bit tedious to change just for the anchor name, but if the wording of the actual generated notice is inappropriate, I can factor that into the API conventions-specific part of the script along with the anchor name.

oddhack:
This got a weird error inside CI which disappeared when I re-ran it, thankfully. Something about not being able to create a directory that already existed for the specattribs.adoc generated file. Definitely does not happen in my local build but it might be something about make parallelization not having all the right dependencies?

api/appendix_extensions.asciidoc
bashbaug: I don't think this is worth worrying about, thanks for the heads up though!

bashbaug (approved):
If there are no other uses of the `provisional_notice.asciidoc` file - and I don't think there should be? - we should just remove it.  I'll take care of this in a subsequent PR, though.

PR #1115, opened 2024-03-28
Set anchor on `CL_COMMAND_BUFFER_MUTABLE_KHR`

EwanC:
The link in the "New Enums" section of cl_khr_command_buffer_mutable_dispatch doesn't lead anywhere otherwise.


oddhack (approved):
LGTM, thanks! I wouldn't be surprised if there were other missing anchors, converting the markup in the extension specs was pretty tedious.

bashbaug (approved):

PR #1116, opened 2024-03-29
Synchronize scripts with Vulkan, correct extension appendix section nesting

oddhack:
The observable effects of this are

- Pushes the subsection titles in the extension appendices down one level, similar to #1087 but keeping the scripts in sync
- Adds an autogenerated 'API Interactions' section with currently only affects the cl_khr_command_buffer extension, since that's the only one with some APIs tagged in the XML as dependent on a particular core version

Recommend this as a replacement for #1087


bashbaug (approved):
LGTM - I verified that the extensions are now nested properly.

PR #1117, opened 2024-03-31
clarify CL_DEVICE_TYPE_DEFAULT and CL_DEVICE_TYPE_ALL for custom devices

### LINKED ISSUE #300: CL_DEVICE_TYPE_ALL and custom devices
While working on #299 I found that querying `CL_DEVICE_TYPE_ALL` devices has this special case (**emphasis** mine):

> `CL_​DEVICE_​TYPE_​ALL`: All OpenCL devices available in the platform, **except for `CL_DEVICE_TYPE_CUSTOM` devices.**

Note that in the headers we have:

```c
/* cl_device_type - bitfield */
// snip
#ifdef CL_VERSION_1_2
#define CL_DEVICE_TYPE_CUSTOM                       (1 << 4)
#endif
#define CL_DEVICE_TYPE_ALL                          0xFFFFFFFF
```

This makes handling of `CL_DEVICE_TYPE_ALL` and `CL_DEVICE_TYPE_CUSTOM` awkward both for implementers and users - implementers cannot purely look at the bits of the passed-in _device_type_ to determine what is being queried, and users need to make two calls to `clGetDeviceIDs` to truly query for all of the devices in the platform.

Would it make more sense to eliminate the special handling for `CL_DEVICE_TYPE_CUSTOM`, and return **all** devices in the platform for `CL_DEVICE_TYPE_ALL` instead?

bashbaug:
fixes #300 

Fixes `CL_DEVICE_TYPE_ALL` - now it will return all devices in the platform, including custom devices.

Clarifies that `CL_DEVICE_TYPE_DEFAULT` may return a custom device if it is the only device in the platform.

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -395,13 +395,13 @@ include::{generated}/api/version-notes/CL_DEVICE_TYPE_DEFAULT.asciidoc[]
     {clGetDeviceIDs} or to create OpenCL contexts using
     {clCreateContextFromType}, and will never be returned in {CL_DEVICE_TYPE}
     for any OpenCL device.
-    The default OpenCL device must not be a {CL_DEVICE_TYPE_CUSTOM} device.
+    The default OpenCL device must not be a {CL_DEVICE_TYPE_CUSTOM} device
+    unless it is the only device in the platform.
 
 | {CL_DEVICE_TYPE_ALL_anchor}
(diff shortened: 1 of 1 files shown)


aharon-abramson (approved):

kpet (approved):

bashbaug:
Merging as discussed in the July 9th teleconference.

PR #1119, opened 2024-03-31
add and link error codes for extensions

### LINKED ISSUE #1095: inconsistent error codes in Appendix F
After #950, the error code appendix table has an error code for `cl_khr_icd`, but no error codes for other extensions.  Do we want to include error codes for extensions in this table, or should it only contain error codes for the core spec?

The primary reason for the error code appendix was to have a place for the error codes to link to (although the text description for each of the error codes is also a nice benefit).  If there is already some other place the extension error codes can link to we can use that instead, although currently many enum error codes do not seem to link to anything.

bashbaug:
Fixes #1095 

Adds error codes for all extensions to the table in appendix F.


oddhack:
LGTM. I'm in favor of having literally every API / header macro described in the API spec, but didn't want to be intrusive about that when doing the extension integration. This is great progress. I think there are still structures and macros in the XML that are not described in the spec.

PR #1122, opened 2024-03-31
cleans up extension metadata

### LINKED ISSUE #1098: check all "new structures" extension sections
Some of the extension descriptions now have a section for "new structures", but most of the types listed in this section do not seem to be structure types.  We should check these sections to determine which ones actually are structures and which should be moved to the "new types" section, instead.

### LINKED ISSUE #1099: do we want to use the "new SPIR-V capabilities" section?
The extension description for `cl_khr_integer_dot_product` now includes a section "New SPIR-V Capabilities", but it is currently marked "TBD".  Do we want to use this section?  If so, what should it contain?  If not, it should be removed.

bashbaug:
fixes #1098 
fixes #1099

* Use a consistent order for new commands, types, enums, etc.
* Add a few missing extension types
* Do not use the new SPIR-V capabilities section
* Use consistent terminology
* Do not include sections where nothing was added

### DIFF
--- a/api/cl_khr_command_buffer_multi_device.asciidoc
+++ b/api/cl_khr_command_buffer_multi_device.asciidoc
@@ -46,6 +46,10 @@ Depending on platform support the mapping of commands to the new target
 device can be done either explicitly by the user, or automatically by the
 OpenCL runtime.
 
+=== New Commands
+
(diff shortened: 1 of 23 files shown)


bashbaug:
Formatting change only, merging immediately.

PR #1124, opened 2024-04-01
fix section titles for sub-group extensions

### LINKED ISSUE #1102: tidy up sub-group built-in function section titles
We should take a pass through the sub-group built-in function section titles to ensure naming consistency.

We should also move the "Sub-Groups Function Mapping and Capabilities" section underneath "Sub-group Functions", rather than having it as its own section.

bashbaug:
fixes #1102 

Also fixes a few other section titles for consistency and one minor typo.


bashbaug:
This is a cosmetic fix, so merging immediately.

PR #1127, opened 2024-04-01
Restore static refpages to the build

### LINKED ISSUE #1121: reference page intro.html not being generated
It looks like the reference page "intro.html" is not being properly generated, so the initial reference page "index.html" is showing a missing page.  There is probably an issue in the build scripts somewhere, but I'm not exactly sure where the problem lies.

I am generating the reference pages with:

`python3 makeSpec -clean -spec khr OUTDIR=out -j32 manhtmlpages`

oddhack:
These were commented out during debugging and not restored. Also factors out a small difference between Vulkan and OpenCL scripts.

Closes #1121


bashbaug (approved):
LGTM - I've verified that the intro page appears after this change.

PR #1135, opened 2024-04-03
remove TODO and TBD placeholders

bashbaug:
Removes a bunch of TODO and TBD placeholders.  We can always add content for these later, if desired.


alycm (approved):
LGTM, seems non-controversial to remove these.

Issue #1044, opened 2024-01-15
cl_khr_fp16 extension consistency issues with core C spec


oddhack:
Some issues that have come up while integrating language from the extension to the core C spec for #950. These mostly arise from the situation that there is a considerable amount of cl_khr_fp16-related functionality in the C spec already, and many footnotes saying "only if cl_khr_fp16 is supported", but by no means all of the extension language. In particular, the functions operating on 'half' data types are not included, nor are compile-time constants. So it is unclear if the extension was already considered to be included in the C spec or not, but fully including it as I've been doing with the other C language extensions raises a bunch of questions:

- The extension spec says "The OpenCL compiler accepts an h and H suffix on floating-point literals, indicating the literal is typed as a half" while the C spec says "The `half` data type can only be used to declare a pointer to a buffer that contains half values", which seems like a major functionality difference. 

- The "Explicit Conversions" section excludes the 'half' data type while the extension says the conversion functions are extended to take 'half' and 'halfn'. Note this is unlike the implicit conversions / explicit casts sections, which exclude 'half' *unless* cl_khr_fp16 is supported, and I assume should be treated similarly.

- The Math Functions table for fract() refers to fmin(x - floor(x), 0x1.fffffep-1f), while the table in the extension refers to fmin(x - floor(x), 0x1.ffcp-1f). Why are the constants different? If the extension constant is intended to be a half-precision version of the fract() in the C spec, (a) shouldn't it have a 'h' suffix and (b) why isn't there also a double-precision constant version for the versions of fract() taking doubles?

- The nextafter() function in the C spec is defined to "Compute the next representable floating-point value following x in the direction of y" but accepts gentype arguments - should it not be "next representable FP value of the precision of its arguments" or similar?

bashbaug:
Hi @oddhack, sorry for the very slow response.

This is all a little tricky because the core OpenCL C specification allows minimal use of the "half" type, but only very minimal.  You need the `cl_khr_fp16` extension to make the "half" type a peer to the "float" or "double" types.  Specifically:

The core OpenCL C specification allows passing pointers to "half", but not declaring "half" variables, directly dereferencing the pointers to "half", or calling built-in functions that work on "half" variables.  The only way to read data out of a "half" pointer is to read-and-up-convert using a built-in function like `vload_half`, and the only way to store data into a "half" pointer is to down-convert-and-store using functions like `vstore_half`.

[...]

PR #1136, opened 2024-04-03
fix fract for double and half arguments

bashbaug:
see #1044, item (3).

Fixes the constant used in the description of the `fract` built-in function, for `double` and `half` arguments.


alycm (approved):
LGTM.

PR #1137, opened 2024-04-03
bugfix: the geometric functions do not operate componentwise

bashbaug:
The geometric functions are vector functions to compute things like dot products and cross products, so they do not operate componentwise.


alycm (approved):
Oops!  LGTM.

Issue #1128, opened 2024-04-01
CLAbot still hanging


oddhack:
@outofcontrol I created PR #1127 a little while ago and it seems that it, too, is suffering the 'license/cla Expected - Waiting for status to be reported' blockage in the checks. It looks like the majority of recently created PRs, and not just mine, are suffering this, which can be seen by going to the Pull Requests tab and looking for the brown dot icon next to the PR name. But not all of them. ATM it's #1127, #1126, #1123, #1119, #1088, while others like #1118, #1117, and #1103 have heard back from CLA.

oddhack:
Ping @bashbaugh since you're handling accepting PRs.

PR #1138, opened 2024-04-03
tidy up cl_khr_create_command_queue

bashbaug:
Removes unneeded special-case text regarding the type accepted for CL_QUEUE_PROPERTIES.  Calling this a cl_bitfield was a bug all along, and using cl_command_queue_properties is fine because it is a bitfield anyhow.

Generalizes some text so it is not specific to OpenCL 2.x devices.

Removes a TODO - no changes required.


alycm (approved):
LGTM, I agree that describing legacy aliases to cl_bitfield isn't worthwhile.

alycm:
CLA bot hasn't run, already documented in https://github.com/KhronosGroup/OpenCL-Docs/issues/1128, there are no CLA issues with this PR, merging.

PR #1144, opened 2024-04-04
update the quick reference list for finalized extensions

bashbaug:


bashbaug:
Cosmetic change, merging immediately.

PR #1146, opened 2024-04-05
Add multi-device wording to clCommandBarrierWithWaitListKHR

EwanC:
`clCommandBarrierWithWaitListKHR` does not have wording for the `command_queue` parameter regarding when
`cl_khr_command_buffer_multi_device` is enabled, and the error wording also does not take the multi-device extension into account. Taken wording from the other command recording entry-points.


bashbaug (approved):
LGTM!

bashbaug:
Merging as discussed in the July 16th teleconference.

PR #1159, opened 2024-05-07
cl_ext_buffer_device_address

pjaaskel:
The basic cl_mem buffer API doesn't enable access to the underlying raw pointers in the device memory, preventing its use in host side data structures that need pointer references to objects. This API adds a minimal increment on top of cl_mem that provides such capabilities.

The version 0.1.0 is implemented in PoCL and rusticl for prototyping, but everything's still up for discussion. chipStar is the first client that uses the API.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -617,6 +617,11 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT.asciido
    allocated buffers can be computed by adding the sub-buffer origin to the
    device-specific start address.
 
+   If the device supports SVM and {clCreateBufferWithProperties} is called with a pointer
+   returned by {clSVMAlloc} as its _host_ptr_ argument, and {CL_MEM_USE_HOST_PTR} is
+   set in its _flags_ argument, the device-side address is guaranteed to match
+   the _host_ptr_.
+
 endif::cl_ext_buffer_device_address[]
 
 |====
@@ -10954,7 +10959,8 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_KERNEL} if _kernel_ is a not a valid kernel object.
-  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_SVM_PTRS} if no devices in
+  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_SVM_PTRS} and
+     {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} if no devices in
      the context associated with _kernel_ support SVM.
 ifdef::cl_ext_buffer_device_address[]
   * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT} if no
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -602,7 +602,7 @@ ifdef::cl_ext_buffer_device_address[]
 
 include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT.asciidoc[]
   | {cl_bool_TYPE}
-      | When set to CL_TRUE, specifies that the buffer must have a single fixed
+      | When set to {CL_TRUE}, specifies that the buffer must have a single fixed
       device-side address for its lifetime, and the address can be queried via {clGetMemObjectInfo}.
 
    Each device in the context can have their own (fixed) device-side address and
@@ -692,9 +692,10 @@ ifdef::cl_khr_external_memory[]
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
 ifdef::cl_ext_buffer_device_address[]
-  * {CL_INVALID_DEVICE}
+  * {CL_INVALID_OPERATION}
   ** If _properties_ includes {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} and there
-  is at least one device in the context that doesn't support such allocation.
+  are no devices in the context that support the {cl_ext_buffer_device_address_EXT}
+  extension.
 endif::cl_ext_buffer_device_address[]
 
 [[memory-flags-table]]
@@ -6503,9 +6504,10 @@ ifdef::cl_ext_buffer_device_address[]
 include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
   |  {cl_mem_device_address_EXT_TYPE}[]
     | If _memobj_ was created using {clCreateBufferWithProperties} with
-      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to CL_TRUE,
+      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to {CL_TRUE},
       returns a list of device addresses for the buffer, one for each
-      device in the context.
+      device in the context in the same order as the list of devices
+      passed to {clCreateContext}.
 
 endif::cl_ext_buffer_device_address[]
 
@@ -6519,16 +6521,16 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_MEM_OBJECT} if _memobj_ is a not a valid memory object.
+ifdef::cl_ext_buffer_device_address[]
+  * {CL_INVALID_OPERATION} is returned for the {CL_MEM_DEVICE_ADDRESS_EXT} query if
+     the {cl_ext_buffer_device_address_EXT} is not supported or if the
+     buffer was not allocated with {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}.
+endif::cl_ext_buffer_device_address[]
   * {CL_INVALID_VALUE} if _param_name_ is not one of the supported values, or
     if the size in bytes specified by _param_value_size_ is less than size of
     the return type specified in the
     <<mem-info-table, Memory Object Queries>> table
     and _param_value_ is not `NULL`.
-ifdef::cl_ext_buffer_device_address[]
-  ** Returned for the {CL_MEM_DEVICE_ADDRESS_EXT} query if
-     the {cl_ext_buffer_device_address_EXT} is not supported or if the
-     buffer was not allocated with {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}.
-endif::cl_ext_buffer_device_address[]
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
     by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
@@ -10854,16 +10856,17 @@ include::{generated}/api/version-notes/clSetKernelArgDevicePointerEXT.asciidoc[]
     the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property. The pointer value specified as
     the argument value can be the pointer to the beginning of the buffer or any offset into
     the buffer region. The device pointer value must be naturally aligned according to
-    the argument's type.
+    the argument's type. It should be noted that it's legal to pass invalid
+    pointers as the value (similarly to C/C++ function calls with pointer arguments) as
+    long as the kernel doesn't dereference the pointer.
 
 {clSetKernelArgDevicePointerEXT} returns {CL_SUCCESS} if the argument was set
 successfully. Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_KERNEL} if _kernel_ is not a valid kernel object.
   * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
-    the device pointer.
+    the {cl_ext_buffer_device_address_EXT} extension.
   * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
-  * {CL_INVALID_ARG_VALUE} if _arg_value_ specified is not a valid value.
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
     by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
@@ -10964,7 +10967,8 @@ Otherwise, it returns one of the following errors:
      the context associated with _kernel_ support SVM.
 ifdef::cl_ext_buffer_device_address[]
   * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT} if no
-  device in the context associated with _kernel_ support device pointers.
+  device in the context associated with _kernel_ support the {cl_ext_buffer_device_address_EXT}
+  extension.
 endif::cl_ext_buffer_device_address[]
   * {CL_INVALID_VALUE} if _param_name_ is not valid, if _param_value_ is
     `NULL` or if the size specified by _param_value_size_ is not valid.
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -595,6 +595,39 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_HANDLE_LIST_KHR.asciidoc[]
         {CL_MEM_DEVICE_HANDLE_LIST_END_KHR_anchor}) to associate with the
         external memory handle.
 endif::cl_khr_external_memory[]
+
+ifdef::cl_ext_buffer_device_address[]
+
+| {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT.asciidoc[]
+  | {cl_bool_TYPE}
+      | When set to CL_TRUE, specifies that the buffer must have a single fixed
+      device-side address for its lifetime, and the address can be queried via {clGetMemObjectInfo}.
+
+   Each device in the context can have their own (fixed) device-side address and
+   a copy of the created buffer which are synchronized
+   implicitly by the runtime.
+
+   The flag might imply that the buffer will be "pinned" permanently to
+   a device's memory, but might not be necessarily so, as long as the address
+   range of the buffer remains constant.
+
+   The device addresses of sub-buffers derived from {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}
+   allocated buffers can be computed by adding the sub-buffer origin to the
+   device-specific start address.
+
+| {CL_MEM_DEVICE_SHARED_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_SHARED_ADDRESS_EXT.asciidoc[]
+  | {cl_bool_TYPE}
+      | When set to CL_TRUE, the buffer has otherwise the same properties as
+      when allocated using the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT_anchor} flag,
+      but with an additional property that the buffer's address is the same across
+      all the devices in the context.
+
+endif::cl_ext_buffer_device_address[]
+
 |====
 
 ifdef::cl_khr_external_memory[]
@@ -662,6 +695,15 @@ ifdef::cl_khr_external_memory[]
      {CL_MEM_DEVICE_HANDLE_LIST_KHR} is specified as part of _properties_.
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
+ifdef::cl_ext_buffer_device_address[]
+  * {CL_INVALID_DEVICE}
+  ** If _properties_ includes either {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} or
+  {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} and there is at least one device in
+  the context that doesn't support such allocation.
+  * {CL_INVALID_VALUE}
+  ** If _properties_ includes both {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} and
+  {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} at the same time.
+endif::cl_ext_buffer_device_address[]
 
 [[memory-flags-table]]
 .List of supported memory flag values
@@ -6463,6 +6505,21 @@ include::{generated}/api/version-notes/CL_MEM_D3D11_RESOURCE_KHR.asciidoc[]
       returns the _resource_ argument specified when _memobj_ was created.
 endif::cl_khr_d3d11_sharing[]
 
+ifdef::cl_ext_buffer_device_address[]
+| {CL_MEM_DEVICE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
+  |  {cl_mem_device_address_EXT_TYPE}[]
+    | If _memobj_ was created using {clCreateBufferWithProperties} with
+      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to CL_TRUE,
+      returns a list of device addresses for the buffer, one for each
+      device in the context. If the buffer was allocated
+      with the {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} property,
+      only one device address is returned.
+
+endif::cl_ext_buffer_device_address[]
+
+
 |====
 
 // refError
@@ -6477,6 +6534,12 @@ Otherwise, it returns one of the following errors:
     the return type specified in the
     <<mem-info-table, Memory Object Queries>> table
     and _param_value_ is not `NULL`.
+ifdef::cl_ext_buffer_device_address[]
+  ** Returned for the {CL_MEM_DEVICE_ADDRESS_EXT} query if
+     the {cl_ext_buffer_device_address_EXT} is not supported or if the
+     buffer was not allocated with neither {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} or
+     {CL_MEM_DEVICE_SHARED_ADDRESS_EXT}.
+endif::cl_ext_buffer_device_address[]
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
     by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
@@ -10778,6 +10841,48 @@ Otherwise, it returns one of the following errors:
     required by the OpenCL implementation on the host.
 --
 
+ifdef::cl_ext_buffer_device_address[]
+[open,refpage='clSetKernelArgDevicePointerEXT',desc='Set a device pointer as the argument value for a specific argument of a kernel.',type='protos']
+--
+To set a device pointer as the argument value for a specific argument of a
+kernel, call the function
+
+include::{generated}/api/protos/clSetKernelArgDevicePointerEXT.txt[]
+include::{generated}/api/version-notes/clSetKernelArgDevicePointerEXT.asciidoc[]
+
+  * _kernel_ is a valid kernel object.
+  * _arg_index_ is the argument index.
+    Arguments to the kernel are referred by indices that go from 0 for the
+    leftmost argument to _n_ - 1, where _n_ is the total number of arguments
+    declared by a kernel.
+  * _arg_value_ is the device pointer that should be used as the argument value for
+    argument specified by _arg_index_.
+    The device pointer specified is the value used by all API calls that enqueue
+    _kernel_ ({clEnqueueNDRangeKernel} and {clEnqueueTask}) until the argument
+    value is changed by a call to {clSetKernelArgDevicePointerEXT} for _kernel_.
+    The device pointer can only be used for arguments that are declared to be a
+    pointer to `global` memory allocated with {clCreateBufferWithProperties} with
+    either the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} or {CL_MEM_DEVICE_SHARED_ADDRESS_EXT}
+    property. The pointer value specified as the argument value
+    can be the pointer to the beginning of the buffer or any offset into
+    the buffer region. The device pointer value must be naturally aligned according to
+    the argument's type.
+
+{clSetKernelArgDevicePointerEXT} returns {CL_SUCCESS} if the argument was set
+successfully. Otherwise, it returns one of the following errors:
+
+  * {CL_INVALID_KERNEL} if _kernel_ is not a valid kernel object.
+  * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
+    the device pointer.
+  * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
+  * {CL_INVALID_ARG_VALUE} if _arg_value_ specified is not a valid value.
+  * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
+    by the OpenCL implementation on the device.
+  * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
+    required by the OpenCL implementation on the host.
+--
+endif::cl_ext_buffer_device_address[]
+
 [open,refpage='clSetKernelExecInfo',desc='Set additional execution information for a kernel.',type='protos']
 --
 To set additional execution information for a kernel, call the function
@@ -10844,6 +10949,19 @@ include::{generated}/api/version-notes/CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM
         If {clSetKernelExecInfo} has not been called with a value for
         {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM}, the default value is
         {CL_TRUE}.
+
+ifdef::cl_ext_buffer_device_address[]
+| {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT.asciidoc[]
+  | {cl_mem_device_address_EXT_TYPE}[]
+      | Device pointers must reference locations contained entirely within
+        buffers that are passed to kernel as arguments, or that are passed
+        through the execution information. Non-argument device pointers accessed
+        by the kernel must be specified by passing pointers to those buffers
+        via this {clSetKernelExecInfo} option.
+endif::cl_ext_buffer_device_address[]
+
 |====
 
 // refError
@@ -10853,7 +10971,14 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_KERNEL} if _kernel_ is a not a valid kernel object.
-  * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support SVM.
+  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_SVM_PTRS} if no devices in
+     the context associated with _kernel_ support SVM.
+ifdef::cl_ext_buffer_device_address[]
+  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT} if no
+  device in the context associated with _kernel_ support device pointers.
+endif::cl_ext_buffer_device_address[]
+  * {CL_INVALID_VALUE} if _param_name_ is not valid, if _param_value_ is
+    `NULL` or if the size specified by _param_value_size_ is not valid.
   * {CL_INVALID_OPERATION} if _param_name_ is
     {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} and _param_value_ is {CL_TRUE}
     and no devices in the context associated with _kernel_ support fine-grain
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -617,15 +617,6 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT.asciido
    allocated buffers can be computed by adding the sub-buffer origin to the
    device-specific start address.
 
-| {CL_MEM_DEVICE_SHARED_ADDRESS_EXT_anchor}
-
-include::{generated}/api/version-notes/CL_MEM_DEVICE_SHARED_ADDRESS_EXT.asciidoc[]
-  | {cl_bool_TYPE}
-      | When set to CL_TRUE, the buffer has otherwise the same properties as
-      when allocated using the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT_anchor} flag,
-      but with an additional property that the buffer's address is the same across
-      all the devices in the context.
-
 endif::cl_ext_buffer_device_address[]
 
 |====
@@ -697,12 +688,8 @@ ifdef::cl_khr_external_memory[]
 endif::cl_khr_external_memory[]
 ifdef::cl_ext_buffer_device_address[]
   * {CL_INVALID_DEVICE}
-  ** If _properties_ includes either {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} or
-  {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} and there is at least one device in
-  the context that doesn't support such allocation.
-  * {CL_INVALID_VALUE}
-  ** If _properties_ includes both {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} and
-  {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} at the same time.
+  ** If _properties_ includes {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} and there
+  is at least one device in the context that doesn't support such allocation.
 endif::cl_ext_buffer_device_address[]
 
 [[memory-flags-table]]
@@ -6513,9 +6500,7 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
     | If _memobj_ was created using {clCreateBufferWithProperties} with
       the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to CL_TRUE,
       returns a list of device addresses for the buffer, one for each
-      device in the context. If the buffer was allocated
-      with the {CL_MEM_DEVICE_SHARED_ADDRESS_EXT} property,
-      only one device address is returned.
+      device in the context.
 
 endif::cl_ext_buffer_device_address[]
 
@@ -6537,8 +6522,7 @@ Otherwise, it returns one of the following errors:
 ifdef::cl_ext_buffer_device_address[]
   ** Returned for the {CL_MEM_DEVICE_ADDRESS_EXT} query if
      the {cl_ext_buffer_device_address_EXT} is not supported or if the
-     buffer was not allocated with neither {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} or
-     {CL_MEM_DEVICE_SHARED_ADDRESS_EXT}.
+     buffer was not allocated with {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}.
 endif::cl_ext_buffer_device_address[]
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
     by the OpenCL implementation on the device.
@@ -10862,9 +10846,8 @@ include::{generated}/api/version-notes/clSetKernelArgDevicePointerEXT.asciidoc[]
     value is changed by a call to {clSetKernelArgDevicePointerEXT} for _kernel_.
     The device pointer can only be used for arguments that are declared to be a
     pointer to `global` memory allocated with {clCreateBufferWithProperties} with
-    either the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} or {CL_MEM_DEVICE_SHARED_ADDRESS_EXT}
-    property. The pointer value specified as the argument value
-    can be the pointer to the beginning of the buffer or any offset into
+    the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property. The pointer value specified as
+    the argument value can be the pointer to the beginning of the buffer or any offset into
     the buffer region. The device pointer value must be naturally aligned according to
     the argument's type.
 
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -3741,7 +3741,7 @@ server's OpenCL/api-docs repository.
             <proto><type>cl_int</type>                                  <name>clSetKernelArgDevicePointerEXT</name></proto>
             <param><type>cl_kernel</type>                               <name>kernel</name></param>
             <param><type>cl_uint</type>                                 <name>arg_index</name></param>
-            <param>const <type>cl_mem_device_address_ext</type>*        <name>arg_value</name></param>
+            <param><type>cl_mem_device_address_ext</type>               <name>arg_value</name></param>
         </command>
         <command suffix="CL_API_SUFFIX__VERSION_2_0">
             <proto><type>cl_int</type>                                  <name>clSetKernelExecInfo</name></proto>
@@ -7204,7 +7204,7 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
-        <extension name="cl_ext_buffer_device_address" revision="1.0.0" supported="opencl" depends="CL_VERSION_3_0">
+        <extension name="cl_ext_buffer_device_address" revision="1.0.2" supported="opencl" depends="CL_VERSION_3_0">
             <require>
                 <type name="cl_mem_device_address_ext"/>
             </require>
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -720,8 +720,6 @@ server's OpenCL/api-docs repository.
         <enum value="((cl_device_partition_property_ext)0 - 1)"    name="CL_PARTITION_BY_NAMES_LIST_END_EXT"/>
         <enum value="-1"            name="CL_PARTITION_BY_NAMES_LIST_END_INTEL"/>
         <enum value="0"             name="CL_MEM_DEVICE_HANDLE_LIST_END_KHR"/>
-        <enum value="1"             name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
-        <enum value="2"             name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
         <enum value="0"             name="CL_SEMAPHORE_DEVICE_HANDLE_LIST_END_KHR"/>
         <enum value="0"             name="CL_SEMAPHORE_EXPORT_HANDLE_TYPES_LIST_END_KHR"/>
     </enums>
@@ -914,7 +912,7 @@ server's OpenCL/api-docs repository.
         <enum bitpos="38"           name="CL_MEM_RESERVED1_QCOM"/>
         <enum bitpos="39"           name="CL_MEM_RESERVED2_QCOM"/>
         <enum bitpos="40"           name="CL_MEM_RESERVED3_QCOM"/>
-            <unused start="43" end="63"/>
+            <unused start="41" end="63"/>
     </enums>
 
     <enums name="cl_map_flags" vendor="Khronos" type="bitmask">
@@ -1633,8 +1631,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x1108"        name="CL_MEM_OFFSET"/>
         <enum value="0x1109"        name="CL_MEM_USES_SVM_POINTER"/>
         <enum value="0x110A"        name="CL_MEM_PROPERTIES"/>
-        <enum value="0x110B"        name="CL_MEM_DEVICE_ADDRESS_EXT"/>
-             <unused start="0x110D" end="0x110F" comment="Reserved for cl_mem_info"/>
+            <unused start="0x110B" end="0x110F" comment="Reserved for cl_mem_info"/>
         <enum value="0x1110"        name="CL_IMAGE_FORMAT"/>
         <enum value="0x1111"        name="CL_IMAGE_ELEMENT_SIZE"/>
         <enum value="0x1112"        name="CL_IMAGE_ROW_PITCH"/>
@@ -1727,8 +1724,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x11B8"        name="CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT"/>
         <enum value="0x11B9"        name="CL_KERNEL_MAX_NUM_SUB_GROUPS"/>
         <enum value="0x11BA"        name="CL_KERNEL_COMPILE_NUM_SUB_GROUPS"/>
-        <enum value="0x11BB"        name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
-            <unused start="0x11BC" end="0x11CF" comment="Reserved for cl_kernel_info / cl_kernel_work_group_info / cl_kernel_exec_info / cl_kernel_sub_group_info"/>
+            <unused start="0x11BB" end="0x11CF" comment="Reserved for cl_kernel_info / cl_kernel_work_group_info / cl_kernel_exec_info / cl_kernel_sub_group_info"/>
         <enum value="0x11D0"        name="CL_EVENT_COMMAND_QUEUE"/>
         <enum value="0x11D1"        name="CL_EVENT_COMMAND_TYPE"/>
         <enum value="0x11D2"        name="CL_EVENT_REFERENCE_COUNT"/>
@@ -2318,6 +2314,12 @@ server's OpenCL/api-docs repository.
 
     <enums start="0x42B0" end="0x4FFF" name="enums.42B0" comment="Reserved for vendor extensions. Allocate in groups of 16.">
             <unused start="0x42B0" end="0x4FFF"/>
+    </enums>
+
+    <enums start="0x5000" end="0x500F" name="enums.5000" comment="For cl_ext_buffer_device_address">
+        <enum value="0x5000"        name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
+        <enum value="0x5001"        name="CL_MEM_DEVICE_ADDRESS_EXT"/>
+        <enum value="0x5002"        name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
     </enums>
 
     <enums start="0x10000" end="0x10FFF" name="enums.10000" vendor="Khronos" comment="Experimental range for internal development only. Do not allocate.">
@@ -3739,7 +3741,7 @@ server's OpenCL/api-docs repository.
             <proto><type>cl_int</type>                                  <name>clSetKernelArgDevicePointerEXT</name></proto>
             <param><type>cl_kernel</type>                               <name>kernel</name></param>
             <param><type>cl_uint</type>                                 <name>arg_index</name></param>
-            <param>const <type>void</type>*                             <name>arg_value</name></param>
+            <param>const <type>cl_mem_device_address_ext</type>*        <name>arg_value</name></param>
         </command>
         <command suffix="CL_API_SUFFIX__VERSION_2_0">
             <proto><type>cl_int</type>                                  <name>clSetKernelExecInfo</name></proto>
@@ -7202,12 +7204,14 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
-        <extension name="cl_ext_buffer_device_address" revision="0.9.0" supported="opencl" depends="CL_VERSION_3_0" provisional="true">
+        <extension name="cl_ext_buffer_device_address" revision="0.9.1" supported="opencl" depends="CL_VERSION_3_0" provisional="true">
+            <require>
+                <type name="cl_mem_device_address_ext"/>
+            </require>
             <require>
                 <command name="clSetKernelArgDevicePointerEXT"/>
             </require>
             <require comment="cl_mem_properties">
-                <enum name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
                 <enum name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
             </require>
             <require comment="cl_mem_info">
@@ -7215,9 +7219,6 @@ server's OpenCL/api-docs repository.
             </require>
             <require comment="cl_kernel_exec_info">
                 <enum name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
-            </require>
-            <require>
-                <type name="cl_mem_device_address_ext"/>
             </require>
         </extension>
         <extension name="cl_khr_command_buffer" revision="0.9.7" supported="opencl" depends="CL_VERSION_1_2" ratified="opencl" provisional="true">
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -7204,7 +7204,7 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
-        <extension name="cl_ext_buffer_device_address" revision="0.9.1" supported="opencl" depends="CL_VERSION_3_0" provisional="true">
+        <extension name="cl_ext_buffer_device_address" revision="1.0.0" supported="opencl" depends="CL_VERSION_3_0">
             <require>
                 <type name="cl_mem_device_address_ext"/>
             </require>
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -255,6 +255,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_platform_command_buffer_capabilities_khr</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mutable_dispatch_asserts_khr</name></type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_kernel_clock_capabilities_khr</name>;</type>
+        <type category="define">typedef <type>cl_ulong</type>      <name>cl_mem_device_address_ext</name>;</type>
 
             <comment>Structure types</comment>
         <type category="struct" name="cl_dx9_surface_info_khr">
@@ -719,6 +720,8 @@ server's OpenCL/api-docs repository.
         <enum value="((cl_device_partition_property_ext)0 - 1)"    name="CL_PARTITION_BY_NAMES_LIST_END_EXT"/>
         <enum value="-1"            name="CL_PARTITION_BY_NAMES_LIST_END_INTEL"/>
         <enum value="0"             name="CL_MEM_DEVICE_HANDLE_LIST_END_KHR"/>
+        <enum value="1"             name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
+        <enum value="2"             name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
         <enum value="0"             name="CL_SEMAPHORE_DEVICE_HANDLE_LIST_END_KHR"/>
         <enum value="0"             name="CL_SEMAPHORE_EXPORT_HANDLE_TYPES_LIST_END_KHR"/>
     </enums>
@@ -911,7 +914,7 @@ server's OpenCL/api-docs repository.
         <enum bitpos="38"           name="CL_MEM_RESERVED1_QCOM"/>
         <enum bitpos="39"           name="CL_MEM_RESERVED2_QCOM"/>
         <enum bitpos="40"           name="CL_MEM_RESERVED3_QCOM"/>
-            <unused start="41" end="63"/>
+            <unused start="43" end="63"/>
     </enums>
 
     <enums name="cl_map_flags" vendor="Khronos" type="bitmask">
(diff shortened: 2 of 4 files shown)


extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,289 @@
...
+| {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT.asciidoc[]
+  | {cl_mem_device_address_EXT_TYPE}
+      | Device pointers must reference locations contained entirely within
+        buffers that are passed to kernel as arguments, or that are passed
+        through the execution information.
+
+        Non-argument device pointers accessed by the kernel must be specified
+        by passing pointers to those buffers via {clSetKernelExecInfo}.
```
karolherbst, line 258: I think we could discuss whether we want to require applications to always pass in the full list of buffers, because that would allow drivers to not having to keep track of `bda` allocations.  But that also depends on if "this memory allocation is used in a dispatch" being relevant for the most/some OpenCL implementations in the first place.

But one can also argue that it should be convenient for users to use this extension, so they don't have to keep track themselves, but that makes the hot path (launching many kernels) more expensive as at launch time the latest you need to map back from pointer to buffer.

But I think it's also fine to keep it like this, because it's closer to how it's done for SVM.

karolherbst:
any motivation to get this merged? Or anything else needed to discuss before merging this? Could also try to bring it up at the CL WG if needed.

pjaaskel:
> any motivation to get this merged? Or anything else needed to discuss before merging this? Could also try to bring it up at the CL WG if needed.

Yep, this is still being discussed in the WG. I personally think it's useful as is and shouldn't harm anything if merged as it even has 2 implementations now.

xml/cl.xml
```diff
@@ -7194,6 +7211,18 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
+        <extension name="cl_ext_buffer_device_address" supported="opencl">
```
SunSerega, line 7214: This fails with the latest state of the `main` branch:

> element "extension" missing required attribute "revision"

SunSerega:
```suggestion
        <extension name="cl_ext_buffer_device_address" revision="0.1.0" supported="opencl">
```

SunSerega: ([from .asciidoc spec](https://github.com/KhronosGroup/OpenCL-Docs/pull/1159/files#diff-cd49946e4c7aaedc4d5adc5bd2f624d351c550d06177b1c66ad48a04b6c4196f)

SunSerega: And with this small fix, XML works, but .asciidoc fails just like [here](https://github.com/KhronosGroup/OpenCL-Docs/pull/1171#issuecomment-2308994847), but seemingly on a later call:
https://github.com/SunSerega/OpenCL-Docs/actions/runs/10550058760/job/29225797422#step:8:126

xml/cl.xml
```diff
@@ -255,6 +255,8 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_platform_command_buffer_capabilities_khr</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mutable_dispatch_asserts_khr</name></type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_kernel_clock_capabilities_khr</name>;</type>
+        <type category="define">typedef <type>cl_ulong</type>      <name>cl_mem_device_address_EXT</name>;</type>
+        <type category="define">typedef struct _cl_mem_device_address_pair_EXT* <name>cl_mem_device_address_pair_EXT</name>;</type>
```
SunSerega, line 259: This type is defined below as a `struct`, so an opaque handle definition is redundant and conflicting.
```suggestion
```

xml/cl.xml
```diff
@@ -7194,6 +7211,18 @@ server's OpenCL/api-docs repository.
...
+            <require>
+                <command name="clSetKernelArgDevicePointerEXT"/>
+                <enum name="CL_MEM_DEVICE_ADDRESS_EXT"/>
+                <enum name="CL_MEM_DEVICE_PRIVATE_EXT"/>
+                <enum name="CL_MEM_DEVICE_PTR_EXT"/>
+                <enum name="CL_MEM_DEVICE_PTRS_EXT"/>
+                <enum name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
+                <type name="cl_mem_device_address_EXT"/>
+                <type name="cl_mem_device_address_pair_EXT"/>
+            </require>
```
SunSerega, line 7224: All other extensions separate their `<require>` blocks like this:
```suggestion
            <require>
                <command name="clSetKernelArgDevicePointerEXT"/>
            </require>
            <require comment="cl_mem_flags">
                <enum name="CL_MEM_DEVICE_ADDRESS_EXT"/>
                <enum name="CL_MEM_DEVICE_PRIVATE_EXT"/>
            </require>
            <require comment="cl_mem_info">
                <enum name="CL_MEM_DEVICE_PTR_EXT"/>
                <enum name="CL_MEM_DEVICE_PTRS_EXT"/>
            </require>
            <require comment="cl_kernel_exec_info">
                <enum name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
            </require>
            <require>
                <type name="cl_mem_device_address_EXT"/>
                <type name="cl_mem_device_address_pair_EXT"/>
            </require>
```

pjaaskel:
Thanks @SunSerega

SunSerega:
Alright, and now the problem I found in #1171 is visible here because the `Presubmit` workflow has been properly rerun.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,289 @@
...
+.List of supported memory flags by {clCreateBuffer}
+[width="100%",cols="<50%,<50%",options="header"]
+|====
+| Memory Flags | Description
+| {CL_MEM_DEVICE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
+  | This flag specifies that the buffer must have a single fixed address
+   for its lifetime and the address should be unique at least across the devices
+   of the context, but not necessarily withing the host (virtual) memory.
```
karolherbst, line 150: I think support for this flag needs to be optional, as otherwise this extension will be difficult to implement on current Vulkan implementations.

With Vulkan you can use bda (optional in 1.2, mandatory in 1.3) to have a fixed address for a memory allocation, but a CL runtime wouldn't be able to guarantee multiple devices to get the same address.

karolherbst: or maybe it's better if it's optional for multi device contexts, so applications and libraries could support the same code on single device contexts as well.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,289 @@
...
+  * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
+    the device pointer.
+  * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
+  * {CL_INVALID_ARG_VALUE} if _arg_value_ specified is not a valid value.
+  * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
+    by the OpenCL implementation on the device.
+  * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
+    required by the OpenCL implementation on the host.
+
+Add a new flag to clSetKernelExecInfo for setting indirect device pointer access info <<clSetKernelExecInfo, List of supported param_name stable>>:
```
karolherbst, line 244: `clSetKernelExecInfo` needs to be adjusted a little more. Currently it states: `CL_INVALID_OPERATION if no devices in the context associated with kernel support SVM.` This needs to change to account for devices not supporting SVM to weaken the condition. E.g.

`CL_INVALID_OPERATION if CL_KERNEL_EXEC_INFO_SVM_PTRS or CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM are specified and no device in the context associated with the kernel supports SVM`

`CL_INVALID_OPERATION if CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT is specified and no device in the context associated with the kernel supports bda`

karolherbst:
One thing I'm wondering about is how should `clCreateSubBuffer` behave when being executed on a bda memory object? Should it fail or succeed? I'm fine with either, just wondering if some of the behavior needs to be formalized.

`CL_MEM_DEVICE_PTR_EXT` and `CL_MEM_DEVICE_PTRS_EXT` comes to my mind, which should probably return the address + the sub buffer offset, at least that's how I have implemented it so far.

Also the sub buffer will inherit the flags from the parent buffer, but I think that's fine.

pjaaskel:
> One thing I'm wondering about is how should `clCreateSubBuffer` behave when being executed on a bda memory object? Should it fail or succeed? I'm fine with either, just wondering if some of the behavior needs to be formalized.
> 
> `CL_MEM_DEVICE_PTR_EXT` and `CL_MEM_DEVICE_PTRS_EXT` comes to my mind, which should probably return the address + the sub buffer offset, at least that's how I have implemented it so far.

Yes, this was the idea. I'll add a mention in the next update.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,289 @@
...
+   CL_MEM_DEVICE_ADDRESS_EXT. If the buffer was not created with the flag
+   or there are multiple devices in the context and the buffer address is
+   not the same for all of them, it returns CL_INVALID_MEM_OBJECT.
+
+| {CL_MEM_DEVICE_PTRS_EXT_anchor}
+include::{generated}/api/version-notes/CL_MEM_DEVICE_PTRS_EXT.asciidoc[]
+  | {cl_mem_device_address_pair_EXT_TYPE}
+      | Returns the device-address pairs for all devices in the context.
+      The per-device addresses might differ when the buffer was allocated
+      with the CL_MEM_DEVICE_PRIVATE_EXT enabled.
```
karolherbst, line 198: What should happen if you mix devices supporting bda with those who don't? Is 0 returned for the devices without support for bda or should the entire query fail?

karolherbst: Nevermind, the creation of the memory should have failed, so such a `cl_mem` object shouldn't exist.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,315 @@
...
+   this type of allocations, an error (CL_INVALID_VALUE) is returned.
+
+   The device addresses of sub-buffers derived from CL_MEM_DEVICE_ADDRESS_EXT
+   allocated buffers can be computed by adding the sub-buffer origin to the
+   start address.
+
+| {CL_MEM_DEVICE_PRIVATE_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_EXT.asciidoc[]
+  | If this flag is combined with CL_MEM_DEVICE_ADDRESS_EXT, each device in
```
karolherbst, line 171: I think it makes more sense to remove the `combined with CL_MEM_DEVICE_ADDRESS_EXT` part, then those two flags are independent.

Then a `CL_DEVICE_SUPPORTS_MULTI_DEVICE_ADDRESS` query could be added to `clGetDeviceInfo` to indicate if `CL_MEM_DEVICE_ADDRESS_EXT` is supported when the device is used in a context with other devices.

This should be enough to make this extension supportable on vulkan as mentioned earlier.

karolherbst (approved):
I've already implemented the extension fully in rusticl/mesa (including sharing the same address across devices) and I think it's fine, however I'd still urge to address the concerns I have for layered implementations implementing it on top of Vulkan. I've already considered the constraints when implementing it, however I think it's better to provide clients to query if address sharing across multiple devices is supported or not.

xml/cl.xml
```diff
@@ -7186,7 +7202,26 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
-        <extension name="cl_khr_command_buffer" revision="0.9.5" supported="opencl" depends="CL_VERSION_1_2" ratified="opencl" provisional="true">
```
SunSerega, line 7189: It seems you've accidentally deleted the first line of the `cl_khr_command_buffer` definition.

xml/cl.xml
```diff
@@ -7186,7 +7202,26 @@ server's OpenCL/api-docs repository.
...
+        <extension name="cl_ext_buffer_device_address" revision="0.1.0" supported="opencl">
+            <require>
+                <command name="clSetKernelArgDevicePointerEXT"/>
+            </require>
+            <require comment="cl_mem_flags">
+                <enum name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
+                <enum name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
```
SunSerega, line 7211: Also, you forgot to update enum names elsewhere in the XML after renaming them.

pjaaskel:
@SunSerega thanks!

pjaaskel:
@karolherbst I asked about this in the CL/memory WG. We need to submit CTS tests and this might be good to go then with this one. Do you have good tests in Rusticl side we could use? The test in PoCL is quite basic (and needs to be updated), but can be used as a starting point also.

karolherbst:
> @karolherbst I asked about this in the CL/memory WG. We need to submit CTS tests and this might be good to go then with this one. Do you have good tests in Rusticl side we could use? The test in PoCL is quite basic (and needs to be updated), but can be used as a starting point also.

I haven't written any more tests and only used the one in pocl. But I can probably help out with writing tests here.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+   CL_MEM_DEVICE_SHARED_ADDRESS_EXT. If the buffer was not created with the flag
+   or there are multiple devices in the context and the buffer address is
+   not the same for all of them, it returns CL_INVALID_MEM_OBJECT.
+
+| {CL_MEM_DEVICE_PTRS_EXT_anchor}
+include::{generated}/api/version-notes/CL_MEM_DEVICE_PTRS_EXT.asciidoc[]
+  | {cl_mem_device_address_pair_EXT_TYPE}
+      | Returns the device-address pairs for all devices in the context.
+      The per-device addresses might differ when the buffer was allocated
+      with the CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT enabled.
```
kpet, line 207: It might be simpler to just return an array of device addresses, with the ordering matching that of the devices in the context. You could also avoid using two separate queries. The array could contain multiple times the same value.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+    leftmost argument to _n_ - 1, where _n_ is the total number of arguments
+    declared by a kernel.
+  * _arg_value_ is the device pointer that should be used as the argument value for
+    argument specified by _arg_index_.
+    The device pointer specified is the value used by all API calls that enqueue
+    _kernel_ ({clEnqueueNDRangeKernel} and {clEnqueueTask}) until the argument
+    value is changed by a call to {clSetKernelArgDevicePointer} for _kernel_.
+    The device pointer can only be used for arguments that are declared to be a
+    pointer to `global` memory allocated with clCreateBuffer() with the
+    CL_MEM_DEVICE_SHARED_ADDRESS_EXT flag. The pointer value specified as the argument value
```
kpet, line 233: Presumably those created with `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT` can be used too, can't they?

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+// refError
+
+Change the descriptions for when returning CL_INVALID_OPERATION from {clSetKernelExecInfo}
+as follows:
+
+  * {CL_INVALID_OPERATION} if passing {CL_KERNEL_EXEC_INFO_SVM_PTRS} or
+  {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} with _param_value_ set to CL_TRUE
+  and no device in the context associated with _kernel_ support SVM.
+  * {CL_INVALID_OPERATION} if passing {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT} and no
+  device in the context associated with _kernel_ support device pointers.
```
kpet, line 279: Please rewrite this extension proposal such that it is integrated into the unified API specification. The Khronos OpenCL working group has decided to move away from using separate extension documents to facilitate the development of extensions that interact with each other. Work has begun to fold all the existing EXT extensions into the unified specification. We should not at this point introduce new standalone extension documents. Thanks!

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+
+== Interactions with Other Extensions
+
+This extension is targeted to complement the OpenCL SVM extension and/or the
+Intel Unified Shared Memory extension by providing an additional lower-end
+step in the spectrum of type of pointers/buffers OpenCL can allocate. The
+extension can be seen as a simplification of the USM Device allocation type
+which drops the need to map the device buffer's address range to the same
+position in the host memory or to implement platform-wide VM.
```
kpet, line 288: Strictly speaking this is not really an interaction. Consider folding this (useful) text into the (already very helpful) overview.

xml/cl.xml
```diff
@@ -255,6 +255,7 @@ server's OpenCL/api-docs repository.
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_platform_command_buffer_capabilities_khr</name>;</type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_mutable_dispatch_asserts_khr</name></type>
         <type category="define">typedef <type>cl_bitfield</type>      <name>cl_device_kernel_clock_capabilities_khr</name>;</type>
+        <type category="define">typedef <type>cl_ulong</type>      <name>cl_mem_device_address_EXT</name>;</type>
```
kpet, line 258:
```suggestion
        <type category="define">typedef <type>cl_ulong</type>      <name>cl_mem_device_address_ext</name>;</type>
```

xml/cl.xml
```diff
@@ -304,6 +305,10 @@ server's OpenCL/api-docs repository.
             <member><type>size_t</type>              <name>origin</name></member>
             <member><type>size_t</type>              <name>size</name></member>
         </type>
+        <type category="struct" name="cl_mem_device_address_pair_EXT">
```
kpet, line 308:
```suggestion
        <type category="struct" name="cl_mem_device_address_pair_ext">
```

xml/cl.xml
```diff
@@ -304,6 +305,10 @@ server's OpenCL/api-docs repository.
...
+            <member><type>cl_device_id</type>        <name>device</name></member>
+            <member><type>cl_mem_device_address_EXT</type>              <name>address</name></member>
```
kpet, line 310:
```suggestion
            <member><type>cl_mem_device_address_ext</type>              <name>address</name></member>
```

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+
+=== New API Enums
+
+Enums for enabling device pointer properties when creating a buffer
+{clCreateBuffer}, see <<clCreateBuffer, the list of supported memory flag values table>>:
+
+[source]
+----
+#define CL_MEM_DEVICE_SHARED_ADDRESS_EXT (1ul << 31)
+#define CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT (1ul << 30)
```
kpet, line 95: The values do not match those defined in the XML.

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+=== Section 5.2.1 - Creating Buffer Objects:
+
+Add new allocation flags <<clCreateBuffer, List of supported memory flag values table>>:
+
+[[list-of-supported-memory-flag-values-adds]]
+.List of supported memory flags by {clCreateBuffer}
+[width="100%",cols="<50%,<50%",options="header"]
+|====
+| Memory Flags | Description
+| {CL_MEM_DEVICE_SHARED_ADDRESS_EXT_anchor}
```
kpet, line 144: Not all implementations will be able to support this. Maybe support could be reported using a capabilities device query?

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+  cl_device_id device;
+  cl_mem_device_address_EXT address;
+} cl_mem_device_address_pair_EXT;
+----
+
+== Modifications to the OpenCL API Specification
+
+=== Section 5.2.1 - Creating Buffer Objects:
+
+Add new allocation flags <<clCreateBuffer, List of supported memory flag values table>>:
```
kpet, line 137: flags are in relatively short supply and it seems this feature would only ever apply to buffers and likely be superseded by the upcoming USM extension. Have you considered using buffer properties instead (this approach would have the drawback of requiring OpenCL 3.0)?

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+----
+#define CL_MEM_DEVICE_PTR_EXT 0xff01
+----
+
+Enums for setting information of indirect device pointer accesses to kernels <<clSetKernelExecInfo, the list of supported param_names table>>. This is for OpenCL 2.0 and above. When implementing the
+extension on an older OpenCL version, indirect device pointer access is not supported.
+
+[source]
+----
+#define CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT 0x11B8
```
kpet, line 110:
```suggestion
#define CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT 0x11BB
```

To match the XML

extensions/cl_ext_buffer_device_address.asciidoc
```diff
@@ -0,0 +1,320 @@
...
+----
+#define CL_MEM_DEVICE_SHARED_ADDRESS_EXT (1ul << 31)
+#define CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT (1ul << 30)
+----
+
+Enums for querying the device pointer from the cl_mem <<clGetMemObjectInfo, the list of supported param_names table>>:
+
+[source]
+----
+#define CL_MEM_DEVICE_PTR_EXT 0xff01
```
kpet, line 102: This value does not match that defined in the XML. `CL_MEM_DEVICE_PTRS_EXT` should be mentioned here too.

kpet (changes requested):
Thanks for this proposal!

xml/cl.xml
```diff
@@ -3725,6 +3735,12 @@ server's OpenCL/api-docs repository.
             <param><type>cl_uint</type>                                 <name>arg_index</name></param>
             <param>const <type>void</type>*                             <name>arg_value</name></param>
         </command>
+        <command suffix="CL_API_SUFFIX__VERSION_1_0">
+            <proto><type>cl_int</type>                                  <name>clSetKernelArgDevicePointerEXT</name></proto>
+            <param><type>cl_kernel</type>                               <name>kernel</name></param>
+            <param><type>cl_uint</type>                                 <name>arg_index</name></param>
+            <param>const <type>void</type>*                             <name>arg_value</name></param>
```
kpet, line 3742:
```suggestion
            <param>const <type>cl_mem_device_address_EXT</type>                             <name>arg_value</name></param>
```

xml/cl.xml
```diff
@@ -7186,6 +7202,26 @@ server's OpenCL/api-docs repository.
                 <command name="clSetContentSizeBufferPoCL"/>
             </require>
         </extension>
+        <extension name="cl_ext_buffer_device_address" revision="0.1.0" supported="opencl">
```
kpet, line 7205:
```suggestion
        <extension name="cl_ext_buffer_device_address" revision="0.3.0" supported="opencl">
```

Guessing you want the XML to report the latest version.

kpet (changes requested):

extensions/cl_ext_buffer_device_address.asciidoc
karolherbst: I wonder if USM can be implemented on top of Vulkan effectively. I haven't really checked in greater detail how much of the SVM semantics are required by USM. But the main issue is that the vulkan driver can allocate GPU vm addresses as they want, which makes it hard to guarantee that allocations can be accessible on the host with the same address.

kpet: It's not possible to guarantee support on Vulkan currently. I proposed https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6024 as a cheap of supporting coarse-grain semantics on Vulkan but have not had time to chase this recently.

karolherbst: mhhh, but that won't work with multiple devices.

Maybe I should chime in and share what kind of API I settled with within mesa to implement SVM semantics through a a more generic API, not relying on drivers doing much.

The tldr: the consumer can reserve VM ranges and assign addresses to memory allocations directly instead of the driver doing that itself. Most Vulkan drivers do VMA assignment in userspace anyway, so it's not too hard to add support for that.

And drivers not doing Usermode vm management should probably move towards that regardless.

That also allows to e.g. reserve the same range across multiple devices and allow implementing SVM supporting multiple devices across vendors.

kpet: Yes, that's a limitation of my proposal, single device only. TBH, I see this proposal more as a stop-gap, I suggested it because it's very cheap for Arm (and likely others) to support and provides a strong guarantees in some useful scenarios.

What's hard with the approach you followed is to guarantee that you can get matching ranges on all devices and the host. It might work well in practice though. I was surprised to see how far I got when I prototyped SVM in clvk just using a shadow host allocation created with `mmap` at the same address as the device allocation. Horrible by design but it works on several devices (tm).

karolherbst: yeah.. I Intel's SVM implementation just hopes for the best and doesn't even reserve any memory on the host.

Large part of the host VM ranges never gets used, unless you use something like libasan. But the only reason I made drivers advertise those ranges is simply because drivers in mesa use certain ranges for private allocations. `iris` (Intel) generally uses the first 2 or 3 GiB, so there is plenty of space. And I also exclude the ranges where address canonicalization might be needed (high bits set). So that still gives a huge amount of space across most drivers.

pjaaskel:
Thanks @kpet for the feedback. Implemented most of it.

extensions/cl_ext_buffer_device_address.asciidoc
pjaaskel: Good idea. Changed.

extensions/cl_ext_buffer_device_address.asciidoc
pjaaskel: Did. Looks much nicer in the unified format.

extensions/cl_ext_buffer_device_address.asciidoc
pjaaskel: Not sure if we need this. The implementation can return CL_INVALID_DEVICE if it cannot  support it for the devices in this context. This property is inherently multi-device, it doesn’t make much sense for a single device context.

extensions/cl_ext_buffer_device_address.asciidoc
pjaaskel: Changed to a clCreateBufferWithProperties() property.

pjaaskel:
After @karolherbst and @kpet are happy with this, we'll implement in PoCL and @franz will add a CTS pull request. Then we can mark this 1.0.0 and merge, I think.

karolherbst:
> After @karolherbst and @kpet are happy with this, we'll implement in PoCL and @franz will add a CTS pull request. Then we can mark this 1.0.0 and merge, I think.

Yeah, should update my implementation of it according to the latest version here and will give it another review while at it. Maybe I get to it next week.

xml/cl.xml
```diff
@@ -719,6 +720,8 @@ server's OpenCL/api-docs repository.
         <enum value="((cl_device_partition_property_ext)0 - 1)"    name="CL_PARTITION_BY_NAMES_LIST_END_EXT"/>
         <enum value="-1"            name="CL_PARTITION_BY_NAMES_LIST_END_INTEL"/>
         <enum value="0"             name="CL_MEM_DEVICE_HANDLE_LIST_END_KHR"/>
+        <enum value="1"             name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
+        <enum value="2"             name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
```
SunSerega, line 724: These 2 enums are now defined in 2 places.

pjaaskel: Good spot, I forgot the flags. Removed.

SunSerega: They aren't supposed to be flags? But below, they are added to `cl_mem_flags`:
```
            <require comment="cl_mem_flags">
                <enum name="CL_MEM_DEVICE_SHARED_ADDRESS_EXT"/>
                <enum name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
            </require>
```
From the definition in spec, it looks like they should be in `cl_mem_properties` instead. But then it's not obvious to me which type these properties have.

Meanwhile, in "Version history", they are called "allocation flags". If so, they are supposed to have their own enum type, and placing them in the `MiscNumbers` block is wrong.

pjaaskel: Thanks. No, I converted them from flags to properties after the good @kpet suggestion. I fixed the wrong type, it's now cl_mem_properties. I'm a bit confused where the cl_mem_properties enum values should be defined, so I added it where the existing one was. The type of these properties is boolean, I believe it's described in the spec text.

SunSerega: Ah, ok; you can add them to a separate `<enums>` block, but it's not as important if these are not flags.

Still, if these are properties, what are the types of values they describe? I don't see this described in the spec you've added with this PR.

Personally, I find it most clear when new enums from the same type are added to a table in the spec, with enum, value type, and description columns. Like in [table 18 in core spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#external-buffer-memory-properties-table).

SunSerega: Also, maybe it's better to allocate new unique enum values instead of using `1` and `2`, which would intersect with many things. Currently, if another extension tries to add its own enums to `cl_mem_properties` and its authors wouldn't notice that your extension does the same, it might also use values `1` and `2` for its enums, and that would make their extension arbitrarily incompatible with this one, That's why most `<enum>` tags are sorted by their value, so new enums can be added either at the end or in the gaps (`<unused>` tags, though most are reserved by specific vendors).

pjaaskel: > Ah, ok; you can add them to a separate `<enums>` block, but it's not as important if these are not flags.
> 
> Still, if these are properties, what are the types of values they describe? I don't see this described in the spec you've added with this PR.
> 
> Personally, I find it most clear when new enums from the same enum type (`cl_mem_properties` in our case) are added to a table in the spec, with enum, value type, and description columns. Here is how it looks in [table 18 of core spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#external-buffer-memory-properties-table).

Hm. I added it [right there](https://github.com/KhronosGroup/OpenCL-Docs/pull/1159/files#diff-6916ed17983ed1950fed3520537beb85d6a0153574436dde334f4911015a030aR603)? I added the enum below the existing (only) mem property that was already there (CL_MEM_DEVICE_HANDLE_LIST_KHR). I'm not sure what you mean with your improvement. Do you mean making all enums globally unique, regarless of their use/"type"?

SunSerega: > Hm. I added it [right there](https://github.com/KhronosGroup/OpenCL-Docs/pull/1159/files#diff-6916ed17983ed1950fed3520537beb85d6a0153574436dde334f4911015a030aR603)?

I see, rich diff is somewhat broken for that file - but I can now see it in source text.

> Do you mean making all enums globally unique, regarless of their use/"type"?

Yes, most OpenCL enums already have globally unique values, and their names are also not re-used in different contexts.

The only enums that aren't globally unique are special values like `0` and `-1`, as well as all the flags.

pjaaskel:
@karolherbst @pvelesko Do you see uses for CL_MEM_DEVICE_SHARED_ADDRESS? If not, perhaps we can drop that to simplify this extension even further? hipMalloc() / hipMemCpy() shouldn't need it as the addresses do not need to be unique even in multi-device due to the stateful device switch API.

karolherbst:
> @karolherbst @pvelesko Do you see uses for CL_MEM_DEVICE_SHARED_ADDRESS? If not, perhaps we can drop that to simplify this extension even further? hipMalloc() / hipMemCpy() shouldn't need it as the addresses do not need to be unique even in multi-device due to the stateful device switch API.

Not sure tbh. I doubt that any implementation of `CL_MEM_DEVICE_SHARED_ADDRESS` would be significantly different from coarse grain SVM, because in either case you need strict control over the devices' virtual memory. Maybe the one thing making `CL_MEM_DEVICE_SHARED_ADDRESS` easier to implement is, that you can ignore mapping at the same address on the host.

I think it mostly depends on what OpenCL implementations would be willing to implement. But it looks like most already implemented coarse-grain SVM and it's mostly implementations layered on top of Vulkan benefiting from this extension here.

So maybe it does make sense to drop `CL_MEM_DEVICE_SHARED_ADDRESS` for now? However, any runtime implementing SVM could also provide it basically for free as they need to solve the same issue with SVM.

Anyway, the additional work to support `CL_MEM_DEVICE_SHARED_ADDRESS` is pretty insignificant from my perspective here.

pjaaskel:
Added 0.9.1 with the shared address feature removed. @karolherbst @franz @pvelesko Also I tried to fix the enums issue, please check if it looks OK now @SunSerega.

xml/cl.xml
```diff
@@ -1633,7 +1631,6 @@ server's OpenCL/api-docs repository.
         <enum value="0x1108"        name="CL_MEM_OFFSET"/>
         <enum value="0x1109"        name="CL_MEM_USES_SVM_POINTER"/>
         <enum value="0x110A"        name="CL_MEM_PROPERTIES"/>
-        <enum value="0x110B"        name="CL_MEM_DEVICE_ADDRESS_EXT"/>
```
SunSerega: I would be fine with both options: keeping this enum here, and moving it down to your new block.

But note: you've changed the `<unused>` block on the next line when adding enums, but you didn't revert it when removing your enums from here.

pjaaskel: Thanks. Fixed.

xml/cl.xml
```diff
@@ -2310,6 +2312,12 @@ server's OpenCL/api-docs repository.
             <unused start="0x42A0" end="0x4FFF"/>
     </enums>
 
+    <enums start="0x5000" end="0x500F" name="enums.5000" comment="For cl_ext_buffer_device_address">
+        <enum value="0x5000"        name="CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT"/>
+        <enum value="0x5001"        name="CL_MEM_DEVICE_ADDRESS_EXT"/>
+        <enum value="0x5002"        name="CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT"/>
+    </enums>
```
SunSerega, line 2319: I'm fine with this, but a bit of context: AFAICT, the `<enums>` blocks were supposed to be grouping enums by the enum type (like `cl_mem_properties`), but this order was not kept. The `<enums>` block for `cl_device_info` is filled with a ton of random enums from different enum types.
The `comment` attributes on the `<require>` tags below group enums by their type consistently, including in this PR.
And otherwise, I don't see anything of note about this `<enums>` block.

Ah, except you forgot to delete the old definition of `CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT`.

xml/cl.xml
```diff
@@ -1630,7 +1631,7 @@ server's OpenCL/api-docs repository.
         <enum value="0x1108"        name="CL_MEM_OFFSET"/>
         <enum value="0x1109"        name="CL_MEM_USES_SVM_POINTER"/>
         <enum value="0x110A"        name="CL_MEM_PROPERTIES"/>
-            <unused start="0x110B" end="0x110F" comment="Reserved for cl_mem_info"/>
+             <unused start="0x110B" end="0x110F" comment="Reserved for cl_mem_info"/>
```
SunSerega, line 1634:
```suggestion
            <unused start="0x110B" end="0x110F" comment="Reserved for cl_mem_info"/>
```

extensions/cl_ext_buffer_device_address.asciidoc
pjaaskel: Removed CL_MEM_DEVICE_SHARED_ADDRESS_EXT in 0.9.1.

karolherbst:
Not really familiar with the build process and such, so I'm wondering how I'd get an updated `cl_ext.h` file I could use to update my implementation?

pjaaskel:
> Not really familiar with the build process and such, so I'm wondering how I'd get an updated `cl_ext.h` file I could use to update my implementation?

Good question. I'm new to it too. @SunSerega ?

pjaaskel:
[PR for CTS](https://github.com/KhronosGroup/OpenCL-CTS/pull/2192).

pjaaskel:
> @pjaaskel do you have local changes, because locally I had to change a few things to get things to compile:

I hadn't local changes, but I didn't try to generate the header (still yet find how), just built the docs. Fixed. Pinging @franz about the CTS for when he comes back to office.

bashbaug:
> I didn't try to generate the header (still yet find how)

The header generation scripts are in the headers repo, specifically:

https://github.com/KhronosGroup/OpenCL-Headers/tree/main/scripts

The CMake file for the headers defines two targets related to header generation: headers_generate, and headers_copy.  headers_generate generates the headers to an output temporary directory and headers_copy copies the genreated headers into the right place.  You will need to tell the header generation scripts where to find the spec XML file, which you can do using the CMake variable OPENCL_HEADERS_XML_PATH.

If you want to do everything in one go, a good workflow is something like:

`make headers_generate headers_copy all test -j`

This generates the headers, copies them to the right place, builds all tests, and runs all tests.

karolherbst:
Now with the `CL_MEM_DEVICE_SHARED_ADDRESS` option remove, I don't think the current interface of clSetKernelArgDevicePointerEXT makes much sense.

The good thing about binding buffer objects via `clSetKernelArg` or setting SVM pointers via `clSetKernelArgSVMPointer` is, that you don't have to change the argument to enqueue the same kernel on a different device.

Now that each device has its own address, one would have to synchronize the set address with the queues launching those kernels, potentially involing a lot of flushing when it wouldn't be necessary.

Though using sub-buffers if one doesn't want to use the base address is also quite the annoying interface. Maybe we need a `clSetKernelArgBuffer`, which comes with some useful features, like an `offset` argument? Might also want to add a couple of more things while at it.

Or we just stick with sub-buffers after all...

pjaaskel:
> Now with the `CL_MEM_DEVICE_SHARED_ADDRESS` option remove, I don't think the current interface of clSetKernelArgDevicePointerEXT makes much sense.

I noticed the same that it's a bit cumbersome, but it maps to the HIP/CUDA programming model where one device is targeted at the time with the hipSetDevice() API.
 
> The good thing about binding buffer objects via `clSetKernelArg` or setting SVM pointers via `clSetKernelArgSVMPointer` is, that you don't have to change the argument to enqueue the same kernel on a different device.

Yes, I agree.

> Now that each device has its own address, one would have to synchronize the set address with the queues launching those kernels, potentially involving a lot of flushing when it wouldn't be necessary.

In practice it means that if you submit the kernel to multiple devices' command queues, you have to set the arg for each submission. The client can still pass the cl_mem handle or subbuffers using the core clSetKernelArg() API if they don't want to touch the per-device pointers or set them one at a time.

> Though using sub-buffers if one doesn't want to use the base address is also quite the annoying interface. Maybe we need a `clSetKernelArgBuffer`, which comes with some useful features, like an `offset` argument? Might also want to add a couple of more things while at it.
> 
> Or we just stick with sub-buffers after all...

If you recall, we discussed and considered this option in the beginning, but it was not good enough for HIP/CUDA due to some subbuffer restrictions (alignment at least, but perhaps something else I forgot) and for the fact that some APIs prefer to use the raw pointer address. It might be doable if we lift those restrictions with the extension and then the runtime has to keep book of what raw address maps to which (sub)buffer when interfacing from APIs with raw pointers, but is that a much better option than this one?

karolherbst:
> > Now that each device has its own address, one would have to synchronize the set address with the queues launching those kernels, potentially involving a lot of flushing when it wouldn't be necessary.
> 
> In practice it means that if you submit the kernel to multiple devices' command queues, you have to set the arg for each submission. The client can still pass the cl_mem handle or subbuffers using the core clSetKernelArg() API if they don't want to touch the per-device pointers or set them one at a time.
> 

Oh right, implementations have to look at the kernel arguments set on `clEnqueueNDRangeKernel` call time, not sure why I thought it's any different.

> > Though using sub-buffers if one doesn't want to use the base address is also quite the annoying interface. Maybe we need a `clSetKernelArgBuffer`, which comes with some useful features, like an `offset` argument? Might also want to add a couple of more things while at it.
> > Or we just stick with sub-buffers after all...
> 
> If you recall, we discussed and considered this option in the beginning, but it was not good enough for HIP/CUDA due to some subbuffer restrictions (alignment at least, but perhaps something else I forgot) and for the fact that some APIs prefer to use the raw pointer address. It might be doable if we lift those restrictions with the extension and then the runtime has to keep book of what raw address maps to which (sub)buffer when interfacing from APIs with raw pointers, but is that a much better option than this one?

I think for alignment it would be much better if we could simply specify the alignment at buffer creation time. Not sure if anything is being worked on in this regard though.

`clSetKernelArgBuffer` could also get a `device` parameter added and then the value is specific to the device. That might be feasible and usable? I think that also allows for better debugability and error checking.

karolherbst:
> > I think for alignment it would be much better if we could simply specify the alignment at buffer creation time. Not sure if anything is being worked on in this regard though.
> 
> I'm not aware. If you want to extend the buffer API, I think it should be a separate extension.
> 
> Another problem in addition to the fixed alignment I recall with the approach of (ab)using sub-buffers for implementing the raw pointer passing is that sub-buffers have a size. We discussed using size 0 sub-buffers for this and it started to seem hacky, therefore we went with this raw pointer API plan, IIRC. I can try to dig up the old notes if you want to go back on the drawing board with this extension.
> 

Oh right yeah, that's impossible to solve with sub-buffers properly without changing semantics.

> > `clSetKernelArgBuffer` could also get a `device` parameter added and then the value is specific to the device. That might be feasible and usable? I think that also allows for better debugability and error checking.
> 
> Do you mean that we could call clSetKernelArgBufferAddress() multiple times for each device holding the buffer? We could, but then again it doesn't differ much from calling it multiple times for each NDRange enqueue of that kernel (to different devices' queues), in which case the device is implied from the queue it is being pushed to. Perhaps I can just add a sentence that highlights the fact that the raw pointer's device is implied from the command queue the kernel is enqueued to?

It could allow you to skip calling `clSetKernelArgBufferAddress` when and argument won't have to change, but that might be niche enough to not matter.

However it kinda matters more with this API, because it's probably not intended to be thread-safe as `clSetKernelArg` isn't. And this extension might want to modify the "Multiple Host Threads" section as well. So if you have arguments which would be the same pointer for a given device you could skip those calls if you set the pointer once for each device.

Like I can see the pattern where an application simply binds all arguments once and then uses commands to update the content of those arguments and never the kernel objects itself. And they can do it from multiple threads concurrently and safely as long as they don't modify the kernel object, like calling `clSetKernelArg*`.

If you require to call `clSetKernelArgBufferAddress` before each enqueue, application will have to synchronize calls to it from multiple threads. Unless of course `clSetKernelArgBufferAddress` will be considered thread-safe, which then might make the problem a much smaller one.

karolherbst:
> I see. What if we simply declare the call thread safe here and add a note of the problem in the spec? Somehow I'm reluctant to add the explicit device argument to the argument setter function as it's not needed in the basic single device use case.

My concern is that most implementations aren't set up to make it thread-safe easily, because it will require reworking how setting arguments is implemented, meaning it's a higher bar to implement this extension. In rusticl it's not an issue, because at the moment it is implemented in a thread-safe way, though I did consider dropping the thread-safety once I get to improve performance more and if it's a significant overhead showing up in applications.

Though one could argue, that all the other `clSetKernelArg*` calls are device independent and the reason they don't have a `device` arg is, because it was never needed, just `clSetKernelArgBufferAddress` being the exception here. Though I don't think it's a huge issue to add it. Nor is it a huge issue not to.

I think the one benefit of adding is, that the runtime could verify if the passed in handle is indeed a valid one. And it could also catch use-after-free use cases when you set a handle, but the buffer gets deallocated later one and the same address becomes valid in a new allocation hiding the bug pretty well.

But if adding it is too much of an issue for chipstar (and other users) then we could also not.

I think it makes sense to wait what others think about this problem. I don't have any strong opinions here either way, just wanted to bring this issue up before the extension is finalized.

bashbaug:
> think for alignment it would be much better if we could simply specify the alignment at buffer creation time. Not sure if anything is being worked on in this regard though.

FYI, there is some work ongoing to specify an alignment at buffer creation time, see internal MR 198.

> `clSetKernelArgBuffer` could also get a `device` parameter added and then the value is specific to the device. That might be feasible and usable? I think that also allows for better debugability and error checking.

> Somehow I'm reluctant to add the explicit device argument to the argument setter function as it's not needed in the basic single device use case.

Perhaps consider doing something like `clBuildProgram`, which accepts a device list to specify which devices the program should be built for, but the device list can be `NULL`, in which case the program is built for all devices?  I don't know if you'd need a device list or if a single argument would be sufficient.

pjaaskel:
> > think for alignment it would be much better if we could simply specify the alignment at buffer creation time. Not sure if anything is being worked on in this regard though.
> 
> FYI, there is some work ongoing to specify an alignment at buffer creation time, see internal MR 198.

Good to know!

> > `clSetKernelArgBuffer` could also get a `device` parameter added and then the value is specific to the device. That might be feasible and usable? I think that also allows for better debugability and error checking.
> 
> > Somehow I'm reluctant to add the explicit device argument to the argument setter function as it's not needed in the basic single device use case.
> 
> Perhaps consider doing something like `clBuildProgram`, which accepts a device list to specify which devices the program should be built for, but the device list can be `NULL`, in which case the program is built for all devices? I don't know if you'd need a device list or if a single argument would be sufficient.

Hmm, if we change the _clSetKernelArgDevicePointerEXT()_ pointer arg to a list of pointers with a pointer per device in the context it would match the buffer info query's return value meaning we could pass its return value directly to this API. The annoying part here is the potential discrepancy with the clBuildProgram's device list: The kernel might not have been built for all kernels in the context whereas the buffers are by default associated with all devices. If we make the list passed to the arg setter to match with the list of devices passed to clBuildProgram, the client has to filter the pointers from the context superset. Likely not a big deal, I suspect the device list in build program is typically not a subset of the context's devices - a rarely if never used feature.

In this sense, Karol's proposal of setting the pointer separately for each device might be actually easier from the client perspective. But then again I don't see the "thread safety" issue of the current proposal a major one as cl_kernels can be cloned etc. In the end I don't have strong preferences and would just go with the current simple option, but can also change it to whatever you prefer.

FYI, Michal implemented the current 0.9.1 version in [PoCL](https://github.com/pocl/pocl/pull/1726).

karolherbst:
I forgot that `clCloneKernel` existed... yeah that's definitely the cheapest option to get around the thread-safety problem.

karolherbst:
Anyway, updated my implementation to 0.9.1 as well, though I still treat `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT` as a flag, because that's also what the current version of the CTS test is doing.

karolherbst:
Another thing I was wondering about today. This extension interacts with SVM in a weird way. One can `clSVMAlloc` memory and then pass it via a host_ptr into `clCreateBuffer` or `clCreateBufferWithProperties`. If `CL_MEM_USE_HOST_PTR` is also specified as a flag when doing so, the `cl_mem` object basically wraps the SVM allocation (the CTS has a couple of tests for this).

Now with this extension one can do a `CL_MEM_USE_HOST_PTR` buffer allocation with an SVM pointer and `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT` set to `CL_TRUE`. That _should_ mean that the addresses returned by `CL_MEM_DEVICE_ADDRESS_EXT` also match the SVM pointer, no? I wonder if this should explicitly be stated or not. And if the CTS should test for this.

This extension could also just disallow this behavior if this sounds too much of an edge case nobody is going to care about.

Just wanted to point out that this is a real possibility, which implementations might have to face supporting both (as I'm now).

pjaaskel:
> This extension could also just disallow this behavior if this sounds too much of an edge case nobody is going to care about.
> 
> Just wanted to point out that this is a real possibility, which implementations might have to face supporting both (as I'm now).

This is a good point. I'd just disallow this as if (CG) SVM is supported by the implementation, then this extension is rather pointless as this extension is supposed to be a simplification to CG SVM.

pjaaskel:
...but having said that, if the implementation does support CG SVM, it might still support BDA for legacy/compatibility reasons and in that case the other behavior (the "device ptr" = SVM pointer) would make sense. Other opinions?

pjaaskel:
Right. I'll add the other option, returning the SVM pointer in this case, in the next specification revision.

pjaaskel:
When adding the sentence about SVM, is this also ready to be marked 1.0.0 and merged in (wondering should I do it with the same commit)?

karolherbst:
No concerns in regards to that from my side. I think from a technical perspective it's in a good shape to land, though I don't want to rule out that more clarifications might be needed once others implement it as well.

api/opencl_runtime_layer.asciidoc
```diff
@@ -595,6 +595,35 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_HANDLE_LIST_KHR.asciidoc[]
...
+   range of the buffer remains constant.
+
+   The device addresses of sub-buffers derived from {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}
+   allocated buffers can be computed by adding the sub-buffer origin to the
+   device-specific start address.
+
+   If the device supports SVM and {clCreateBufferWithProperties} is called with a pointer
+   returned by {clSVMAlloc} as its _host_ptr_ argument, and {CL_MEM_USE_HOST_PTR} is
+   set in its _flags_ argument, the device-side address is guaranteed to match
+   the _host_ptr.
```
karolherbst, line 623: missing `_` in `_host_ptr_`

api/opencl_runtime_layer.asciidoc
```diff
@@ -10689,6 +10741,47 @@ Otherwise, it returns one of the following errors:
...
+    Arguments to the kernel are referred by indices that go from 0 for the
+    leftmost argument to _n_ - 1, where _n_ is the total number of arguments
+    declared by a kernel.
+  * _arg_value_ is the device pointer that should be used as the argument value for
+    argument specified by _arg_index_.
+    The device pointer specified is the value used by all API calls that enqueue
+    _kernel_ ({clEnqueueNDRangeKernel} and {clEnqueueTask}) until the argument
+    value is changed by a call to {clSetKernelArgDevicePointerEXT} for _kernel_.
+    The device pointer can only be used for arguments that are declared to be a
+    pointer to `global` memory allocated with {clCreateBufferWithProperties} with
```
karolherbst, line 10764: should it say `global` and `constant` or is this all restricted to `global` memory? I'm fine either way, though it might be good to point it out if it's not supported for `constant` memory.

For some hardware/implementations it's more or less the same, so might be better to be more explicit about it before one implementation supports it for `constant` and another doesn't.

karolherbst (approved):
Read through it again as I'm updating my implementation. I left a nit and a comment, but the later shouldn't block this.

api/opencl_runtime_layer.asciidoc
```diff
@@ -10764,7 +10870,14 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_KERNEL} if _kernel_ is a not a valid kernel object.
-  * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support SVM.
+  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_SVM_PTRS} if no devices in
```
karolherbst, line 10873: does `and {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM}` needed to be added here?

api/opencl_runtime_layer.asciidoc
pjaaskel: Hmm. Is it even possible to allocate buffers from the constant space and assign them as arguments?

karolherbst: It's a bit complicated. Nvidia GPUs have a hardware path for constant memory which doesn't really allow global addresses, however in command submission you'd still just upload from a global address, so it might be better to not allow it.

However modern nvidia GPUs can do the same with bindless constant buffers where the global address can be used (though the size of the entire access would be needed, but the runtime could handle it internally).

I don't think other hardware has a similar restriction as it's often simply a global load instruction with a special caching mode.

Though I think I'm leaning towards not allowing it for now, because it might make it a performance trade of for some implementation.

pjaaskel: I meant from the OpenCL API perspective. Surely the constant can be mapped to whatever memory physically in HW if wanted.

karolherbst: > Surely the constant AS can be mapped to whatever memory physically in HW if wanted.

It depends. If you make use of Nvidia's const buffers, then the answer is no. Or you add a lot of compiler smartness to make it somehow happen, but it's quite a bit of work. Modern nvidia GPUs (last 5 years) can deal a bit better, but it's still a performance trade-off nonetheless. It's quite a different thing on the ISA level and it's a huge perf gain to not use virtual addresses for constant memory at all as the instruction pulling from constant memory behave more like pulling data from registers instead of VRAM.

There are also push constants in other vendors hardware which could be used to implement the constant AS, and for those similar restrictions apply.

From an API perspective it doesn't matter much however.

karolherbst: You can have kernel arguments being pointers to the constant AS and you bind them the same way you bind pointers to the global AS.

karolherbst: you simply use `clSetKernelArg` and bind `cl_mem` objects as normal kernel arguments, there `constant` and `global` pointer arguments are equal. With this extension one could simply assume that storing such a pointer to a struct in the global AS _could_ mean that they could reuse the pointer in a future invocation:

> If the argument is declared to be a pointer of a built-in scalar or vector type, or a user defined structure type in the global or constant address space, the memory object specified as argument value must be a buffer object (or NULL). If the argument is declared with the constant qualifier, the size in bytes of the memory object cannot exceed [CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE) and the number of arguments declared as pointers to constant memory cannot exceed [CL_DEVICE_MAX_CONSTANT_ARGS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_CONSTANT_ARGS).

The more I think about the `constant` AS here, the more problems I see... but anyway, I think the best action here is to simply explicitly state that none of the guarantees this extension adds for pointers to the global AS hold for pointers to the constant AS.

karolherbst: Well.. this extension doesn't really add any guarnatees in regards to whatever is legal in OpenCL C, so maybe it's fine.

karolherbst: There is nothing special about `global` and `constant` there as both simply use cl_mem buffers. Constant in the kernels are a bit different, as you can also have constant pointers point to them, but to set the value of e.g. a `__constant float *tmpF` kernel argument, you simply call `clSetKernelArg(kernel, 0, sizeof(cl_mem), &cl_mem);` with a ordinary `cl_mem` created through `clCreateBuffer`.

karolherbst: > Doesn't it mean that constant and global address spaces cannot be disjoint but actually map to the same address space, otherwise how the device can arbitrate between these address spaces?

it depends on a lot of things. As explained above, there are hardware paths a compiler could use to make the constant address space faster, but at least from a GPU programming perspective those things are initialized either from constant data the host sent or from a buffer in the global address space.

But the semantics in the hardware shaders/kernels are entirely disjoint, so if the hw paths are chosen, you can't access the constant AS with global operations directly (would need to pull the address from somewhere else).

On nvidia it's e.g. 16 or more buffers of 64kb size, and the index is a 0 based vec2 (index + offset) and it's all bindful (meaning you program the individual slots when launching the shader/kernel, so you don't guarantee that the constant address remains stable across invocations at all).

So most of the constant weirdness is just part of the command submission when launching the kernel and can even happen on the GPU independently from the host, e.g. you can write to a global buffer and use it as a hw constant buffer in the next kernel, without the host having to do anything to update the contents of the constant buffer slots as it happens all on the GPU.

OpenCL implementations might also use those hardware buffers for in source constants, especially if they are indirectly accessed or huge tables.

So from an ISA perspective it can look entirely different, while from an API perspective it looks almost the same. However, runtimes _could_ make the global backing storage of a hardware constant buffer available to the kernel, if the compiler needs to access constant data through the global AS (e.g. through an internal driver buffer mapping from constant buffer index to global AS).

There are also instructions with can be hybrid (e.g. on nvidia there is `ld.constant` or bindless constant buffers needing the global address + the size of the buffer, but that's a relatively new feature), like using global addressing, but making use of the constant buffer hardware and aggressively cache data.

Of course an implementation can also simply use the hw global AS for both API global and API constant if they don't care about the performance benefits (rusticl atm doesn't use the hw constant buffers for kernel arguments, because I haven't gotten to it yet).

pjaaskel: OK, I see now what you mean, thanks. The runtime could optimize in this case and move/allocate the buffer to a constant memory if there is a separate one, the kernel arg qualifier is constant and then ISA in the kernel would always access the constant space if the HW has a disjoint memory for constant.

api/opencl_runtime_layer.asciidoc
```diff
@@ -595,6 +595,35 @@ include::{generated}/api/version-notes/CL_MEM_DEVICE_HANDLE_LIST_KHR.asciidoc[]
...
         external memory handle.
 endif::cl_khr_external_memory[]
+
+ifdef::cl_ext_buffer_device_address[]
+
+| {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT.asciidoc[]
+  | {cl_bool_TYPE}
+      | When set to CL_TRUE, specifies that the buffer must have a single fixed
```
bashbaug, line 605:
```suggestion
      | When set to {CL_TRUE}, specifies that the buffer must have a single fixed
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -662,6 +691,11 @@ ifdef::cl_khr_external_memory[]
      {CL_MEM_DEVICE_HANDLE_LIST_KHR} is specified as part of _properties_.
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
+ifdef::cl_ext_buffer_device_address[]
+  * {CL_INVALID_DEVICE}
+  ** If _properties_ includes {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} and there
+  is at least one device in the context that doesn't support such allocation.
```
bashbaug, line 697: What are the conditions when a device "doesn't support such (an) allocation"?  Is the only relevant condition whether the device supports `cl_ext_buffer_device_address`, or is there something else?

If this is the only condition then I think it would be clearer to enumerate it explicitly.

I think this also means that it is impossible to use the buffer device address feature in a multi-device context if any of the devices do not support the buffer device address feature - is that correct and is this the desired behavior?

Note that for other OpenCL APIs it is usually only an error if none of the devices in the context support the feature, and the error code in this case is usually `CL_INVALID_OPERATION` rather than `CL_INVALID_DEVICE`: for example, creating images, creating programs from IL, etc.

api/opencl_runtime_layer.asciidoc
```diff
@@ -6374,6 +6408,19 @@ include::{generated}/api/version-notes/CL_MEM_D3D11_RESOURCE_KHR.asciidoc[]
       returns the _resource_ argument specified when _memobj_ was created.
 endif::cl_khr_d3d11_sharing[]
 
+ifdef::cl_ext_buffer_device_address[]
+| {CL_MEM_DEVICE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
+  |  {cl_mem_device_address_EXT_TYPE}[]
+    | If _memobj_ was created using {clCreateBufferWithProperties} with
+      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to CL_TRUE,
```
bashbaug, line 6417:
```suggestion
      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to {CL_TRUE},
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -6374,6 +6408,19 @@ include::{generated}/api/version-notes/CL_MEM_D3D11_RESOURCE_KHR.asciidoc[]
...
+      returns a list of device addresses for the buffer, one for each
+      device in the context.
```
bashbaug, line 6419: Do we need to say anything about the order in which the device addresses are returned?  For example, are the device addresses returned in the same order as the list of devices passed to `clCreateContext` and/or returned by `CL_CONTEXT_DEVICES`?

api/opencl_runtime_layer.asciidoc
```diff
@@ -6388,6 +6435,11 @@ Otherwise, it returns one of the following errors:
     the return type specified in the
     <<mem-info-table, Memory Object Queries>> table
     and _param_value_ is not `NULL`.
+ifdef::cl_ext_buffer_device_address[]
+  ** Returned for the {CL_MEM_DEVICE_ADDRESS_EXT} query if
+     the {cl_ext_buffer_device_address_EXT} is not supported or if the
+     buffer was not allocated with {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT}.
```
bashbaug, line 6441: I'd consider returning `CL_INVALID_OPERATION` instead of `CL_INVALID_VALUE` in the case where the buffer was not allocated with `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT`.  Rationale is: in this scenario the value is valid, it is just invalid in combination with this specific buffer.

api/opencl_runtime_layer.asciidoc
```diff
@@ -10689,6 +10741,47 @@ Otherwise, it returns one of the following errors:
...
+    the argument's type.
+
+{clSetKernelArgDevicePointerEXT} returns {CL_SUCCESS} if the argument was set
+successfully. Otherwise, it returns one of the following errors:
+
+  * {CL_INVALID_KERNEL} if _kernel_ is not a valid kernel object.
+  * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
+    the device pointer.
+  * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
+  * {CL_INVALID_ARG_VALUE} if _arg_value_ specified is not a valid value.
```
bashbaug, line 10777: It might be good to clarify what is considered "not a valid value", especially given the `CL_INVALID_OPERATION` error condition above.

Is zero considered a valid value, to set a device pointer to `NULL`?

This discussion for `cl_khr_unified_svm` could also be relevant: https://github.com/KhronosGroup/OpenCL-Docs/pull/1282#discussion_r1913868626

pjaaskel:
Good points @bashbaug, I changed these and made it rev 1.0.1. Pinging @karolherbst @franz

api/opencl_runtime_layer.asciidoc
```diff
@@ -6374,6 +6409,20 @@ include::{generated}/api/version-notes/CL_MEM_D3D11_RESOURCE_KHR.asciidoc[]
...
+ifdef::cl_ext_buffer_device_address[]
+| {CL_MEM_DEVICE_ADDRESS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_MEM_DEVICE_ADDRESS_EXT.asciidoc[]
+  |  {cl_mem_device_address_EXT_TYPE}[]
+    | If _memobj_ was created using {clCreateBufferWithProperties} with
+      the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property set to {CL_TRUE},
+      returns a list of device addresses for the buffer, one for each
+      device in the context in the same order as the list of devices
+      passed to {clCreateContext}.
```
karolherbst, line 6421: Now that `CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT` can be passed if some devices don't support the extension, will this query return `NULL` or an undefined value for devices not supporting it? Or would one have to filter those out leaving no gaps?

pjaaskel: I think it has to be "undefined" because NULL might be a valid value in the device, unless we define it to be not.

karolherbst: `NULL` doesn't imply it being 0, just that it's a NULL pointer, so runtimes could return whatever value represents `NULL` for that device.

Though I don't mind it being undefined either.

pjaaskel: Yes, but then we'd need to have an API for figuring out what is the "NULL" for each device.. I think it might be simplest and clearest to leave it undefined (likely 0).

karolherbst: yeah, undefined is fine. One could argue that `NULL` issue exists with `clSetKernelArgDevicePointerEXT`, but one can use `clSetKernelArg` to set a `NULL` pointer, so yeah, probably better not to get into the `NULL` topic if we can avoid it.

pjaaskel: Yes and I even expanded the semantics to allow invalid pointers of any kind (like C/C++ calls do), as long as they are not dereferenced.

xml/cl.xml
```diff
@@ -3730,6 +3737,12 @@ server's OpenCL/api-docs repository.
             <param><type>cl_uint</type>                                 <name>arg_index</name></param>
             <param>const <type>void</type>*                             <name>arg_value</name></param>
         </command>
+        <command suffix="CL_API_SUFFIX__VERSION_3_0">
+            <proto><type>cl_int</type>                                  <name>clSetKernelArgDevicePointerEXT</name></proto>
+            <param><type>cl_kernel</type>                               <name>kernel</name></param>
+            <param><type>cl_uint</type>                                 <name>arg_index</name></param>
+            <param>const <type>cl_mem_device_address_ext</type>*        <name>arg_value</name></param>
```
bashbaug, line 3744: I don't think this is correct, and this doesn't match the function signature in POCL.  I think it should be:

```suggestion
            <param><type>cl_mem_device_address_ext</type>               <name>arg_value</name></param>
```

It'd probably be a good idea to generate the headers from this XML file to make sure everything looks as expected.

bashbaug: Hmm, the signature in the XML file *does* seem to match the function signature in Mesa though.

@karolherbst and @pjaaskel, which one is right?

pjaaskel: I believe what you @bashbaug propose is right. I don't see point passing a pointer to the pointer value instead of passing it by value. @karolherbst do you agree?

karolherbst: I think we had a discussion on this and potentially considered passing a list of addresses one per device. And I assumed this change was made, so it's easier to switch to that model when decided so in a future version of this extension.

Though I don't mind either way.

pjaaskel:
I believe I've now addressed the feedback. @franz pls check that this 1.0.2 matches the PoCL implementation and the CTS.

bashbaug:
Would it be possible to create a draft PR with headers for this extension, generated from the XML file in this PR?  Sometimes I've found that that can be a good way to identify XML issues, and we'll need to do this anyhow before merging the CTS changes.

pjaaskel:
Ah, I was planning to simply attach the generated cl_ext.h but forgot. Would this work:
[cl_ext.h.gz](https://github.com/user-attachments/files/18660156/cl_ext.h.gz)

karolherbst:
> Would it be possible to create a draft PR with headers for this extension, generated from the XML file in this PR? Sometimes I've found that that can be a good way to identify XML issues, and we'll need to do this anyhow before merging the CTS changes.

I can attach a diff, because I'll need one for mesa anyway.

bashbaug:
Here's a draft PR with header changes generated from the XML file in this PR: https://github.com/KhronosGroup/OpenCL-Headers/pull/273

xml/cl.xml
pjaaskel: Reverted to value passing.

pjaaskel:
This is ready to go as far as I'm concerned. I forgot to resolve the conversations, but I addressed those in the latest commits.

bashbaug:
Great.  @kpet you still have some requested changes, but I think they've been addressed?

Let's see if we can get this merged in next week's call, if not before.

If we can get a review approval on the headers then we can merge the headers, also.  https://github.com/KhronosGroup/OpenCL-Headers/pull/273

api/opencl_runtime_layer.asciidoc
```diff
@@ -10778,6 +10832,48 @@ Otherwise, it returns one of the following errors:
...
+    argument specified by _arg_index_.
+    The device pointer specified is the value used by all API calls that enqueue
+    _kernel_ ({clEnqueueNDRangeKernel} and {clEnqueueTask}) until the argument
+    value is changed by a call to {clSetKernelArgDevicePointerEXT} for _kernel_.
+    The device pointer can only be used for arguments that are declared to be a
+    pointer to `global` memory allocated with {clCreateBufferWithProperties} with
+    the {CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT} property. The pointer value specified as
+    the argument value can be the pointer to the beginning of the buffer or any offset into
+    the buffer region. The device pointer value must be naturally aligned according to
+    the argument's type. It should be noted that it's legal to pass invalid
```
kpet, line 10859:
```suggestion
    the argument's type. It should be noted that it is legal to pass invalid
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -10844,6 +10940,19 @@ include::{generated}/api/version-notes/CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM
...
+| {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT_anchor}
+
+include::{generated}/api/version-notes/CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT.asciidoc[]
+  | {cl_mem_device_address_EXT_TYPE}[]
+      | Device pointers must reference locations contained entirely within
+        buffers that are passed to kernel as arguments, or that are passed
+        through the execution information. Non-argument device pointers accessed
+        by the kernel must be specified by passing pointers to those buffers
+        via this {clSetKernelExecInfo} option.
+endif::cl_ext_buffer_device_address[]
```
kpet, line 10954: I'm guessing we don't want to require this be called before or after argument setting. We probably ought to test both.

kpet (approved):
LGTM

bashbaug:
Merging as discussed in the February 25th teleconference.

PR #1161, opened 2024-05-08
Add revision to XML extensions

EwanC:
This PR actions https://github.com/KhronosGroup/OpenCL-Headers/issues/248 to mandate a revision field in the XML tag for extensions. This allows a version macro to be generated with headers PR https://github.com/KhronosGroup/OpenCL-Headers/pull/251

To reduce the scope of this change only KHR extensions are given a revision,  based on the semantic version defined in their respective specifications. For EXT and vendor extensions where using semantic versioning is less standardized, a placeholder `0.0.0` value is used until they can be updated by the owner as part of https://github.com/KhronosGroup/OpenCL-Docs/issues/1168

The XML schema is updated to make the revision field mandatory in the XML entry for extensions, in the future this means that the rendered extension specifications could generated the revision based on this - as tracked by https://github.com/KhronosGroup/OpenCL-Docs/issues/1169 

The existence of the macro is also advertised as a note in the versioning sectioning of the spec.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -5389,7 +5389,7 @@ server's OpenCL/api-docs repository.
 
     <!-- SECTION: OpenCL extension interface definitions -->
     <extensions>
-        <extension name="cl_khr_d3d10_sharing" supported="opencl" ratified="opencl">
+        <extension name="cl_khr_d3d10_sharing" revision="1.0.0" supported="opencl" ratified="opencl">
             <require>
                 <type name="d3d10.h"/>
                 <type name="CL/cl.h"/>
@@ -5436,7 +5436,7 @@ server's OpenCL/api-docs repository.
                 <command name="clEnqueueReleaseD3D10ObjectsKHR"/>
             </require>
         </extension>
-        <extension name="cl_khr_d3d11_sharing" supported="opencl" ratified="opencl">
+        <extension name="cl_khr_d3d11_sharing" revision="1.0.0" supported="opencl" ratified="opencl">
             <require>
                 <type name="d3d11.h"/>
                 <type name="CL/cl.h"/>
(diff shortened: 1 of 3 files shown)


xml/cl.xml
```diff
@@ -6983,7 +6983,7 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_4x8BIT_PACKED_KHR"/>
             </require>
         </extension>
-        <extension name="cl_khr_semaphore" revision="0.9.1" supported="opencl" depends="CL_VERSION_1_2" ratified="opencl">
```
SunSerega, line 6986: Are you sure this is intentional? There are a few more extensions that already had a revision attribute, and have been reset to `1.0.0`.

EwanC: Yes this is intentional - I manually looked through the extensions while I was doing this and changed the revisions in the xml tag to match the spec. Several of the semaphore and external memory family of extension tags needed bumped to 1.0.0. Saying that, because I did this manually I still could have made a mistake so double checking these revisions are correct should be verified in review.

xml/cl.xml
```diff
@@ -5622,7 +5622,7 @@ server's OpenCL/api-docs repository.
                 <command name="clIcdGetPlatformIDsKHR"/>
             </require>
         </extension>
-        <extension name="cl_loader_layers" supported="opencl">
+        <extension name="cl_loader_layers" revision="1.0.0" supported="opencl">
```
EwanC, line 5625: I've set this to `1.0.0` but can find the spec for it

bashbaug: I think the latest spec is here: https://github.com/KhronosGroup/OpenCL-Docs/blob/main/extensions/cl_loader_layers.asciidoc

Looks like version `1.0.0` should be fine.

@Kerilk should we build an HTML version of this spec and post it on the registry?  We published an HTML spec for [cl_loader_info](https://registry.khronos.org/OpenCL/extensions/loader/cl_loader_info.html), but it looks like we haven't published cl_loader_layers.

bashbaug (approved):
I think we can merge this as-is to unblock progress.  We can always make addtional tweaks later.

In particular, I think it'd be best for most vendor extensions to go from `0.0.0` to `1.0.0`, at least for extensions that are supported by newer drivers that report version `1.0.0` or newer for the supported extension version.  This will avoid awkward situations where the reported extension version is "newer" than the version in the headers.  I think it'd be fine for each of the vendors to do this for their own extensions, though, or we can do this in one follow-on PR if all vendors are comfortable bumping their extensions to version `1.0.0`.

bashbaug:
Merging as discussed in the May 21st teleconference.

PR #1163, opened 2024-05-10
Define extension macros in environment specification

kpet:
This was missed during the spec unification work.

Change-Id: I23107b104431abdac2eaf8a3b1ebedd10303fc12


bashbaug (approved):
This change is fine, but out of curiosity: is there anything that is currently broken by not including this file in the OpenCL SPIR-V environment specification?

kpet:
Not that I know of. I found it was missing as part of drafting a new extension.

Issue #1129, opened 2024-04-01
Can an implementation diverge on device info query and compiler feature macro for  atomic_scope_all_devices


bcalidas:
This issue is related to https://github.com/KhronosGroup/OpenCL-Docs/issues/1047 . We have a specific question.

Can an implementation not report  CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES  for the clGetDeviceInfo query CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES but still define the compiler feature macro - __opencl_c_atomic_scope_all_devices

In this case the implementation does report __opencl_c_atomic_scope_all_devices for the 	clGetDeviceInfo query: CL_DEVICE_OPENCL_C_FEATURES

From the spec - "When used on a fine-grained non-atomic SVM buffer, a coarse-grained
SVM buffer, or a non-SVM buffer, operations parameterized with memory_scope_all_svm_devices
will behave as if they were parameterized with memory_scope_device"

This would imply that it is ok for an implementation to report __opencl_c_atomic_scope_all_devices but not CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES since the behavior of kernels on such an implementation is defined in the spec.

There are some conformance tests which check for consistency between feature macro and device info queries. These will need to be adapted pending the outcome of this discussion. It does bring up the larger question of when it is appropriate to expect these queries to match.

An additional consideration is how the compiler options -cl-std=CL3.0 and -cl-std=CL2.0 should affect compiler  behavior in this case,

bashbaug:
In the scenario above what are the SVM capabilities for this device?  Specifically, does it support fine-grain SVM with atomics?

bcalidas:
The device supports coarse grain SVM only.

bashbaug:
> We pass with the following configuration [...]

This matches what we report for our coarse-grain SVM GPUs also, see e.g. https://opencl.gpuinfo.org/displayreport.php?id=2215.

lakshmih:
[...]

PR #1171, opened 2024-05-21
platform: Clarify behavior for ATOMIC_SCOPE_ALL_DEVICES #1129

lakshmih:
Clarify behavior of above capability value on devices that don't support fine grained SVM

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1606,6 +1606,9 @@ include::{generated}/api/version-notes/CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES.asci
         {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
         {CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP}
 
+        A device that does not support CL_DEVICE_SVM_ATOMICS (and hence does not support CL_MEM_SVM_ATOMICS) may still support CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES. On these devices, an atomic operation with
+        memory_scope_all_svm_devices will behave the same as if the scope were memory_scope_device - refer to <<memory-consistency-model, memory consistency model>>
+
 | {CL_DEVICE_ATOMIC_FENCE_CAPABILITIES_anchor}
 
 include::{generated}/api/version-notes/CL_DEVICE_ATOMIC_FENCE_CAPABILITIES.asciidoc[]
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1606,8 +1606,8 @@ include::{generated}/api/version-notes/CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES.asci
         {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
         {CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP}
 
-        A device that does not support CL_DEVICE_SVM_ATOMICS (and hence does not support CL_MEM_SVM_ATOMICS) may still support CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES. On these devices, an atomic operation with
-        memory_scope_all_svm_devices will behave the same as if the scope were memory_scope_device - refer to <<memory-consistency-model, memory consistency model>>
+        A device that does not support {CL_DEVICE_SVM_ATOMICS} (and hence does not support {CL_MEM_SVM_ATOMICS}) may still support {CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES}. On these devices, an atomic operation with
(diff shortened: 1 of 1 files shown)


api/opencl_platform_layer.asciidoc
```diff
@@ -1606,6 +1606,8 @@ include::{generated}/api/version-notes/CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES.asci
         {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
         {CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP}
 
+        A device that only supports {CL_DEVICE_SVM_COARSE_GRAIN_BUFFER_anchor} as per {CL_DEVICE_SVM_CAPABILITIES_anchor} query can still report support for {CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES_anchor} in queries for {CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES_anchor} and {CL_DEVICE_ATOMIC_FENCE_CAPABILITIES_anchor}. Operations parameterized with memory_scope_all_svm_devices on such a device will behave as if they were parameterized with memory_scope_device.
```
bashbaug, line 1609: I think it would be best to be consistent with [Section 3.3.7](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#memory-consistency-model) and describe this behavior in terms of [CL_MEM_SVM_ATOMICS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_SVM_ATOMICS) and/or [CL_DEVICE_SVM_ATOMICS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_SVM_ATOMICS).

How about something like:

> A device that does not support `CL_DEVICE_SVM_ATOMICS` (and hence does not support `CL_MEM_SVM_ATOMICS`) may still support `CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES`.  On these devices, an atomic operation with **memory_scope_all_svm_devices** will behave the same as if the scope were **memory_scope_device** - refer to (nice link to Section 3.3.7).

How do you feel about putting this text into a footnote?  The nice thing about a footnote is that we can include the same text both for both `CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES` and `CL_DEVICE_ATOMIC_FENCE_CAPABILITIES`.

PS: The "anchor" attribute should only be used once, since it will define the target of the link for the enum.  Any other attributes - like those added by this PR - can just use the name of the attribute, instead.  I should really start writing that style guide...

bashbaug: @lakshmih did you have a chance to look at this review comment?

SunSerega:
This PR fails to process .asciidoc files when run merged with the latest state of `main`:
https://github.com/SunSerega/OpenCL-Docs/actions/runs/10550074212/job/29225831755#step:6:73
(this is a merge commit GitHub automatically generates to check if the PR has conflicts, but it doesn't automatically re-run the workflows on this commit)

bashbaug:
I think this PR is intended to fix issue #1129 so I linked the issue and the PR.

@lakshmih let us know if there's anything more we can do to help this get across the finish line - thanks!

api/opencl_platform_layer.asciidoc
```diff
@@ -1606,6 +1606,9 @@ include::{generated}/api/version-notes/CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES.asci
         {CL_DEVICE_ATOMIC_ORDER_RELAXED} \| +
         {CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP}
 
+        A device that does not support CL_DEVICE_SVM_ATOMICS (and hence does not support CL_MEM_SVM_ATOMICS) may still support CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES. On these devices, an atomic operation with
+        memory_scope_all_svm_devices will behave the same as if the scope were memory_scope_device - refer to <<memory-consistency-model, memory consistency model>>
```
bashbaug, line 1610: Thanks!  Latest text LGTM.  Here are my suggested formatting changes.  If these look good, just accept the suggestion, and I'll merge.

```suggestion
        A device that does not support {CL_DEVICE_SVM_ATOMICS} (and hence does not support {CL_MEM_SVM_ATOMICS}) may still support {CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES}. On these devices, an atomic operation with
        *memory_scope_all_svm_devices* will behave the same as if the scope were *memory_scope_device* - refer to the <<memory-consistency-model, memory consistency model>>.
```

api/opencl_platform_layer.asciidoc
bashbaug: I suspect we should have similar text for `CL_DEVICE_ATOMIC_FENCE_CAPABILITIES` also, but this is a definite improvement, so let's cross that bridge if we come to it.

bashbaug (approved):

PR #1173, opened 2024-05-21
fix dictionary generation script to generate links for extension APIs

### LINKED ISSUE #1170: extension API functions are still not links
When we generated asciidoc dictionaries for extension API functions previously we unconditionally generated them without links, allowing them to be used in the API specification even though they were described in the extensions specification.  Now that we have unified the specifications we can generate the extension APIs with links.

bashbaug:
Fixes #1170.

Updates the dictionary generation script to generate links for extension APIs in addition to core APIs, now that the extensions are merged into the core spec.


bashbaug:
Closing - #1179 has the same fix.

Issue #1086, opened 2024-03-22
Clarify behavior in case of multiple clEnqueueAcquireExternalMemObjectsKHR calls on the same cl_mem


nikhiljnv:
Clarify behavior in case of multiple clEnqueueAcquireExternalMemObjectsKHR calls on the same cl_mem without corresponding clEnqueueReleaseExternalMemObjectsKHR.

This came up as part of https://github.com/KhronosGroup/OpenCL-CTS/pull/1899

nikhiljnv:
Check how implementations behave if we do multiple acquires with earlier interops (GL/DX) as well as external sharing.

nikhiljnv:
Some options -
1. Leave the behavior as implementation defined
2. Return an error
3. Recommend acquire to be used only once, but leave multiple acquires/release behavior implementation defined.

bcalidas:
Option 3 looks to be best.

bashbaug:
This was fixed by #1176.  Closing.

PR #1176, opened 2024-05-21
Clarify Acquire/Release behavior for external memory

### LINKED ISSUE #1078: Clarify the scope of clEnqueueAcquireExternalMemObjectsKHR and clEnqueueReleaseExternalMemObjectsKHR
The spec doesn't say clearly if the acquire and release calls are supposed to be applicable for external memobj usage across command queues, devices of the context or are limited to the usage on the particular command queue it is acquired/released on. Need to discuss and clarify one way or the other.

nikhiljnv:
Clarify Acquire/Release behavior for external memory specs to call out the scope of operations as well as
the behavior in case of multiple acquire/release calls.

Fixes #1078, #1086

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
 acquired.
 This is to guarantee that the state of the memory objects is up-to-date and
 they are accessible to OpenCL.
+
+The following restrictions shall apply -
+  * Each memory object must be acquired only once. Acquiring memory object 
+    multiple times without releasing it may result in implementation defined
+    behavior.
+  * The acquire must be performed on a command-queue associated with a device
+    that was one of the devices specified via {CL_MEM_DEVICE_HANDLE_LIST_KHR}
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
+    any device in the context.
+  * The memory object will be acquired on all devices specified
+    via {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory objects
+    will be aquired on all devices in the context.
+ 
 See <<cl_khr_external_memory-Sample-Code, "`Example with Acquire /
 Release`">> for more details on how to use this API.
 
@@ -5503,6 +5520,23 @@ Applications must release the memory objects that are acquired using
 commands in the other API.
 This is to guarantee that the state of memory objects is up-to-date and they
 are accessible to the other API.
+
+The following restrictions shall apply -
+  * Each memory object must be released only once. Releasing memory object 
+    multiple times without acquiring it may result in implementation defined
+    behavior.
+  * The release must be performed on a command-queue associated with a device
+    that was one of the devices specified via {CL_MEM_DEVICE_HANDLE_LIST_KHR}
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
+    any device in the context.
+  * The memory object will be released on all devices specified via
+    {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory object
+    will be released on all devices in the context.
+
 See "`Example with Acquire / Release`" provided in
 <<cl_khr_external_memory-Sample-Code>> for more details on how to use this
 API.
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -5442,7 +5442,7 @@ The following restrictions shall apply -
     via {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
     using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
     If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory object
-    will be aquired for all devices in the context.
+    will be acquired for all devices in the context.
  
 See <<cl_khr_external_memory-Sample-Code, "`Example with Acquire /
 Release`">> for more details on how to use this API.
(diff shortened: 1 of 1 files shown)


bcalidas:
Looks good generally.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
 acquired.
 This is to guarantee that the state of the memory objects is up-to-date and
 they are accessible to OpenCL.
+
+The following restrictions shall apply -
+  * Each memory object must be acquired only once. Acquiring memory object
```
kpet, line 5432: nit: *a* memory object

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
 acquired.
 This is to guarantee that the state of the memory objects is up-to-date and
 they are accessible to OpenCL.
+
+The following restrictions shall apply -
+  * Each memory object must be acquired only once. Acquiring memory object 
+    multiple times without releasing it may result in implementation defined
```
kpet, line 5433: s/may result/results/
s/implementation defined/implementation-defined/ (we tend to hyphenate this in the rest of the spec)

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
...
+    behavior.
+  * The acquire must be performed on a command-queue associated with a device
+    that was one of the devices specified via {CL_MEM_DEVICE_HANDLE_LIST_KHR}
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
+    any device in the context.
+  * The memory object will be acquired on all devices specified
```
kpet, line 5441: Very minor point but I think I'd suggest "for all devices" instead of "on all devices". I think it's better to describe this in terms of the outcome of the command rather than suggest that there might be an operation performed for each device if that makes sense.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
...
+    via {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory objects
+    will be aquired on all devices in the context.
```
kpet, line 5445: s/on/for/ if folks agree with the above.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
...
+  * The acquire must be performed on a command-queue associated with a device
+    that was one of the devices specified via {CL_MEM_DEVICE_HANDLE_LIST_KHR}
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
+    any device in the context.
+  * The memory object will be acquired on all devices specified
+    via {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory objects
```
kpet, line 5444: s/memory objects/memory object/ so the pluralisation is consistent across all the bullet points.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5503,6 +5520,23 @@ Applications must release the memory objects that are acquired using
...
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
+    any device in the context.
+  * The memory object will be released on all devices specified via
+    {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory object
+    will be released on all devices in the context.
+
```
kpet, line 5539: Most of the comments above apply to these additions too.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5503,6 +5520,23 @@ Applications must release the memory objects that are acquired using
...
+The following restrictions shall apply -
+  * Each memory object must be released only once. Releasing memory object 
+    multiple times without acquiring it may result in implementation defined
+    behavior.
+  * The release must be performed on a command-queue associated with a device
+    that was one of the devices specified via {CL_MEM_DEVICE_HANDLE_LIST_KHR}
+    when the memory object was imported using {clCreateBufferWithProperties} or
+    {clCreateImageWithProperties}. If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not
+    specified, the acquire can be performed on a command-queue associated with
```
bcalidas, line 5532: the release?

api/opencl_runtime_layer.asciidoc
```diff
@@ -5427,6 +5427,23 @@ handle is used by an OpenCL command queued to a command-queue without being
...
+    any device in the context.
+  * The memory object will be acquired for all devices specified
+    via {CL_MEM_DEVICE_HANDLE_LIST_KHR} when the memory object was imported
+    using {clCreateBufferWithProperties} or {clCreateImageWithProperties}.
+    If {CL_MEM_DEVICE_HANDLE_LIST_KHR} was not specified, the memory object
+    will be aquired for all devices in the context.
```
kpet, line 5445:
```suggestion
    will be acquired for all devices in the context.
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -5503,6 +5520,23 @@ Applications must release the memory objects that are acquired using
 commands in the other API.
 This is to guarantee that the state of memory objects is up-to-date and they
 are accessible to the other API.
+
+The following restrictions shall apply -
+  * Each memory object must be released only once. Releasing a memory object 
+    multiple times without acquiring it results in implementation-defined
+    behavior.
```
kpet, line 5527: Strictly speaking this is ambiguous (and so is the mirror statement above for acquire operations). There might be a better formulation, possibly describing this in terms of number of acquire/release operations but the current text is a clear step forward so fine if you want to merge it as is.

nikhiljnv:
Summarizing discussion from memory subgroup call of July 9th 2024.
Per Joshua's comment, using CL_INVALID_SEMAPHORE error code as a property value is confusing.
Ben suggested two options 
- Clarify that CL_INVALID_SEMAPHORE which is an error code is actually an invalid value that can be passed to test error condition. - Alternatively, pass a different value instead of CL_INVALID_SEMAPHORE to avoid the confusion.

bashbaug (approved):
Merging as discussed in the July 16th memory subgroup.

Issue #396, opened 2020-08-04
Behaviour of CL_*_PROPERTIES queries with pointer values


kpet:
On the 2020/08/04 teleconference, we decided that we'd resolve https://github.com/KhronosGroup/OpenCL-Docs/issues/317  by having `CL_*_PROPERTIES` queries return the values provided by the user as is. We also identified a special case that warranted further discussion: that of pointer property values. Specifically:
- Are applications required to keep the memory pointed to by pointer values accessible for the lifetime of objects created with a pointer value passed as property? This does not seem desirable.
- Should implementations return the pointer value that was passed knowing that it may no longer point to valid memory at the time where it is returned?
- Should implementations make a copy of the memory referenced by pointer property values and replace it with the pointer to the copy they maintain when properties are queried?

kpet:
Discussed on the 2020/09/08 teleconference:
- Having the behaviour mandated in the core specfication makes it possible to build generic tooling around the properties.
- We need to decide the behaviour we want when those pointers are OpenCL objects, probably increment/decrement the object's refcount as the using object is created/destroyed.
- We should consider modifying the extension template to recommend that extensions always add individual `clGet{Image,MemObject}Info` queries for the new properties they add.
- There seemed to be a preference for option 2 or 3.

bashbaug:
We will draft a PR with option (3), to make a deep copy of the properties array.

PR #1177, opened 2024-05-21
Add mechanism to specify named win32 NT handles

### LINKED ISSUE #943: Importing external memory via named NT handle reference along with pointer reference
[cl_khr_external_memory ](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_external_memory)
allows importing external memory specified via NT handle through CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_KHR 
The spec says "CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_KHR specifies an NT handle ..."
This is mostly assuming reference to external memory is specified via void* NT handle pointer.
However, NT handle references can also be specified via named string.

This issue tracks discussion wrt supporting named NT handle references.

nikhiljnv:
Currently, cl_khr_external_semaphore and cl_khr_external_memory define properties to specify external win32 NT handle via pointer. Win32 NT handles can also be specified via named strings.

Add properties to specify external Win32 NT handles via named strings too.

Fixes #943


xml/cl.xml
```diff
@@ -1894,7 +1894,9 @@ server's OpenCL/api-docs repository.
         <enum value="0x2062"        name="CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_KMT_KHR"/>
             <unused start="0x2063" end="0x2066" comment="Reserved to Khronos for interop"/>
         <enum value="0x2067"        name="CL_EXTERNAL_MEMORY_HANDLE_DMA_BUF_KHR"/>
-            <unused start="0x2068" end="0x2FFF" comment="Reserved to Khronos for interop"/>
+        <enum value="0x2068"        name="CL_SEMAPHORE_HANDLE_OPAQUE_WIN32_NAME_KHR"/>
+        <enum value="0x2069"        name="CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_NAME_KHR"/>
```
SunSerega, line 1898: I think these should also be added to the `<extension name="cl_khr_external_memory_win32"` block below, to associate them with this extension.

api/opencl_runtime_layer.asciidoc
```diff
@@ -5619,6 +5619,14 @@ buffer or an image memory object from an external handle:
...
     all memory objects associated with it are destroyed.
 endif::cl_khr_external_memory_win32[]
+  * {CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_NAME_KHR_anchor} specifies an NT handle name that
+    has only limited valid usage outside of OpenCL and other compatible
+    APIs.
+    It must be compatible with the functions `DuplicateHandle`,
+    `CloseHandle`, `CompareObjectHandles`, `GetHandleInformation`, and
+    `SetHandleInformation`.
+    It owns a reference to the underlying memory resource represented by its
+    memory object.
```
bashbaug, line 5629: I think these additions should be within the `ifdef::cl_khr_external_memory_win32` block, above.

Do we need to say anything about how the NT handle name is specified?  For reference, the Vulkan spec says for its "name" that "name is NULL or a null-terminated UTF-16 string naming the payload to import."

Presumably the property value for `CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_NAME_KHR` is a pointer to the null-terminated string.  If an application queries `CL_MEM_PROPERTIES` does it get the exact same pointer that was passed when the memory object was created, which may no longer be valid?  Or, is there an expectation that the implementation makes a copy of the name and returns it?

Same questions apply to semaphores, also.

bashbaug (changes requested):
One more comment, sorry: I think this needs to bump the extension version also, probably to v1.1?

api/opencl_runtime_layer.asciidoc
bashbaug: Regarding the query, see also: https://github.com/KhronosGroup/OpenCL-Docs/issues/396

(Discussed in the June 4th teleconference.)

bashbaug: Thanks, I think this is partially fixed now.

Vulkan allows a NULL handle name, but I think this is only because the same structure is used to pass either the handle itself or the handle name.  What would it mean to pass NULL for `CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_NAME_KHR`?

Note, we do not currently allow a NULL handle for `CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_KHR` - should we?

We also still need to decide what should happen for the properties query.

bashbaug:
Reviewed in the June 18th memory TSG.

The updated text looks good.
The extension version needs to be updated in the XML file.
It would also be good to do a rebase or merge to pick up the latest XML changes.

bashbaug (approved):
My changes have been addressed - thanks!

Note, I did fix one small typo, and added `CL_SEMAPHORE_HANDLE_OPAQUE_WIN32_NAME_KHR` to the list of new enums for `cl_khr_external_semaphore_win32`.

bashbaug:
Merging as discussed in the September 3rd teleconference.

PR #1179, opened 2024-06-04
Fix links to commands in API specification

kpet:
Links were not created for commands defined by extensions. With this changes all uses of e.g. {clCreateSemaphoreWithPropertiesKHR} link to the definition of the command which makes navigating the specification much easier.

Change-Id: I4a9458609f4ba3229b66e3d169a68cb4564e2538

### DIFF
--- a/scripts/gen_dictionaries.py
+++ b/scripts/gen_dictionaries.py
@@ -99,7 +99,7 @@ if __name__ == "__main__":
 
             numberOfFuncs = numberOfFuncs + 1
 
-    # Add extension API functions without links:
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
This is the same fix as #1173, which is good.  Ideally we would avoid duplicating work, though!

I'll just merge this one and I'll close my PR.  Your comments are better, I think 😄

kpet:
I hadn't seen that one fly by, sorry. I noticed the issue in today's teleconference and fixed it in the background. Thanks!

PR #1189, opened 2024-06-19
further clarify a clCreateBuffer with SVM pointer error condition

karolherbst:
It should also be invalid to use SVM host_ptr smaller than `size` in the CL_MEM_COPY_HOST_PTR case


karolherbst:
Not sure if this has been specifically discussed in the OpenCL WG, but it feels like an oversight as this is almost guaranteed to cause segmentation faults if not caught.

bashbaug (commented):
Linking with #875, which I believe prompted the current text.

This seems reasonable to me, though if we merge these changes we really ought to have a negative CTS test to check for the new error behavior..

kpet (approved):

bashbaug (approved):

bashbaug:
Merging as discussed in the August 13th teleconference.

PR #1195, opened 2024-06-28
New Intel extension cl_intel_subgroup_buffer_prefetch

gwawiork:


bashbaug (approved):

bashbaug:
I'll plan to merge this by the end of the week (July 12th), unless somebody has comments or decides to merge it first.  Thanks!

bashbaug (changes requested):
@gwawiork can you please add this extension to the XML file also?  Because this is an OpenCL C-only extension this can probably be done with a single line like the following, based on the other OpenCL C-only extensions in the XML file:

```xml
<extension name="cl_intel_subgroup_buffer_prefetch" revision="1.0.0" supported="opencl" comment="OpenCL C only"/>
```

bashbaug (approved):
Actually, on second thought, I think we should merge this PR as-is.

We do include OpenCL C-only and SPIR-V-only extensions in the XML file, but currently only for KHR extensions.  I do not see any EXT or vendor-specific OpenCL C or SPIR-V extensions in the XML file, so we should probably add them all in one go, in a separate PR.

PR #1196, opened 2024-07-01
Use Khronos asciidoctor-spec Docker image in CI

### LINKED ISSUE #1192: Presubmit workflow fails on `Install required packages` stage
I fast-forwarded `main` in my fork and got an email pointing to this:

https://github.com/SunSerega/OpenCL-Docs/actions/runs/9651543782/job/26619662691

It seems it fails to download `ghostscript` and its dependencies.

oddhack:
Per discussion with @bashbaug

N.b. at present the CI script has less parallelism than it could, at least as I understand Actions. Some of the 'steps' could be split off into 'jobs'. Might try that next once the basic build is working. Net performance is still somewhat faster than current CI since it's generally faster to load the container than to add needed packages at each invocation, and the spec build is pretty fast, so there's not much to be gained - might reduce runtime from 3 to 1.5 minutes, at best.

There was odd error behavior from shifting to the container which I have never seen in Vulkan CI, having to do with mixed ownership of files in the checked-out repository. I inserted a brute-force workaround right after the checkout action.

Closes #1192

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -14,8 +14,10 @@ jobs:
   build:
     name: Build spec artifacts
     runs-on: ubuntu-latest
-    # needs: checkout
-    container: khronosgroup/docker-images@sha256:bd30a83285a2ea062598f053b5bd8ebc843e16c639c0e4cd88ab4bbb4e63ead3
+    # Refer to the build container by its SHA instead of the name, to prevent
+    # caching problems when updating the image.
+    # container: khronosgroup/docker-images:asciidoctor-spec.20240701
+    container: khronosgroup/docker-images@sha256:1175e55feeaca36d8c53b3628372cd371078473a162a558c3e89ffafdde40676
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -17,7 +17,7 @@ jobs:
     # Refer to the build container by its SHA instead of the name, to
     # prevent caching problems when updating the image.
     # container: khronosgroup/docker-images:asciidoctor-spec.20240702
-    container: khronosgroup/docker-images@sha256:91a1b62399f9804e17110e89d85d62e90bc8d2ad7b94a329bee87784158368a9
+    container: khronosgroup/docker-images@sha256:4aab96a03ef292439c9bd0f972adfa29cdf838d0909b1cb4ec2a6d7b2d14a37f
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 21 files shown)


oddhack:
@bashbaug BTW, building the spec with this toolchain reports some errors in tables with the wrong numbers of cells on some rows, and maybe some others. Not sure if those were present before, but the updated asciidoctor apparently does do some more comprehensive error reporting as @gmlueck noticed the same thing in SYCL-Docs.

.github/workflows/presubmit.yml
```diff
@@ -3,47 +3,58 @@ name: Presubmit
...
+# These jobs are potentially parallelizeable
 jobs:
   build:
-    name: Build all specs
+    name: Build spec artifacts
     runs-on: ubuntu-latest
+    # Refer to the build container by its SHA instead of the name, to
+    # prevent caching problems when updating the image.
+    # container: khronosgroup/docker-images:asciidoctor-spec.20240702
+    container: khronosgroup/docker-images@sha256:91a1b62399f9804e17110e89d85d62e90bc8d2ad7b94a329bee87784158368a9
```
bashbaug, line 20: I'm not a docker expert, but this SHA appears to be from [asciidoctor-spec.20240627](https://hub.docker.com/layers/khronosgroup/docker-images/asciidoctor-spec.20240627/images/sha256-91a1b62399f9804e17110e89d85d62e90bc8d2ad7b94a329bee87784158368a9?context=explore) and not [asciidoctor-spec.20240702](https://hub.docker.com/layers/khronosgroup/docker-images/asciidoctor-spec.20240702/images/sha256-4aab96a03ef292439c9bd0f972adfa29cdf838d0909b1cb4ec2a6d7b2d14a37f?context=explore).  Should the comment be updated or should the SHA be updated, or am I completely misunderstanding how docker hub works?

.github/workflows/presubmit.yml
```diff
@@ -3,47 +3,58 @@ name: Presubmit
...
       - name: Generate core specs (HTML and PDF)
         run: |
-          python3 makeSpec -clean -spec core OUTDIR=out.core -j 5 api c env ext cxx4opencl
-          
+          python3 makeSpec -clean -spec core OUTDIR=out.core -j 5 -O api c env ext cxx4opencl
+
       - name: Generate core + extension specs (HTML)
         run: |
-          python3 makeSpec -clean -spec khr OUTDIR=out.khr -j 12 html
+          python3 makeSpec -clean -spec khr OUTDIR=out.khr -j -O html
```
bashbaug, line 56: Checking: In addition to adding `-O` this switches from `-j 12` to just plain `-j`.  Was this intended?

bashbaug (approved):
LGTM, just one question about the docker SHA.

Issue #1191, opened 2024-06-22
typo in description for clReImportSemaphoreSyncFdKHR


bashbaug:
There is a typo (probably a copy-paste error) in the description for clReImportSemaphoreSyncFdKHR, where the name of the function is listed as clGetSemaphoreHandleForTypeKHR instead of clReImportSemaphoreSyncFdKHR.

> clGetSemaphoreHandleForTypeKHR returns [CL_SUCCESS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_SUCCESS) if the semaphore handle is re-imported successfully. Otherwise, it returns one of the following errors [...]

bashbaug:
The function clGetSemaphoreHandleForTypeKHR is also listed in the summary for cl_khr_external_semaphore_sync_fd also, whereas it should be listed in the summary for cl_khr_external_semaphore.

PR #1208, opened 2024-07-10
Fix typos in description of clReImportSemaphoreSyncFdKHR

kpet:
Change-Id: I80b1d25368c2cefb0967120bdc429a8187c518c7


bashbaug (approved):
Thanks!

FYI, this is also a partial fix for #1191, which I haven't gotten around to yet, so double-thanks!

PR #1210, opened 2024-07-11
xml: require CL/GL types for cl_khr_gl_sharing

kpet:
So the type definitions are included in the generated cl_gl.h.

Split out of #960 following the discussion on https://github.com/KhronosGroup/OpenCL-Headers/pull/257

Change-Id: I65a666dde8066958897acf13fb755ae2a3f3b52d


bashbaug (approved):
Thanks!

Issue #1165, opened 2024-05-10
Generate the "New types/enums/etc" sections of the OpenCL Extensions appendix


kpet:
Following the initial step towards a unified OpenCL specification that describes core features and extensions, a new appendix giving an overview of all the extensions included in the specification was added (currently Appendix I). For each individual extension, a list of the new enums, features, structures, etc is given. These lists are currently authored manually, which is error prone and does not enable large-scale changes to how the information is presented. These sections should be generated automatically from the XML specification.

Issue #1111, opened 2024-03-28
Remove OpenCL Extensions spec and move remaining content into API / C specs?


oddhack:
With khr extensions integrated into the core specs now, there is very little content left in the extensions spec - a big list of links into the extension appendices, and some framing material on how to enable and use extensions. ISTM the framing material could be moved into the API or C specs as appropriate, and the separate extension spec removed.

bashbaug:
I think this is a good long-term goal, but it might be a good idea to keep the standalone extension spec around for one more release, to give time for any links to transition to the main API spec instead.  Maybe we should aim to do this for the next-next release?

PR #1213, opened 2024-07-17
Begin integration of EXT extensions into the unified specification

kpet:
- Add a khr+ext spec type to makeSpec and cover in CI
- Document the version and dependencies of all EXT extensions in the XML
- Integrate cl_ext_cxx_for_opencl into the unified specification
- Add placeholder descriptions for all EXT extensions linking to either core/KHR features that supersede old extensions that were never part of the specification or the OpenCL Extensions document.

This enables us to integrate EXT extensions incrementally and integrate future EXT extensions directly into the unified specification.

Change-Id: Ic634ce000ad3ebfb56e56bce91f9c0de3e786383

### DIFF
--- a/api/cl_ext_float_atomics.asciidoc
+++ b/api/cl_ext_float_atomics.asciidoc
@@ -12,7 +12,8 @@ include::{generated}/meta/{refprefix}cl_ext_float_atomics.txt[]
 
 === Description
 
-This extension is described in the OpenCL Extensions document.
+The latest published specification for this extension is available on
+the https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_float_atomics.html[OpenCL registry].
 
 === Version History
 
--- a/api/cl_ext_image_from_buffer.asciidoc
+++ b/api/cl_ext_image_from_buffer.asciidoc
@@ -12,7 +12,8 @@ include::{generated}/meta/{refprefix}cl_ext_image_from_buffer.txt[]
 
 === Description
 
-This extension is described in the OpenCL Extensions document.
+The latest published specification for this extension is available on
+the https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_image_from_buffer.html[OpenCL registry].
 
 === Version History
 
--- a/api/cl_ext_image_raw10_raw12.asciidoc
+++ b/api/cl_ext_image_raw10_raw12.asciidoc
@@ -12,7 +12,8 @@ include::{generated}/meta/{refprefix}cl_ext_image_raw10_raw12.txt[]
 
 === Description
 
-This extension is described in the OpenCL Extensions document.
+The latest published specification for this extension is available on
+the https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_image_raw10_raw12.html[OpenCL registry].
 
 === Version History
 
--- a/api/cl_ext_image_requirements_info.asciidoc
+++ b/api/cl_ext_image_requirements_info.asciidoc
@@ -12,7 +12,8 @@ include::{generated}/meta/{refprefix}cl_ext_image_requirements_info.txt[]
 
 === Description
 
-This extension is described in the OpenCL Extensions document.
+The latest published specification for this extension is available on
+the https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_image_requirements_info.html[OpenCL registry].
 
 === Version History
 


api/cl_ext_cxx_for_opencl.asciidoc
```diff
@@ -0,0 +1,59 @@
...
+
+=== New Enums
+
+  * {cl_device_info_TYPE}
+  ** {CL_DEVICE_CXX_FOR_OPENCL_NUMERIC_VERSION_EXT}
+
+=== New build option
+
+This extension adds support for a new `CLC++` value to be passed to the
+`-cl-std` build option accepted by *clBuildProgram* and *clCompileProgram*.
```
EwanC, line 38: I've not double checked how this looks, but since we're integrating with the unified spec, I think this is the best way to reference entry-points
```suggestion
`-cl-std` build option accepted by {clBuildProgram} and {clCompileProgram}.
```

kpet: Thanks!

EwanC (approved):
LGTM

After building this my main thought would be whether it's worth splitting the appendix of supported extensions into separate KHR and EXT lists, as the single list is already quite long (although it is ordered to make the EXTs easy to find in that list)

api/cl_ext_float_atomics.asciidoc
```diff
@@ -0,0 +1,20 @@
...
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2020-08-12
+*IP Status*::
+    No known IP claims.
+
+=== Description
+
+This extension is described in the OpenCL Extensions document.
```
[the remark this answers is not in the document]
kpet: Agree but it's hard to provide a link that's valid regardless of the spec you're looking at. Links would be different for published PDF or HTML versions of the spec and different again for the dev versions. I felt this was a bit too hard for a "temporary" (tm) problem.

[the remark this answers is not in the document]
EwanC: cool, we can leave this out of it isn't trivial.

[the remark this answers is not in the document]
bashbaug: Note, the EXT extensions actually aren't currently in the OpenCL Extensions document, and they're documented in standalone extension docs on the registry, similar to vendor extensions.  For example:

https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_float_atomics.html

Until this changes, maybe it is possible to put a stable link to the appropriate extension document?

api/cl_ext_migrate_memobject.asciidoc
```diff
@@ -0,0 +1,35 @@
...
+*IP Status*::
+    No known IP claims.
+
+=== Description
+
+Precursor to {clEnqueueMigrateMemObjects}.
+
+=== New Commands
+
+  * {clEnqueueMigrateMemObjectEXT}
```
bashbaug, line 19: We don't currently define anchors for any of these EXT commands, types, or enums, so they're currently dead links in the spec.  I suppose we could add these to the spec and define anchors for them, but given that there are no devices that support this extension in the opencl.gpuinfo.org database, maybe the best thing to do is to just link to the extension doc like most of the other EXT extensions in this PR and call things good?

bashbaug (commented):
Thank you for doing this!

api/cl_ext_float_atomics.asciidoc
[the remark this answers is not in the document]
kpet: I was looking at the PDF when writing this. The spec tooling generates:
- `out/pdf/OpenCL_Ext.pdf` titled "The OpenCL Extension Specification"
- `out/pdf/extensions.pdf` titled "OpenCL Extensions"

It's the latter that this refers to and that one does include EXT extensions. I agree there is potential for confusion and the real fix IMO is to get rid of both these documents :).

We could do what you propose but this would always link to the *latest pulished* version of the specification in HTML form. It would likely be useful to a good number of users but would at the same time potentially point to the wrong document for some others.

api/cl_ext_migrate_memobject.asciidoc
kpet: There was no mention of this extension in the extension doc either. We may never have had a spec document for it. The need to include an entry in the extension appendix was driven by its existence in the XML. As soon as we start automatically generating the list of new commands, types, etc in the extension appendix the dead links will reappear. The description I've added does link to the equivalent core functionality which AFAICT and sadly is the best documentation we've ever had for this extension. My inclination is to keep the document as is but maybe create an issue to add the descriptions to the spec, maybe as aliases of the core functionality if we can. Note that we'll have a similar problem for cL_ext_device_fission.

bashbaug: There is a spec for this extension, though it's quite old (and still plaintext): https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_migrate_memobject.txt

I think if we start automatically generating the list of new commands / types / etc. in the extension appendices I'd prefer to have a "skip list" for older extensions like these where the ROI just isn't high enough to pull them into the spec.  Regardless, for now, I'd really prefer not to knowingly introduce "broken links" into the spec.  If we want to include these EXT commands, types, and enums then I think we should also add appropriate anchors for them, also.

api/cl_ext_float_atomics.asciidoc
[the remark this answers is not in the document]
bashbaug: The tooling generates "extensions.pdf", but AFAIK we don't publish it.  We still do publish "OpenCL_Ext.pdf", but we might not do so for long - see #1111.

What's the disadvantage of linking to the HTML spec?  The way I look at it, if we don't put any link, readers looking for more information about the extension are either going to get frustrated and give up, or they're going to search for the name of the extension, which will (hopefully!) point them to the HTML spec on the registry.  If that's the case, why not give them the link directly?

api/cl_ext_migrate_memobject.asciidoc
kpet: Right, I hadn't looked at the registry. Good to know.

We could add a "skip list" or we could do the integration. Unless the behaviour specified in the extension diverges from what we have in core, it should be easy to just add anchors and aliases and get rid of more cruft from the registry.

For now, I've removed the dangling links. We'll have to deal with the issue when we work on #1165.

api/cl_ext_float_atomics.asciidoc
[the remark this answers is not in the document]
kpet: Beyond having PDFs link to the HTML specs, the main issue I have with adding these links is that the version linked would not be the one aligned to the document linking to it. It's not an issue for published versions of the specification but if I package a set of PDFs generated from an arbitrary SHA for example, the link would potentially point to the wrong spec text. This mostly affects IHVs so maybe not a good reason to deprive OpenCL users of a helpful link even if only temporary so. I'll add the links with text stating that they point to the "latest published" version of the specifications.

kpet:
@bashbaug I think I've addressed all your comments, PTAL. Note that we need to merge https://github.com/KhronosGroup/OpenCL-Registry/pull/156 too (seems we never published that one on the registry; can't wait for this to become unnecessary :))

bashbaug (approved):

bashbaug:
Merging as discussed in the August 13th teleconference.

xml/cl.xml
```diff
@@ -6764,7 +6764,7 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_QUEUE_COMPUTE_UNIT_LIMIT_ARM"/>
             </require>
         </extension>
-        <extension name="cl_ext_cxx_for_opencl" revision="0.0.0" supported="opencl">
+        <extension name="cl_ext_cxx_for_opencl" revision="1.0.0" depends="CL_VERSION_3_0,CL_VERSION_2_0+cl_khr_extended_versioning" supported="opencl">
```
SunSerega, line 6767: Can you please explain what it means, to have 2 different versions of core in the `depends` attribute?

Given coma, IG it means this extension can be supported if either:
- OpenCL is 2.0 and `cl_khr_extended_versioning` is supported.
- OpenCL is 3.0 is supported.

This is my best guess, but this is the first time coma was used in `depends` and `registry.rnc` explanation is still pretty surface-lvl:
```
# Boolean expression of core version and extension names using (),+ operators
Depends = attribute depends { text }
```
Before I thought the coma here is not an operator but a separator between `()` and `+` operators...

---

Also below:
```
        <extension name="cl_ext_float_atomics" revision="1.0.0" depends="CL_VERSION_2_0,CL_VERSION_3_0" supported="opencl">
```
Wouldn't `CL_VERSION_3_0` here just be redundant?

Issue #1207, opened 2024-07-10
Do we want a path to provide properties for all command types defined by cl_khr_command_buffer?


kpet:
The provisional specification for cl_khr_command_buffer defines a path to provide additional properties for NDRange kernel commands (clCommandNDRangeKernelKHR/cl_ndrange_kernel_command_properties_khr). This enables future extensibility which cl_khr_command_buffer_mutable_dispatch is already using. Do we want a similar path for all command types?

EwanC:
I think that makes sense, it's a minor increase in verbosity but gives a significant extensibility benefit.

EwanC:
My main question before drafting this up is, do we want a generic property type for all commands?

Right now `cl_ndrange_kernel_command_properties_khr` is tied to the `clCommandNDRangeKernelKHR` command, and if we followed that precedent then each command would have its own `cl_foo_command_properties_khr` property type - which feels a bit excessive. Alternatively, we could have a generic `cl_command_properties_khr` property type for all commands (including changing `clCommandNDRangeKernelKHR` to use it) It would mean that we would need error/valid-usage wording around which commands each property define for it can be used with.

kpet:
A single `cl_command_properties_khr` property type sounds like a better direction to me. It's not like our approach of properties is type safe anyway...

Issue #1175, opened 2024-05-21
Table 58. Summary of command-buffer creation configurations title is rendering incorrectly


bashbaug:
The table title for:

> Table 58. Summary of command-buffer creation configurations, for the [cl_khr_command_buffer_multi_device](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_khr_command_buffer_multi_device) extension

... is rendering incorrectly in the HTML spec, at least for the version of asciidoctor that I am using to build the specs.  The PDF spec looks fine.  

We should either figure out what's happening with it, or rephrase the table title so it renders correctly.

EwanC:
What asciidoctor version are you using? Looking at a HTML rendered locally with Asciidoctor 2.0.16 (the version suggested by the repo README) the title renders correctly

bashbaug:
> What asciidoctor version are you using?

Odd, I'm using 2.0.16:

```sh
$ asciidoctor -V
Asciidoctor 2.0.16 [https://asciidoctor.org]
Runtime Environment (ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

What command line are you using to build the specs locally?  I'm basically using a command line like the one from our CI: `python3 makeSpec -clean -spec khr OUTDIR=out.khr -j 12 html`.

EwanC:
My setup is
```bash
$ asciidoctor -V
Asciidoctor 2.0.16 [https://asciidoctor.org]
Runtime Environment (ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

I was using, `$ make html`, but with `$python3 makeSpec -clean -spec khr OUTDIR=out.khr -j 12 html` I still see an okay rendering: 
![image](https://github.com/user-attachments/assets/81398341-1d42-41bb-9966-e718b8a5e86b)

@alycm Could you try this out?

bashbaug:
@EwanC, @alycm (and any others), do you have any more thoughts what might be causing the difference?

I have a workflow that does automated spec builds using the Khronos docker image and this also seems to render the table title incorrectly, for whatever that's worth.

bashbaug:
[...]

PR #1215, opened 2024-07-24
Add properties parameter to all command-buffer commands

EwanC:
See Issue https://github.com/KhronosGroup/OpenCL-Docs/issues/1207 discussion how it could make the `cl_khr_command_buffer` extension more extensible if we allowed all the command-recording parameters to have a `properties` parameter that could be used.

Closes https://github.com/KhronosGroup/OpenCL-Docs/issues/1207 

Related PRs:
* https://github.com/KhronosGroup/OpenCL-CLHPP/pull/305
* https://github.com/KhronosGroup/OpenCL-Headers/pull/260
* https://github.com/KhronosGroup/OpenCL-CTS/pull/2050

### DIFF
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -142,7 +142,7 @@ support their intended use cases.
 The command recording entry-points allow a `properties` parameter of
 new type {cl_command_properties_khr_TYPE} to be passed. No properties are
 defined in {cl_khr_command_buffer_EXT}, but the parameter enables layered
-extensions to define characterisitcs of the individual commands.
+extensions to define characteristics of the individual commands.
 
 For example, {cl_khr_command_buffer_mutable_dispatch_EXT} defines properties
 that can be set when appending a kernel command with {clCommandNDRangeKernelKHR}.
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -4,7 +4,7 @@
 include::{generated}/meta/{refprefix}cl_khr_command_buffer.txt[]
 
 // *Revision*::
-//    0.9.4
+//    0.9.5
 // *Extension and Version Dependencies*::
 //     This extension requires OpenCL 1.2 or later.
 //     Buffering of SVM commands requires OpenCL 2.0 or later.
@@ -12,7 +12,7 @@ include::{generated}/meta/{refprefix}cl_khr_command_buffer.txt[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2023-03-31
+    2024-07-24
 *IP Status*::
     No known IP claims.
 *Contributors*::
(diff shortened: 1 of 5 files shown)


api/cl_khr_command_buffer.asciidoc
```diff
@@ -137,13 +137,15 @@ retain its provisional extension status until other layered extensions are
...
 
-The {clCommandNDRangeKernelKHR} entry-point defines a `properties` parameter of
-new type {cl_ndrange_kernel_command_properties_khr_TYPE}. No properties are defined
-in {cl_khr_command_buffer_EXT}, but the parameter enables layered extensions like
-{cl_khr_command_buffer_mutable_dispatch_EXT} to define properties that inform
-the characteristics of the kernel command.
+The command recording entry-points allow a `properties` parameter of
+new type {cl_command_properties_khr_TYPE} to be passed. No properties are
+defined in {cl_khr_command_buffer_EXT}, but the parameter enables layered
+extensions to define characterisitcs of the individual commands.
```
bashbaug, line 145:
```suggestion
extensions to define characteristics of the individual commands.
```

bashbaug (commented):
This is a pretty big breaking change that will affect pretty much all code that uses command buffers, but the extension still is provisional and I agree it would be a good idea to build in extensibility.

bashbaug:
Merging after final review, as discussed in the September 3rd teleconference.

kpet (approved):

api/opencl_runtime_layer.asciidoc
```diff
@@ -15110,7 +15158,7 @@ ifdef::cl_khr_command_buffer_mutable_dispatch[]
     The {cl_khr_command_buffer_EXT} extension does not define any
     properties, but supported properties defined by extensions are defined
     in the <<ndrange-kernel-properties-table, List of supported properties
-    by {clCommandNDRangeKernelKHR}>> table.
+    by clCommandNDRangeKernelKHR>> table.
```
bashbaug, line 15161: Curious: did use of the asciidoctor attribute here cause problems?

This is fine for now, but at some point we might want to rename this table because using "clCommandNDRangeKernelKHR" without the asciidoctor attribute will get flagged by some of the spec checkers.

api/opencl_runtime_layer.asciidoc
```diff
@@ -16098,20 +16158,17 @@ include::{generated}/api/version-notes/CL_MUTABLE_COMMAND_COMMAND_TYPE_KHR.ascii
 | {CL_MUTABLE_DISPATCH_PROPERTIES_ARRAY_KHR_anchor}
 
 include::{generated}/api/version-notes/CL_MUTABLE_DISPATCH_PROPERTIES_ARRAY_KHR.asciidoc[]
-  | {cl_ndrange_kernel_command_properties_khr_TYPE}[]
-    | Return the properties argument specified on _command_ recording with
-      {clCommandNDRangeKernelKHR}.
+  | {cl_command_properties_khr_TYPE}[]
```
bashbaug, line 16161: Should we consider renaming this query to something a bit more general now that other command buffer commands can also have optional properties arguments?  For example, should we consider `CL_MUTABLE_COMMAND_PROPERTIES_ARRAY_KHR` rather than the mutable dispatch-specific `CL_MUTABLE_DISPATCH_PROPERTIES_ARRAY_KHR`?

bashbaug (approved):
I'm fine merging this ASAP - it's probably better to get these changes in sooner rather than later.

api/opencl_runtime_layer.asciidoc
EwanC: Interesting, in my local build I'm seeing this
image
But that doesn't seem to be a problem in the hosted render. Wonder if this is another data point for https://github.com/KhronosGroup/OpenCL-Docs/issues/1175 on inconsistent rendering 

Happy to remove to keep this change as preferred.

api/opencl_runtime_layer.asciidoc
EwanC: That's a good idea, I've made that change as a new commit so it can be reviewed in isolation.

bashbaug:
Merging as discussed in the September 3rd teleconference + email.

Issue #1214, opened 2024-07-17
Zero-width spaces in spec html


svenvh:
Various function names in the spec have zero-width spaces in them, e.g. `get_&#8203;sub_&#8203;group_&#8203;size`. It seems this was done deliberately in #950 :
```
+// Note: the Unicode "zero width space" (&#8203;) is used in some places to
+// cause long function names to break much more sensibly.
```

The zero-width spaces end up in the rendered html, which interferes with copy/pasting a function name from the html:
- Searches for that copied function name will not match actual occurrences of that function name in for example source code files.
- Pasting the copied function name into a source file includes the zero-width spaces, which will need to be removed.

bashbaug:
I'll fix this - this was something I did a long time ago as part of the "extended subgroups" extension spec, probably before we observed that using the unicode "zero width spaces" in the HTML specs has some problems (see #393),

PR #1228, opened 2024-08-18
generate an OpenCL C built-in functions dictionary

bashbaug:
Possible fix for #1214 

This PR generalizes the script used to generate OpenCL C feature dictionaries from #1212 so it can be used for other types of dictionaries also.

Then:

1. It adds a list of OpenCL C built-in functions and uses the script to generate a dictionary for them.
2. Replaces the built-in functions with hard-coded unicode zero-width spaces with the asciidoctor attributes instead.

I believe this fixes the issue.

A few additional notes:

* I've only added the built-in functions with the hard-coded zero-width spaces.  If desired, we could add additional built-in functions, also.
* I haven't included any formatting in the generated built-in functions dictionary because some functions use **bold** formatting and other functions use `monospace` formatting.  We ought to unify this at some point.

I've tried copy-pasting the function names from the HTML spec with both Chrome and FireFox and they both seem to work fine.  I'm able to copy-paste the function names from the PDF spec, also.


bashbaug:
I found a few formatting issues I need to fix.  Converting to draft.

bashbaug:
The formatting issue is fixed, so this is now ready to go.

scripts/gen_dictionary_from_file.py
```diff
@@ -11,40 +11,41 @@
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
 
-    parser.add_argument('-features', action='store',
+    parser.add_argument('-file', action='store',
```
kpet, line 14:
```suggestion
    parser.add_argument('-i', action='store',
```
or `-inputfile` maybe.

kpet (approved):

scripts/gen_dictionary_from_file.py
bashbaug: Done - I switched to `-i` everywhere in the script, and the Makefile.

kpet (approved):

PR #1230, opened 2024-08-18
clarify cl_mem_flags to not affect copies

### LINKED ISSUE #770: how do cl_mem_flags affect fills and copies
For e.g. [clEnqueueReadBuffer](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueReadBuffer) there is an error condition if the buffer being read from does not support the proper cl_mem_flags:

> CL_INVALID_OPERATION if clEnqueueReadBuffer is called on buffer which has been created with CL_MEM_HOST_WRITE_ONLY or CL_MEM_HOST_NO_ACCESS.

There are no such error condition for fills and copies, however, and this seems like an omission.  The lack of error conditions means that we cannot reliably allocate memory that is guaranteed to be immutable from the host or on the device.

Proposal:

(edit: this is the old proposal, see update below: https://github.com/KhronosGroup/OpenCL-Docs/issues/770#issuecomment-1068729209)

* For memory object fills ([clEnqueueFillBuffer](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueFillBuffer) and [clEnqueueFillImage](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueFillImage)), treat this as access from the host.  Add an error condition if a fill is called on a memory object created with CL_MEM_HOST_READ_ONLY or CL_MEM_HOST_NO_ACCESS.

* For memory object copies (there are a lot of these!  [clEnqueueCopyBuffer](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueCopyBuffer), [clEnqueueCopyImage](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueCopyImage), plus copies between buffers and images, and the rect variants), treat this as access from the device.  Add an error condition if the copy source was created with CL_MEM_WRITE_ONLY, and an error condition if the copy destination was created with CL_MEM_READ_ONLY.  Note that there isn't a standard CL_MEM_NO_ACCESS flag to indicate no access on the device, though there is a version added by an extension.

* For SVM fills ([clEnqueueSVMMemFill](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMemFill)), there will be no error condition because SVM is always accessible on the host.  Revisit if we change the error condition for memory object fills.

* For SVM copies ([clEnqueueSVMMemcpy](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMemcpy)), add error conditions similar to memory object copies.  Revisit if we change the error condition for memory object copies.

* If we go with the behavior above we will also want to update the descriptions for [CL_MEM_READ_ONLY](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_READ_ONLY) and [CL_MEM_WRITE_ONLY](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_WRITE_ONLY) because these flags will also affect copies and not just access "inside a kernel".  

For completeness, [USM](https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_unified_shared_memory.html) allocations  are currently not subject to additional cl_mem_flags, but we would want to add similar error conditions for USM fills and copies if we did eventually support cl_mem_flags for USM allocations.

bashbaug:
Fixes #770 

Clarifies that the cl_mem_flags for the source and destination memory objects for copies are ignored, for both commands enqueued into a command-queue and commands recorded into a command-buffer.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -1254,6 +1254,11 @@ include::{generated}/api/version-notes/clEnqueueCopyBuffer.asciidoc[]
     If _event_wait_list_ and _event_ are not `NULL`, _event_ must not refer
     to an element of the _event_wait_list_ array.
 
+The usage information which indicates whether the memory object can be read
+or written by a kernel and/or the host and is given by the {cl_mem_flags_TYPE}
+argument value specified when _src_buffer_ or _dst_buffer is created is ignored by
+{clEnqueueCopyBuffer}.
+
 // refError
 
 {clEnqueueCopyBuffer} returns {CL_SUCCESS} if the function is executed
@@ -1380,6 +1385,11 @@ If _src_buffer_ and _dst_buffer_ are the same buffer object, _src_row_pitch_
 must equal _dst_row_pitch_ and _src_slice_pitch_ must equal
 _dst_slice_pitch_.
 ====
+
+The usage information which indicates whether the memory object can be read
+or written by a kernel and/or the host and is given by the {cl_mem_flags_TYPE}
+argument value specified when _src_buffer_ or _dst_buffer_ is created is ignored by
+{clEnqueueCopyBufferRect}.
 
 // refError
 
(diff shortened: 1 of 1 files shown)


lakshmih (commented):
LGTM

kpet (approved):
I note that the information is presented as a note for all command buffer commands  but not for enqueue commands (which is in line with the existing spec text for enqueue commands). Fine with the change as is but I'm wondering whether we want to align them all.

bashbaug:
Merging as discussed in the October 29th teleconference.

PR #1233, opened 2024-08-22
Extension proposal for command-buffer internal buffer property

linehill:
This patch proposes a new buffer creation property for creating internal buffers for command-buffers.

HTML is rendered [here](https://htmlpreview.github.io/?https://github.com/linehill/OpenCL-Docs/blob/cmdbuf-internal-buf/extensions/cl_ext_command_buffer_internal_storage.html).


linehill:
cc @pjaaskel.

EwanC (commented):
Interesting proposal thanks, has similarities to the SYCL extension https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_graph_fusion.asciidoc

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+// Creative Commons Attribution 4.0 International License; see
+// http://creativecommons.org/licenses/by/4.0/
+
+:data-uri:
+:icons: font
+include::../config/attribs.txt[]
+//include::{generated}/api/api-dictionary.asciidoc[]
+:source-highlighter: coderay
+
+= cl_khr_command_buffer_internal_buffer
```
pjaaskel, line 11: Having "buffer" twice in the name looks funny. Perhaps "data" or "storage"?

pjaaskel: Also best not use 'khr' before it's khr. Let's use 'ext' or 'exp' for now.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+by a single command-buffer the buffer is associated with. The contents
+of the buffer with this property are not accessible nor observable by
+the host and non-recording commands. The property potentially enables
+runtimes to potentially optimize command-buffers to:
+
+* free space by deallocating the "internal buffers" while the associated
+  command-buffers are not executed and reallocate them when needed.
+
+* reduce memory usage by sharing data storage among the internal
+  buffers. In C analogy:
```
pjaaskel, line 63: Not sure if we need the "C analogy" as the idea is pretty clear.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+void a_command_buffer() {
+  cl_mem buf0 = ..., buf1 = ...;
+  kernelA(in, buf0);
+  kernelB(buf0, buf1);
+  kernelC(buf1, buf0)
+  kernelD(buf0, out);
+}
+----
+* fuse kernels together as intermediate results do not need to be
+  preserved. In C analogy:
```
pjaaskel, line 85: Same here. The idea is clear.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+  cl_mem buf0 = ...;
+  convolutionKernel(in, w, buf0);
+  reluKernel(buf0, out);
+}
+// -->
+void a_command_buffer() {
+  convolutionPlusReluKernel(in, w, out);
+}
+----
+
```
pjaaskel, line 100: Perhaps mention about the OpenVX counterpart, just for reference?

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+// End (Modify Section 5.2.1, *Creating Buffer Objects*)
+
+(Add to the list of error codes for *clEnqueueReadBuffer*, *clEnqueueWriteBuffer*, *clEnqueueReadBufferRect*, *clEnqueueWriteBufferRect*, *clEnqueueCopyBuffer*, *clEnqueueCopyBufferRect*, *clEnqueueFillBuffer* and *clEnqueueMapBuffer*) ::
++
+--
+* *CL_INVALID_MEM_OBJECT* if a memory object passed to this function
+  is a buffer object or references a buffer object created with
+  *CL_MEM_COMMAND_BUFFER_INTERNAL_KHR* property.
+
+// "references a buffer": E.g. sub-buffers.
```
pjaaskel, line 174: Do we want to allow sub-buffers to be created of these internal buffers? I wonder if there's a use case for that.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+--
+* *CL_INVALID_MEM_OBJECT* if a new kernel argument value is a buffer
+  object or references a buffer object created with
+  *CL_MEM_COMMAND_BUFFER_INTERNAL_KHR* property and _command_buffer_
+  is not same as the command-buffer the buffer is associated with.
+--
+
+== Issues
+
+. Should we add memory object query for returning the associated
```
pjaaskel, line 229: I think so, yes.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
```diff
@@ -0,0 +1,256 @@
...
+command-buffer handle?
++
+--
+*UNRESOLVED*
+--
+
+. Should we add a command-buffer query for returning total internal
```
pjaaskel, line 236: Perhaps not. I'd leave this as an implementation internal.

pjaaskel (changes requested):

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
linehill: Removed the analogy.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
linehill: Removed the analogy.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
linehill: Added reference to OpenVX.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
linehill: Perhaps, we want to because OpenVX supports sub-objects of virtual objects.

extensions/cl_khr_command_buffer_internal_buffer.asciidoc
linehill: Added query.

pjaaskel:
> Interesting proposal thanks, has similarities to the SYCL extension https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_graph_fusion.asciidoc

Good that you pointed the SYCL's graph extension out. I'm not familiar with it. So far we have looked mostly on ML/AI formats, OpenVX and CUDA/HIP (task) graphs. Do you think cmdbuf extended with this simple extension can be used to implement a the SYCL graph and the fusion extension here? Do you see gaps that we cannot yet fill efficiently?

EwanC:
> Good that you pointed the SYCL's graph extension out. I'm not familiar with it. So far we have looked mostly on ML/AI formats, OpenVX and CUDA/HIP (task) graphs. Do you think cmdbuf extended with this simple extension can be used to implement the SYCL graph and the fusion extension here? Do you see gaps that we cannot yet fill efficiently?

The graph fusion extension is based on the same underlying implementation this IWOCL talk https://www.youtube.com/watch?v=s39awJ_-W_k 

@sommerlukas Do you have any insights into whether it would be feasible to layer Graph fusion ontop of this for buffer only use-cases.

sommerlukas:
> > Good that you pointed the SYCL's graph extension out. I'm not familiar with it. So far we have looked mostly on ML/AI formats, OpenVX and CUDA/HIP (task) graphs. Do you think cmdbuf extended with this simple extension can be used to implement the SYCL graph and the fusion extension here? Do you see gaps that we cannot yet fill efficiently?
> 
> The graph fusion extension is based on the same underlying implementation this IWOCL talk https://www.youtube.com/watch?v=s39awJ_-W_k
> 
> @sommerlukas Do you have any insights into whether it would be feasible to layer Graph fusion ontop of this for buffer only use-cases.

I think there's definitely some overlap in ideas here. We have similar properties that in combination express similar semantics as the property proposed here. 

In the prototype implementation, we also act on that property by internalizing the dataflow if the property is present. This corresponds to:
> fuse kernels together as intermediate results do not need to be preserved.

I think the biggest difference is the lifetime of the buffer: Our extension and properties do not restrict the buffer's lifetime to the graph. They only indicate that the updates made to the buffer by kernels in the fused graph may not be observable to kernels outside the graph or on the host. That part is identical to what is proposed here.

However, we allow the buffer to be used outside the graph. So, while our properties imply a part of the semantics proposed here, I think additional properties (or other mechanisms) would be required on the SYCL side to apply the proposed property to a buffer.

Issue #1244, opened 2024-08-29
does CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM also apply to kernel arguments?


bashbaug:
Found while working on the fix for #1152.

The description for [`CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM`](https://github.com/KhronosGroup/OpenCL-Docs/issues/new#CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM) is currently (**emphasis** mine):

> This flag indicates whether the kernel uses pointers that are fine grain system SVM allocations. These fine grain system SVM pointers **may be passed as arguments** or defined in SVM buffers that are passed as arguments to kernel.

Additionally, in the "note" below:

> [`CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM) = [`CL_TRUE`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_TRUE) indicates that the OpenCL implementation must assume that system pointers **might be passed as kernel arguments** and/or stored inside SVM allocations passed as kernel arguments.

So, it _seems_ like the setting for `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` should also apply to kernel arguments, but I don't see any description how the setting has any affect on or interaction with kernel arguments.

Here are a few specific questions (pseudocode):

```c++
// Assume that CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM for kernel is initially CL_TRUE:

// Allocate some memory.
void* ptr = malloc(...);

// Assume setting a system SVM kernel argument is initially successful:
clSetKernelArgSVMPointer(kernel, ptr);

// Now disable CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM:
clSetKernelExecInfo(kernel, CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM, CL_FALSE);

// Now is setting a system SVM kernel argument an error?  If so, which error?
// (see also note below)
clSetKernelArgSVMPointer(kernel, ptr);

[...]

PR #1245, opened 2024-08-29
rephrase and correct the descriptions for clSetKernelExecInfo

### LINKED ISSUE #1152: valid values for clSetKernelExecInfo
One of our engineers raised a question internally: if a set of indirectly accesseed SVM (or USM) pointers are set using `clSetKernelExecInfo(CL_KERNEL_EXEC_INFO_SVM_PTRS)`, how are they "un-set"?  To "un-set" all SVM pointers should an empty set be passed to `clSetKernelExecInfo`, specifically one where `param_value_size` is zero (and, possibly, `param_value` is `nullptr`)?

For reference:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetKernelExecInfo

The spec describes an error condition for invalid values, but never describes what values are considered invalid.

> CL_INVALID_VALUE if param_name is not valid, if param_value is NULL or if the size specified by param_value_size is not valid.

I think we should clarify:

1. The empty set is valid and is the proper way to "un-set" any previously set SVM (or USM) pointers.  Specifically:
2. When `param_name` is `CL_KERNEL_EXEC_INFO_SVM_PTRS`, passing zero for `param_value_size` is valid.
3. When `param_name` is `CL_KERNEL_EXEC_INFO_SVM_PTRS`, passing `nullptr` for `param_value` is valid if and only if `param_value_size` is zero.

TBD: When `param_name` is `CL_KERNEL_EXEC_INFO_SVM_PTRS`, is it valid to pass zero for `param_value_size`, and a non-`nullptr` value for `param_value`, or is this an error?

bashbaug:
fixes #1152 

Rephrases and corrects the descriptions for multiple parts of clSetKernelExecInfo.

Removes the cumbersome "note" below the API description and moves the content into the API description instead.  I don't think I lost any important content, but this would be a good area for a careful review.

Explicitly states that an empty set of SVM pointers for CL_KERNEL_EXEC_INFO_SVM_PTRS is valid.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -10433,16 +10433,16 @@ Otherwise, it returns one of the following errors:
     required by the OpenCL implementation on the host.
 --
 
-[open,refpage='clSetKernelExecInfo',desc='Pass additional information other than argument values to a kernel.',type='protos']
+[open,refpage='clSetKernelExecInfo',desc='Set additional information specifying how a kernel will execute.',type='protos']
 --
-To pass additional information other than argument values to a kernel, call
+To set additional information specifying how a kernel will execute, call
 the function
 
 include::{generated}/api/protos/clSetKernelExecInfo.txt[]
 include::{generated}/api/version-notes/clSetKernelExecInfo.asciidoc[]
 
-  * _kernel_ specifies the kernel object being queried.
-  * _param_name_ specifies the information to be passed to kernel.
+  * _kernel_ is a valid kernel object.
+  * _param_name_ specifies the type of information to set.
     The list of supported _param_name_ types and the corresponding values passed
     in _param_value_ is described in the <<kernel-exec-info-table,Kernel
     Execution Properties>> table.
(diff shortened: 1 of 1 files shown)


bashbaug:
Depending how we resolve #1244 we may need to make additional clarifications.

bashbaug:
Converting to draft to work out system SVM updates.

bashbaug:
This is ready for review again.

api/opencl_runtime_layer.asciidoc
```diff
@@ -10460,22 +10459,45 @@ include::{generated}/api/version-notes/clSetKernelExecInfo.asciidoc[]
...
-      | This flag indicates whether the kernel uses pointers that are fine
-        grain system SVM allocations.
-        These fine grain system SVM pointers may be passed as arguments or
-        defined in SVM buffers that are passed as arguments to _kernel_.
+      | Specifies whether the kernel may use pointers to system allocations
+        that are not set directly as kernel arguments.
+
+        When {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} is {CL_FALSE}, behavior
+        is undefined if the kernel acceses a system allocation that is not set as
+        a kernel argument.
```
bashbaug, line 10491: Typo:
```suggestion
        When {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} is {CL_FALSE}, behavior
        is undefined if the kernel accesses a system allocation that is not set as
        a kernel argument.
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -11434,10 +11395,10 @@ Otherwise, it returns one of the following errors:
...
     are not valid events.
   * {CL_INVALID_OPERATION} if SVM pointers are passed as arguments to a kernel
-    and the device does not support SVM or if system pointers are passed as
-    arguments to a kernel and/or stored inside SVM allocations passed as
-    kernel arguments and the device does not support fine grain system SVM
-    allocations.
+    and the device does not support SVM, or if system pointers are passed as
+    arguments to a kernel and the device does not support fine-grain system SVM,
+    or if {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} is {CL_TRUE} and the device
+    does not support fine-grain system SVM.
```
karolherbst, line 11401: I'm not sure if the `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` part is really needed here. An client could use a kernel on multiple devices, some not supporting system SVM. Should we now require applications to flip `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` each time they enqueue the kernel on a different device?

It kinda feels more like an inconvenience, because for the runtime it's more or less meaningless as well. If an application uses a system SVM pointer on a device not supporting it indirectly, having `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` set to `CL_TRUE` or `CL_FALSE` doesn't really change anything about the abnormal behavior the client will see.

There is another aspect to this, by default `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` is `CL_TRUE`/`CL_FALSE` depending on the device. An application _might_ set it to `CL_FALSE` for devices supporting system SVM to be nice to the runtime, but maybe they want to revert that, so now they'll need to toggle, which earlier they probably wouldn't had to.

I think it would probably make more sense to simply state, that `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` is assumed to be always `CL_FALSE` on devices the kernel gets enqueued not supporting system SVM or that this flag only has an impact on kernels for devices supporting system SVM.

Though to be fair, I doubt there is any applications doing that, but so far it just sounds like an inconvenience gets added for no real benefit.

bashbaug: Full disclosure: I've never liked `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM`.  Having the default state dependent on the capabilities of the device always seemed especially problematic.

This is an interesting idea:

> I think it would probably make more sense to simply state, that `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` is assumed to be always `CL_FALSE` on devices the kernel gets enqueued not supporting system SVM or that this flag only has an impact on kernels for devices supporting system SVM.

I think this could solve the default state problem as well, since then the default could be `CL_TRUE` if **any** device supports system SVM, and it'd just be ignored for devices that do not support system SVM.

If we're OK with this, I think I'd just need to make a few small edits to remove this error condition and to tidy up the `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` description.

bashbaug: I think I've done this now, please take a look.  In short:

1. To use system SVM allocations indirectly within a kernel, the device must support system SVM, and `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` must be `CL_TRUE`.  If either of these conditions is false, accessing a system allocation within a kernel is undefined behavior.
    * Note specifically: There is no requirement to return an error if either condition is false.  Because behavior is undefined, returning an error is allowed, but not required.
3. The default value for `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` is `CL_TRUE`, no change here.
4. Trying to set `CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM` to `CL_TRUE` if none of the devices in _context_ support system SVM is an error, no change here.

karolherbst (approved):
looks good!

alycm (approved):
LGTM.

PR #1249, opened 2024-09-03
Add error when memory objects or semaphores are created with more than one external handle

### LINKED ISSUE #1246: Behaviour of `clCreate{Buffer,Image}WithProperties` when passing multiple external memory handles
As far as I can tell, there does not seem to be a specification rule that forbids an application from passing multiple external memory handles, of the same or different types, when creating a buffer or an image. This feels like an oversight. How about adding an error condition like the following:
>`CL_INVALID_PROPERTY`
◦ if `properties` includes more than one external memory handle.

kpet:
Also link the description of external semaphore handles to the spec for clCreateSemaphoreWithPropertiesKHR.

Fixes #1246

Change-Id: Ifb4c02795c6d4db8aee9b5f14b10fecd26992fd5

### DIFF
--- a/api/cl_khr_external_memory.asciidoc
+++ b/api/cl_khr_external_memory.asciidoc
@@ -6,7 +6,7 @@ include::{generated}/meta/{refprefix}cl_khr_external_memory.txt[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2024-03-15
+    2024-09-03
 *IP Status*::
     No known IP claims.
 *Contributors*::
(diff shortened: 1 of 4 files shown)


bashbaug (approved):
Changes LGTM.

I'd be comfortable calling this a patch revision to version 1.0.0 since there was no previously defined behavior for this scenario (and no CTS test for it, also), but if folks  prefer a minor revision update I'm fine with that too.

bashbaug:
Noting for completeness: #1256 updated error behavior and we're currently calling it a "patch" revision.  I think this is making a similar update?  If so, we should either call this a "patch" revision also, or change #1256 to be a "minor" revision update.

lakshmih (approved):

PR #1252, opened 2024-09-04
add nofooter by default when building the OpenCL specs

### LINKED ISSUE #1222: Consider Disabling Asciidoctor Footer to Reduce Diffs
The SPIR-V working group is considering a PR that disables the asciidoctor "footer" to reduce diffs.

https://github.com/KhronosGroup/SPIRV-Registry/pull/268

We should consider doing the same in the OpenCL spec toolchain as well.  I know this will simplify reviews for the generated reference pages considerably, for example.

@oddhack, is this somethign that's been considered for other working groups (Vulkan)?

bashbaug:
fixes #1222

This will minimize diffs in the future, especially for things like the reference pages that may change infrequently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,7 @@ COMMONATTRIBOPTS	= -a revdate="$(SPECDATE)" \
 			  -a stem=latexmath \
 			  -a generated=$(GENERATED) \
 			  -a sectnumlevels=5 \
+			  -a nofooter \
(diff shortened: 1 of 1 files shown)


oddhack (approved):
LGTM. I think this is a relatively recent addition and I should probably do it for Vulkan, too - thanks!

PR #1256, opened 2024-09-09
unify the CL_INVALID_COMMAND_QUEUE behavior for semaphore signals and waits

### LINKED ISSUE #1217: Spec of negative result for waiting on and signaling semaphores differs
Spec states that for `clEnqueueSignalSemaphoresKHR`, negative result `CL_INVALID_COMMAND_QUEUE` may happen if:

`if one or more of sema_objects belong to a context that does not contain a device associated with command_queue`

For similar `clEnqueueWaitSemaphoresKHR` this case is missing - rest of negative spec is identical, is that intentional ? I went briefly through the graph of relations between context-device-queue-semaphore and it seems like this case is redundant and fells into `CL_INVALID_CONTEXT` case.

bashbaug:
Fixes #1217 

Unifies the CL_INVALID_COMMAND_QUEUE behavior for clEnqueueSignalSemaphoresKHR and clEnqueueWaitSemaphoresKHR.

Also bumps the patch revision, as described in the suggested resolution.

### DIFF
--- a/api/cl_khr_semaphore.asciidoc
+++ b/api/cl_khr_semaphore.asciidoc
@@ -255,3 +255,6 @@ while (true) {
      {CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR} (provisional).
   * Revision 1.0.0, 2024-03-15
   ** First non-provisional version.
+  * Revision 1.0.1, 2024-09-08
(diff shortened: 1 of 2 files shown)


alycm (approved):
LGTM, change is one of the two possible resolutions on the ticket, I see no problem with taking this one.

PR #1257, opened 2024-09-09
clarify clGetSemaphoreHandleForTypeKHR is part of cl_khr_external_semaphore

### LINKED ISSUE #1191: typo in description for clReImportSemaphoreSyncFdKHR

bashbaug:
fixes #1191 

* Puts clGetSemaphoreHandleForTypeKHR in the summary for cl_khr_external_semaphore, not cl_khr_external_semaphore_sync_fd.
* Adds the version note for clGetSemaphoreHandleForTypeKHR.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -12940,6 +12940,7 @@ Please refer to handle specific documentation for more details on transference r
 To export an external handle from a semaphore, call the function
 
 include::{generated}/api/protos/clGetSemaphoreHandleForTypeKHR.txt[]
+include::{generated}/api/version-notes/clGetSemaphoreHandleForTypeKHR.asciidoc[]
 
   * _sema_object_ specifies a valid semaphore object with exportable
     properties.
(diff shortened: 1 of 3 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -12940,6 +12940,7 @@ Please refer to handle specific documentation for more details on transference r
 To export an external handle from a semaphore, call the function
 
 include::{generated}/api/protos/clGetSemaphoreHandleForTypeKHR.txt[]
+include::{generated}/api/version-notes/clGetSemaphoreHandleForTypeKHR.asciidoc[]
```
kpet, line 12943: I'm wondering whether we should include the version notes in the generated proto snippet. It seems we mostly (always?) include them right after the proto snippet.

kpet (approved):

api/opencl_runtime_layer.asciidoc
alycm: I honestly can't remember if there was a reason I put them in a separate file, rather than adding them to the proto.

alycm (approved):
LGTM.

PR #1260, opened 2024-09-15
[XML] Fix type defined by `cl_intel_driver_diagnostics`

SunSerega:
https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_driver_diagnostics.txt

XML says this extension defines `cl_diagnostics_verbose_level`, but the spec says it's `cl_diagnostic_verbose_level_intel` (and with base type `cl_bitfield` rather than `cl_uint` - which checks out in favour of .txt spec)


bashbaug (approved):
Thank you for fixing these!

PR #1263, opened 2024-09-15
[XML] `ARM` => `Arm`

SunSerega:
In the cl.xml file, in 4 places, it was `ARM`; in 5, `Arm`.

A minor thing, but all other vendor names are consistent, and this was tripping sanity checks in my bindings, so I fixed it a while ago. And now when reviewing - I found that I've never reported this back.


kpet (approved):
LGTM, thanks!

PR #1273, opened 2024-10-22
clarify the minimum value for CL_DEVICE_HALF_FP_CONFIG

### LINKED ISSUE #1267: minimum values for CL_DEVICE_HALF_FP_CONFIG should not be tied to an OpenCL version
The current minimum value for [CL_DEVICE_HALF_FP_CONFIG](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_HALF_FP_CONFIG) has a dependency on OpenCL 2.0 or newer:

> If half-precision is supported by the device, then the minimum half-precision floating-point capability for OpenCL 2.0 or newer devices is: [...]

I don't think this was intended and was introduced when the extensions were incorporated into the core specification.  Note that there is no similar requirement in the last separate extension doc v3.0.15.

We should remove "for OpenCL 2.0 or newer devices", and check that there are no similar requirements for any other fp16 queries.

bashbaug:
fixes #1267

The minimum value for CL_DEVICE_HALF_FP_CONFIG applies to devices supporting all OpenCL versions, not just for OpenCL 2.0 or newer devices.

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -1929,8 +1929,7 @@ include::{generated}/api/version-notes/CL_DEVICE_HALF_FP_CONFIG.asciidoc[]
         addition, subtraction, multiplication) are implemented in software
 
         If half-precision is supported by the device, then the minimum
-        half-precision floating-point capability for OpenCL 2.0 or newer
-        devices is:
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM.

PR #1280, opened 2024-11-06
Add specification for cl_ext_immutable_memory_objects

### LINKED ISSUE #1110: The current specification makes it impossible to support a fully read-only image format (and related `cl_mem_flags` issues)
As part of adding support for a new image format for which we can only support read access, we're bumping into the following rule (there are similar notes for **clEnqueueFillBuffer**, **clCommandFillBufferKHR**, and **clCommandFillImageKHR**; these notes have been there since the beginning of our git history for **clEnqueueFill\*** commands):

>The usage information which indicates whether the memory object can be read or written by a kernel and/or the host and is given by the `cl_mem_flags` argument value specified when image is created is ignored by **clEnqueueFillImage**.

Looking at _Table 19. List of supported memory flag values_, it seems access control flags fall into two groups:

* `CL_MEM_READ_WRITE`, `CL_MEM_READ_ONLY`, and `CL_MEM_WRITE_ONLY` specify what accesses kernels are allowed to perform on a memory object.
* `CL_MEM_HOST_WRITE_ONLY`, `CL_MEM_HOST_READ_ONLY`, and `CL_MEM_HOST_NO_ACCESS` specify what accesses the host is allowed to perform on a memory object.

1. The current state of the specification looks inconsistent
  * Do we need notes similar to the ones quoted above for copy commands? They are neither kernels nor require host access, which   Table 19 suggests applies to read, write, or map commands.
  * Should the notes for fill commands be removed? There is nothing specific to fill commands that I can spot that justifies the presence of the notes in the absence of similar notes for copy commands for example.
2. This strikes me as an odd split of capabilities. On many devices access to memory is controlled using settings that apply to all device accesses that often would cover fill commands too. This means that it is impossible on these devices to enforce read-only access permissions for kernels if fill commands run on the device and need write permissions to the same memory. One possible change we could consider would be to change the `CL_MEM_*` flags so they apply to all device commands and remove the notes on fill commands.
3. If neither 2 nor removing notes on fill commands are an acceptable resolution, what mechanism do we think should be used by implementations that wish to support a truly read-only image format?

kpet:
Fixes #1110

Change-Id: I120f66ad20f977c251b77cc42a32021cb407518e

### DIFF
--- a/api/cl_ext_immutable_memory_objects.asciidoc
+++ b/api/cl_ext_immutable_memory_objects.asciidoc
@@ -1,0 +1,43 @@
+// Copyright 2024 The Khronos Group Inc.
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}cl_ext_immutable_memory_objects.txt[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2024-11-06
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Kevin Petit, Arm Ltd.
+  - Ahmed Hesham, Arm Ltd.
+  - Michael Rizkalla, Arm Ltd.
+*Interactions*::
+  - This extension interacts with {cl_khr_egl_image_EXT}.
+
+=== Description
+
+{cl_ext_immutable_memory_objects_EXT} provides a mechanism to create memory
+objects that cannot be modified by the implementation after creation. This,
+for example, enables implementation to support image formats for which no write
+operations can be supported.
+
+=== New Enums
+
+  * {cl_mem_flags_TYPE}
+  ** {CL_MEM_IMMUTABLE_EXT}
+
+=== Issues
+
+1) Can {CL_MEM_READ_ONLY} be used instead of {CL_MEM_IMMUTABLE_EXT}?
+--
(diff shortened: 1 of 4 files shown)


bcalidas:
Looks ok at first glance

bashbaug:
Discussed in the January 10th teleconference:

- Next step is for Arm to push the CTS tests.
- The merge conflict is only because we already merged the bit reservations, so nothing to worry about.

kpet:
CTS tests are now available: https://github.com/KhronosGroup/OpenCL-CTS/pull/2286

bashbaug (approved):
Merging as discussed in the March 4th teleconference.

eyalroz:
I don't understand how this makes sense.

Instead of deciding which flags govern the behavior of copy and fill, or adding flags for whether copy and fill are allowed - a semi-clashing new flag was added. Why should the user believe the being immutable mean anything other than no host write and no kernel write? If your answer is "because copy and fill are not like kernels" - that's not actually stated anywhere; plus, if that were true - where are the `CL_API_FUNC_MEM_READ_WRITE`, `CL_API_FUNC_MEM_READ_ONLY`, `CL_API_FUNC_MEM_WRITE_ONLY` flags? Or at least `CL_API_FUNC_MEM_WRITE`?

I realize this was committed last year, but - this doesn't seem like the right flag to introduce.

PR #1282, opened 2024-11-08
add cl_khr_unified_svm extension

bashbaug:
This PR contains the draft specification for `cl_khr_unified_svm`.  It is intended to provide early access for review and community feedback, and it is not a ratified specification.

`cl_khr_unified_svm` adds additional types of Shared Virtual Memory (SVM) to OpenCL.  Compared to Coarse-Grained and Fine-Grained SVM in OpenCL 2.0 and newer, the additional types of SVM added by this this extension provides:

* Sufficient functionality to implement "Unified Shared Memory" (USM) in other APIs, such as SYCL.

* Additional control over the ownership and accessibility of SVM allocations, to more precisely choose between application performance and programmer convenience.

* A simpler programming model, by automatically migrating more SVM allocations between devices and the host, or by accessing more SVM allocations on the host without needing to map or unmap the allocation.

Specifically, this extension provides:

* Extensible interfaces to support many types of SVM, including the SVM types defined in core OpenCL, in this extension, and additional SVM types defined by other combinations of SVM capabilities.

* Explicit control over memory placement and migration by supporting device-owned SVM allocations for best performance, host-owned SVM allocations for wide visibility, and shared SVM allocations that may migrate between devices and the host.

* The ability to query detailed SVM capabilities for each SVM allocation type supported by a platform and device.

* Additional properties to control how memory is allocated and freed, including properties to associate an SVM allocation with both a device and a context.

* A mechanism to indicate that a kernel may access SVM allocations indirectly, without passing a set of indirectly accessed SVM allocations to the kernel, improving usability and reducing driver overhead for kernels that access many SVM allocations.

* A new query function to query properties of an SVM allocation.

* A new function to suggest an SVM allocation type for a set of SVM capabilities.

Because the interfaces defined by this specification are not final and are subject to change they are not intended to be used by shipping software products.  If you are interested in using this feature in your software product, please let us know!


illuhad (commented):
Great to see at last what you've been working on :)

You asked for feedback, so here are some things I noticed:

- I can see that the ability to request individual SVM features can be very powerful, if implementations are happy to carry the burden of dealing with all of the possible combinations. However, I am a bit concerned with adoption if the "common cases" of device/host/shared SVM that you describe are hidden underneath some not necessarily obvious  combinations of bitfield values. Can we perhaps have convenience shortcuts for these common cases? This could make user code simpler and make it more attractive to use this memory management model, and it could provide guidance for users and vendors if everybody knew clearly which combinations of capabilities to focus their testing on. This could help portability and adoption.
- Can we have the ability to register an existing memory allocation to make it device accessible, similarly to `cudaHostRegister`?

bashbaug:
Thank you for the feedback!

> * I can see that the ability to request individual SVM features can be very powerful, if implementations are happy to carry the burden of dealing with all of the possible combinations. However, I am a bit concerned with adoption if the "common cases" of device/host/shared SVM that you describe are hidden underneath some not necessarily obvious  combinations of bitfield values. Can we perhaps have convenience shortcuts for these common cases? This could make user code simpler and make it more attractive to use this memory management model, and it could provide guidance for users and vendors if everybody knew clearly which combinations of capabilities to focus their testing on. This could help portability and adoption.

This is the intent behind the SVM type "capability macros", which we've defined in the spec and in the headers.  We've currently defined "macros" for the common cases we've been thinking about, and if other combinations become common they can be added with a simple header update.

Specifically, we've defined capability macros for:

* `CL_SVM_TYPE_MACRO_COARSE_GRAIN_BUFFER_KHR`
* `CL_SVM_TYPE_MACRO_FINE_GRAIN_BUFFER_KHR`
* `CL_SVM_TYPE_MACRO_DEVICE_KHR`
* `CL_SVM_TYPE_MACRO_HOST_KHR`
* `CL_SVM_TYPE_MACRO_SINGLE_DEVICE_SHARED_KHR`
* `CL_SVM_TYPE_MACRO_SYSTEM_KHR`

The capability macros provide a very concise and easy mechanism to request these types of SVM, for [example](https://github.com/bashbaug/SimpleOpenCLSamples/blob/6d880789e77444607f1ef572f2ca72e9df87e29a/samples/usvm/100_udmemhelloworld/main.cpp#L74), here is a way to allocate memory equivalent to SYCL "device USM":

```c++
    cl_uint index = CL_UINT_MAX;
    clGetSVMSuggestedTypeIndexKHR(
        context(),
        CL_SVM_TYPE_MACRO_DEVICE_KHR,
        0,
        props,
        0,
        &index);

    cl_uint* d_src = (cl_uint*)clSVMAllocWithPropertiesKHR(
        context(),
        props,
        index,
        gwx * sizeof(cl_uint),
        nullptr );
```

> * Can we have the ability to register an existing memory allocation to make it device accessible, similarly to `cudaHostRegister`?

We have this request recorded (in document issue (5)), but we are not currently considering it in scope for the initial extension revision, primarily because this functionality isn't supported by the Intel USM extension and hence isn't required to implement SYCL.  Does that sound reasonable?

If we don't include it in the initial extension, we could support it via a layered extension, or a subsequent extension version.

illuhad:
Hi Ben!

> The capability macros provide a very concise and easy mechanism to request these types of SVM, for [example](https://github.com/bashbaug/SimpleOpenCLSamples/blob/6d880789e77444607f1ef572f2ca72e9df87e29a/samples/usvm/100_udmemhelloworld/main.cpp#L74), here is a way to allocate memory equivalent to SYCL "device USM":

Ah, I see, thanks for pointing this out. This was not very obvious to me. Perhaps other users might feel the same way?
For me this workflow would certainly work, but then I'm more of a framework vendor rather than an application developer/end user, so sensitivities might be different.

>We have this request recorded (in document issue (5)), but we are not currently considering it in scope for the initial extension revision, primarily because this functionality isn't supported by the Intel USM extension and hence isn't required to implement SYCL. Does that sound reasonable?

I agree that it is not needed for SYCL (at least not today), but I would need it for C++ standard parallelism offloading rather sooner than later in order to make the stack GPU accessible. But I appreciate that the scope of individual extensions must be limited in some way to keep the work manageable in scope and size :)

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1082 @@
...
+| `CL_SVM_CAPABILITY_HOST_READ_KHR`
+  | This type of SVM is readable on the host without needing to map or unmap the allocation.
+| `CL_SVM_CAPABILITY_HOST_WRITE_KHR`
+  | This type of SVM is writeable on the host without needing to map or unmap the allocation.
+| `CL_SVM_CAPABILITY_HOST_MAP_KHR`
+  | This type of SVM is accessible on the host but requires mapping and unmapping the allocation.
+| `CL_SVM_CAPABILITY_DEVICE_READ_KHR`
+  | This type of SVM is accessible on the device for reading.
+| `CL_SVM_CAPABILITY_DEVICE_WRITE_KHR`
+  | This type of SVM is writeable on the device for writing.
```
bashbaug, line 314: We should clarify exactly what is meant by "accessible on the device for reading" and "accessible on the device for writing", given issues like #770.

* Do these capabilities only apply to kernels, or do they apply to other APIs as well?

Specifically:
* What capabilities are needed for clEnqueueSVMMemcpy?  Does it matter if the SVM allocation is the memcpy source or destination?
* What capabilities are needed for clEnqueueSVMMemFill?
* What capabilities are needed for clEnqueueSVMMigrateMem?

Minor, there is a small typo here too: "writeable on the device for writing" should be "accessible on the device for writing".

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1081 @@
...
+
+Accepted values for the _param_name_ parameter to *clSetKernelExecInfo* to enable and disable indirect access to SVM allocations made by *clSVMAllocWithPropertiesKHR* or *clSVMAlloc*:
+
+[source]
+----
+/* cl_kernel_exec_info */
+#define CL_KERNEL_EXEC_INFO_SVM_INDIRECT_ACCESS_KHR 0x11BB
+----
+
+== Modifications to the OpenCL API Specification
```
b-sumner, line 243: What about updates to section 3.3 Memory Model?

bashbaug:
Mentioned in the January 7th teleconference: I have OpenCL Intercept Layer support for unified SVM done and in a branch, which I will try to keep up-to-date with the latest specs.  I'm open to suggestions, but I probably won't merge the branch until the extensions is a bit further along.

The branch can be found here:
https://github.com/bashbaug/opencl-intercept-layer/tree/cl_khr_unified_svm

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1081 @@
...
+--
+*UNRESOLVED*:
+Trending "no" for device and shared USM allocations.
+If the USM allocation is from the same context this could be an error, such as `CL_INVALID_HOST_PTR`.
+If the USM allocation is from a different context then behavior could be undefined.
+
+Trending "yes" for host USM allocations, both when the host USM allocation is from this context and from another context.
+--
+
+. Can a pointer to a device, host, or shared SVM allocation be passed to API functions to read from or write to `cl_mem` objects, such as *clEnqueueReadBuffer* or *clEnqueueWriteImage*?
```
bashbaug, line 912: Discussed briefly in the January 7th memory subgroup.  We may want to consider buffers and images separately.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1081 @@
...
+`RESOLVED`: We decided not to accept any `cl_mem_flags` or `cl_svm_mem_flags`, and added access properties instead.
+--
+
+. Exactly how do the additional SVM types affect the memory model?
++
+--
+*UNRESOLVED*:
+--
+
+. Should it be an error to set an unknown pointer as a kernel argument using *clSetKernelArgSVMPointer* if no devices support shared system allocations?
```
bashbaug, line 948: Discussed in the January 7th memory subgroup.  For SYCL, we generally want to follow C++ semantics, meaning that it is valid to pass any pointer to a function (or kernel) and it is only an problem (specifically, undefined behavior) to dereference a bogus pointer.  This would imply that it should NOT be an error to set a kernel argument that does not point into a memory region returned by clSVMAlloc, even for an implementation that didn't support system SVM.

See: https://github.com/KhronosGroup/OpenCL-Docs/pull/905

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Discussed in the November 29th memory subgroup.  We don't want to repeat the mistakes we made leading to #770, so our intent is that these capabilities apply to all forms of device access, not just kernels.  Therefore:

* What capabilities are needed for clEnqueueSVMMemcpy?
    * DEVICE_READ is needed to be a memcpy source, and DEVICE_WRITE is needed for a memcpy dst.
* What capabilities are needed for clEnqueueSVMMemFill?
    * DEVICE_WRITE is needed to be a memfill dst. 
* What capabilities are needed for clEnqueueSVMMigrateMem?
    * TBD - do we want DEVICE_READ to migrate to the host (via CL_MIGRATE_MEM_OBJECT_HOST), and DEVICE_WRITE to migrate to the device?

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1081 @@
...
+* `CL_INVALID_VALUE` if _flags_ contains an invalid SVM free flag.
+* `CL_INVALID_VALUE` if _ptr_ is not a value returned by *clSVMAlloc*, *clSVMAllocWithPropertiesKHR*, or a `NULL` pointer.
+* `CL_OUT_OF_RESOURCES` if there is a failure to allocate resources required by the OpenCL implementation on the device.
+* `CL_OUT_OF_HOST_MEMORY` if there is a failure to allocate resources required by the OpenCL implementation on the host.
+
+By default, *clSVMFreeWithPropertiesKHR* does not wait for previously enqueued commands that may be using _ptr_ to finish before freeing _ptr_.
+It is the responsibility of the application to make sure enqueued commands that use _ptr_ are complete before freeing _ptr_.
+Behavior is undefined if a previously enqueued command that may be using _ptr_ is still executing.
+Applications should take particular care freeing memory allocations with kernels that may access memory indirectly, since a kernel that accesses memory indirectly may be using any memory allocation of the specified type or types.
+To wait for previously enqueued commands to finish that may be using _ptr_ before freeing _ptr_, use the flag `CL_SVM_FREE_BLOCKING_KHR`.
```
bashbaug, line 545: SYCL is clarifying the description of `sycl::free`, see: https://github.com/KhronosGroup/SYCL-Docs/pull/758

We should consider how this could / should affect unified SVM free behavior.

A few thoughts:

1. The updated SYCL spec explicitly says (as a non-normative note) that the SYCL free may be "blocking" or "non-blocking".  As long as we are OK with a similar behavior in the unified SVM spec, we can replace the explicit `CL_SVM_FREE_BLOCKING_KHR` flag with a similar non-normative note, which would be a nice simplification.
    * This would not preclude adding additional flags or properties to enforce more specific freeing behaviors in the future, if desired.
2. If we remove the `CL_SVM_FREE_BLOCKING_KHR` flag then will not be adding any new SVM free flags or SVM free properties, so strictly speaking we could also remove `clSVMFreeWithPropertiesKHR` from the spec as well.  I'm still inclined to keep it, though, both to add extensibility for the future, and because `clSVMFree` does not return an error code.
3. We should take a closer look at `clEnqueueSVMFree` to ensure it does what we want it to do, to asynchronously free SVM via a command in a command queue.

paulfradgley: There's no mention of command flushing here. Does this function get added to the list of functions that implicitly flush? (blocking clEnqueue commands and clReleaseCommandQueue)

Or must the application ensure that the commands are flushed another way?

bashbaug: I don't think `clSVMFree` or `clSVMFreeWithPropertiesKHR` should _require_ any implicit flushing.

* Our trend has been to reduce requirements to implicitly flush, rather than adding them.
* The existing `clSVMFree` does not have any requirement to implicitly flush.

That said, I think it is _allowable_ for an implementation to flush on a call to `clSVMFree` if required or desired.  Does this make sense?

Note, if we align more closely to the proposed SYCL behavior, then we would be even more permissive, because `clSVMFree` would be allowed to flush and wait for commands to complete.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1069 @@
...
+* `CL_INVALID_PROPERTY` if a memory property name in _properties_ is not a supported property name, if the value specified for a supported property name is not valid, or if the same property name is specified more than once.
+* `CL_INVALID_VALUE` if _flags_ contains an invalid SVM free flag.
+* `CL_INVALID_VALUE` if _ptr_ is not a value returned by *clSVMAlloc*, *clSVMAllocWithPropertiesKHR*, or a `NULL` pointer.
+* `CL_OUT_OF_RESOURCES` if there is a failure to allocate resources required by the OpenCL implementation on the device.
+* `CL_OUT_OF_HOST_MEMORY` if there is a failure to allocate resources required by the OpenCL implementation on the host.
+
+[NOTE]
+====
+Whether *clSVMFree* or *clSVMFreeWithPropertiesKHR* is blocking or non-blocking is unspecified.
+Applications should not rely on *clSVMFree* or *clSVMFreeWithPropertiesKHR* for synchronization, nor assume that *clSVMFree* or *clVMFreeWithPropertiesKHR* cannot cause deadlocks.
```
kpet, line 542: As discussed in the 2025/03/18 memory subgroup TSG, I think this note should be reworded to make it clear that
- Deadlocks can only happen when **clSVMFree** (or **clSVMFreeWithPropertiesKHR**) is called on an allocation that may be used by commands pending execution.
- "blocking" means "waiting for pending commands that may use the allocation to complete execution"

With these changes, this would be allowed by the "behavior is undefined" statement above but, strictly speaking, the current wording for the note suggests that deadlocks could happen in any scenarios.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1081 @@
...
+. At least for HOST and SHARED allocations, should we have separate mem alloc flags for the host and the device?
++
+--
+`RESOLVED`: We removed the _flags_ argument entirely.
+--
+
+. What are invalid values for `ptr` and `size` for *clEnqueueSVMMigrateMem*?
+How about *clEnqueueSVMMemFill* and *clEnqueueSVMMemcpy*?
+Specifically, is `NULL` a valid value for `ptr`?
+Is `size` equal to zero valid?
```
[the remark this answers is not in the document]
bashbaug: Discussed in the July 1st memory subgroup.  Our tentative resolution is:

* A `size` equal to zero is valid. When `size` is zero, the call to *clEnqueueSVMMigrateMem*, *clEnqueueSVMMemFill*, and *clEnqueueSVMMemcpy* trivially succeeds, similar to an enqueued marker. This appears to be the specified behavior for the C `memcpy` and `memset` functions.
* A `ptr` equal to `NULL` is valid if and only if `size` is equal to zero, otherwise it is an error.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Discussed in the July 1st memory subgroup.  Given that we recently updated the Intel USM extension to follow these C++ semantics at the request of our SYCL runtime, I strongly recommend that we adopt this behavior for the unified SVM extension also.  This means it would not be an error to set a bogus kernel argument and only would be an error to dereference a bogus pointer in the kernel.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1139 @@
...
+If a `NULL` pointer is returned then `errno` may be set to an implementation-defined value.
+If a unique non-null pointer is returned then it cannot be dereferenced.
+* Allocating an array of zero elements using `new` must return a non-null pointer, though dereferencing the pointer is undefined.
+
+For reference, the full set of options we considered were:
+
+.. [.line-through]#Allow zero-sized allocations and require returning a non-null pointer that must be freed.#
+.. Allow zero-sized allocations and require returning a `NULL` pointer.
+No error will be generated.
+Note, it is not currently an error to free a `NULL` pointer.
```
bashbaug, line 1254: Discussed in the July 1st memory subgroup; see tentative resolution above.

As a follow-up, while the current OpenCL spec says that it is not an error to free a `NULL` pointer using [clSVMFree](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSVMFree):
> If a NULL pointer is passed in svm_pointer, no action occurs.

Should this apply to the asynchronous [clEnqueueSVMFree](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMFree) also?  This case is not explicitly described.

extensions/cl_khr_unified_svm.asciidoc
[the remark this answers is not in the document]
bashbaug: This has been resolved in the list of issues in the extension doc, as described above, so I am resolving this Github comment also.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: This is resolved in the spec document and is tested in the CTS.  It is not an error to set an unknown pointer as a kernel argument and it is only an error to dereference the unknown pointer in a kernel.  Therefore, I am resolving this GitHub comment also.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: The current extension text aligns very closely with the SYCL text, specifically:

> Whether clSVMFree or clSVMFreeWithPropertiesKHR is blocking or non-blocking is unspecified. Applications should not rely on clSVMFree or clSVMFreeWithPropertiesKHR for synchronization, nor assume that clSVMFree or clSVMFreeWithPropertiesKHR cannot cause deadlocks.

So, I think we can resolve this GitHub comment thread.

Note, there is another open comment thread to discuss refining the text further.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+  | This type of SVM is accessible to other devices in the context that support the SVM type.
+| {CL_SVM_CAPABILITY_HOST_OWNED_KHR}
+  | This type of SVM is owned by the host and is not intended to migrate to a device.
+    Allocations that are owned by the host generally trade off wide accessibility for potentially higher per-access costs.
+| {CL_SVM_CAPABILITY_HOST_READ_KHR}
+  | This type of SVM is readable on the host without needing to map or unmap the allocation.
+| {CL_SVM_CAPABILITY_HOST_WRITE_KHR}
+  | This type of SVM is writeable on the host without needing to map or unmap the allocation.
+| {CL_SVM_CAPABILITY_HOST_MAP_KHR}
+  | This type of SVM is accessible on the host but requires mapping and unmapping the allocation.
```
bashbaug: From a discussion in the March 3rd memory subgroup: The description of this capability is too strong ("requires").  As written, it cannot be combined with, say, the HOST_READ and HOST_WRITE capabilities.

Since this capability is intending to say that this type of SVM supports mapping and unmapping, how about:

```suggestion
| {CL_SVM_CAPABILITY_HOST_MAP_KHR}
  | This type of SVM supports mapping and unmapping for access on the host.
```

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+        Each entry in the returned array must be either a super-set of the entry in the array returned by the platform query {CL_PLATFORM_SVM_TYPE_CAPABILITIES_KHR}, indicating that the SVM type is supported by the device, or zero, indicating that the SVM type is not supported by this device.
+
+        Please refer to the <<svm-capabilities-table,SVM capabilities>> table for valid capability values and their description.
+| {CL_DEVICE_SVM_CONCURRENT_ACCESS_ATOM_SIZE_KHR}
+  | {size_t_TYPE}
+      | Queries the size and alignment in bytes that bound concurrent access to SVM allocations.
+        The returned value must be zero or a power of two.
+
+        If the device does not support concurrent access to SVM allocations, then the query must return `0`, indicating that concurrent access is not supported.
+        If the query for {CL_DEVICE_SVM_CAPABILITIES} includes {CL_DEVICE_SVM_FINE_GRAIN_BUFFER} or {CL_DEVICE_SVM_FINE_GRAIN_SYSTEM}, then the query must return one, indicating that concurrent access occurs at the granularity of individual bytes within allocations.
```
joselopeqti, line 305: Should this refer to {CL_DEVICE_SVM_FINE_GRAIN_BUFFER} with {CL_MEM_SVM_ATOMICS} ?

bashbaug: I don't think so.  Broadly speaking, if an implementation supports fine-grain buffer SVM without SVM atomics, then it should support (fine-grain) concurrent access.  Refer to:

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#shared-virtual-memory

>**Fine-Grained buffer SVM**: Sharing occurs at the granularity of individual loads/stores into bytes within OpenCL buffer memory objects.

Then, CL_MEM_SVM_ATOMICS adds concurrent **atomic** access on top of this.  Kind of make sense?

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+=== Section 4.1 - Querying Platform Info:
+
+Add to Table 3 - List of supported param_names by {clGetPlatformInfo}:
+
+[caption="Table 5. "]
+.List of supported param_names by clGetDeviceInfo
+[width="100%",cols="<30%,<20%,<50%",options="header"]
+|====
+| Device Info | Return Type | Description
+| {CL_PLATFORM_SVM_TYPE_CAPABILITIES_KHR}
```
joselopeqti, line 271: Would a platform be allowed to return an entry with a bitfield set to 0?
If not, is this something we would like to allow?

bashbaug: Gut feeling: in a perfect world, I'd prefer that we did not need to allow this, but in an imperfect world I can see how this might be useful to support "unused" entries.

Would you mind filing a separate issue to discuss in more detail?  Regardless what we decide, we should add something to the unified SVM consistency test, to either disallow this (if we decide to disallow it) or to check for any other requirements (if we decide to allow it).

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Resolving after discussion in the March 24th teleconference.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Discussed in the March 24th memory subgroup.  We can think of reasons why this might be useful, so we're leaning towards allowing entries with a bitfield set to zero.

Tagging @kpet for additional feedback.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+| {CL_SVM_INFO_CAPABILITIES_KHR}
+  | {cl_svm_capabilities_khr_TYPE}
+      | Returns the SVM capabilities for the SVM allocation for the specified _device_.
+        If _device_ is `NULL` and there is a device associated with the SVM allocation, returns the SVM capabilities for the device associated with the SVM allocation.
+        If _device_ is `NULL` and there is no device associated with the SVM allocation, returns the SVM capabilities for all devices in _context_ supporting the SVM allocation.
+
+        Returns `0` if _ptr_ does not point into an SVM allocation returned from {clSVMAllocWithPropertiesKHR} or {clSVMAlloc} for _context_.
+
+| {CL_SVM_INFO_PROPERTIES_KHR}
+  | {cl_svm_alloc_properties_khr_TYPE}
```
bashbaug:
```suggestion
  | {cl_svm_alloc_properties_khr_TYPE}[]
```

extensions/cl_khr_unified_svm.asciidoc
joselopeqti: How about:
```suggestion
| {CL_SVM_CAPABILITY_HOST_MAP_KHR}
  | This type of SVM can be accessed on the host by mapping and unmapping the allocation.
```
I think it makes clearer that you need to map to access on the host if `CL_SVM_CAPABILITY_HOST_READ_KHR` or `CL_SVM_CAPABILITY_HOST_WRITE_KHR` are not present without being as restrictive as "requires". Otherwise, I'm fine with your suggestion.

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+    ** if the value specified for a supported property name is not valid
+    ** if the same property name is specified more than once
+  * {CL_INVALID_VALUE}
+    ** if _required_capabilities_ contains an invalid SVM capability
+    ** if _desired_capabilities_ contains an invalid SVM capability
+    ** if _suggested_svm_type_index_ is `NULL`
+  * {CL_INVALID_BUFFER_SIZE}
+// TODO: update depending on the updated queries for available SVM sizes:
+    ** if _size_ is greater than {CL_DEVICE_MAX_MEM_ALLOC_SIZE} for any OpenCL device in _context_ that supports the specified SVM type
+    ** if _size_ is greater than {CL_DEVICE_MAX_MEM_ALLOC_SIZE} for a device associated with the SVM allocation
```
joselopeqti: Should the `CL_INVALID_BUFFER_SIZE` error be removed?
 As per the resolved issue 28:

> This extension will not define an upper limit on the size of an SVM allocation

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1176 @@
...
+
+  * {CL_INVALID_CONTEXT}
+    ** if _context_ is not a valid OpenCL context
+  * {CL_INVALID_PROPERTY}
+    ** if a memory property name in _properties_ is not a supported property name
+    ** if the value specified for a supported property name is not valid
+    ** if the same property name is specified more than once
+  * {CL_INVALID_VALUE}
+    ** if _required_capabilities_ contains an invalid SVM capability
+    ** if _desired_capabilities_ contains an invalid SVM capability
```
joselopeqti: What would be consider an invalid SVM capability here?
As the last capability right now is `#define CL_SVM_CAPABILITY_INDIRECT_ACCESS_KHR           (1 << 14)`
Any value greater than that?
If yes, we might need to modify the current test.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Yes, I'll go through and tidy up this error condition (and a few others).

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Ah, I'm guessing you're referring to [this test](https://github.com/KhronosGroup/OpenCL-CTS/blob/b1100a52d0667a011b2fa13d5822046d4b6c212b/test_conformance/SVM/test_unified_svm_api_suggested_type_index.cpp#L73) specifically?

You're right that this test is inconsistent with this error description since it's specifically passing some bogus capabilities to test the `CL_UINT_MAX` suggested type index.

What do we want to do here?  Some possibilities:

1. Keep the test behavior and update the spec.  If an application requests an unknown or unsupported required capability, return `CL_UINT_MAX` to indicate that no SVM type index supports this capability.
2. Keep the spec behavior and update the test.  Note that this could make it difficult or impossible to test the `CL_UINT_MAX` behavior depending on the capabilities supported by the implementation.
3. Maybe we could have different behavior for "required" vs. "desired" capabilities, although this sounds confusing, and the more I think about it I'm not sure that it buys us anything.

I think the extension is slightly easier to use if we keep the test behavior and update the spec, since otherwise it's an application requirement to check that the required capabilities are supported by the device to avoid an error.  I don't have a very strong opinion, though.

joselopeqti: > Ah, I'm guessing you're referring to this test specifically?

Yes, I was referring to that test.

> What do we want to do here?

In my opinion, I lean towards 1, removing the error conditions for required and desired capabilities and just return `CL_UINT_MAX` as the suggested type index and `CL_SUCCESS`.

> I think the extension is slightly easier to use if we keep the test behavior and update the spec, since otherwise it's an application requirement to check that the required capabilities are supported by the device to avoid an error.

I agree.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Fixed by https://github.com/KhronosGroup/OpenCL-Docs/pull/1282/commits/97a43e19e3d8aaea306ef253d29d22b37ee1aa3b, resolving.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Fixed by https://github.com/KhronosGroup/OpenCL-Docs/pull/1282/commits/97a43e19e3d8aaea306ef253d29d22b37ee1aa3b.

Note that the this error condition was already removed for `clSVMAllocWithPropertiesKHR`, with a nice note describing the behavior when the allocation size exceeds `CL_DEVICE_MAX_MEM_ALLOC_SIZE`, but I missed this case for `clGetSVMSuggestedTypeIndexKHR` :sob:

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Fixed by https://github.com/KhronosGroup/OpenCL-Docs/pull/1282/commits/0d2663ef72f7fbdc6c8e70c38c6ee5a13bd2c594.

extensions/cl_khr_unified_svm.asciidoc
bashbaug: Coming back to this discussion with fresh eyes:

* For clEnqueueSVMMemcpy and clEnqueueSVMMemFill, what should the behavior be if a source or destination SVM pointer does not have the required capabilities?
    * Should this be undefined behavior?
    * Or, should this be an error?  If so, which error?  Note that we do not currently define error behavior for these cases.
* For clEnqueueSVMMigrateMem, I don't think it makes sense to require any capabilities, for several reasons:
    1. It's unclear whether the device or the host is performing the operation, therefore it's unclear whether the device or host capabilities should apply.
    2. Migrating memory does not affect the functional correctness of an application, only its performance characteristics, therefore if an implementation cannot migrate an allocation it can simply be ignored.

joselopeqti:
Let’s say we had a macro – `CL_SVM_TYPE_MACRO_DEVICE_NON_EXCLUSIVE_KHR = CL_SVM_CAPABILITY_DEVICE_OWNED | CL_SVM_CAPABILITY_DEVICE_READ | CL_SVM_CAPABILITY_DEVICE_WRTIE`. Could we fully support this combination in the base spec? Wouldn't the absence of `CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR` imply that properties would be required in `clEnqueueSVMMemcpyWithPropertiesKHR`?

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1335 @@
...
+  | This type of SVM is accessible on the device for writing.
+| {CL_SVM_CAPABILITY_DEVICE_ATOMIC_ACCESS_KHR}
+  | This type of SVM is accessible on the device using atomic built-in functions.
+| {CL_SVM_CAPABILITY_CONCURRENT_ACCESS_KHR}
+  | This type of SVM supports concurrent access from the host and a device, or from multiple devices.
+    Use {CL_DEVICE_SVM_CONCURRENT_ACCESS_ATOM_SIZE_KHR} to determine the supported granularity for concurrent access.
+| {CL_SVM_CAPABILITY_CONCURRENT_ATOMIC_ACCESS_KHR}
+  | This type of SVM supports concurrent atomic access from the host and a device, or from multiple devices.
+| {CL_SVM_CAPABILITY_INDIRECT_ACCESS_KHR}
+  | This type of SVM supports a single kernel enable to indicate that the kernel may allocate any allocation of this type, rather than passing a list of indirectly accessed allocations to the kernel.
```
joselopeqti: Should this be?

```suggestion
  | This type of SVM supports a single kernel enable to indicate that the kernel may access any allocation of this type, rather than passing a list of indirectly accessed allocations to the kernel.
```

bashbaug: Yep, good catch, fixed!

extensions/cl_khr_unified_svm.asciidoc
```diff
@@ -0,0 +1,1338 @@
...
+* A mechanism to indicate that a kernel may access SVM allocations indirectly, without passing a set of indirectly accessed SVM allocations to the kernel, improving usability and reducing driver overhead for kernels that access many SVM allocations.
+
+* A new query function to query properties of an SVM allocation.
+
+* A new function to suggest an SVM allocation type for a set of SVM capabilities.
+
+== New API Functions
+
+[source]
+----
```
joselopeqti, line 104: `clEnqueueSVMMemcpyWithPropertiesKHR` is missing in this block.

Issue #844, opened 2022-09-14
What happens on cl_sync_point_khr overflow?


jansol:
Didn't find a mention of this in the command buffer base extension so here goes: `cl_sync_point_khr` is defined as `cl_uint`, from this follows that at most `cl_uint_max-1` commands can be unambiguously recorded to a command buffer if the implementation leaves 0 as an invalid init value. On a basic x64 Ubuntu installation `cl_uint` appears to be defined as a 32-bit unsigned integer, which means the maximum value is absurdly high for this purpose.

Now for the sake of argument, let's assume that a pathological application DOES record an absurd number of commands to the point of overflowing the range of unique `cl_sync_point_khr` handles (maybe a platform defines `cl_uint` and by extension `cl_sync_point_khr` to something much smaller than 32 bits, I did not see any requirements for type sizes in the data types appendix of the main OpenCL 3.0 spec.)

What is the expected behavior in such a situation? My first thought was that attempting to record more commands should return `CL_OUT_OF_HOST_MEMORY`.

EwanC:
> maybe a platform defines cl_uint and by extension cl_sync_point_khr to something much smaller than 32 bits

I think `cl_uint` is always defined as 32-bits - https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_platform.h#L155 / https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_platform.h#L261, and we define `#define CL_UINT_MAX         0xffffffffU`

But I agree, introducing some sort of error for when this limit is exceeded makes sense, rather than specifying some behaviour that allows `cl_sync_point_khr` to be ambiguous.

bashbaug:
> But I agree, introducing some sort of error for when this limit is exceeded makes sense...

Serious question, do we really expect we may overflow sync points anytime in the near or distant future, given that a `cl_uint` is 32 bits?  Even if a command in a command buffer is only a handful of bytes - and they would probably be quite a bit more than this in practice - a command buffer that overflows sync points would be very, very large (many gigabytes).

I suppose we could define an error behavior, or choose a 64-bit type like `size_t` or `cl_ulong`, but even with a 32-bit `cl_uint` it seems like we we'll run out of other resources far before they overflow.

bashbaug:
> Then again I don't think anything is preventing implementations from reserving some of those bits to represent arbitrary internal data so long as each sync point remains unique?

[...]

PR #1286, opened 2024-11-21
cl_khr_command_buffer sync-point capacity

EwanC:
Document the behaviour when command-buffer command capacity is reached, and track under "Issues" the possible future use-cases for being able to optimize based on the capacity of a command-buffer.

Closes https://github.com/KhronosGroup/OpenCL-Docs/issues/844

### DIFF
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -109,7 +109,7 @@ command-queue. The command-buffer itself has no inherent in-order/out-of-order
 property, this ordering is inferred from the command-queue used on command
 recording. {clEnqueueCommandBufferKHR} submissions to an out-of-order queue
 have the same execution semantics are other operations enqueued to an
-out-of-order queue, such as {clEnqueueFillBuffer}. Where execution between
+out-of-order queue, such as {clEnqueueFillBuffer}, where execution between
 enqueued operations may happen concurrently unless dependencies between the
 operations are expressed with events.
 
--- a/api/cl_khr_command_buffer.asciidoc
+++ b/api/cl_khr_command_buffer.asciidoc
@@ -43,16 +43,6 @@ Command-buffers enable a reduction in overhead when enqueuing the same
 workload multiple times. By separating the command-queue setup from dispatch,
 the ability to replay a set of previously created commands is introduced.
 
-Device-side _cl_sync_point_khr_ synchronization-points can be used within
-command-buffers to define command dependencies. This allows the commands of a
-command-buffer to execute out-of-order on a single <<compatible, compatible>>
-command-queue. The command-buffer itself has no inherent in-order/out-of-order
-property, this ordering is inferred from the command-queue used on command
-recording. Out-of-order enqueues without event dependencies of both regular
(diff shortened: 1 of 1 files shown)


api/cl_khr_command_buffer.asciidoc
```diff
@@ -110,6 +100,28 @@ following reasons:
...
+
+Device-side {cl_sync_point_khr_TYPE} synchronization-points can be used within
+command-buffers to define command dependencies. This allows the commands of a
+command-buffer to execute out-of-order on a single <<compatible, compatible>>
+command-queue. The command-buffer itself has no inherent in-order/out-of-order
+property, this ordering is inferred from the command-queue used on command
+recording. Out-of-order enqueues without event dependencies of both regular
+commands, such as {clEnqueueFillBuffer}, and command-buffers are allowed to
+execute concurrently, and it is up to the user to express any dependencies using
+events.
```
bashbaug, line 113: I realize this text isn't new and just moved, but when I read this sentence as part of this review I initially thought it was saying that an event dependency is required in all cases to express dependencies.  After re-reading, I think it's only intending to apply to out-of-order queues though, correct?  Not sure if it'd help to have more words or fewer, or if we're OK with the text as-is, mostly noting for completeness.

bashbaug (approved):

api/cl_khr_command_buffer.asciidoc
EwanC: If it's not clear seems like a good opportunity to try improve the text. How about?

> {clEnqueueCommandBuffer} submissions to an out-of-order queue have the same execution semantics are other operations enqueued to an out-of-order queue, such as {clEnqueueFillBuffer}. Where execution between enqueued operations may happen concurrently unless dependencies between the operations are expressed with events.

EwanC: I've pushed that wording as a separate commit, so it can be reviewed standalone.

kpet (approved):
Thanks!

api/cl_khr_command_buffer.asciidoc
```diff
@@ -110,6 +100,29 @@ following reasons:
...
+Device-side {cl_sync_point_khr_TYPE} synchronization-points can be used within
+command-buffers to define command dependencies. This allows the commands of a
+command-buffer to execute out-of-order on a single <<compatible, compatible>>
+command-queue. The command-buffer itself has no inherent in-order/out-of-order
+property, this ordering is inferred from the command-queue used on command
+recording. {clEnqueueCommandBufferKHR} submissions to an out-of-order queue
+have the same execution semantics are other operations enqueued to an
+out-of-order queue, such as {clEnqueueFillBuffer}. Where execution between
+enqueued operations may happen concurrently unless dependencies between the
+operations are expressed with events.
```
bashbaug, line 114: I like the updated text, but I don't think the last sentence is a complete sentence.  How about:

```suggestion
recording. {clEnqueueCommandBufferKHR} submissions to an out-of-order queue
have the same execution semantics are other operations enqueued to an
out-of-order queue, such as {clEnqueueFillBuffer}, where execution between
enqueued operations may happen concurrently unless dependencies between the
operations are expressed with events.
```

PR #1290, opened 2024-12-12
Fix formatting issues in builtins tables

frasercrmck:
1. Add missing newline between double and half versions of 'ldexp'
2. Fix formatting of snippets in 'mix'


frasercrmck:
I also noted some empty table lines in several places - are those intentional? E.g., between `normalize` and `fast_distance`

svenvh (approved):

svenvh:
> I also noted some empty table lines in several places - are those intentional? E.g., between `normalize` and `fast_distance`

I suspect they are intentional.  The table "Built-in Scalar and Vector Relational Functions" has similar empty rows which seem to imply some sort of grouping (e.g. `any` and `all` form a group).

bashbaug (approved):
We may want to revisit the use of whitespace between types at some point, but the blank line between `double` and `half` for `ldexp` is consistent with the other functions.

Issue #1142, opened 2024-04-04
Command-buffer queue comptabaility


EwanC:
The current `clCreateCommandBufferKHR` API takes a list of queues on creation, and enforces that the properties of these queues must match the properties of the matching queues passed in `clEnqueueCommandBufferKHR`. However, this API design does not work well for applications that do not know at the time of command-buffer creation what the properties of the queue that the command-buffer will be executed on are.

In particular this is an issue for [SYCL-Graph](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_graph.asciidoc) layering, where when the command-buffer is created during `graph.finalize()`, the SYCL-RT does not yet know what SYCL queue a user will submit the graph to. The graph is treated as a black box by the SYCL queue that is executed in-order or out-of-order with respect to the other SYCL queue commands depending on the type of queue.

There are two possible approaches to help with this in cl_khr_command_buffer that I've thought about so far:

1. **Optional feature to relax queue compatibility constraint** - Add an optional feature to relax the constraint that the queues passed to `clEnqueueCommandBuffer` must be compatible with those set on creation (or maybe a query to check for "compatible" means for that vendor).  The SYCL-RT could already use the `clRemapCommandBufferKHR()` API with `CL_COMMAND_BUFFER_PLATFORM_REMAP_QUEUES_KHR ` for this to created a remap of the command-buffer before each enqueue during graph submission, but we don't really want to incur the deep copy cost here and the `cl_khr_command_buffer_multi_device` extension also doesn't have a large amount of vendor coverage. However, a drawback of this is that the SYCL-RT would still need to create a placeholder OpenCL queue when the command-buffer is created and commands added, which has a construction overhead.

2. **Remove the queue object from the command-buffer interface enqueue time**. We could replace the command-queue parameters with device parameters (and a context parameter on creation) on command-queue creation/command adding/remap and then only introduce the queues at `clEnqueueCommandBuffer` time. We'd probably still want more command-buffer properties so that a user can say up-front what type of queue they want the command-buffer to run on if they want a more optimized implementation. This is quite an invasive change.

I haven't really thought all this through yet but the design point has come up recently so I think it's worth creating an issue to track the discussion.

bashbaug:
Need to consider interactions with PR #850.

bashbaug:
[...]

PR #1292, opened 2024-12-16
Refactor command-buffer queue compatability

EwanC:
As proposed in https://github.com/KhronosGroup/OpenCL-Docs/issues/1142 the PR changes the semantics of the command-queues parameters used for command-buffer creation and enqueue.

The queues used on command-buffer creation now only inform the device and dependencies of commands, rather than restricting the properties set on the queues used for command-buffer enqueue.

Related PRs:
* OpenCL-Headers - https://github.com/KhronosGroup/OpenCL-Headers/pull/271
* OpenCL-CTS - https://github.com/KhronosGroup/OpenCL-CTS/pull/2230

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -14104,16 +14104,33 @@ of 0 or 1.
 The simultaneous use capability removes this restriction and allows
 command-buffers to have a <<pending_count, Pending Count>> greater than 1.
 
-[[compatible]]
 Command-buffers are created using an ordered list of command-queues that
-commands are recorded to and execute on by default.
-These command-queues can be replaced on command-buffer enqueue with
-different command-queues, provided for each element in the replacement list
-the substitute command-queue is compatible with the command-queue used on
-command-buffer creation.
-A _compatible_ command-queue is defined as a command-queue with
-identical properties targeting the same device and in the same OpenCL
-context.
+commands are recorded to and execute on by default. All these queue objects
+must share the same context, but may be associated with different devices when
+the {cl_khr_command_buffer_multi_device_EXT} extension is supported.
+
+When constructing a command-buffer by appending commands, the queue parameter
+passed for the command being created is used to set the device with which the
+command will be associated with, and also inform the scheduling of the command.
+If the queue is an in-order queue, then an additional dependency is created on the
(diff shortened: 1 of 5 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -13915,16 +13915,33 @@ of 0 or 1.
...
+command will be associated with, and also inform the scheduling of the command.
+If the queue is an in-order queue, then an additional dependency is created on the
+last command appended to the command-buffer using the same queue parameter. If
+the queue is an out-of-order queue, then no extra dependencies on previous
+commands using the same queue are created. All queue properties other than
+{CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE} are ignored for the purposes of command
+creation, with the exception of any vendor extension defined queue properties
+that explcitly define semantics for this purpose.
+
+The command-queues used on command-buffer creation must be replaced on
```
aharon-abramson, line 13934: Why say "must be replaced"? Can't the user use the same queue objects for creation and execution?

EwanC: I'll try reword this. I was trying to communicate that since the queues used on command-buffer creation are really placeholders used to inform the device and dependencies of commands, those placeholders must be replaced with queues which will actually submit work to the device.

aharon-abramson: I understand, but the question is: **_must_** the queues used for recording and replaying be different? If yes, then your wording is OK; if not, you should replace "must be replaced" with "may be replaced".

EwanC: There is no requirement that the queues **must** be different, they only **may** be different as I say in the next sentence, so I'll reword this sentence.

EwanC: Changed this line to say "When enqueuing a command-buffer, a list of command-queues to execute the command-buffer on can be passed by the user, otherwise the command-queues set on command-buffer creation are used by default for execution. A user passed list may contain different command-queues, ..."

aharon-abramson (approved):

kpet (approved):
LGTM

bashbaug:
Merging as discussed in the January 28th teleconference.

PR #1295, opened 2025-01-06
Integrate cl_ext_image_requirements_info into unified specification

kpet:
Change-Id: Ia249f78aa521a8d202dfafbb736c9887574e56f9


bashbaug (commented):
I didn't do a detailed text review, but I did a quick skim and everything looks like it's in the right place.  Other than a few minor editorial issues, LGTM.

kpet:
Thanks for the review and the suggestions, I've applied them both.

bashbaug (approved):

PR #1299, opened 2025-01-15
Integrate cl_ext_image_from_buffer into unified specification

kpet:
Also add version notes for all CL_IMAGE_REQUIREMENTS_* enums.

Change-Id: I7b7b093034121a9215786beff7318b18e7d0c24a


api/opencl_runtime_layer.asciidoc
```diff
@@ -2071,9 +2079,23 @@ returned in _errcode_ret_:
...
+endif::cl_ext_image_from_buffer[]
+ifdef::cl_ext_image_from_buffer[]
+  * {CL_INVALID_IMAGE_FORMAT_DESCRIPTOR} if an image is created from a
+    buffer and the row pitch, or slice pitch, if the
+    {cl_ext_image_from_buffer_EXT} extension is supported, or base address
+    alignment do not follow the rules described for creating an image from a
+    buffer.
+  * {CL_INVALID_IMAGE_SIZE} if the {cl_ext_image_from_buffer_EXT} extension is
+    supported and an image is created from a buffer and the buffer passed in
+    _image_desc->_mem_object_ is too small to be used as a data store for the
```
bashbaug, line 2095: Formatting nitpick: I think this should be:

```suggestion
    __image_desc__->__mem_object__ is too small to be used as a data store for the
```

api/footnotes.asciidoc
```diff
@@ -53,7 +53,7 @@ Note that reading and writing 2D image arrays from a kernel with `image_array_si
 ]
 
 :fn-image-from-buffer: pass:n[ \
-To create a 2D image from a buffer object that share the data store between the image and buffer object. \
+To create an image from a buffer object that share the data store between the image and buffer object. \
```
bashbaug, line 56: Optional / editorial: this wasn't changed in this PR, but I don't this sentence is grammatically correct.  How about:

```suggestion
To create an image from a buffer object that shares the data store between the image and buffer object. \
```

api/cl_ext_image_from_buffer.asciidoc
```diff
@@ -1,19 +1,78 @@
...
+    - Create an image from the buffer
+    - Read the image from a kernel and compare with values read using the buffer and direct addressing. They must match.
+
+. TODO Test access via read/write/map commands?
+
+. TODO Test copy to/from buffer?
+
+. TODO Test fill?
+
+. TODO Test copy to/from another image?
```
bashbaug, line 46: Recording a comment from the January 21st teleconference: do we want to do anything with these TODOs, or are they ok as-is?

bashbaug (commented):
LGTM, just a few comments.

I'd be OK merging this as-is, but we really ought to fix the formatting issue first.

api/cl_ext_image_from_buffer.asciidoc
kpet, line 46:
```suggestion
//. TODO Test access via read/write/map commands?

//. TODO Test copy to/from buffer?

//. TODO Test fill?

//. TODO Test copy to/from another image?
```

kpet: As discussed on the call, I'd rather make them comments in the asciidoctor source instead. Now done.

kpet:
Thanks for the review and suggestions. I've applied them. PTAL.

bashbaug (approved):
LGTM.

We should try to remember to update the link to this extension on the OpenCL extension registry when we publish a new version of the unified spec with this extension integrated.

PR #1301, opened 2025-01-23
update the spec source copyright dates to 2025

bashbaug:
New year, new copyright headers.  This was done with the same script as before.


alycm (approved):
LGTM.

PR #1319, opened 2025-02-18
Improve the description of error codes for clSetKernelArg

### LINKED ISSUE #1317: Is calling `clSetKernelArg` with an `arg_size` of 0 valid for arguments in the `local` address space?
Some implementations accept calls to `clSetKernelArg` with an `arg_size` of 0 for arguments in the `local` address space, others return `CL_INVALID_ARG_SIZE`.

Should this be allowed? I don't see why it should not. We should also use this discussion as an opportunity to check that we have language that correctly describes what memory accesses are valid for kernel arguments in the `local` address space.

Lastly, regardless of the decision, we need a new CTS test.

kpet:
This would likely have prevented me from creating #1317 after reading the spec a bit too quickly.

Fixes #1317

Change-Id: I1ed3737335edc07747e872821210e5785ae6635a


bashbaug (approved):
LGTM.  I have some ideas about applying this change more broadly, but I think it'd be better to have that conversation in an issue than in this PR.

PR #1321, opened 2025-02-20
Remove assume linear

joshqti:
The default behavior of an image import operation should be well defined behavior.  If the implementation is capable of importing non-linear images from a known external API, this is best as query, because this functionality is currently implementation defined behavior.


bashbaug (changes requested):
I don't think we want to do this.

1. It changes a final (non-provisional) extension in a non-backwards compatible way.
2. It essentially inverts the current query, so it doesn't add any new functionality currently.
3. It is not as extensible in the future.  With the current "assume linear" query, we've given ourselves freedom to describe any of the handle types that are not assumed to be linear with some other query.

I know it's a long thread, but I would strongly suggest reading through https://github.com/KhronosGroup/OpenCL-Docs/issues/861 to understand why we ended up with the current "assume linear" query.

If you still think there is a problem, could you please start by describing what it is?  Thanks!

joshqti:
This proposal will not alter the behavior of the extension, it will simply make it more precise as to when OpenCL can import an image correctly.  The current default behavior is undefined.  When creating an image and importing a handle, it is implementation defined behavior, and not robust. I believe this will lead to non-portable apps and an explosion of vendor extensions.  Our test_vulkan relies on this behavior, as well as matches Vulkan and OpenCL's UUID, which is also undefined. 

Here is the problem I am facing.  We did not think assuming opaque_fds from Vulkan was robust, so we marked it as assume_linear, and introduced a vendor extension along the lines of VULKAN_OPAQUE_FD, which will assume the fd is from our Vulkan driver, and we can import non-linear images safely.

I am facing a similar situation with NT handles, which would need another vendor extension.  Instead we are considering a third vendor extension, such that the existence will enable the default behavior, of making assumptions about the source of the handle.  We could not have our handle in assume_linear in this case.  It seems assume_linear is not very useful, it simply enables the ability to define vendor specific handle types, such that the app knows the Khronos handle is not useful.

Due to the convenience of simply importing a handle from Vulkan or DirectX, and have the implementation assume its from a known driver, I think most apps will be written that way.  If an implementation does not want to leave this to undefined behvior, the app will have to be aware of vendor extensions to do the same job, leading to portability problems.

We like the ability to assume a handle comes from Vulkan, but would like the spec to be more robust.  Right now it is too vague, and the test is incorrect in assuming matching UUIDs is sufficient to be conformant.

joshqti:
I am withdrawing this issue

Issue #1325, opened 2025-02-26
printf format string


aharon-abramson:
I want to clarify something about `printf`'s `format` argument.
The spec says:

> The format is in the constant address
space and must be resolvable at compile time, i.e. cannot be dynamically created by the executing
program itself.

Does it mean that `format` may be a kernel argument rather than a string literal? For example:

`__kernel void f(__constant char* arg) {
  printf(arg);
}`

I don't see any such code in the CTS. I prefer this code to be disallowed. If you agree, I'll upload a change to the spec to clarify this.

bashbaug:
It's been a while, but I think the intent behind "resolvable at compile time" is that `format` cannot be a kernel argument, therefore the code snip in your example is disallowed.

aharon-abramson:
https://github.com/KhronosGroup/OpenCL-Docs/pull/1327

PR #1327, opened 2025-02-27
clarify that format must be a string literal in OpenCL_C.txt

aharon-abramson:
The intent behind "resolvable at compile time" is that `format` cannot be a kernel argument (see https://github.com/KhronosGroup/OpenCL-Docs/issues/1325#issuecomment-268542016)

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -9571,9 +9571,7 @@ The format is composed of zero or more directives: ordinary characters (not
 specifications, each of which results in fetching zero or more subsequent
 arguments, converting them, if applicable, according to the corresponding
 conversion specifier, and then writing the result to the output stream.
-The format is in the constant address space and must be resolvable at
-compile time, i.e. cannot be dynamically created by the executing program
-itself.
+The format must be a string literal.
 
 Each conversion specification is introduced by the character *%*.
 After the *%*, the following appear in sequence:


SunSerega:
I think the old text allowed a constant variable containing format, not just a string literal.

That way, we can define these formats at the top and reuse them for multiple `printf` calls without `#define`-s.

OpenCL_C.txt
```diff
@@ -9571,9 +9571,7 @@ The format is composed of zero or more directives: ordinary characters (not
 specifications, each of which results in fetching zero or more subsequent
 arguments, converting them, if applicable, according to the corresponding
 conversion specifier, and then writing the result to the output stream.
-The format is in the constant address space and must be resolvable at
-compile time, i.e. cannot be dynamically created by the executing program
-itself.
+The format must be a string literal.
```
bashbaug, line 9574: I agree with @SunSerega that this is a bit too strong, at least based on the current Clang behavior, which does not require the format string to be a string literal: https://godbolt.org/z/b33jKTe7K

Is there a C or C++ term we could use instead?

aharon-abramson: I see that Clang allows this, but it isn't happy either:
> warning: format string is not a string literal (potentially insecure)

Doesn't this warning hint that format should be restricted to string literals?

bashbaug: The warning is a security-related warning, because in the absence of the constant address space (say, for CPU host code), a malicious actor could modify the content of the format string to do devious things.

I don't think it's relevant to the OpenCL C case, because the string is in the constant address space.  Nevertheless, the warning can be fixed with one more `const`: https://godbolt.org/z/3MYTWWs1P

aharon-abramson:
Added the last commit here by mistake.

Issue #1308, opened 2025-02-03
cl_khr_external_memory_dma_buf is underspecified


joshqti:
It is not clear from the spec if the app or the implementation is required to call DMA_BUF_IOCTL_SYNC/DMA_BUF_IOCTL_END  when using clEnqueueMap(Image/Buffer).  The CTS PR for the test does not make any sync calls, which seems to imply clEnqueueMap is expected to do it.

kpet:
Discussed in the 2025/02/04 memory TSG: the specification does not adequately describe when an application is allowed to perform operations on a dma_buf allocation without using OpenCL calls. There seems to be a consensus to forbid applications from performing such operations after a memory object created from a dma_buf allocation is acquired and prior to that memory object being released.

bashbaug:
Our current spec text says, for [clEnqueueAcquireExternalMemObjectsKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueAcquireExternalMemObjectsKHR):

> Applications must acquire the memory objects that are created using external handles before they can be used by any OpenCL commands queued to a command-queue. Behavior is undefined if a memory object created from an external memory handle is used by an OpenCL command queued to a command-queue without being acquired. This is to guarantee that the state of the memory objects is up-to-date and they are accessible to OpenCL.

And likewise, for [clEnqueueReleaseExternalMemObjectsKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueReleaseExternalMemObjectsKHR):

> Applications must release the memory objects that are acquired using [clEnqueueReleaseExternalMemObjectsKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueReleaseExternalMemObjectsKHR) before using them through any commands in the other API. This is to guarantee that the state of memory objects is up-to-date and they are accessible to the other API.

The part for clEnqueueAcquireExternalMemObjectsKHR seems to say roughly what I'd expect it to say - the external memory object must be acquired before it may be mapped via clEnqueueMapBuffer.  Because there is no documented requirement to do anything more with the dma_buf, I think your assessment is correct, and if additional IOCTLs are required then they should be done by the implementation.  Is this a problem?

[...]

PR #1332, opened 2025-03-04
clarify clEnqueueReleaseExternalMemObjectsKHR

bashbaug:
partial fix for #1308 

This changes the text in the description of clEnqueueReleaseExternalMemObjectsKHR from "in the other API" to "outside of OpenCL".

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -5875,10 +5875,10 @@ include::{generated}/api/protos/clEnqueueReleaseExternalMemObjectsKHR.txt[]
     command to complete.
 
 Applications must release the memory objects that are acquired using
-{clEnqueueReleaseExternalMemObjectsKHR} before using them through any
-commands in the other API.
-This is to guarantee that the state of memory objects is up-to-date and they
-are accessible to the other API.
+{clEnqueueReleaseExternalMemObjectsKHR} before using them through any commands
+outside of OpenCL.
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -5875,10 +5875,10 @@ include::{generated}/api/protos/clEnqueueReleaseExternalMemObjectsKHR.txt[]
     command to complete.
 
 Applications must release the memory objects that are acquired using
-{clEnqueueReleaseExternalMemObjectsKHR} before using them through any
-commands in the other API.
-This is to guarantee that the state of memory objects is up-to-date and they
-are accessible to the other API.
+{clEnqueueReleaseExternalMemObjectsKHR} before using them through any commands
+outside of OpenCL.
```
bashbaug, line 5879: Possible additional improvements:

1. Disambiguate "them", because the memory objects do not have any meaning outside of OpenCL.  Should this be referring to the external memory handles used to create the memory objects instead?  Should it be the memory that those handles represent?
2. Disambiguate "any commands".  Should this be "functions" instead?  What is the right term?

bashbaug:
I've tried to improve this by removing the previous ambiguity.  Let me know what you think - thanks!

joshqti (commented):
This looks reasonable.  We should merge

joshqti (approved):
Merging as discussed in conference on April 1 2025

PR #1334, opened 2025-03-06
cl_intel_subgroup_2d_block_io

bashbaug:
This PR adds the spec source for the cl_intel_subgroup_2d_block_io extension.

This is an OpenCL C-only extension that does not add any new APIs, so no XML changes are required (I think?).

### DIFF
--- a/extensions/cl_intel_subgroup_2d_block_io.asciidoc
+++ b/extensions/cl_intel_subgroup_2d_block_io.asciidoc
@@ -1,0 +1,1211 @@
+:data-uri:
+:sectanchors:
+:icons: font
+:source-highlighter: coderay
+
+= cl_intel_subgroup_2d_block_io
+
+== Name Strings
+
+`cl_intel_subgroup_2d_block_io`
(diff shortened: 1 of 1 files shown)


kpet:
>This is an OpenCL C-only extension that does not add any new APIs, so no XML changes are required (I think?).

Don't you want an entry similar to the following?
```
 <extension name="cl_khr_subgroup_clustered_reduce" revision="1.0.0" supported="opencl" ratified="opencl" comment="OpenCL C only"/>
```

bashbaug:
> Don't you want an entry similar to the following? [...]

Interesting.  It seems that it should, but I do not see any OpenCL C-only vendor extensions listed.  Would you be OK if I added all of the Intel extensions at least in a separate PR?

kpet:
Yes, that works for me. It'll be useful to have these if/when we do get to integrating all the vendor extensions into the unified specification.

bashbaug:
Merging - vendor extension.

PR #1335, opened 2025-03-11
cl_khr_external_memory_dma_buf: Clarify usage

joshqti:


api/opencl_runtime_layer.asciidoc
```diff
@@ -6135,7 +6135,12 @@ handles, and adds as a property that may be specified when creating a buffer
 or an image memory object from an external handle:
 
   * {CL_EXTERNAL_MEMORY_HANDLE_DMA_BUF_KHR_anchor} is a file descriptor for a Linux
-    dma_buf.
+    dma_buf. Once a memory object created from a dma_buf has been acquired with 
+    clEnqueueAcquireExternalMemObjectsKHR, all access, including host access should 
+    be done through OpenCL, until a call to clEnqueueReleaseExternalMemObjectsKHR has completed.
+    Any operations external to OpenCL on the dma_buf while it is acquired 
+    may result in undefined behavior.
```
bashbaug, line 6142:
```suggestion
    dma_buf. Once a memory object created from a dma_buf has been acquired with 
    {clEnqueueAcquireExternalMemObjectsKHR}, all access, including host access should 
    be done through OpenCL, until a call to {clEnqueueReleaseExternalMemObjectsKHR} has completed.
    Any operations external to OpenCL on the dma_buf while it is acquired 
    may result in undefined behavior.
```

We may want to say "operations outside of OpenCL" instead of "operations external to OpenCL", both to avoid possibly confusing the "external" term with "external memory", and to align closer to the proposed terminology in #1332.

bashbaug (approved):

bashbaug:

PR #1337, opened 2025-03-20
Add support for experimental extensions to spec tooling

kpet:
Change-Id: Id2bc90ca7d9a3c1523080f30f5d6b60cd01d54f2


bashbaug (approved):
This looks reasonable to me, but it would be good to get input from @oddhack for the script changes, and from @neiltrevett for the experimental extensions boilerplate text.

PR #1339, opened 2025-03-22
rephrase table title to avoid incorrect rendering

### LINKED ISSUE #1175: Table 58. Summary of command-buffer creation configurations title is rendering incorrectly

bashbaug:
Fixes #1175 

The table title with the extension name link rendered incorrectly in some browsers, so rephrase it so it renders correctly everywhere. Also, adjust table column widths.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -14716,8 +14716,8 @@ endif::cl_khr_command_buffer_mutable_dispatch[]
     If _errcode_ret_ is `NULL`, no error code is returned.
 
 ifdef::cl_khr_command_buffer_multi_device[]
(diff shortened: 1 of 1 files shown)


bashbaug:
Editorial change, merging immediately.

PR #1340, opened 2025-03-22
explicitly list the data types supported for explicit conversions

### LINKED ISSUE #1190: Should explicit conversion with half exist?
Documentation says that no explicit conversion is provided for half:
- https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#explicit-conversions
- https://github.com/KhronosGroup/OpenCL-Docs/blob/main/man/static/convert_T.txt

But the header in clang is defining it in both direction (input and output) for every type:
- https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/opencl-c.h

I see half conversion being used in tensorflow lite:
- https://github.com/tensorflow/tensorflow/blob/6dda17ce00d3042e1e2a6c2be252e06bbc3b3846/tensorflow/lite/delegates/gpu/cl/cl_operation.cc#L86

### LINKED ISSUE #1239: confusing phrasing for fp16
This description is confusing and should be cleaned up:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#cl_khr_fp16

> The cl_khr_fp16 extension was promoted to OpenCL C 1.2 as an optional feature, and to OpenCL 3.0 as the optional cl_khr_fp16 feature. The extension provides 16-bit precision scalar and vector floating-point data types and extends many functions to accept these types.

fp16 is still currently an extension and there is no cl_khr_fp16 feature or feature test macro.

The fp64 description below it is a little confusing too and could possibly be tidied up also, depending what we decide to do for fp16.

bashbaug:
Fixes #1190
Fixes #1239

Instead of listing the data types that are NOT supported for explicit conversions, list the data types that ARE supported.  This includes the `half` data type, when the `cl_khr_fp16` extension is supported and enabled.

Also, tidies up the text for the half-precision and double-precision data types.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -376,10 +376,8 @@ ifdef::cl_khr_fp16[]
 [[cl_khr_fp16,cl_khr_fp16]]
 ==== Half-Precision Floating-Point
 
-The {cl_khr_fp16_EXT} extension was promoted to OpenCL C 1.2 as an optional
-feature, and to OpenCL 3.0 as the optional {cl_khr_fp16_EXT} feature.
-The extension provides 16-bit precision scalar and vector floating-point
-data types and extends many functions to accept these types.
+The {cl_khr_fp16_EXT} extension provides 16-bit half-precision scalar and vector
+floating-point data types and extends many functions to accept these types.
 endif::cl_khr_fp16[]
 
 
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM, I'll merge immediately.

PR #1341, opened 2025-03-22
allow abnormal termination for in-flight commands when context is deleted

### LINKED ISSUE #1238: exiting program without waiting for the completion of all commands
The spec is unclear about what happens when function `main` returns while commands are still in flight, even if their queue has been released.
The problem is that the finalization flow in C/C++ after `main` has returned is not well-defined and is platform-dependent. That might lead to things like crashes, deadlocks, or resource leaks.
I propose adding a note in the spec that applications must ensure all enqueued commands have been completed before exiting `main`, otherwise the behavior is undefined.

bashbaug:
Fixes #1238

When a context is deleted, any in-flight commands (commands that have not completed) may be abnormally terminated.


kpet (approved):
LGTM. Merging as discussed in 2025-04-01 teleconference.

PR #1343, opened 2025-03-22
clarify CL_INVALID_IMAGE_SIZE error conditions

### LINKED ISSUE #1277: Discrepancy in error codes between legacy and new image creation API
Under error codes for[ clCreateImage/clCreateImageWithProperties](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateImage), the API spec says:

 - CL_INVALID_IMAGE_DESCRIPTOR if values specified in image_desc are not valid or if image_desc is NULL.

- CL_INVALID_IMAGE_SIZE if image dimensions specified in image_desc exceed the maximum image dimensions described in the Device Queries table for all devices in context.

However for the legacy [clCreateImage3D](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateImage3D), spec says:
CL_INVALID_IMAGE_SIZE if image_width or image_height are 0 or if image_depth ≤ 1, or if they exceed the maximum values specified in CL_DEVICE_IMAGE3D_MAX_WIDTH, CL_DEVICE_IMAGE3D_MAX_HEIGHT or CL_DEVICE_IMAGE3D_MAX_DEPTH respectively for all devices in context, or if values specified by image_row_pitch and image_slice_pitch do not follow rules described in the argument description above.
This means, for e.g., that when trying to create a 3D image with depth 0 (and hence invalid), the new APIs are required to return CL_INVALID_IMAGE_DESCRIPTOR  whereas the legacy API would return CL_INVALID_IMAGE_SIZE. It would be ideal to reconcile the error descriptions so we use/expand CL_INVALID_IMAGE_SIZE for this case as well rather than restrict it to exceeding the maximum dimension supported.

bashbaug:
fixes #1277

For clCreateImageWithProperties, allow implementations to return CL_INVALID_IMAGE_SIZE when image dimensions are invalid, not just when image dimensions exceed the maximum image dimensions supported by al devices in the context.  This is consistent with the behavior of clCreateImage2D, where creating a 2D image with width or height equal to zero returns CL_INVALID_IMAGE_SIZE, and with clCreateImage3D, where creating a 3D image with width, height, or depth equal to zero returns CL_INVALID_IMAGE_SIZE.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -2786,28 +2786,35 @@ include::{generated}/api/structs/cl_image_desc.txt[]
     {CL_MEM_OBJECT_IMAGE1D_ARRAY}, {CL_MEM_OBJECT_IMAGE2D},
     {CL_MEM_OBJECT_IMAGE2D_ARRAY}, or {CL_MEM_OBJECT_IMAGE3D}.
   * _image_width_ is the width of the image in pixels.
-    For a 2D image and image array, the image width must be a value {geq} 1 and
-    {leq} {CL_DEVICE_IMAGE2D_MAX_WIDTH}.
-    For a 3D image, the image width must be a value {geq} 1 and {leq}
-    {CL_DEVICE_IMAGE3D_MAX_WIDTH}.
-    For a 1D image buffer, the image width must be a value {geq} 1 and {leq}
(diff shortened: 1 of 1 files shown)


lakshmih (approved):
Thanks for writing this up!

Issue #1323, opened 2025-02-25
Command buffers and thread safety


bashbaug:
Creating a sub-issue based on https://github.com/KhronosGroup/OpenCL-Docs/issues/1309:

@joshqti asked:

> clUpdateMutableCommandsKHR seems to be defined in a similar manner as clSetKernelArg. Is it not thread safe like clSetKernelArg? Clarification on how updates apply for simultaneous use required. If its not, we need to update Appendix A.

We had a long discussion in the February 18th teleconference.  Some takeaways:

* We may want to use a different term than "thread safe".  Vulkan uses the term "externally synchronized", with a precise definition, for example, see: https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#fundamentals-threadingbehavior
* Whatever we document, it should be in the main spec (normative), not in an appendix (informative).
* Any API that "sets" or "modifies" or "mutates" an OpenCL object is something that we should consider to be non-"thread safe" or "externally synchronized".  There are some exceptions to this rule though, for example there are some APIs that modify program objects that are still thread safe, such as `clBuildProgram`.
* Most command buffer APIs would remain "thread safe", and only the APIs that modify command buffers would need to be non-"thread safe".  Aside from  `clUpdateMutableCommandsKHR`, one of the other command buffer APIs we discussed was `clFinalizeCommandBufferKHR`.
* Can we put information about thread safety into the XML file and auto-generate the right text into the spec?

We will continue discussion on this issue.

bashbaug:
Discussed in the March 18th teleconference:

* We will split the changes into two.
* The short-term change will simply add command buffer APIs to the existing appendix.
* The longer-term fix will move the description into the main spec and potentially adopt a more precise term like "externally synchronized" and record this information in the XML file.
* The APIs we will add to the appendix for the short-term fix are `clFinalizeCommandBufferKHR` and `clUpdateMutableCommandsKHR`.

EwanC:
Now that the short term fix in https://github.com/KhronosGroup/OpenCL-Docs/pull/1345 is merged, should we create a separate issue for the long term fix and close this one?

Reducing the amount of open "cl_khr_command_buffer" labelled issues will help focus our attention on outstanding command-buffer issues for discussion, and the long term fix could be forgotten as a single bullet point inside this larger issue thread.

EwanC:
[...]

PR #1345, opened 2025-03-25
document command buffer APIs that are not thread-safe

bashbaug:
This PR has the first short-term fix for #1323.

It documents that the following command buffer APIs are not thread-safe: clFinalizeCommandBufferKHR and clUpdateMutableCommandsKHR.

Additionally:
* I took out the reentrant description.  Since the behavior of OpenCL APIs called from an interrupt or signal handler is implementation-defined, I suspect any reentrancy behavior is implementation-defined also.
* I changed the "OpenCL callback functions" link to a more appropriate location.
* Minor grammar and punctuation updates.

### DIFF
--- a/api/appendix_a.asciidoc
+++ b/api/appendix_a.asciidoc
@@ -30,22 +30,44 @@ how to correctly order operations that change the state of a shared object.
 
 == Multiple Host Threads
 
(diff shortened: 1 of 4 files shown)


EwanC (approved):

bashbaug:

PR #1347, opened 2025-03-30
add a section describing document conventions

### LINKED ISSUE #1300: add a section to better document string representations
Currently, there is minimal information in the OpenCL specification about strings that are passed into and out of OpenCL, and it is in a footnote:

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_footnotedef_3

> A null terminated string is returned by OpenCL query function calls if the return type of the information being queried is a char[].

We should better document the behavior for strings passed into and out of OpenCL, such as the expected string encoding, whether strings are null-terminated, etc.  As one possibility, we could add "string representations" section similar to the one in Vulkan specification:

https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#fundamentals-strings

### LINKED ISSUE #1320: Improve documentation when multiple error conditions produce the same error code
_Originally posted by @bashbaug in https://github.com/KhronosGroup/OpenCL-Docs/pull/1319#pullrequestreview-2625434899_

Should our best practice in general be to describe separate error conditions like these using sub-bullets?  I agree this makes it easier to separate the different error conditions.

We could also replicate the name of the error.  I don't think we do this anywhere currently today, but this would be a bit more concise (one fewer line) and a bit friendlier when copy-pasting.  Example:

```asciidoc
* `CL_INVALID_ARG_SIZE` if if _arg_size_ does not match the size of the data type for an argument that is not a memory object.
* `CL_INVALID_ARG_SIZE` if the argument is a memory object and _arg_size_ != `sizeof(cl_mem)`.
```

I don't mind making this change "globally" if we think this is the right thing to do.

bashbaug:
fixes #1300 
fixes #1320 

Adds a section describing document conventions.  This section currently includes:

* Normative terminology - how the OpenCL specification uses terms such as "must", "may", and "should".
* Error codes - conditions when an OpenCL implementation returns an error code.
* String representations - how strings are passed to or returned form OpenCL API functions.

Large parts of this section were adapted from the similar section in the Vulkan specification.

### DIFF
--- a/api/introduction.asciidoc
+++ b/api/introduction.asciidoc
@@ -10,7 +10,7 @@ envelope, Central Processing Units (CPUs) now improve performance by adding
 multiple cores.
 Graphics Processing Units (GPUs) have also evolved from fixed function
 rendering devices into programmable parallel processors.
-As todays computer systems often include highly parallel CPUs, GPUs and
+As today's computer systems often include highly parallel CPUs, GPUs and
 other types of processors, it is important to enable software developers to
 take full advantage of these heterogeneous processing platforms.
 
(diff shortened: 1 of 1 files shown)


alycm (approved):
LGTM, no changes to suggest.

Issue #1164, opened 2024-05-10
Merge SPIR-V environment specification into API specification


kpet:
The specification tooling now has some support for writing unified specification that comprise both core features and extensions. A next logical step seems to be to merge the SPIR-V environment specification into the API specification, likely as a new chapter or normative appendix. As an example the Vulkan specification fully describes the SPIR-V environment rules in the same document as the API. This makes it much easier to cross-reference the two and make the documents generally easier to use for implementers and application writes.

PR #1348, opened 2025-03-31
Fix image conversion rules for signed and unsigned image channel data types

kpet:
The returned values are always 32-bit integers. Pixel values are sign- or zero-extended to 32 bits when the data type is smaller than 32-bit.

Also fix a typo in the description of single precision FP values to normalized integer values (the description erroneously referred to half precision floating point).

Change-Id: Idaa61a77949530e18442df680175de04347af9fa


kpet:
We should consider de-duplicating all those rules between the environment and OpenCL C specs. Adding a new channel data type means describing essentially the same rules in two places. Maybe this becomes a bit more natural to do once we've merged the environment specification into the unified specification? (see https://github.com/KhronosGroup/OpenCL-Docs/issues/1164)

bashbaug (approved):

bashbaug:

Issue #1269, opened 2024-10-02
remove unavoidable error conditions


bashbaug:
Philosophically, it should be possible for a well-written OpenCL program to execute without generating any OpenCL errors.  This means that no error conditions should be "unavoidable", where the only way to determine if an operation will be successful is to try making an API call and then check the returned error code.

This issue is to document places in the OpenCL spec where there are currently "unavoidable" error conditions and to document how to remove them.  This will likely need to be done in a minor OpenCL API revision, since it is a change to specified behavior.

1. `clGetPlatformIDs` returns `CL_PLATFORM_NOT_FOUND_KHR` if the `cl_khr_icd` extension is supported and zero platforms are available.
    * Why couldn't this just return `CL_SUCCESS` and `num_platforms` equal to zero, instead?
2. `clGetDeviceIDs` returns `CL_DEVICE_NOT_FOUND` if no devices in the platform match the requested device type.
    * This is true even for `CL_DEVICE_TYPE_ALL`.
    * Why couldn't this just return `CL_SUCCESS` and `num_devices` equal to zero, instead?

PR #1350, opened 2025-04-01
OpenCL ICD Loader Deinitialization Support

Kerilk:
This is the follow up of https://github.com/KhronosGroup/OpenCL-Docs/pull/1005 and the continuation of the implementation of https://github.com/KhronosGroup/OpenCL-Docs/issues/1003

This implements exposes the required functionalities to implement OpenCL ICD Loader de-initialization.

### DIFF
--- a/api/cl_khr_icd.asciidoc
+++ b/api/cl_khr_icd.asciidoc
@@ -274,10 +274,17 @@ Returned by {clGetPlatformIDs} when no platforms are found:
 
   * {CL_PLATFORM_NOT_FOUND_KHR}
 
-New in version 2.0.0, used as a value in the pointers to `clGetPlatformIDs`
-and `clUnloadCompiler` in the dispatch structure:
+New in version 2.0.0.
+
+Used as a value in the pointers to `clGetPlatformIDs` and `clUnloadCompiler`
+in the dispatch structure:
 
   * `CL_ICD2_TAG_KHR`
+
+Used as a parameter name to query if a platform can be unlaoded by the ICD
+Loader:
+
+  * {CL_PLATFORM_UNLOADABLE_KHR}
 
 === Issues
 
--- a/extensions/cl_loader_layers.asciidoc
+++ b/extensions/cl_loader_layers.asciidoc
@@ -21,6 +21,7 @@ interception layers (Layer ICDs) for OpenCL.
 |====
 | *Date*     | *Version* | *Description*
 | 2020-11-04 | 1.0.0     | First assigned version.
+| 2024-09-04 | 2.0.0     | Deinitialization support.
 |====
 
 ==== Contributors
@@ -45,6 +46,8 @@ cl_int clInitLayer(cl_uint                 num_entries,
                    const cl_icd_dispatch  *target_dispatch,
                    cl_uint                *num_entries_ret,
                    const cl_icd_dispatch **layer_dispatch_ret);
+
+cl_int clDeinitLayer(void);
 ----
 
 [[cl_loader_layers-new-api-types]]
@@ -72,12 +75,13 @@ Accepted as _param_name_ to the function *clGetLayerInfo*:
 === New API Tokens
 
 Returned by *clGetLayerInfo* when supplied *CL_LAYER_API_VERSION*
-and the corresponding layer implements version 1.0.0 of the layer
-API:
+and the corresponding layer implements version 1.0.0 or 2.0.0 of
+the layer API:
 
 [source,opencl]
 ----
 #define CL_LAYER_API_VERSION_100 100
+#define CL_LAYER_API_VERSION_200 200
 ----
 
 [[cl_loader_layers-new-environment-variables]]
@@ -172,6 +176,18 @@ Otherwise, it returns one of the following errors.
   * *CL_INVALID_VALUE* if _num_entries_ is insufficient for the layer or if
     _target_dispatch_ is a `NULL` value, or _num_entries_ret_ is a `NULL`
     value, or _layer_dispatch_ret_ is a `NULL` value.
+
+==== Layer Deinitialization
+
+[open,refpage='clDeinitLayer',desc='Deinitialize an OpenCL layer',type='protos']
+Deinitilization of a layer can be achieved with the function:
+[source,opencl]
+
+----
+cl_int clInitLayer(void);
+----
+
+*clDeinitLayer* returns *CL_SUCCESS* if the function is executed successfully.
 
 [[cl_loader_layers-source-code]]
 === Source Code
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -523,6 +523,7 @@ server's OpenCL/api-docs repository.
 
     <enums name="Constants.cl_loader_layers" vendor="Khronos" comment="Defined Layer API version, in cl_layer.h">
         <enum value="100"           name="CL_LAYER_API_VERSION_100"/>
+        <enum value="200"           name="CL_LAYER_API_VERSION_200"/>
     </enums>
 
     <enums start="0" end="-999" name="ErrorCodes.0" vendor="Khronos" comment="Error codes start at 0 and decrease">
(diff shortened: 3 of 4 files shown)


api/cl_khr_icd.asciidoc
```diff
@@ -274,11 +274,18 @@ Returned by {clGetPlatformIDs} when no platforms are found:
...
+
+Used as a value in the pointers to `clGetPlatformIDs` and `clUnloadCompiler`
+in the dispatch structure:
 
   * `CL_ICD2_TAG_KHR`
 
+Used as a parameter name to query if a platform can be unlaoded by the ICD
+Loader:
+
+  * {CL_PLATFORM_UNLOADABLE_KHR}
```
bashbaug: I'm trying to think of the best way to handle this.  I think we want the following to be true:

1. Shipping ICDs supporting v1.0.0 to continue to be correct and conformant.
2. ICDs to be able to report that they are unloadable without supporting v2.0.0.
3. Shipping ICDs supporting v2.0.0 (do these exist in the wild?) to continue to be correct and conformant.

Options to consider:

1. Retroactively add this query to v1.0.0.  Implementations that either don't support the query or return `CL_FALSE` for the query will not be unloaded.  Implementations that return `CL_TRUE` may be unloaded.  This is the easiest solution and matches what is currently implemented in the OpenCL ICD loader, but it could impact shipping ICDs, which will see an unknown query and generate an error (see also: https://github.com/KhronosGroup/OpenCL-Docs/issues/1269).
2. Add the query to a new v1.1.0.  Update the OpenCL ICD loader to check for this version and only perform the query if the version is v1.1.0 or newer.  This is a bit more work in the OpenCL ICD loader, and adds a dependency on OpenCL 3.0 or `cl_khr_extended_versioning` to get the ICD extension version, but it would avoid the error, for shipping v1.0.0 ICDs at least.  There still might be an error for shipping v2.0.0 ICDs though, which do not support the query even though they support a version newer than v1.1.0.
3. I can think of some other more creative options using new extension APIs, but that'd be an even bigger change.

What do we think?  Is option (1) sufficient, or do we need something fancier?

If we do decide to keep this text in some form, note the typo:


```suggestion
Used as a parameter name to query if a platform can be unloaded by the ICD
Loader:

  * {CL_PLATFORM_UNLOADABLE_KHR}
```

extensions/cl_loader_layers.asciidoc
```diff
@@ -173,6 +177,18 @@ Otherwise, it returns one of the following errors.
...
     value, or _layer_dispatch_ret_ is a `NULL` value.
 
+==== Layer Deinitialization
+
+[open,refpage='clDeinitLayer',desc='Deinitialize an OpenCL layer',type='protos']
+Deinitilization of a layer can be achieved with the function:
+[source,opencl]
+
+----
+cl_int clInitLayer(void);
```
EwanC: Think there's a copy paste typo here:
```suggestion
cl_int clDeinitLayer(void);
```

extensions/cl_loader_layers.asciidoc
```diff
@@ -173,6 +177,18 @@ Otherwise, it returns one of the following errors.
...
+----
+
+*clDeinitLayer* returns *CL_SUCCESS* if the function is executed successfully.
```
EwanC: If we say "if the function executed successfully" then do we also need to specify what is returned if the function executes unsuccessfully. For `clUnloadCompiler` we specified the return code as "will always return CL_SUCCESS", I'm not sure if that's an analogous entry-point to take wording from.

api/cl_khr_icd.asciidoc
Kerilk: I like 1. To avoid using the error path, we could define an extension string `cl_khr_icd_uloadable` that would signal the presence of the query. If this extension string is not present, the platform would not be deemed unloadable. If the string is present, we can query if the platform supports it.

What do you think?

Kerilk:
@bashbaug @EwanC Thanks for the reviews. I will update the layer doc with the solution we implemented in the ICD Loader PR, as the current text is out of date.
I will also rebase it.

extensions/cl_loader_layers.asciidoc
Kerilk: That is a good question.  I am fine with always returning success, but it could be a valuable information that the layer returned a non success error code. We could add a trace in the loader in this case.
So the two options are:
 * Rewording to "*clDeinitLayer* will always return *CL_SUCCESS*."
 * Rewording to "*clDeinitLayer* returns *CL_SUCCESS* if the function is executed successfully, an error code otherwise."

EwanC: Yeah that's true, If the layer was trying to do something interesting on deinit then the presence of a return code would be an indication to the user that things didn't go as expected. Even if the return code was something generic in itself, it would flag to the user to debug the layer further to see why it wasn't `CL_SUCCESS`. So fine with option 2

Kerilk:
```suggestion
*clDeinitLayer* returns *CL_SUCCESS* if the function is executed successfully, an error code otherwise.
```

api/cl_khr_icd.asciidoc
bashbaug: I like the idea of adding the `CL_PLATFORM_UNLOADABLE_KHR` to a separate `cl_khr_icd_unloadable` extension.  This avoids the "unavoidable error condition", doesn't require any of the extension versioning support in the unlikely event that a very old ICD wants to support unloading, and is supportable by both 1.0 and 2.0 ICDs.

I was initially worried that the ICD loader wouldn't be able to query for the `cl_khr_icd_unloadable` extension, but we should already have a platform by this point, so we should also have a dispatch table to call through.

If we really wanted to, we could even use the presence (or absence) of the extension as the only way to determine whether an ICD is unloadable (or not), and remove the `CL_PLATFORM_UNLOADABLE_KHR` query, but debatably that's a step too far.

bashbaug: Re-enumerating the options we discussed in the teleconference on December 2nd:

1. Retrofit the `CL_PLATFORM_UNLOADABLE_KHR` query into an existing extension.  Do not add a new extension `cl_khr_icd_unloadable`.
     * Pros: Matches the WIP implementation, minimal changes.
     * Cons: No way for the ICD loader to avoid generating an error for older ICDs, awkward versioning.
2. Add `CL_PLATFORM_UNLOADABLE_KHR` in a new extension `cl_khr_icd_unloadable`.
     * Pros: Clean solution, does not generate an error for older ICDs, does not require any retrofits or awkward versioning.
     * Cons: More involved solution, requires a new extension in the extension string, requires parsing an extension string in the ICD loader.
3. Add a new extension `cl_khr_icd_unloadable` to indicate that the ICD is unloadable.
     * Pros: Similar to (2), with slightly fewer changes required.
     * Cons: Similar to (2), plus a bit awkward if we ever try to make this a core feature and not an extension.
4. None of the above.

I lean slightly towards (2), but honestly any of the options could work.

xml/cl.xml
```diff
@@ -5919,7 +5925,7 @@ server's OpenCL/api-docs repository.
                 <command name="clIcdSetPlatformDispatchDataKHR"/>
             </require>
         </extension>
-        <extension name="cl_loader_layers" revision="1.0.0" supported="opencl">
+        <extension name="cl_loader_layers" revision="2.0.0" supported="opencl">
```
EwanC: I'm not sure we're bumping the version correctly here, in the spec changes  `cl_loader_layers` goes to `1.0.1` and `cl_khr_icd` it's that goes to `2.0.1`.

Could also add `require` tags for `clInitLayerWithProperties` and associated `CL_LAYER_PROPERTIES_LIST_END` & `cl_layer_properties`

api/cl_khr_icd.asciidoc
EwanC: I would probably fall off the fence on the side of (2) as well. It seems most aligned to our design principles in terms of not breaking older implementations and adding a queryable capabilities.

xml/cl.xml
Kerilk: Nice catch!

Kerilk: This should be addressed, let me know i missed something.

api/cl_khr_icd.asciidoc
bashbaug: We discussed the options one last time in the December 9th teleconference and agreed to go with option (2): Add `CL_PLATFORM_UNLOADABLE_KHR` in a new extension `cl_khr_icd_unloadable`.

Kerilk: I implemented the proposed solution here:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/250/commits/41a93d583d7969eb6a8f9f41f5f604b17c6ad985

For the specification, does the extension need to be in a new file or should be part of the cl_khr_icd extension file?

bashbaug: > For the specification, does the extension need to be in a new file or should be part of the cl_khr_icd extension file?

Let's go with a separate file for `cl_khr_icd_unloadable`.  This seems cleaner, since we don't seem to describe multiple extensions in any of the other files, for either OpenCL or Vulkan.

Kerilk: Sounds good.
Implemented.

PR #1351, opened 2025-04-01
add a github action to release spec PDFs when a tag is pushed

bashbaug:
Adds a GitHub action to release spec PDFs when a tag is pushed.  This replaces the old Travis script, that is no longer working.

I'll delete it when this PR is merged, but for now, an example what this looks like can be found on my fork, here:

https://github.com/bashbaug/OpenCL-Docs/releases

The release is created as a draft to allow manual editing before publishing.

There is one minor issue with this action that I wasn't quite able to fix.  For tag pushes, GitHub actions is checking out the tag, which does not have a branch associated with it.  This means that the spec toolchain `git symbolic-ref --short HEAD` fails, hence the git branch information is "Git branch not available" vs. the expected git branch "main".  This only affects what gets printed on the spec title pages and does not affect the spec content, so I am considering this a nice-to-have and not a must-have.


bashbaug:
I created an issue to track improvements to the spec tooling to improve building from a tag, where a git branch may not be available.  Merging.

PR #1360, opened 2025-04-27
editorial: fix linking for several extension APIs and enums

bashbaug:


bashbaug:
Editorial changes only, merging immediately.

PR #1364, opened 2025-05-01
fix wording inconsistency for device and host timer synchronization

### LINKED ISSUE #1363: Inconsistent CL_PLATFORM_HOST_TIMER_RESOLUTION description
In Chapter 4, 4.1. Querying Platform Info:

> `CL_PLATFORM_HOST_TIMER_RESOLUTION`
> ...
> This value must be 0 for devices that do not support device and host timer synchronization.

In Appendix H, Device and Host Timer Synchronization:
> **clGetPlatformInfo**, passing `CL_PLATFORM_HOST_TIMER_RESOLUTION`
> ...
> May return 0, indicating that *platform* does not support device and host timer synchronization.

Note the difference `devices` vs. `platform`.

bashbaug:
The query for CL_PLATFORM_HOST_TIMER_RESOLUTION is a platform query, not a device query, so the description should refer to platforms that do or do not support the query, not devices.

fixes #1363

### DIFF
--- a/api/opencl_platform_layer.asciidoc
+++ b/api/opencl_platform_layer.asciidoc
@@ -256,7 +256,7 @@ include::{generated}/api/version-notes/CL_PLATFORM_HOST_TIMER_RESOLUTION.asciido
 
         Support for device and host timer synchronization is required for
         platforms supporting OpenCL 2.1 or 2.2.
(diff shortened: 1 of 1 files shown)


bashbaug:
Merging as discussed in the June 17th teleconference.

PR #1367, opened 2025-05-03
publish cl_khr_external_memory_android_hardware_buffer

bashbaug:
Publish the cl_khr_external_memory_android_hardware_buffer extension (experimental).

### DIFF
--- a/api/cl_khr_external_memory_android_hardware_buffer.asciidoc
+++ b/api/cl_khr_external_memory_android_hardware_buffer.asciidoc
@@ -82,7 +82,7 @@ cl_mem_properties props[] = {
     (cl_mem_properties) ahb_handle,
     0};
 cl_mem imported_buffer = clCreateBufferWithProperties(context, props, CL_MEM_READ_WRITE,
-                            size, NULL, &error);
+                            0 /* size */, NULL, &error);
 If (error == CL_SUCCESS)
 {
     clEnqueueAcquireExternalMemObjectsKHR(queue, 1, &imported_buffer, ...);


api/opencl_runtime_layer.asciidoc
```diff
@@ -702,6 +702,15 @@ ifdef::cl_khr_external_memory[]
      {CL_MEM_DEVICE_HANDLE_LIST_KHR} is specified as part of _properties_.
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+  * {CL_INVALID_BUFFER_SIZE} 
+  ** if _size_ is non-zero and greater than the AHardwareBuffer when importing external memory 
+     using {cl_khr_external_memory_android_hardware_buffer_EXT}.
```
bashbaug, line 708: Minor: Should this refer to the handle type `CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR` rather than the name of the extension?

api/opencl_runtime_layer.asciidoc
```diff
@@ -2266,6 +2275,21 @@ ifdef::cl_khr_external_memory[]
      {CL_MEM_DEVICE_HANDLE_LIST_KHR} is specified as part of _properties_.
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
+
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+  * {CL_INVALID_IMAGE_FORMAT_DESCRIPTOR} 
+  ** if _image_format_ is not `NULL` when using {cl_khr_external_memory_android_hardware_buffer_EXT}.
+
+  * {CL_INVALID_IMAGE_DESCRIPTOR}
+  ** if _image_desc_ is not `NULL` when using {cl_khr_external_memory_android_hardware_buffer_EXT}.
```
bashbaug, line 2284: Similarly, I think these error conditions should refer to the handle type `CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR` rather than the name of the extension.

api/opencl_runtime_layer.asciidoc
```diff
@@ -2266,6 +2275,21 @@ ifdef::cl_khr_external_memory[]
...
+
+  * {CL_IMAGE_FORMAT_NOT_SUPPORTED}
+  ** if the AHardwareBuffer format is not supported.
+
+  * {CL_INVALID_HOST_PTR} 
+  ** if _host_ptr_ is not `NULL`.
```
bashbaug, line 2290: For these error conditions, I think we need to add a phrase like "if the image is created with a `CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR` handle type and..."

For the second one especially, the current text implies that the _host_ptr_ must be `NULL` in all cases.

api/opencl_runtime_layer.asciidoc
```diff
@@ -702,6 +702,15 @@ ifdef::cl_khr_external_memory[]
...
   ** if _properties_ includes more than one external memory handle.
 endif::cl_khr_external_memory[]
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+  * {CL_INVALID_BUFFER_SIZE} 
+  ** if _size_ is non-zero and greater than the AHardwareBuffer when importing external memory 
+     using {cl_khr_external_memory_android_hardware_buffer_EXT}.
+  * {CL_INVALID_OPERATION}
+  ** if the AHardwareBuffer format is not `AHARDWAREBUFFER_FORMAT_BLOB`.
+  * {CL_INVALID_HOST_PTR}
+  ** if _host_ptr_ is not `NULL`.
```
bashbaug, line 712: I think these error conditions need an additional phrase, also.

xml/cl.xml
```diff
@@ -1899,7 +1899,9 @@ server's OpenCL/api-docs repository.
         <enum value="0x2067"        name="CL_EXTERNAL_MEMORY_HANDLE_DMA_BUF_KHR"/>
         <enum value="0x2068"        name="CL_SEMAPHORE_HANDLE_OPAQUE_WIN32_NAME_KHR"/>
         <enum value="0x2069"        name="CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_WIN32_NAME_KHR"/>
-            <unused start="0x206A" end="0x2FFF" comment="Reserved to Khronos for interop"/>
+            <unused start="0x206A" end="0x206F" comment="Reserved to Khronos for interop"/>
+        <enum value="0x2070"        name="CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR"/>
```
bashbaug, line 1903: Minor: Could this enum use value `0x206A`, or is it too late to change this?

bashbaug (commented):
Here are a few comments and questions I found while editing this specification.

api/opencl_runtime_layer.asciidoc
joshqti: We agree, we should make this change

api/opencl_runtime_layer.asciidoc
joshqti: agree

api/opencl_runtime_layer.asciidoc
joshqti: agree

xml/cl.xml
bashbaug: Discussed in the May 6th teleconference.  We will keep the existing token assignment.

api/opencl_runtime_layer.asciidoc
bashbaug: Fixed.

api/opencl_runtime_layer.asciidoc
bashbaug: Fixed.

api/opencl_runtime_layer.asciidoc
bashbaug: Fixed.

api/opencl_runtime_layer.asciidoc
bashbaug: Fixed.

bashbaug:
I ended up completely refactoring the error conditions for `clCreateBufferWithProperties` and `clCreateImageWithProperties` as part of this PR, following the guideance from https://github.com/KhronosGroup/OpenCL-Docs/issues/1320.  This makes for bigger diffs, but I think the end result is much more readable.  Please take a look.

api/opencl_runtime_layer.asciidoc
```diff
@@ -2184,101 +2203,120 @@ is created successfully.
...
-  * {CL_INVALID_VALUE} if {CL_MEM_IMMUTABLE_EXT} is set in _flags_ and none
-    of the following conditions are met
-  ** {CL_MEM_COPY_HOST_PTR} or {CL_MEM_USE_HOST_PTR} is set in _flags_.
+    ** if {CL_MEM_IMMUTABLE_EXT} is set in _flags_ and {CL_MEM_READ_WRITE},
+    {CL_MEM_WRITE_ONLY}, or {CL_MEM_HOST_WRITE_ONLY} is set in _flags_
+    ** if {CL_MEM_IMMUTABLE_EXT} is not set in _flags_ and _mem_object_ was
+    created with {CL_MEM_IMMUTABLE_EXT}
+    ** if {CL_MEM_IMMUTABLE_EXT} is set in _flags_ and none of the following conditions are met:
+    *** {CL_MEM_COPY_HOST_PTR} or {CL_MEM_USE_HOST_PTR} is set in _flags_
+    // TODO: What about creating an image from a buffer or another image?
```
bashbaug: Check this: Should it be possible to create an immutable image from an immutable buffer or image memory object?

api/opencl_runtime_layer.asciidoc
```diff
@@ -2184,101 +2203,120 @@ is created successfully.
...
+    ** if values specified in _image_format_ are not valid
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+    ** if _properties_ includes an AHardwareBuffer external memory handle and
+    _image_format_ is not `NULL`
+endif::cl_khr_external_memory_android_hardware_buffer[]
+ifndef::cl_ext_image_from_buffer[]
+    // TODO: It seems like this error condition should be invalid image descriptor instead.
+    ** if a 2D image is created from a buffer and the row pitch and base address
+    alignment does not follow the rules described for creating a 2D image from a
+    buffer
```
bashbaug, line 2272: Check this: This faithfully follows the error code for this condition in the current spec, but it seems like it would be more appropriate to return `CL_INVALID_IMAGE_DESCRIPTOR` rather than `CL_INVALID_IMAGE_FORMAT_DESCRIPTOR` here.  Is this a mistake in teh current spec?

api/opencl_runtime_layer.asciidoc
bashbaug: Discussed in the June 3rd memory subgroup.  Will add this condition.

api/opencl_runtime_layer.asciidoc
bashbaug: Discussed in the June 3rd memory subgroup.  Will create a separate issue for additional discussion.

api/cl_khr_external_memory_android_hardware_buffer.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+AHardwareBuffer_allocate(&desc,  &ahb_handle);
+cl_mem_properties props[] = {
+    (cl_mem_properties) CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR,
+    (cl_mem_properties) ahb_handle,
+    0};
+cl_mem imported_buffer = clCreateBufferWithProperties(context, props, CL_MEM_READ_WRITE,
+                            NULL, &error);
+If (error == CL_SUCCESS)
+{
+     clEnqueueCopyBuffer(queue, imported_buffer, ...);
```
joshqti: I think we need an acquire?

api/opencl_runtime_layer.asciidoc
```diff
@@ -660,65 +660,84 @@ is created successfully.
...
+  * {CL_INVALID_HOST_PTR}
+    ** if _host_ptr_ is `NULL`, and {CL_MEM_USE_HOST_PTR} or {CL_MEM_COPY_HOST_PTR} are set in _flags_
+    ** if _host_ptr_ is not `NULL` but {CL_MEM_COPY_HOST_PTR} or {CL_MEM_USE_HOST_PTR} are not set in _flags_
+ifdef::cl_khr_external_memory[]
+    ** if _properties_ includes a supported external memory handle and
+    _host_ptr_ is not `NULL`
+endif::cl_khr_external_memory[]
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+    ** if _properties_ includes an AHardwareBuffer external memory handle and
+    _host_ptr_ is not `NULL`
```
dcrawleyqti: We don't need a specific case for AHardwareBuffer as that's covered in line 714's "supported external memory handle"

api/opencl_runtime_layer.asciidoc
```diff
@@ -2184,101 +2203,120 @@ is created successfully.
...
+  * {CL_INVALID_HOST_PTR}
+    ** if _host_ptr_ is `NULL` and {CL_MEM_USE_HOST_PTR} or {CL_MEM_COPY_HOST_PTR} are set in _flags_
+    ** if _host_ptr_ is not `NULL` but {CL_MEM_COPY_HOST_PTR} or {CL_MEM_USE_HOST_PTR} are not set in _flags_
+ifdef::cl_khr_external_memory[]
+    ** if _properties_ includes a supported external memory handle and
+    _host_ptr_ is not `NULL`
+endif::cl_khr_external_memory[]
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+    ** if _properties_ includes an AHardwareBuffer external memory handle and
+    _host_ptr_ is not `NULL`
```
dcrawleyqti: Same as above, we don't need a specific case for AHardwareBuffer as that's covered in line 2299's "supported external memory handle"

api/opencl_runtime_layer.asciidoc
```diff
@@ -660,65 +660,84 @@ is created successfully.
...
 endif::cl_ext_immutable_memory_objects[]
+  * {CL_INVALID_BUFFER_SIZE}
+    ** if _size_ is zero
+    ** if _size_ is greater than {CL_DEVICE_MAX_MEM_ALLOC_SIZE} for all devices in _context_
+    ** if {CL_MEM_USE_HOST_PTR} or {CL_MEM_COPY_HOST_PTR} is set in _flags_, and
+    _host_ptr_ is a pointer returned by {clSVMAlloc}, and _size_ is greater than
+    the size passed to {clSVMAlloc}
+ifdef::cl_khr_external_memory_android_hardware_buffer[]
+    ** if _properties_ includes an AHardwareBuffer external memory handle and
+    _size_ is non-zero and greater than the AHardwareBuffer
```
bashbaug: I have a few additional questions about this error code:

1. Elsewhere in this PR we say for the AHardwareBuffer case: "If importing as a buffer, the implementation will infer the size of the buffer."  What does this mean for the passed-in size?  Specifically, what happens if the passed-in size is smaller than the size of the AHardwareBuffer?  Does the inferred size replace the passed-in size?
2. In the AHardwareBuffer case is a size of zero valid?  Note that this is a `CL_INVALID_BUFFER_SIZE` error in general (without AHardwareBuffer), so we would need to explicitly allow this if a size of zero is valid.

api/cl_khr_external_memory_android_hardware_buffer.asciidoc
bashbaug: Fixed!

Note:
* I added a missing release, also.
* I fixed a couple of other typos, also.

api/opencl_runtime_layer.asciidoc
bashbaug: Added the case:

"the image is being created from another memory object (buffer or image)"

api/opencl_runtime_layer.asciidoc
bashbaug: Good observation!  I removed this specific AHardwareBuffer case.

api/opencl_runtime_layer.asciidoc
bashbaug: Filed https://github.com/KhronosGroup/OpenCL-Docs/issues/1389 for additional discussion.

api/opencl_runtime_layer.asciidoc
joselopeqti: We agree that for the AHardwareBuffer the passed-in size should be always zero and the buffer size inferred, so we should remove the `CL_INVALID_BUFFER_SIZE` error for this case

bashbaug: Great.  I think I have clarified this now - please take a look.

joselopeqti: Yeah, I think now it is good. The sample code of the extension should also be changed as the size is still being passed to clCreateBufferWithProperties

bashbaug: Thanks, I fixed the one example code snip I found.  Please take a look - is this the only one?  dbe13e7

Jeremy-Kemp (approved):

PR #1374, opened 2025-05-13
Fix fast_distance parameter type

frasercrmck:
This builtin was incorrectly described as `fast_distance(float, floatn)`.


bashbaug (approved):
LGTM!  Obvious fix, merging immediately.

PR #1380, opened 2025-05-31
clarify the image row pitch requirements

### LINKED ISSUE #1333: Contradictory wording of cl_image_desc specification
[Description](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#image-descriptor) of `image_row_pitch` (the same refers to `image_slice_pitch`) seems mutually contradictory. First spec says:

`The image_row_pitch must be zero if ... . If image_row_pitch is zero and the image is created from an external memory handle, then the image row pitch is implementation-defined.`

which implies that in some circumstances `image_row_pitch` may be zero, in others it must be zero. In the same paragraph spec says:

`The image row pitch must be ≥ image_width × the size of an image element in bytes, and must be a multiple of the size of an image element in bytes.`

Shouldn't this sentence begin with: "Otherwise, the image pitch must ..." ?

bashbaug:
fixes #1333 

Clarifies the conditions when the image row pitch must be zero and what the requirements are when the image row pitch is nonzero.  I thought it was helpful to add line breaks to separate the different conditions, but I'm flexible and I can remove these or convert them to bullet points or similar, if desired.

Here are how these changes render in the spec:

> _image_row_pitch_ is the scan-line pitch in bytes. The _image_row_pitch_ must be zero if host_ptr is NULL, the image is not an image created from an external memory handle, and the image is not a 2D image created from a buffer.
If _image_row_pitch_ is zero and host_ptr is not NULL, then the image row pitch is calculated as image_width times the size of an image element in bytes.
If _image_row_pitch_ is zero and the image is created from an external memory handle, then the image row pitch is implementation-defined.
If _image_row_pitch_ is non-zero, then it must be greater than or equal to the image_width times the size of an image element in bytes, and must be a multiple of the size of an image element in bytes.
For an image created from a buffer, the _image_row_pitch_ must also:
> * be a multiple of the [CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT](https://github.com/KhronosGroup/OpenCL-Docs/compare/main...bashbaug:fix-image-row-pitch?expand=1#CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT) value for the image_format, image_type, and flags used to create the image, if the [cl_ext_image_requirements_info](https://github.com/KhronosGroup/OpenCL-Docs/compare/main...bashbaug:fix-image-row-pitch?expand=1#cl_ext_image_requirements_info) extension is supported, or
> * be a multiple of the maximum of the [CL_DEVICE_IMAGE_PITCH_ALIGNMENT](https://github.com/KhronosGroup/OpenCL-Docs/compare/main...bashbaug:fix-image-row-pitch?expand=1#CL_DEVICE_IMAGE_PITCH_ALIGNMENT) value for all devices in the context that support images, otherwise.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -2864,28 +2864,28 @@ include::{generated}/api/structs/cl_image_desc.txt[]
 ifdef::cl_khr_external_memory[]
     the image is not an image created from an external memory handle,
 endif::cl_khr_external_memory[]
-    and the image is not a 2D image created from a buffer,
+    and the image is not a 2D image created from a buffer. +
     If _image_row_pitch_ is zero and _host_ptr_ is not `NULL`, then the
-    image row pitch is calculated as _image_width_ {times} the size of an
-    image element in bytes.
+    image row pitch is calculated as _image_width_ times the size of an
+    image element in bytes. +
 ifdef::cl_khr_external_memory[]
     If _image_row_pitch_ is zero and the image is created from an external
-    memory handle, then the image row pitch is implementation-defined.
+    memory handle, then the image row pitch is implementation-defined. +
 endif::cl_khr_external_memory[]
-    The image row pitch must be {geq} _image_width_ {times} the size of an
-    image element in bytes, and must be a multiple of the size of an image
-    element in bytes.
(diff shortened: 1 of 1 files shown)


bashbaug:

PR #1383, opened 2025-06-03
Replace all instances of "provisional" with "experimental"

kpet:
As agreed in the 2025/05/03 teleconference.

Update the headers to remove references to the Khronos IP framework.

Change-Id: I2d0685d28310b840df70f83dd6e72aef86108c09

### DIFF
--- a/api/appendix_extensions.asciidoc
+++ b/api/appendix_extensions.asciidoc
@@ -34,6 +34,12 @@ publicly available to enable review and feedback
 from the community. While an extension is experimental, features may be
 added, removed, or changed in non-backward compatible ways.
 
+[NOTE]
+====
+Previous revisions of this specification used the term _provisional_ to denote
+ratified extensions with an experimental status.
+====
+
 If you have feedback on an experimental extension, please create an issue on
 the link:https://github.com/KhronosGroup/OpenCL-Docs/[OpenCL-Docs
 repository].


api/appendix_extensions.asciidoc
```diff
@@ -26,21 +26,18 @@ Within each group, extensions are listed in alphabetical order by their
 names.
 
 
-[[boilerplate-provisional-header]]
-== Provisional Extensions
+[[boilerplate-experimental-header]]
+== Experimental Extensions
```
alycm, line 30: Should there be a comment in this section that the term provisional used to be used?

kpet: We could do but we would also have to explain that the new "experimental" is not a strict equivalent of the old "provisional". Specifically, "provisional" only applied to ratified extensions but "experimental" applies to all extensions. Here's a straw-man:

>Previous revisions of this specification used the term "provisional" to denote ratified extensions with an experimental status.

api/appendix_extensions.asciidoc
```diff
@@ -26,21 +26,18 @@ Within each group, extensions are listed in alphabetical order by their
...
 
-_Provisional_ OpenCL extensions described in this appendix have been
-Ratified under the Khronos Intellectual Property Framework.
-They are being made publicly available as provisional extensions to enable
-review and feedback from the community.
-While an extension is provisional, features may be added, removed, or
-changed in non-backward compatible ways.
+_Experimental_ OpenCL extensions described in this appendix are being made
+publicly available as experimental extensions to enable review and feedback
+from the community. While an extension is experimental, features may be
```
bashbaug: This sentence reads a little strangely now, since it uses "experimental" twice.  How about:

```suggestion
_Experimental_ OpenCL extensions described in this appendix are being made
publicly available to enable review and feedback
from the community. While an extension is experimental, features may be
```

api/appendix_extensions.asciidoc
```diff
@@ -26,21 +26,18 @@ Within each group, extensions are listed in alphabetical order by their
...
+added, removed, or changed in non-backward compatible ways.
 
-If you have feedback on a provisional extension, please create an issue on
+If you have feedback on a experimental extension, please create an issue on
```
bashbaug:
```suggestion
If you have feedback on an experimental extension, please create an issue on
```

extensions/cl_intel_mem_force_host_memory.asciidoc
```diff
@@ -46,7 +46,7 @@ Final Draft
 
 == Dependencies
 
-This extension is written against the OpenCL Specification Version 3.0.1-Provisional.
+This extension is written against the OpenCL Specification Version 3.0.1-Experimental.
```
bashbaug: FWIW, this version of the specification was released as `3.0.1-Provisional`, so it might be best to leave this one as-is (even though you have to dig pretty deep into OpenCL-Registry git history to even find this version of the spec).

bashbaug (commented):
LGTM, aside from a few stylistic comments.  I built the specs with these changes and everything looks like it's in the right place.

I'll need to make a few changes to the header generation scripts as a result of these changes.  I'll take care of that tomorrow.

api/appendix_extensions.asciidoc
bashbaug: +1, I think it would be useful to have an informative note like the one you suggested, in case readers wonder what happened to the "provisional" term.

api/appendix_extensions.asciidoc
kpet: Thanks, missed that one.

extensions/cl_intel_mem_force_host_memory.asciidoc
kpet: I've restored the reference to `3.0.1-Provisional`

api/appendix_extensions.asciidoc
kpet: I've added a note with the text I proposed above.

bashbaug:

PR #1385, opened 2025-06-04
publish cl_khr_spirv_queries

bashbaug:
Please publish the `cl_khr_spirv_queries` extension.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -7617,7 +7617,7 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_UNORM_INT14X2_EXT"/>
             </require>
         </extension>
-        <extension name="cl_khr_spirv_queries" revision="1.0.0" supported="opencl">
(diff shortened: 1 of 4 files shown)


xml/cl.xml
```diff
@@ -7614,5 +7617,15 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_UNORM_INT14X2_EXT"/>
             </require>
         </extension>
+        <extension name="cl_khr_spirv_queries" revision="1.0.0" supported="opencl" ratified="opencl">
+            <require>
+                <type name="CL/cl.h"/>
+            </require>
```
kpet, line 7623: Do we really need these lines?

kpet (approved):

xml/cl.xml
bashbaug: I'm honestly not sure.

They're present for every extension, though, so if we decide that we don't need them, I'd prefer to remove them all in a separate PR.

bashbaug:

PR #1399, opened 2025-06-24
improve error code consistency, part one

bashbaug:
This is a PR to partially address the way we document error conditions, see #1320 and in particular https://github.com/KhronosGroup/OpenCL-Docs/issues/1320#issuecomment-2745919727.

This doesn't modify the entire spec, but:

* It's already a very large PR.
* I think it's a nice improvement over what we have currently.

We should decide whether we merge these changes in their partial state or whether we wait to review and merge the changes for the entire spec in the same PR.

Regardless, it would be beneficial to decide what to do with this PR sooner rather than later, because any changes to error conditions will likely result in merge conflicts that will need to be resolved.


bashbaug:
All of these changes have now been merged individually, so there is nothing left in this PR :tada:

PR #1402, opened 2025-07-01
Add vec16 types for short types

Maetveis:
Match cl_intel_subgroups_short version 1.1 and add vec16 types for block reads and writes for shorts.

### DIFF
--- a/extensions/cl_intel_subgroup_local_block_io.asciidoc
+++ b/extensions/cl_intel_subgroup_local_block_io.asciidoc
@@ -115,6 +115,12 @@ void  intel_sub_group_block_write_us( __local ushort* p, ushort data )
 void  intel_sub_group_block_write_us2( __local ushort* p, ushort2 data )
 void  intel_sub_group_block_write_us4( __local ushort* p, ushort4 data )
 void  intel_sub_group_block_write_us8( __local ushort* p, ushort8 data )
+
+/* Extension version 1.1 adds the functions: */
+
+ushort16 intel_sub_group_block_read_us16( const __local ushort* p )
+
+void intel_sub_group_block_write_us16( __local ushort* p, ushort16 data )
 ----
 
 If `cl_intel_subgroups_long` is supported, add variants of the `ulong` subgroup block read and write functions that support loading from and storing to pointers to the `+__local+` address space:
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
Thanks for updating this!

Since I was confused for a bit: the `cl_intel_subgroups_short` extension added support for 16-wide vectors via https://github.com/KhronosGroup/OpenCL-Docs/pull/906.  This PR updates `cl_intel_subgroup_local_block_io` as well, to explicitly document that 16-wide vectors are supported for pointers to the local address space, also.

I'm wondering if we need to explicitly rev the `cl_intel_subgroup_local_block_io` extension to add this support, though, based on:

> This extension extends the subgroup block read and write functions defined by cl_intel_subgroups (and, when supported, cl_intel_subgroups_char, cl_intel_subgroups_short, and cl_intel_subgroups_long) to support reading from and writing to pointers to the __local memory address space in addition to pointers to the __global memory address space.

It also looks like the [commit](https://github.com/intel/intel-graphics-compiler/commit/93c59b8eb95b5aab8cf50629ebe73af345bbab2a) that added support for 16-wide vectors added it for both the global address space and the local address space.

If so, perhaps:

1. We can keep this extension (`cl_intel_subgroup_local_block_io`) as version 1.0, perhaps revving only the patch version (so 1.0.1).
2. Clarify that "For extension version 1.1 or newer" is referring to the `cl_intel_subgroups_short` extension and not this extension.

Then, drivers do not need to be updated to safely access this functionality.

What do you think?

Maetveis:
> Thanks for updating this!
> 
> Since I was confused for a bit: the `cl_intel_subgroups_short` extension added support for 16-wide vectors via #906. This PR updates `cl_intel_subgroup_local_block_io` as well, to explicitly document that 16-wide vectors are supported for pointers to the local address space, also.

Sorry for being so brief on the description, I wanted to get this posted yesterday, but didn't have the time to properly explain.

> I'm wondering if we need to explicitly rev the `cl_intel_subgroup_local_block_io` extension to add this support, though, based on:
> 
> > This extension extends the subgroup block read and write functions defined by cl_intel_subgroups (and, when supported, cl_intel_subgroups_char, cl_intel_subgroups_short, and cl_intel_subgroups_long) to support reading from and writing to pointers to the __local memory address space in addition to pointers to the __global memory address space.
> 
> It also looks like the [commit](https://github.com/intel/intel-graphics-compiler/commit/93c59b8eb95b5aab8cf50629ebe73af345bbab2a) that added support for 16-wide vectors added it for both the global address space and the local address space.
> 
> If so, perhaps:
> 
> 1. We can keep this extension (`cl_intel_subgroup_local_block_io`) as version 1.0, perhaps revving only the patch version (so 1.0.1).
> 2. Clarify that "For extension version 1.1 or newer" is referring to the `cl_intel_subgroups_short` extension and not this extension.
> 
> Then, drivers do not need to be updated to safely access this functionality.
> 
> What do you think?

Sounds good to me. For context, I was looking into adding the declarations of these functions to the clang builtin header `opencl-c.h` and that's when I spotted that the `__local` 16 wide variants are not explicitly documented.

PR #1408, opened 2025-07-10
publish cl_img_safety_mechanisms extension

AhmedAmraniAkdi:


xml/cl.xml
```diff
@@ -6933,6 +6946,30 @@ server's OpenCL/api-docs repository.
             <require comment="cl_device_info">
                 <enum name="CL_DEVICE_MEMORY_CAPABILITIES_IMG"/>
             </require>
+        </extension>
+         <extension name="cl_img_safety_mechanisms" revision="0.0.0" supported="opencl">
```
bashbaug: Do you want this to be version 1.0.0 to match the version in the extension doc?

```suggestion
         <extension name="cl_img_safety_mechanisms" revision="1.0.0" supported="opencl">
```

bashbaug (approved):
Have you tried generating extension headers with these XML changes?  This is something you'll probably want to do anyhow, and I've found that it's a good way to check that the XML changes have all been made properly.

xml/cl.xml
AhmedAmraniAkdi: Thanks Ben, I've changed the revision to 1.0.0.

PR #1382, opened 2025-06-03
setting kernel arguments before recording it in command buffer

aharon-abramson:
Add an image for the life cycle in mutable command buffers and specify that there is no need to set all the kernel arguments before recording it.

Issue #1311, opened 2025-02-10
Clarify the necessity of CL_COMMAND_BUFFER_STATE_KHR


joshqti:
The query for command buffer state via CL_COMMAND_BUFFER_STATE_KHR seems unnecessary. The application can deduce if a command buffer is PENDING or EXECUTABLE by checking the state of the cl_event returned by clEnqueueCommandBufferKHR.   Querying the state of a command buffer will incur a performance penalty in the implementation, especially in the case of simultaneous use.

Can we remove this query?  I can see the utility of a query for if a command buffer has been finalized.  Perhaps we could replace CL_COMMAND_BUFFER_STATE_KHR  with CL_COMMAND_BUFFER_HAS_BEEN_FINALIZED, which will report if a command buffer has been finalized or not.

EwanC:
I think this is somewhat related to the discussion in https://github.com/KhronosGroup/OpenCL-Docs/issues/891#issuecomment-2524736323 around clarifying the state diagram

bashbaug:
Discussed in the February 11th teleconference:

-	Eliminating direct application access to the command buffer state may lead to more efficient implementations.
-	If we removed direct application access to the command buffer state what would break?  We do not believe the command buffer CTS is using this query, outside of targeted tests.  We do not believe that the SYCL graph implementation will use the command buffer state query, either.
-	We will revisit this question next week as well, when more vendors who have implemented command buffers are present.

bashbaug:
Recording a possibly related thought: We currently have an error condition for [clEnqueueCommandBufferKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueCommandBufferKHR) that refers to the pending state:

> [CL_INVALID_OPERATION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_OPERATION) if command_buffer was not created with the [CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR) flag and is in the [Pending](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#pending) state.

If the query for the command buffer state is expensive, is checking for this error condition expensive, also?

[...]

Issue #891, opened 2023-03-09
Revisit command-buffer simultaneous-use


EwanC:
During the course of writing CTS test for cl_khr_command_buffer, in particular https://github.com/KhronosGroup/OpenCL-CTS/pull/1584, questions have arisen about the behaviour of [CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_modifications_to_section_4_2_of_the_opencl_api_specification) and how usable it is for developers.

E.g - Do the following examples exhibit simultaneous-use behaviour?

```cpp
// Case 1: queue is an out-of-order queue.
// Note: No event dependencies!
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
clFinish(queue);

// Case 2: queue0 and queue1 are in-order queues.
clEnqueueCommandBufferKHR(1, &queue0, cmdbuf, 0, NULL, NULL);
clEnqueueCommandBufferKHR(1, &queue1, cmdbuf, 0, NULL, NULL);
clFlush(queue0); // these flushes may not be required
clFlush(queue1);
clFinish(queue0);
clFinish(queue1);
```

In my mind, a lot of the behaviour of simultaneous-use actually depends on whether a vendor implements queues eagerly or lazily. In lazy execution (what happens in Codeplay' implementation for example) we wouldn't start submitting anything until the clFinish()/cFlush(), so the pending count keeps on increasing until then, even for in-order queues. In an eager queue implementation, the first enqueue would be run straight away, and may indeed have finished by the time the host code calls the second command buffer.

Depending on an unreported implementation detail is hard for a user developing code to reason about
though, so maybe we can do better here. Before the cl_khr_command_buffer extension becomes final from provisional we should confirm if the currently specified behaviour is what we want, or whether we should revisit this property.

bashbaug:
I think this is another case to consider, especially for implementations that implement queues lazily (AKA "batched").  I think this would be considered simultaneous use?

```c++
// Case A: queue is an in-order queue.
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
[...]

PR #1411, opened 2025-07-11
Redefine command-buffer simultaneous-use

EwanC:
As discussed in https://github.com/KhronosGroup/OpenCL-Docs/issues/891 the current definition of [simultaneous use](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_command_buffers) is hard for users to reason about.

Instead a better model is one where simultaneous-use isn't a constraint on when a command-buffer can be submitted for execution, but the execution itself. That is, simultaneous-use occurs when a command-buffer is enqueued for execution while any previous submission of the command-buffer is still in-flight, and there any no scheduling dependencies expressed to serialize execution on the in-flight submission.

This means that pipelined submissions of a command-buffer, where there is an in-order queue, barrier, or cl_events to serialize execution, is always valid usage without the optional simultaneous-use device feature.

To avoid the runtime having to incur overheads from monitoring when simultaneous-use occurs so it can throw an error, violating this valid usage for non simultaneous-use command-buffers is UB. Following from this and relation to https://github.com/KhronosGroup/OpenCL-Docs/issues/1311, the pending state has also been removed from the command-buffer lifecycle, and there is only left the binary states of recording and executable (note that this conflicts with the change proposed in https://github.com/KhronosGroup/OpenCL-Docs/pull/1382, but it I believe it is straightforward to resolve). This is because a user can use the existing OpenCL mechanisms of host waits and event queries to inspect the state of individual command-buffer enqueues to avoid simultaneous-use, and having this stored as a command-buffer state incurs the runtime overhead of tracking a previous submissions. 

The implications for updating a command-buffer is that the error behavior is removed from update, so simultaneous-use is a property of execution rather than enqueue. In fact update is the only scenario where simultaneous use now adds value, as without it there will be race conditions during execution. But with command-buffer update the command-buffer operations can work on different data. As a result i've moved the simultaneous capability from the base extension to mutable-dispatch.

I've outlined the CTS changes that would be required for this change in https://github.com/KhronosGroup/OpenCL-CTS/issues/2473 and drafted them in https://github.com/KhronosGroup/OpenCL-CTS/pull/2477


api/cl_khr_command_buffer.asciidoc
```diff
@@ -126,11 +126,16 @@ from the sync-point values returned is implementation defined.
...
+The optional <<simultaneous-use, simultaneous use>> capability was added to the
+extension so that vendors could support concurrent execution of the same
+command-buffer. However, simultaneous use may result in command-buffers having
+a larger overhead to implement, so the capability is optional to enable
+optimizations when this usage isn't required by a user.
+
+Instead the goal for the base level of functionality provided by the extension
+is to support pipelined workflows. Where a command-buffer is repeatedly
+enqueued, with each enqueue expressing dependencies on any previous submissions,
+without the enqueue calls blocking in user code.
```
bashbaug: Nitpick: I found this paragraph difficult to parse and I'd consider removing or rephrasing it.

If we decide to keep it, what is it trying to say that isn't covered in the first paragraph?

api/opencl_runtime_layer.asciidoc
```diff
@@ -14618,13 +14618,14 @@ on one or more command-queues without any application code interaction.
...
-The simultaneous use capability removes this restriction and allows
-command-buffers to have a <<pending_count, Pending Count>> greater than 1.
+Upon creation a command-buffer is defined as being in the <<recording,
+Recording>> state, in order for the command-buffer to be enqueued
+it must first be finalized using {clFinalizeCommandBufferKHR} after which no
+further commands can be recorded. A command-buffer is submitted for execution
+on command-queues with a call to {clEnqueueCommandBufferKHR}. It is
+always valid to call {clEnqueueCommandBufferKHR} with a command-buffer that
+has previously been enqueued, provided the call doesn't violate the definition
+of <<simultaneous-use, simultaneous use>>.
```
bashbaug: Editorial:

```suggestion
Upon creation a command-buffer is in the <<recording,
Recording>> state.  In order for the command-buffer to be enqueued
it must first be finalized using {clFinalizeCommandBufferKHR}, after which no
further commands can be recorded. A command-buffer is enqueued for execution
on command-queues with a call to {clEnqueueCommandBufferKHR}. It is
always valid to call {clEnqueueCommandBufferKHR} with a command-buffer that
has previously been enqueued, provided the call doesn't violate the definition
of <<simultaneous-use, simultaneous use>>.
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -14690,6 +14691,24 @@ If using layered extension {cl_khr_command_buffer_mutable_dispatch_EXT},
...
+execution without a prerequisite on all the previous submissions of the same
+command-buffer which are not in the {CL_COMPLETE} state.
+
+An example of simultaneous use would be two submissions of the same
+command-buffer to a single out-of-order queue, without any events or barriers
+used to express a dependency between the two enqueue calls. Using a single
+in-order queue, events, or barriers to express dependencies between submissions
+of the same command-buffer would each be ways to avoid simultaneous use and are
+valid usage of command-buffers created without the
+{CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR} flag.
```
bashbaug: Is this normative or informative?  Consider:

```suggestion
[NOTE]
====
An example of simultaneous use would be two submissions of the same
command-buffer to a single out-of-order queue, without any events or barriers
used to express a dependency between the two enqueue calls. Using a single
in-order queue, events, or barriers to express dependencies between submissions
of the same command-buffer would each be ways to avoid simultaneous use and are
valid usage of command-buffers created without the
{CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR} flag.
====
```

api/cl_khr_command_buffer.asciidoc
```diff
@@ -4,15 +4,15 @@
 include::{generated}/meta/{refprefix}cl_khr_command_buffer.txt[]
 
 // *Revision*::
-//    0.9.6
+//    0.9.7
```
bashbaug: This is revision 0.9.8.

Can we just remove this commented out text, though?  It's all auto-generated now.

api/cl_khr_command_buffer.asciidoc
EwanC: I've just gone and removed it

api/opencl_runtime_layer.asciidoc
EwanC: good shout, it's non-normative. Added the `[NOTE]`

api/cl_khr_command_buffer.asciidoc
EwanC: removed it

bashbaug (approved):

api/opencl_runtime_layer.asciidoc
```diff
@@ -16553,8 +16552,7 @@ include::{generated}/api/version-notes/clRemapCommandBufferKHR.asciidoc[]
   * _errcode_ret_ returns an appropriate error code.
     If _errcode_ret_ is `NULL`, no error code is returned.
 
-The returned command-buffer has the same state as the input command-buffer,
-unless the input command-buffer is in the <<pending, Pending>> state, in
+The returned command-buffer has the same state as the input command-buffer.
 which case the returned command-buffer has state <<executable, Executable>>.
```
joselopeqti: This line should be removed as well

`which case the returned command-buffer has state <<executable, Executable>>.`

api/opencl_runtime_layer.asciidoc
```diff
@@ -14813,9 +14821,9 @@ include::{generated}/api/version-notes/CL_COMMAND_BUFFER_FLAGS_KHR.asciidoc[]
       | This is a bitfield and can be set to a combination of the following values:
 
         {CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR_anchor} - Allow multiple
-        instances of the command-buffer to be submitted to the device for
-        execution.
-        If set, devices must support
+        instances of the command-buffer to be scheduled for execution on the
+        device in a usage pattern that exhibits <<simultaneous-use,
+        simultaneous use>>. If set, devices must support
         {CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR}.
```
joselopeqti, line 14833: The description for `CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR` should also be changed in opencl_platform_layer.asciidoc, currently it is:

`{CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR_anchor} Device
        supports the command-buffers having a <<pending_count, Pending
        Count>> that exceeds 1.`

EwanC: good catch, thanks. updated now.

EwanC:
@bashbaug I've requested you re-review the PR, as since you approved it I've pushed a commit that moves the simultaneous-use capability from the base extension to cl_khr_command_buffer_mutable_dispatch. While updating the CTS tests I realized I couldn't think of a scenario without update where it adds value, as any concurrent execution of the command-buffer would probably be race condition. Whereas with mutable-dispatch you can update your kernel to new inputs/outputs to use that would allow two concurrent executions to operate on different data. We haven't discussed this in the group yet, so I can revert it if we think it's undesirable.

api/opencl_runtime_layer.asciidoc
```diff
@@ -14690,6 +14691,32 @@ If using layered extension {cl_khr_command_buffer_mutable_dispatch_EXT},
 usage>>.
 ====
 
+Simultaneous use is defined using the _prerequisite_ terminology from the
+<<_execution_model, execution model>>.
+ifndef::cl_khr_command_buffer_multi_device[]
+A command-buffer exhibits undefined behavior if a simultaneous use usage
```
aharon-abramson: The expression "use usage" seems redundant. I believe the word "usage" can be omitted.

bashbaug (approved):
Merging as discussed in the August 26th teleconference.

PR #1419, opened 2025-07-25
clean up inconsistencies for clSetKernelExecInfo error conditions

### LINKED ISSUE #1413: clSetKernelExecInfo error code inconsistency
I think when we merged support for cl_ext_buffer_device_address in https://github.com/KhronosGroup/OpenCL-Docs/pull/1159 we accidentally introduced an error code inconsistency for clSetKernelExecInfo.

Specifically, for [CL_KERNEL_EXEC_INFO_SVM_PTRS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_EXEC_INFO_SVM_PTRS) we say:

> To specify that no SVM allocations will be accessed by a kernel other than those set as kernel arguments, specify an empty set by passing param_value_size equal to zero and param_value equal to NULL.

But then, we also describe the error condition:

> [CL_INVALID_VALUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_VALUE) if param_name is not valid, if param_value is NULL or if the size specified by param_value_size is not valid.

I think this error condition is unnecessary and should be removed, especially given the additional `CL_INVALID_VALUE` error description following the quoted description, where it is only an error if _param_value_ is `NULL` **and** _param_value_size_ is greater than zero.

We should also check that this error condition that was added by the PR for cl_ext_buffer_device_address is still applicable, or whether it can be removed, also:

> [CL_INVALID_OPERATION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_OPERATION) for [CL_KERNEL_EXEC_INFO_SVM_PTRS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_EXEC_INFO_SVM_PTRS) and [CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM) if no devices in the context associated with kernel support SVM.

bashbaug:
fixes #1413

It is only a `CL_INVALID_VALUE` error when _param_value_ is `NULL` if _param_value_size_ is greater than zero.

Also tidies up a few of the `CL_INVALID_OPERATION` error conditions.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -11235,20 +11235,18 @@ successfully.
 Otherwise, it returns one of the following errors:
 
   * {CL_INVALID_KERNEL} if _kernel_ is a not a valid kernel object.
-  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_SVM_PTRS} and
-     {CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM} if no devices in
-     the context associated with _kernel_ support SVM.
-ifdef::cl_ext_buffer_device_address[]
-  * {CL_INVALID_OPERATION} for {CL_KERNEL_EXEC_INFO_DEVICE_PTRS_EXT} if no
-  device in the context associated with _kernel_ support the {cl_ext_buffer_device_address_EXT}
-  extension.
-endif::cl_ext_buffer_device_address[]
-  * {CL_INVALID_VALUE} if _param_name_ is not valid, if _param_value_ is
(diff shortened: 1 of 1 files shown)


joselopeqti:
Do we want this change to apply only when ext_buffer_device_address is supported by the implementation?

bashbaug:
Merging as discussed in the September 16th teleconference.

PR #1420, opened 2025-07-25
clarifies the change log appendix text

### LINKED ISSUE #1410: editorial: change log appendix text
The text we use in the change log appendix is confusing since it describes the version that was **changed**, but not the version **with the changes**.  We should try to improve it.

An easy fix might be to change e.g.:

> Changes from v3.0.5:

to e.g.:

> Changes from v3.0.5 to v3.0.6:

We should also probably change the change log appendix section title:

> Summary of Changes from OpenCL 3.0

Though, I'm less sure what this should change to.  Maybe:

> Summary of Changes to OpenCL 3.0

We would probably want to change this again when we release a new OpenCL revision, but we'd likely want to change the section title anyhow so this doesn't seem like a problem.

bashbaug:
fixes #1410

### DIFF
--- a/api/appendix_e.asciidoc
+++ b/api/appendix_e.asciidoc
@@ -457,11 +457,11 @@ conformance process:
 
   * {CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED}
 
-== Summary of Changes from OpenCL 3.0
+== Summary of Changes to OpenCL 3.0
 
 The first non-experimental version of the OpenCL 3.0 specifications was *v3.0.5*.
 
-Changes from *v3.0.5*:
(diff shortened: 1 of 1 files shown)


api/appendix_e.asciidoc
```diff
@@ -457,11 +457,11 @@ conformance process:
 
   * {CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED}
 
-== Summary of Changes from OpenCL 3.0
+== Summary of Changes to OpenCL 3.0
```
EwanC, line 460: If you're looking for suggestions, could also say "Changes **after** OpenCL 3.0", but I'm not sure if that's any better.

PR #1424, opened 2025-08-04
Fix markup in cl_img_safety_mechanisms extension

kpet:
Change-Id: I0f9c12e034a611b6acc8c6de60b9c7f47d89188c


kpet:
We could move to `--failure-level WARNING` and fix all warnings but some of them can be a bit annoying.

bashbaug:
Ah, coffee hadn't kicked in yet and I misunderstood what `--failure-level` did.

Do you have some examples of annoying warnings?  I don't see any other warnings generated in our CI job, at least...

PR #1426, opened 2025-08-06
XML: add cl_properties to OpenCL 1.0 requirements

RocketRide9:
`cl_queue_properties` is defined as `cl_properties`, so I think OpenCL 1.0 should require `cl_properties` too.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -4429,6 +4429,7 @@ server's OpenCL/api-docs repository.
             <type name="cl_sampler"/>
             <type name="cl_bool" comment="WARNING! Unlike cl_ types in cl_platform.h, cl_bool is not guaranteed to be the same size as the bool in kernels."/>
             <type name="cl_bitfield"/>
+            <type name="cl_properties"/>
             <type name="cl_device_type"/>
             <type name="cl_platform_info"/>
             <type name="cl_device_info"/>


bashbaug (approved):
Thanks, LGTM!

FWIW, I don't think that `cl_properties` necessarily needs to be exposed in OpenCL 1.0 due to `cl_queue_properties`, because `cl_queue_properties` is only used by `clCreateCommandQueueWithProperties`, which is an OpenCL 2.0 API.  However:

* The typedef for `cl_properties` is unguarded and exposed for OpenCL 1.0 in the [headers](https://github.com/KhronosGroup/OpenCL-Headers/blob/8a97ebc88daa3495d6f57ec10bb515224400186f/CL/cl.h#L47).
* `cl_properties` is used by many extensions that do not require OpenCL 2.0 or newer.
* `cl_properties` is not listed for any other OpenCL version.

RocketRide9:
Should these types also be required by some OpenCL version? Probably by OpenCL 1.0 too, because they are unguarded in [cl_platform.h](https://github.com/KhronosGroup/OpenCL-Headers/blob/8a97ebc88daa3495d6f57ec10bb515224400186f/CL/cl_platform.h#L150)
```c
cl_double
cl_float
cl_short
cl_int
cl_long
cl_char
cl_uchar
cl_half
cl_ushort
cl_uint
cl_ulong
```

bashbaug:
There was some discussion about this previously, here: https://github.com/KhronosGroup/OpenCL-Docs/pull/960

(I see this had the `cl_properties` change also, which is a little frustrating...)

I think the hangup was mostly about the `CL_HALF_xxx` constants, and everything else was non-contentious.  Maybe we can try merging everything but the `CL_HALF_xxx` constant changes to make forward progress?

CC @kpet

PR #1435, opened 2025-08-08
cl.xml Add values for cl_img_unified_svm_external_memory_dma_buf

paulfradgley:


xml/cl.xml
```diff
@@ -7680,5 +7682,14 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_DEVICE_SPIRV_CAPABILITIES_KHR"/>
             </require>
         </extension>
+        <extension name="cl_img_unified_svm_external_memory_dma_buf" revision="1.1.0" supported="opencl" depends="cl_khr_unified_svm" comment="requires cl_khr_unified_svm 0.2.0 or later">
```
bashbaug: I'd suggest calling this extension "experimental" if it's based on cl_khr_unified_svm, which is also currently "experimental".  Note, this will mean that it will be guarded by `CL_ENABLE_BETA_EXTENSIONS`, so please be sure this is the behavior that you want.

```suggestion
        <extension name="cl_img_unified_svm_external_memory_dma_buf" revision="1.1.0" supported="opencl" depends="cl_khr_unified_svm" experimental="true" comment="requires cl_khr_unified_svm 0.2.0 or later">
```

We should probably also have a discussion when we should add cl_khr_unified_svm support to the upstream headers, since it's currently only in a branch.

bashbaug (approved):

PR #1444, opened 2025-08-16
XML: Add a few missing requires

RocketRide9:
Selection of non contentious changes from #960. Also see #1426

- OpenCL 1.0 requires 
- - `cl_char`, `cl_int`, ... data types
- - `CL_M_*` constants
- cl_loader_layers extension requires `cl_icd_dispatch`

Unlike #960, it doesn't include `cl_half` type and `CL_HALF_*` constants. I also moved `cl_icd_dispatch` into the group with other types.


xml/cl.xml
```diff
@@ -5711,6 +5749,7 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl_icd.h"/>
             </require>
             <require>
+                <type name="cl_icd_dispatch"/>
```
bashbaug, line 5752: I was initially surprised to see this, but since `cl_icd_dispatch` is used by the `clInitLayer` function I think it makes sense to add it.

FWIW, the `cl_layer.h` includes `cl_icd.h`, which contains the definition of the `cl_icd_dispatch` structure, but listing `cl_icd_dispatch` here won't hurt anything.  The header generation scripts will generate a forward declaration of the structure, but this is still valid after the structure definition.

CC @Kerilk for awareness.

bashbaug:
Merging as discussed in the August 19th teleconference.

xml/cl.xml
Kerilk: Missed this one, sorry about that. Would require removing the typedef from `icd.h`.

PR #1450, opened 2025-08-26
error code consistency for buffers

bashbaug:
This is a PR to partially address the way we document error conditions, see https://github.com/KhronosGroup/OpenCL-Docs/issues/1320 and in particular https://github.com/KhronosGroup/OpenCL-Docs/issues/1320#issuecomment-2745919727.

It contains a subset of the changes from https://github.com/KhronosGroup/OpenCL-Docs/pull/1399, specifically for buffer memory objects.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -918,43 +918,39 @@ include::{generated}/api/version-notes/CL_BUFFER_CREATE_TYPE_REGION.asciidoc[]
 successfully.
 Otherwise, it returns one of the following errors in _errcode_ret_:
 
-  * {CL_INVALID_MEM_OBJECT} if _buffer_ is not a valid buffer object or is a
-    sub-buffer object.
-  * {CL_INVALID_VALUE} if _buffer_ was created with {CL_MEM_WRITE_ONLY} and
-    _flags_ specifies {CL_MEM_READ_WRITE} or {CL_MEM_READ_ONLY}, or if _buffer_
-    was created with {CL_MEM_READ_ONLY} and _flags_ specifies
-    {CL_MEM_READ_WRITE} or {CL_MEM_WRITE_ONLY}, or if _flags_ specifies
-    {CL_MEM_USE_HOST_PTR} or {CL_MEM_ALLOC_HOST_PTR} or {CL_MEM_COPY_HOST_PTR}.
-  * {CL_INVALID_VALUE} if _buffer_ was created with {CL_MEM_HOST_WRITE_ONLY} and
-    _flags_ specify {CL_MEM_HOST_READ_ONLY}, or if _buffer_ was created with
-    {CL_MEM_HOST_READ_ONLY} and _flags_ specify {CL_MEM_HOST_WRITE_ONLY}, or if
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -1210,57 +1207,52 @@ then be reused by the application.
...
+    ** if the context associated with _command_queue_ and _buffer_ are not the same
+    ** if the context associated with _command_queue_ and events in _event_wait_list_ are not the same
+  * {CL_INVALID_MEM_OBJECT}
+    ** if _buffer_ is not a valid buffer object
+  * {CL_INVALID_VALUE}
+    ** if _buffer_origin_, _host_origin_, or _region_ is `NULL`
+    ** if the region being read or written specified by _buffer_origin_, _region_, _buffer_row_pitch_, and _buffer_slice_pitch_ is out of bounds
+    ** if any _region_ array element is zero
+    ** if _buffer_row_pitch_ is not equal to zero and is less than _region_[0]
+    ** if _buffer_slice_pitch_ is not equal to zero and is less than the_region_[1] times _buffer_row_pitch_
```
EwanC:
```suggestion
    ** if _buffer_slice_pitch_ is not equal to zero and is less than _region_[1] times _buffer_row_pitch_
```

there was also `{times}` rather than `times` before, not sure if that was intentional.

EwanC (approved):
LGTM, this one is a nice readability improvement I feel

api/opencl_runtime_layer.asciidoc
bashbaug: Yeah, the change from `{times}` to `times` was intentional.

We have a number of different "conventions" throughout the spec, ranging from `*` to `X` to `{times}` (which expands to something like `X`) to the text `times`.  I think text descriptions like `times` is the most descriptive, so I've been trying to switch to it more consistently, though I'm sure this will be an ongoing process.

api/opencl_runtime_layer.asciidoc
```diff
@@ -1629,35 +1620,43 @@ argument value specified when _buffer_ is created is ignored by
...
-    {CL_DEVICE_MEM_BASE_ADDR_ALIGN} value for device associated with _queue_.
+  * {CL_INVALID_COMMAND_QUEUE}
+    ** if _command_queue_ is not a valid host command-queue
+  * {CL_INVALID_CONTEXT}
+    ** if the context associated with _command_queue_ and _buffer_ are not the same
+    ** if the context associated with _command_queue_ and events in _event_wait_list_ are not the same
+  * {CL_INVALID_MEM_OBJECT}
+    ** if _buffer_ is not a valid buffer object
+  * {CL_INVALID_VALUE}
+    ** if the region being read or written specified by _offset_ and _size_ is out of bounds
```
joselopeqti:
```suggestion
    ** if the region being filled specified by _offset_ and _size_ is out of bounds
```

bashbaug: Gah, too much copy-paste.  Thanks!

bashbaug:
Merging as discussed in the September 9th teleconference.

Issue #1407, opened 2025-07-09
Unclear semantics of clReImportSemaphoreSyncFdKHR in regards to clEnqueueWaitSemaphoresKHR and clEnqueueSignalSemaphoresKHR


karolherbst:
At the moment the spec doesn't specify if `clEnqueueSignalSemaphoresKHR` and `clEnqueueWaitSemaphoresKHR` operate on the semaphore already imported at call time or the currently imported semaphore once the queue reaches those commands.

I'd assume it's meant to operate on the already imported semaphore and `clReImportSemaphoreSyncFdKHR` just affects future calls to those. But I think this is a valid assumption to make just reading the specification.

I think "Calling clReImportSemaphoreSyncFdKHR is equivalent to destroying sema_object and re-creating it with the original sema_props from clCreateSemaphoreWithPropertiesKHR" can be read this way, but I think it's better to explicitly state this.

Any thoughts?

joselopeqti:
I think the application should make sure that any pending `clEnqueueSignalSemaphoresKHR` or `clEnqueueWaitSemaphoresKHR` have been completed before doing a re-import, and not doing it would cause undefined behavior.

PR #1454, opened 2025-09-09
Clear clReImportSemaphoreSyncKhr semantics for use after signal/wait

joselopeqti:
Fixes https://github.com/KhronosGroup/OpenCL-Docs/issues/1407

Currently the spec doesn't specify what would be the behavior if you call clReImportSemaphoreSyncKhr after signaling or importing that semaphore. 
As it is stated that "Calling clReImportSemaphoreSyncFdKHR is equivalent to destroying sema_object and re-creating it with the original sema_props from clCreateSemaphoreWithPropertiesKHR", this PR requires an interleaving wait, so no semaphore payload is left behind.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -13842,6 +13842,10 @@ _sema_object_ and re-creating it with the original _sema_props_ from
 be imported.
 The semaphore _sema_object_ must have originally imported an external handle
 of type {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR}.
+Calling {clReImportSemaphoreSyncFdKHR} after signaling the _sema_object_ or
+calling {clReImportSemaphoreSyncFdKHR} after creating the _sema_object_
+with {clCreateSemaphoreWithPropertiesKHR} without an interleaving wait may
+lead to undefined behavior.
 
 // refError
 


api/opencl_runtime_layer.asciidoc
```diff
@@ -13842,6 +13842,8 @@ _sema_object_ and re-creating it with the original _sema_props_ from
 be imported.
 The semaphore _sema_object_ must have originally imported an external handle
 of type {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR}.
+Calling {clReImportSemaphoreSyncFdKHR} when the last operation performed on
+_sema_object_ was not a wait may lead to undefined behavior.
```
kpet:
```suggestion
_sema_object_ was not a wait causes undefined behavior.
```
or
```suggestion
_sema_object_ was not a wait will result in undefined behavior.
```

api/opencl_runtime_layer.asciidoc
```diff
@@ -13842,6 +13842,8 @@ _sema_object_ and re-creating it with the original _sema_props_ from
 be imported.
 The semaphore _sema_object_ must have originally imported an external handle
 of type {CL_SEMAPHORE_HANDLE_SYNC_FD_KHR}.
+Calling {clReImportSemaphoreSyncFdKHR} when the last operation performed on
```
kpet, line 13845: 2025/09/16 memory TSG teleconference: do we need a more precise definition of what constitutes performing an operation? Is a query an operation? Does "perform[ing]" mean executing a command or enqueuing a command? This needs to be clarified.

kpet (changes requested):

api/opencl_runtime_layer.asciidoc
joselopeqti: In this context "perform" covers both enqueue and execution. We are open to proposals that would help to clarify the meaning of the text.

bashbaug: Does it make sense to refer to the queried value for [CL_SEMAPHORE_PAYLOAD_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_SEMAPHORE_PAYLOAD_KHR) here?  This isn't sufficient to cover cases where the semaphore has pending signals or waits enqueued, so perhaps additional text is still required, but if the (binary) semaphore payload is zero to indicate that it is in the un-signaled state then I think it is a more precise way of saying "the last operation performed on _sema_object_ was a wait".

joselopeqti: We believe that referring to the queried value might not be enough due to possible pending signals and waits. We are leaning towards a condition like:
"The last enqueued command to reference the semaphore should be clEnqueueWaitSemaphoresKHR and the semaphore should be in the un-signaled state."

Issue #1455, opened 2025-09-09
XML File and cl_icd_dispatch


bashbaug:
#1444 updated the XML file to add `cl_icd_dispatch` as a requirement for the `cl_loader_layers` extension.

This seemed like the right thing to do - see https://github.com/KhronosGroup/OpenCL-Docs/pull/1444/files#r2283135496 - but it causes issues with the generated headers, because it causes another typedef to be generated after the structure definition, which is not valid for some C versions.

```
In file included from /__w/OpenCL-Headers/OpenCL-Headers/tests/test_cl_layer.h.c:25:
/__w/OpenCL-Headers/OpenCL-Headers/CL/cl_layer.h:59:33: error: redefinition of typedef 'cl_icd_dispatch' [-Werror=pedantic]
   59 | typedef struct _cl_icd_dispatch cl_icd_dispatch;
      |                                 ^~~~~~~~~~~~~~~
In file included from /__w/OpenCL-Headers/OpenCL-Headers/CL/cl_layer.h:24:
/__w/OpenCL-Headers/OpenCL-Headers/CL/cl_icd.h:332:3: note: previous declaration of 'cl_icd_dispatch' with type 'cl_icd_dispatch' {aka 'struct _cl_icd_dispatch'}
  332 | } cl_icd_dispatch;
      |   ^~~~~~~~~~~~~~~
```

How do we want to fix this?  Some options to consider:

1. Remove the requirement from the XML file, so the duplicate typedef will not be generated in the extension headers.
2. Remove the typedef from the structure definition in `cl_icd.h`, so the generated typedef is not a duplicate.  Note: This could break code that is including `cl_icd.h` but not `cl_layer.h` and expecting the typedef to be present.
3. Do not include `cl_icd.h` from `cl_layer.h`.  Note: While this would probably work for some usages, it'd still be a problem if both headers are included in the "wrong" order, and most layers will want to include both headers.
4. Add a special case in the header generation script to skip generating the duplicate typedef.

I don't like any of these options, so I'm hoping there's another alternative to consider.  Thanks!

CC: @Kerilk @RocketRide9

Kerilk:
My view: I think the `require` should be in a formal XML definition of `cl_icd.h` (yes, I know we did not want to specify this) that formally bring the `cl_icd_dispatch` definition. Since the inclusion is described in the xml, the `cl_loader_layer` would get it transitively. 
```xml
            <require>
[...]

PR #1458, opened 2025-09-10
Remove require of cl_icd_dispatch to avoid double typedef

Kerilk:
Removing from icd.h would break codes that only include this file.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -5902,7 +5902,8 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl_icd.h"/>
             </require>
             <require>
-                <type name="cl_icd_dispatch"/>
+                <!-- Defined in icd.h -->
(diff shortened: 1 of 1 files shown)


xml/cl.xml
```diff
@@ -5902,7 +5902,8 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl_icd.h"/>
             </require>
             <require>
-                <type name="cl_icd_dispatch"/>
+                <!-- Defined in icd.h -->
```
bashbaug:
```suggestion
                <!-- Defined in cl_icd.h -->
```

bashbaug (approved):
Seems like the best short-term solution to unblock progress.

We can use #1455 to track a longer-term solution.

xml/cl.xml
Kerilk: Thanks Ben!

PR #1460, opened 2025-09-15
Generate extension interfaces for all EXT extensions

kpet:
- Add script to automatically generate the list of new commands, types, enums, and error codes.
- Use generated output for all EXT extensions.
- This fixes missing new enums for cl_ext_image_raw10_raw12.

Change-Id: I5a7edd837d39a3a0c6f86876ec7faa97a01d736e

### DIFF
--- a/scripts/gen_extension_interfaces.py
+++ b/scripts/gen_extension_interfaces.py
@@ -16,7 +16,7 @@ def parse_xml(path):
         return tree
 
 def GetHeader():
-    return """// Copyright 2017-2025 The Khronos Group.
+    return """// Copyright 2025 The Khronos Group.
 // SPDX-License-Identifier: CC-BY-4.0
 
 """
--- a/scripts/gen_extension_interfaces.py
+++ b/scripts/gen_extension_interfaces.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2019-2025 The Khronos Group Inc.
+# Copyright 2025 The Khronos Group Inc.
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
--- a/scripts/gen_extension_interfaces.py
+++ b/scripts/gen_extension_interfaces.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+
+# Copyright 2019-2025 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import os
+import urllib
+import xml.etree.ElementTree as etree
+import urllib.request
+
+def parse_xml(path):
+    file = urllib.request.urlopen(path) if path.startswith("http") else open(path, 'r')
+    with file:
+        tree = etree.parse(file)
+        return tree
+
+def GetHeader():
+    return """// Copyright 2017-2025 The Khronos Group.
+// SPDX-License-Identifier: CC-BY-4.0
+
+"""
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('-registry', action='store',
+                        default='cl.xml',
+                        help='Use specified registry file instead of cl.xml')
+    parser.add_argument('-o', action='store', dest='directory',
+                        default='.',
+                        help='Create target and related files in specified directory')
+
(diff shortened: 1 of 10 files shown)


scripts/gen_extension_interfaces.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+
+# Copyright 2019-2025 The Khronos Group Inc.
```
bashbaug: Does this copyright date need to go back so far?  If not:

```suggestion
# Copyright 2025 The Khronos Group Inc.
```

scripts/gen_extension_interfaces.py
```diff
@@ -0,0 +1,84 @@
...
+import urllib.request
+
+def parse_xml(path):
+    file = urllib.request.urlopen(path) if path.startswith("http") else open(path, 'r')
+    with file:
+        tree = etree.parse(file)
+        return tree
+
+def GetHeader():
+    return """// Copyright 2017-2025 The Khronos Group.
```
bashbaug: Same here:

```suggestion
    return """// Copyright 2025 The Khronos Group.
```

scripts/gen_extension_interfaces.py
```diff
@@ -0,0 +1,84 @@
...
+        for req in requires:
+            enums = req.findall('enum')
+            first_for_require = True
+            for e in enums:
+                if first_enum:
+                    iface_file.write('=== New enums\n\n')
+                    first_enum = False
+                if first_for_require:
+                    comment = req.get('comment')
+                    if comment == 'Error codes':
```
bashbaug, line 76: This whole section feels a bit brittle and we should probably find a more formal way to distinguish between error codes and enums.  For now, though, could we go down the "error code" path if the comment has "error" in any form, case insensitive?  It looks like we're more disciplined about this than I was fearing, but I do see several instances of "Error codes" (with no trailing colon), as well as several examples with additional trailing text.  I don't see any instances of "error codes" (non-capitalized) currently, but I could easily see somebody writing a comment like "New error codes" and be surprised when the script didn't pick it up.

bashbaug (commented):
Mostly LGTM - thank you for doing this!

I'm guessing there is a reason why you didn't apply this to KHR extensions also?  Is this something we should consider as a follow-on step?

scripts/gen_extension_interfaces.py
kpet: I did copy/paste code from a file that had this copyright but let me apply your suggestion so we don't waste time discussing this :).

scripts/gen_extension_interfaces.py
kpet: Agree it's brittle and we will want to do better. We should introduce dedicated attributes and stop using `comment` for various things and/or add a checker script. I don't really see the need to change this code though. As far as I can tell, there is only one `require` tag for error enums that does not use `Error codes` as spelled here (cl_khr_egl_image, checked with a case-insensitive grep for `require.*error`). I'd rather we change the XML in these cases and not make the script more tolerant. A more tolerant script would encourage lower consistency.

kpet:
Thanks for the review!

>I'm guessing there is a reason why you didn't apply this to KHR extensions also? Is this something we should consider as a follow-on step?

It was simply more work than I had time for. I don't expect there to be any KHR-shaped dragons hiding :). I really wanted to put a stake in the ground so we can start using the generation in new extensions. Completing the migration can be done in steps and we should absolutely do it.

If there are no other significant comments, I suggest we just merge this and create two issues:
1. Migrate all KHR extensions to use the generation.
2. Introduce a checker script for the `comment` attribute on `require` tags and/or change the schema to either replace `comment` with new and more solid attributes or add these new attributes alongside `comment` and that we'd retain for genuine comments. (Maybe this is two issues?)

bashbaug (approved):

bashbaug:
Merging as discussed in the September 16th teleconference.

kpet:
Created the two issues I proposed above:
1. https://github.com/KhronosGroup/OpenCL-Docs/issues/1461
2. https://github.com/KhronosGroup/OpenCL-Docs/issues/1462

PR #1463, opened 2025-09-17
cl_mobileye_reservation_sets

aharon-abramson:
Add cl_mobileye_reservation_sets extension and related API definitions


bashbaug:
You may want to rebase so this PR only includes your changes.  It looks like it picked up a bunch of other commits, which is confusing the CLA bot.  There is a merge conflict that needs to be resolved now, also.  Thanks!

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+  * _reservation_set_ specifies the reservation-set object to be retained.
+
+The _reservation_set_ reference count is incremented.
+
+*clRetainReservationSetMOBILEYE* returns `CL_SUCCESS` if the function is
+executed successfully.
+Otherwise, it returns one of the following errors:
+
+  * `CL_INVALID_COMMAND_BUFFER_KHR` if _command_buffer_ is not a valid command
+    -buffer.
```
EwanC, line 268: Think it's a bit confusing using `CL_INVALID_COMMAND_BUFFER_KHR` for the error code with reservation set retain/release. You seem to be using `CL_INVALID_RESERVATION_SET_MOBILEYE`  for a new `clCommandNDRangeKernelReservationSetMOBILEYE` error, so you could use that here too.

Also here "not a valid command-buffer" -> "not a valid reservation-set object"

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+After the _reservation_set_ reference count becomes zero and the command-buffer
+associated with _reservation_set_ is not executing, the reservation-set object
+is deleted.
+--
+
+[open,refpage='clCommandNDRangeKernelReservationSetMOBILEYE',desc='Record an ND-range kernel command into a reservation-set.',type='protos']
+--
+To record an ND-range kernel command into a reservation-set, call the function:
+include::{generated}/api/protos/clCommandNDRangeKernelReservationSetMOBILEYE.txt[]
+include::{generated}/api/version-notes/clCommandNDRangeKernelReservationSetMOBILEYE.asciidoc[]
```
EwanC, line 306: Just a thought, but instead of defining a new entry-point you could define a new `cl_command_properties_khr` property for `clCommandNDRangeKernelKHR` that lets the user specify the reservation set. Then rather than having to copy all the spec language for `clCommandNDRangeKernelKHR`  and keep it up to date with any changes to `clCommandNDRangeKernelKHR`, you would automatically be in sync.

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+    `cl_mutable_dispatch_config_khr` struct to update the command configuration
+    between recordings. The lifetime of this handle is tied to the parent
+    command-buffer, such that freeing the command-buffer will also free this
+    handle.
+
+*clCommandNDRangeKernelReservationSetMOBILEYE* returns `CL_SUCCESS` if the
+function is executed successfully. Otherwise, it returns the errors defined by
+*clEnqueueNDRangeKernel* except::
+
+  * `CL_INVALID_RESERVATION_SET_MOBILEYE` if _reservation_set_ is not a valid
```
EwanC, line 342: You should grab a new enum value for this in the xml https://github.com/KhronosGroup/OpenCL-Docs/blob/696e82027d2baa650c894fb1b2524afef717f52e/xml/cl.xml#L788

EwanC (commented):
I'm not an expert on pipes, but left a couple of minor command-buffer related thoughts.

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+[open,refpage='clCreateReservationSetMOBILEYE',desc='Create a reservation-set for ND-range kernel commands inside a command buffer.',type='protos']
+--
+To create a reservation-set for ND-range kernel commands inside a command
+buffer, call the function:
+
+include::{generated}/api/protos/clCreateReservationSetMOBILEYE.txt[]
+include::{generated}/api/version-notes/clCreateReservationSetMOBILEYE.asciidoc[]
+
+  * _command_buffer_ is a valid command buffer object.
+  * _affinity_domain_ is the level of the cache hierarchy that the work-groups
```
pjaaskel, line 215: I'm not sure if this is useful. Isn't it implicit from declaring kernels to the reservation set that the implementation wants to maximize locality between the connected work-groups? What if the implementation has, say, too small L2 but large enough L3. Or doesn't even have dynamic caches but pieces of scratch pad memory? Should it fail in that case when the affinity is defined?

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+    was created with the `CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR`
+    property with `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` and
+    _local_work_size_ is `NULL`, or if _properties_ includes the
+    `CL_MUTABLE_DISPATCH_ASSERTS_KHR` property with
+    `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` and
+    _local_work_size_ is `NULL`.
+endif::cl_khr_command_buffer_mutable_dispatch[]
+--
+
+[open,refpage='clCreateReservationSetPipeMOBILEYE',desc='Create a reservation-set pipe.',type='protos']
```
pjaaskel, line 403: For a sake of avoiding more spec fragmentation and reusing good spec groundwork, I strongly suggest to expand the old [pipe spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_pipes) with the concept of that if they are used to connect kernels in a reservation set, they get one instance for each concurrently executing work-group. A new flag in pipe creation could do?

extensions/cl_mobileye_reservation_sets.asciidoc
```diff
@@ -0,0 +1,545 @@
...
+
+Within a reservation-set, implementations can run work-groups from different
+commands with the same ID on the compute units that share the specified part of
+the cache hierarchy in parallel. These work-groups can access the same tiles of
+inputs and outputs, which are more likely to stay in the cache of these compute
+units and not be evicted until the work-groups are complete.
+
+All ND-range kernel commands inside a reservation set must have the same number
+of dimensions and the same number of work-groups in each dimension.
+
```
pjaaskel, line 105: The concurrency semantics need to be defined a bit clearer: What is expected to make progress concurrently? If WG 0 of kernel A starts running, also WG 0 of  _all_ other kernels in the set should make progress or just those that are connected with a pipe to declare a producer-consumer relation?

extensions/cl_mobileye_reservation_sets.asciidoc
pjaaskel: We could add a similar "not used outside a command buffer" property to the pipe as we've outlined [here](https://github.com/KhronosGroup/OpenCL-Docs/pull/1233) to the basic buffers to cover the problem of needing to preserve the state outside the command buffer.

pjaaskel (changes requested):

extensions/cl_mobileye_reservation_sets.asciidoc
aharon-abramson: The implementation doesn't know if a certain cache level is big enough for the tiles used by the kernels; it's determined by the kernels' logic.
If the platform doesn't have the desired cache level, the API should return an error. Also, when recording an ND-range kernel to a device that doesn't share this cache level.

pjaaskel: In terms of portability it might be a good idea to consider add a bit more freedom, perhaps by making it a hint, or at least by providing a default value which gives the implementation the freedom. AFAIU, the most essential information is that that there are kernels collaborating and should concurrently execute, thus the implementation should use this information for optimization and place them close to each other in terms of the memory hierarchy. In fact, I think the kernels have to make concurrent progress otherwise deadlocks might occur, right?

Issue #1465, opened 2025-09-30
Simultaneous update capability to mutable dispatch


joselopeqti:
Currently the specification for `cl_khr_command_buffer_mutable_dispatch` does not specify whether `clUpdateMutableCommandsKHR` can be called while an instance of the command-buffer is being executed.
Some implementations may not support this behavior and require the command-buffer to not being executing in order to update its commands arguments.

I suggest adding similar flag and capability as the ones added for simultaneous use https://github.com/KhronosGroup/OpenCL-Docs/issues/891. And require the developers to wait on the previous submissions of the command buffer to reach `CL_COMPLETE` state before calling `clUpdateMutableCommandsKHR` if the simultaneous update capability is not supported.

E.g. with simultaneous update
```
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, &event);
clUpdateMutableCommandsKHR(cmdbuf, 1, &config_type, &config);
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 1, &event, NULL);
```

E.g. without simultaneous update
```
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
clFinish(queue)
clUpdateMutableCommandsKHR(cmdbuf, 1, &config_type, &config);
clEnqueueCommandBufferKHR(1, &queue, cmdbuf, 0, NULL, NULL);
```

PR #1466, opened 2025-09-30
Add simultaneous update capability

joselopeqti:
This PR addresses https://github.com/KhronosGroup/OpenCL-Docs/issues/1465.
It follows a similar pattern to the one used for simultaneous update https://github.com/KhronosGroup/OpenCL-Docs/pull/1411.

It adds the `CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_UPDATE_KHR` capability, which allows developers to query if the implementation supports updating a command-buffer while an instance of the same is being ran.
If the capability is not supported, applications must wait for all previous submissions of the command-buffer to be finished before calling `clUpdateMutableCommandsKHR`.

It also fixes a wrong `ifdef` that was checking for `cl_khr_command_buffer_multi_device` instead of `cl_khr_command_buffer_mutable_dispatch`.

### DIFF
--- a/api/cl_khr_command_buffer_mutable_dispatch.asciidoc
+++ b/api/cl_khr_command_buffer_mutable_dispatch.asciidoc
@@ -6,7 +6,7 @@ include::{generated}/meta/{refprefix}cl_khr_command_buffer_mutable_dispatch.txt[
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2025-08-08
+    2025-09-30
 *IP Status*::
     No known IP claims.
 *Contributors*::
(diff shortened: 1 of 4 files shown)


EwanC:
Thanks for opening this, it's important to clarify to users if this is allowed, as it's an important use case to enable.

The current CTS test for non simultaneous-use workflow uses simultaneous-update behaviour by calling [update without a blocking wait on the previous command-buffer enqueue](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp#L231). So all cl_khr_command_buffer_mutable_dispatch implementations that currently run and pass the `mutable_dispatch_in_order`/`mutable_dispatch_out_of_order`/`mutable_dispatch_cross_queue`
tests should already support this behavior.

It would be interesting if that's been painful for any vendors and we should have this optional capability for simultaneous-update, or we can just make simultaneous-update required functionality for the extension and better specify that it is allowed.

PR #1471, opened 2025-10-07
Fixup command-buffer enqueue error after PR-1382

EwanC:
https://github.com/KhronosGroup/OpenCL-Docs/pull/1382 added support for deferring setting arguments on a kernel command which is updatable. To achieve this we added an extra command-buffer state, "finalized" which is entered when a command-buffer has been finalized but doesn't yet have all it's arguments set.

However, if a user tries to enqueue a command-buffer in this state it shouldn't be valid. Therefore update our current wording about when an command-buffer can be enqueued to say the specifically the executable state, when all the arguments are set. Rather than defined as only after the finalization operation, which could result in the command-buffer being in either the finalization or executable state.

We have a NOTE a few lines above to this effect already, which I'm not really sure

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -15122,12 +15122,6 @@ include::{generated}/api/version-notes/clEnqueueCommandBufferKHR.asciidoc[]
     application to wait on this command or query it for profiling
     information.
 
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Nice catch!

PR #1482, opened 2025-10-19
error code consistency for SVM

bashbaug:
This is another PR to partially address the way we document error conditions, see https://github.com/KhronosGroup/OpenCL-Docs/issues/1320 and in particular https://github.com/KhronosGroup/OpenCL-Docs/issues/1320#issuecomment-2745919727.

It contains a subset of the changes from https://github.com/KhronosGroup/OpenCL-Docs/pull/1399, specifically for SVM commands.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -8327,25 +8327,30 @@ context from which _command_queue_ was created the behavior is undefined.
 successfully.
 Otherwise, it returns one of the following errors:
 
-  * {CL_INVALID_COMMAND_QUEUE} if _command_queue_ is not a valid host
-    command-queue.
-  * {CL_INVALID_OPERATION} if the device associated with _command_queue_ does not support SVM.
-  * {CL_INVALID_CONTEXT} if the context associated with _command_queue_ and
-    events in _event_wait_list_ are not the same.
-  * {CL_INVALID_EVENT_WAIT_LIST} if _event_wait_list_ is `NULL` and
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -8656,24 +8675,28 @@ could result in undefined results.
...
+    ** if _command_queue_ is not a valid host command-queue
+  * {CL_INVALID_OPERATION}
+    ** if the device associated with _command_queue_ does not support SVM
+  * {CL_INVALID_CONTEXT}
+    ** if the context associated with _command_queue_ and events in _event_wait_list_ are not the same
+  * {CL_INVALID_VALUE}
+    ** if _num_svm_pointers_ is zero
+    ** if _svm_pointers_ is `NULL`
+    ** if _flags_ is not zero and is not a valid combination of the values
+    described in the <<migration-flags-table, Memory migration Flags>> table
```
EwanC, line 8688: This makes sense to me, but just acknowledging that this condition is new.

bashbaug: Yeah, there is a similar error condition for [clEnqueueMigrateMemObjects](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueMigrateMemObjects), and I can't think of any other behavior that would make sense, so hopefully this is non-contentious.

If someone has strong feelings about this I'll move it to a separate PR, but I'd prefer to just call it a related bugfix and move on.

PR #1483, opened 2025-10-19
clarify description of sub-group functions

### LINKED ISSUE #1429: Potential OpenCL Spec Error in 3.2.4
OpenCL specification says the following in section 3.2.4:

`For example, if a work-group function occurs within a loop, the work-items must encounter the same sub-group function in the same loop iterations.`

Is it supposed to say `sub-group function` instead of `work-group function` or is that intended?

bashbaug:
Fixes #1429

Removes a copy-paste error and rephrases to reduce duplication.

### DIFF
--- a/api/opencl_architecture.asciidoc
+++ b/api/opencl_architecture.asciidoc
@@ -578,12 +578,12 @@ Consider the following three domains of synchronization in OpenCL:
     for execution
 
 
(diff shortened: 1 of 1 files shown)


EwanC (approved):
Agree this is NFC and makes the text a bit more readable

PR #1487, opened 2025-10-23
cl_intel_subgroup_matrix_multiply_accumulate_tf32

bashbaug:
This PR has the extension source for the cl_intel_subgroup_matrix_multiply_accumulate_tf32 extension.

### DIFF
--- a/extensions/cl_intel_subgroup_matrix_multiply_accumulate_tf32.asciidoc
+++ b/extensions/cl_intel_subgroup_matrix_multiply_accumulate_tf32.asciidoc
@@ -1,0 +1,196 @@
+:data-uri:
+:sectanchors:
+:icons: font
+:source-highlighter: coderay
+
+ifdef::backend-html5[]
+:cl_intel_subgroup_matrix_multiply_accumulate_tf32: cl_intel_{zwsp}subgroup_{zwsp}matrix_{zwsp}multiply_{zwsp}accumulate_tf32
+:cl_intel_subgroup_matrix_multiply_accumulate: cl_intel_{zwsp}subgroup_{zwsp}matrix_{zwsp}multiply_{zwsp}accumulate
+:SPV_INTEL_subgroup_matrix_multiply_accumulate: SPV_INTEL_{zwsp}subgroup_{zwsp}matrix_{zwsp}multiply_{zwsp}accumulate
+:SPV_INTEL_tensor_float32_conversion: SPV_INTEL_{zwsp}tensor_{zwsp}float32_{zwsp}conversion
+endif::[]
+ifndef::backend-html5[]
+:cl_intel_subgroup_matrix_multiply_accumulate_tf32: cl_intel_&#8203;subgroup_&#8203;matrix_&#8203;multiply_&#8203;accumulate_tf32
+:cl_intel_subgroup_matrix_multiply_accumulate: cl_intel_&#8203;subgroup_&#8203;matrix_&#8203;multiply_&#8203;accumulate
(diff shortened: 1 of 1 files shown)


bashbaug:
https://opencl.gpuinfo.org/listreports.php?extension=cl_intel_subgroup_matrix_multiply_accumulate_tf32

Please note, since this is an OpenCL C-only extension, I have not modified the XML file.  I still have an action item to add all of the Intel OpenCL C-only functions to the XML file, which I will do as a single PR.

extensions/cl_intel_subgroup_matrix_multiply_accumulate_tf32.asciidoc
```diff
@@ -0,0 +1,196 @@
...
+
+* `M` may be equal to 1, 2, 4, or 8.
+* `N` must be equal to 16.
+In other words, the only supported subgroup size is 16.
+* The supported floating-point matrix types for `a` and `b` are 32-bit `float` data that is interpreted as `tf32` data when performing the matrix multiplication operation.
+For these `tf32` matrices, `K` must be equal to 8.
+The accumulation value `acc` and result value are 32-bit `float` values.
+* Because `N` must be equal to 16 and `K` must be equal to 8, each work-item contributes every other row of the matrix `a`.
+For `M` equal to one, only the first `K` work-items contribute to the matrix `a`, and contributions from the remaining work-items are ignored.
+For other values of `M`, the first `K` work-items contribute the even rows of the matrix `a`, and the remaining work-items contribute the odd rows of the matrix `a`.
```
BKoscielak, line 141: ~I think it is other way around. First 8 work items contribute odd rows, and other work items contribute even~

BKoscielak: I double checked and I am wrong, this commit is correct.

extensions/cl_intel_subgroup_matrix_multiply_accumulate_tf32.asciidoc
```diff
@@ -0,0 +1,196 @@
...
+----
+float intel_convert_tfloat32_as_float(float source);
+float2 intel_convert_tfloat322_as_float2(float2 source);
+float3 intel_convert_tfloat323_as_float3(float3 source);
+float4 intel_convert_tfloat324_as_float4(float4 source);
+float8 intel_convert_tfloat328_as_float8(float8 source);
+float16 intel_convert_tfloat3216_as_float16(float16 source);
+----
+
+=== Add a new Section 6.13.X.Y - `tf32` Subgroup Matrix Multiply Accumulate Functions
```
EwanC, line 120: This is such a nitpick, but I think Section 6.13 of the live hosted OpenCL-C spec (which says V3.0.19) is "Attribute Qualifiers" and not sure if this should be "Section 6.15" for "Built-In Functions"

bashbaug: I think you're right, but I also think I'd prefer to keep this unchanged.  My rationale is: The base [cl_intel_subgroup_matrix_multiply_accumulate](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroup_matrix_multiply_accumulate.html) extension was written against the v3.0.10 spec, and the built-in functions section appears to have been 6.13 back then, so it addded section 6.13.X.  Since this extension is adding a new sub-section for tf32, I think it is clearer to keep the same section numbering 6.13.X.Y than it would be to renumber the section 6.15.X.Y.

I suppose I could update cl_intel_subgroup_matrix_multiply_accumulate also, but I'm a bit worried how far the changes will ripple, so I'm inclined to just leave things as-is.

EwanC: Makes sense to me :+1:

PR #1492, opened 2025-11-14
Add missing error case for clSetKernelArgDevicePointerEXT

kpet:
It is never valid to call clSetKernelArgDevicePointerEXT for an argument that is not a pointer to the global address space. Since it is the choice of argument that is wrong in this case, CL_INVALID_ARG_INDEX looks like the most natural error code.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -11173,7 +11173,10 @@ successfully. Otherwise, it returns one of the following errors:
   * {CL_INVALID_KERNEL} if _kernel_ is not a valid kernel object.
   * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
     the {cl_ext_buffer_device_address_EXT} extension.
-  * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
+  * {CL_INVALID_ARG_INDEX}
+  ** if _arg_index_ is not a valid argument index.
+  ** if _arg_index_ corresponds to an argument that is not a pointer to the `global`
+     address space.
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
     by the OpenCL implementation on the device.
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -11173,7 +11173,10 @@ successfully. Otherwise, it returns one of the following errors:
   * {CL_INVALID_KERNEL} if _kernel_ is not a valid kernel object.
   * {CL_INVALID_OPERATION} if no devices in the context associated with _kernel_ support
     the {cl_ext_buffer_device_address_EXT} extension.
-  * {CL_INVALID_ARG_INDEX} if _arg_index_ is not a valid argument index.
+  * {CL_INVALID_ARG_INDEX}
+  ** if _arg_index_ is not a valid argument index.
+  ** if _arg_index_ corresponds to an argument that is not a pointer to the `global`
+     address space.
```
bashbaug, line 11179: I suspect this should also be valid for pointers to the `constant` address space?

Should we also have a similar error condition for `clSetKernelArgSVMPointer`?

pjaaskel: I say "yes" to both.

bashbaug: Discussed in the November 18th teleconference.

* We're going to keep this for the `global` address space only.
* We will file an issue to investigate expanding to the `constant` address space.
* We will also file an issue for `clSetKernelArgSVMPointer`.  We may need to tie this error condition to unified SVM though, or to a future minor OpenCL version.

kpet: Created the following two issues:
- https://github.com/KhronosGroup/OpenCL-Docs/issues/1494
- https://github.com/KhronosGroup/OpenCL-Docs/issues/1495 (added new minor version and USM labels)

bashbaug (approved):

PR #1493, opened 2025-11-18
refactor the description for clSetKernelArg

bashbaug:
Refactors the description of clSetKernelArg.

Key changes:

1. Removes the example kernel.  I did not find it useful, and it breaks the description of the API in a way that I found confusing.
2. Documents the requirements for the kernel argument value more precisely, based on the what the kernel argument is.
    * Note, the descriptions do not use the OpenCL C types, since the API works both for OpenCL C kernels and kernels created from SPIR-V or other mechanisms.
3. Updates the error code descriptions to use the new layout convention.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -11039,7 +11039,6 @@ Otherwise, it returns one of the following errors:
     ** if _arg_value_ points to the data to be used as the kernel argument value
     and _arg_size_ does not match the size of the data type for the argument
   * {CL_INVALID_ARG_VALUE}
-    ** if _arg_value_ is not a valid value.
     ** if the argument is an image declared with the `read_only` qualifier and
     _arg_value_ points to an image object created with the memory flag
     {CL_MEM_WRITE_ONLY}


api/opencl_runtime_layer.asciidoc
```diff
@@ -10911,112 +10911,90 @@ include::{generated}/api/protos/clSetKernelArg.txt[]
...
+set to `NULL`.
+
+If the kernel argument being set is a pointer to the `constant` address space,
+then the size in bytes of the memory object cannot exceed
+{CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE}.
+
+// TODO: Should CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE be checked when the kernel
+// argument is set or when the kernel is enqueued?
+// If it is checked when the kernel argument is set, then what is the error code
+// if the size is exceeded?
```
bashbaug: Note that this requirement was documented previously in text, but it was not described as any specific error condition.

If we check for this error when the kernel argument is set, what should the expected behavior be for a multi-device context, where the size is exceeded for one of the devices, but not for the other?

Even though it would mean detecting the error later, we may want to check for this error when the kernel is enqueued rather than when the kernel argument is set.

api/opencl_runtime_layer.asciidoc
```diff
@@ -10911,112 +10911,90 @@ include::{generated}/api/protos/clSetKernelArg.txt[]
...
-Behavior is undefined if the same memory object is passed as both a `read_only`
-image and a `write_only` image, or as a `read_write` image and either a
-`read_only` image or a `write_only` image.
-
-For all other kernel arguments, the _arg_value_ entry must be a pointer to
-the actual data to be used as argument value.
+All objects set as kernel arguments must be created from the same context as the
+kernel object.
+
+// TODO: Should this be CL_INVALID_CONTEXT, or should the object be considered invalid?
```
bashbaug: This is another case where the requirement was documented previously in text but was not described as a specific error condition.

api/opencl_runtime_layer.asciidoc
```diff
@@ -11036,55 +11014,61 @@ the {cl_mem_TYPE} backing store used with {CL_MEM_USE_HOST_PTR}.
...
+    ** if _arg_value_ is not a valid value.
+    ** if the argument is an image declared with the `read_only` qualifier and
+    _arg_value_ points to an image object created with the memory flag
+    {CL_MEM_WRITE_ONLY}, or
+    ** if the argument is an image declared with the `write_only` qualifier and
+    _arg_value_ points to an image object created with the memory flag
+    {CL_MEM_READ_ONLY}
+ifdef::cl_ext_immutable_memory_objects[]
+    or {CL_MEM_IMMUTABLE_EXT}
+endif::cl_ext_immutable_memory_objects[]
```
bashbaug, line 11050: This case was not described previously, but I think this should be an error condition.  We could also generalize the description to something like "with a memory flag that disallows writing", but I think the precision is helpful.

api/opencl_runtime_layer.asciidoc
```diff
@@ -11036,55 +11014,61 @@ the {cl_mem_TYPE} backing store used with {CL_MEM_USE_HOST_PTR}.
...
-     _arg_value_ refers to an image object created with _cl_mem_flags_ of
-     {CL_MEM_WRITE_ONLY}, or
-  ** if the image argument is declared with the `write_only` qualifier and
-     _arg_value_ refers to an image object created with _cl_mem_flags_ of
-     {CL_MEM_READ_ONLY}.
-  * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
-    by the OpenCL implementation on the device.
-  * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
-    required by the OpenCL implementation on the host.
+    ** if _arg_value_ is not a valid value.
```
bashbaug: This was a previously documented error condition, but I am not sure what would be considered "not a valid value" that is not documented elsewhere.  Perhaps if this is a local memory kernel argument and _arg_value_ is NOT equal to `NULL`?  Are there any other cases?

api/opencl_runtime_layer.asciidoc
```diff
@@ -11036,55 +11014,61 @@ the {cl_mem_TYPE} backing store used with {CL_MEM_USE_HOST_PTR}.
...
```
asudarsa-qti: May be we can replace CL_INVALID_ARG_VAL with CL_MEM_QUAL_FLAG_MISMATCH (if we are going to remove catch-all phrase?)

Thanks

api/opencl_runtime_layer.asciidoc
bashbaug: Discussed in the November 18th teleconference and decided to remove the text from clSetKernelArg.

I filed https://github.com/KhronosGroup/OpenCL-Docs/issues/1496 to track checking for this condition at enqueue time, instead.

api/opencl_runtime_layer.asciidoc
bashbaug: I've made this a `CL_INVALID_CONTEXT` error, as discussed in the November 18th teleconference.

api/opencl_runtime_layer.asciidoc
bashbaug: I was going to file a CTS issue to add a test for this case, but it looks like a negative test already exists.  So, it should be safe to add this error condition.  See:

https://github.com/KhronosGroup/OpenCL-CTS/blob/ec546b80c5995c981e1d92bf88e29ade01014cd5/test_conformance/api/test_kernels.cpp#L746-L752

api/opencl_runtime_layer.asciidoc
bashbaug: Note, strictly speaking this is a new error condition.  Are we OK adding this to OpenCL 3.0?

api/opencl_runtime_layer.asciidoc
bashbaug: I've removed this error condition as discussed in the November 18th teleconference, since we couldn't come up with any specific conditions that might generate it.

I filed https://github.com/KhronosGroup/OpenCL-Docs/issues/1497 to track the question about a local memory kernel argument.

api/opencl_runtime_layer.asciidoc
bashbaug: I'm not familiar with the error code `CL_MEM_QUAL_FLAG_MISMATCH` - is this part of a vendor extension?

Regardless, which error conditions would generate this error code?  Thanks!

PR #1501, opened 2025-12-09
error code consistency for clEnqueueNDRangeKernel

bashbaug:
This is another PR to partially address the way we document error conditions, see https://github.com/KhronosGroup/OpenCL-Docs/issues/1320 and in particular https://github.com/KhronosGroup/OpenCL-Docs/issues/1320#issuecomment-2745919727.

It contains a subset of the changes from https://github.com/KhronosGroup/OpenCL-Docs/pull/1399, specifically for clEnqueueNDRangeKernel.

Also, for clGetKernelSuggestedLocalWorkSizeKHR, while we're at it, since they share a lot of the same error conditions.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -11975,12 +11975,12 @@ Otherwise, it returns one of the following errors:
     _command_queue_.
     This error code is <<unified-spec, missing before>> version 1.1.
   * {CL_INVALID_IMAGE_SIZE}
-    ** if a kernel argument for _kernel_ is an image and the dimensions for the
+    ** if a kernel argument for _kernel_ is an image and the dimensions of the
     image, such as the image width or image height, are not supported by the
     device associated with _command_queue_
   * {CL_IMAGE_FORMAT_NOT_SUPPORTED}
-    ** if a kernel argument for _kernel_ is an image and the format the _image_,
-    such as the image channel order or image channel data type, are not
+    ** if a kernel argument for _kernel_ is an image and the format of the
+    image, such as the image channel order or image channel data type, are not
     supported by the device associated with _command_queue_
   * {CL_INVALID_OPERATION}
     ** if SVM pointers are set as arguments for _kernel_ and the device associated with _command_queue_ does not support SVM
@@ -12172,12 +12172,12 @@ Otherwise, it returns one of the following errors:
     _command_queue_.
     This error code is <<unified-spec, missing before>> version 1.1.
   * {CL_INVALID_IMAGE_SIZE}
-    ** if a kernel argument for _kernel_ is an image and the dimensions for the
+    ** if a kernel argument for _kernel_ is an image and the dimensions of the
     image, such as the image width or image height, are not supported by the
     device associated with _command_queue_
   * {CL_IMAGE_FORMAT_NOT_SUPPORTED}
-    ** if a kernel argument for _kernel_ is an image and the format the _image_,
-    such as the image channel order or image channel data type, are not
+    ** if a kernel argument for _kernel_ is an image and the format of the
+    image, such as the image channel order or image channel data type, are not
     supported by the device associated with _command_queue_
   * {CL_MEM_OBJECT_ALLOCATION_FAILURE}
     ** if there is a failure to allocate memory for the data store associated with any buffer or image object kernel arguments for _kernel_


api/opencl_runtime_layer.asciidoc
```diff
@@ -11942,45 +11942,57 @@ local work size.
...
-    required by the OpenCL implementation on the device.
-  * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
-    required by the OpenCL implementation on the host.
+    This error code is <<unified-spec, missing before>> version 1.1.
+  * {CL_INVALID_IMAGE_SIZE}
+    ** if a kernel argument for _kernel_ is an image and the dimensions for the
+    image, such as the image width or image height, are not supported by the
+    device associated with _command_queue_
+  * {CL_IMAGE_FORMAT_NOT_SUPPORTED}
+    ** if a kernel argument for _kernel_ is an image and the format the _image_,
```
joselopeqti: I think there is something missing in this line, like `the format of the _image_` or `the format values of the _image_`

api/opencl_runtime_layer.asciidoc
```diff
@@ -12115,90 +12127,86 @@ The starting local ID is always (0, 0, ..., 0).
...
-  * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources required
-    by the OpenCL implementation on the device.
-  * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
-    required by the OpenCL implementation on the host.
+  * {CL_INVALID_IMAGE_SIZE}
+    ** if a kernel argument for _kernel_ is an image and the dimensions for the
+    image, such as the image width or image height, are not supported by the
+    device associated with _command_queue_
+  * {CL_IMAGE_FORMAT_NOT_SUPPORTED}
+    ** if a kernel argument for _kernel_ is an image and the format the _image_,
```
joselopeqti: Same

api/opencl_runtime_layer.asciidoc
```diff
@@ -12115,90 +12127,86 @@ The starting local ID is always (0, 0, ..., 0).
...
+  * {CL_MEM_OBJECT_ALLOCATION_FAILURE}
+    ** if there is a failure to allocate memory for the data store associated with any buffer or image object kernel arguments for _kernel_
+  * {CL_INVALID_EVENT_WAIT_LIST}
+    ** if _event_wait_list_ is `NULL` and _num_events_in_wait_list_ is greater than zero
+    ** if _event_wait_list_ is not `NULL` and _num_events_in_wait_list_ is zero
+    ** if event objects in _event_wait_list_ are not valid events
+  * {CL_INVALID_OPERATION}
+    ** if SVM pointers are set as arguments for _kernel_ and the device associated with _command_queue_ does not support SVM
+    ** if system pointers are set as arguments for _kernel_ and the device associated with _command_queue_ does not support fine-grain system SVM
+// TODO: Do we still need these explicit examples?
```
joselopeqti, line 12191: In my opinion, this can be removed as it is already specified in the footnote 7: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_footnotedef_7

api/opencl_runtime_layer.asciidoc
bashbaug: Whoops, not sure what happened there.  Changed to:

> if a kernel argument for _kernel_ is an image and the format of the image, such as the image channel order or image channel data type, are not supported by the device associated with _command_queue_

api/opencl_runtime_layer.asciidoc
bashbaug: Fixed - same text as above.

api/opencl_runtime_layer.asciidoc
bashbaug: Thanks for the reminder!

I agree footnote 7 is a good example where these out-of-resources cases are described elsewhere in the spec.  So, I'm inclined to keep the same boilerplate text here for `CL_OUT_OF_RESOURCES` for `clEnqueueNDRangeKernel` vs. including the explicit examples.

If we really want to keep text like the following, which I don't believe is documented anywhere else in the spec...

> For example, the explicitly specified _local_work_size_ causes a failure to execute the kernel because of insufficient resources such as registers or local memory.

... perhaps it should move to a footnote, also?  Or, we can just remove it completely, as is done in this PR.

joselopeqti (approved):

PR #1514, opened 2026-01-08
remove OpenCL 1.2 dependency for Intel extensions

bashbaug:
There is no OpenCL 1.2 requirement for these extensions, either for the OpenCL device or the version of OpenCL C a program is compiled for.


wenju-he (approved):
LGTM

PR #1516, opened 2026-01-13
remove the OpenCL extension specification

### LINKED ISSUE #437: Should clGetExtensionFunctionAddressForPlatform be in the API specification?
Currently `clGetExtensionFunctionAddressForPlatform` is described in the extension specification, but the API entry point itself is not an extension.   Should the description actually be moved to the API specification?

E.g the extension specification section 1.3 "Getting OpenCL API Extension Function Pointers", up to the paragraph beginning "Function pointer typedefs must be declared ..." could be in the API specification, possibly at the end of section 4.1 "Querying Platform Info".

I was thinking about this because I realized that the old `clGetExtensionFunctionAddress` API is not described anywhere because it wasn't covered while unifying the API specification.

Putting this in 3.0 project because although it represents no functional change moving things between documents is quite a large change to a reader.

### LINKED ISSUE #1111: Remove OpenCL Extensions spec and move remaining content into API / C specs?

bashbaug:
fixes #437 
fixes #1111

Moves the remaining content in the OpenCL extension specification into the main OpenCL API specification and the OpenCL C specification and removes the standalone OpenCL extension specification from the spec toolchain.  Also updates the OpenCL reference pages and removes remaining links to the OpenCL extension specification.

Specifically:

* Moves the description of clGetExtensionFunctionAddressForPlatform and the older clGetExtensionFunctionAddress into the OpenCL API specification, as a sub-section of The OpenCL Platform Layer.
* Moves the extension naming conventions and header file conventions into the OpenCL API specification, as a sub-section of the existing OpenCL Extensions appendix.
* Moves the information about extension defines and pragmas into the OpenCL C specification, as a sub-section of the existing Preprocessor Directives and Macros section.


man/static/deadLinks.txt
```diff
@@ -19,16 +19,12 @@ cross-reference was created, even though the cross-reference target is not
 part of the OpenCL 3.0 specifications. Details are explained individually
 below for each such link.
 
-:clext: https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html
 :clenv: https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Env.html
 
 [cols="1a,1a", options="header"]
 |====
 | Cross-Reference               | Why It Resolves Here
 | `cles_khr_int64`              | See the link:{clenv}#_cles_khr_int64[OpenCL Environment specification]
```
bashbaug: I couldn't find a link to `cles_khr_int64` anymore.  If it doesn't exist, maybe this entire deadLinks page can be removed.

OpenCL_C.txt
```diff
@@ -4587,9 +4665,6 @@ compile-time constant expression.
 
 The attribute syntax can be extended for standard language extensions and
 vendor specific extensions.
```
bashbaug, line 4666: I've simply removed the reference to the extension spec here for now, since I'm not sure what it should refer to.  I also filed https://github.com/KhronosGroup/OpenCL-Docs/issues/1518 to revisit this section more broadly.

api/appendix_extensions.asciidoc
```diff
@@ -26,6 +26,119 @@ Within each group, extensions are listed in alphabetical order by their
 names.
 
 
+[[naming-convention-for-optional-extensions]]
+== Naming Convention for Optional Extensions
+
+OpenCL extensions approved by the OpenCL working group use the following
+naming convention:
+
+  * A unique _name string_ of the form `*cl_khr_<__name__>*` or
```
oddhack, line 37: Maybe add a note that exts not approved by the WG are still found separately?

api/appendix_extensions.asciidoc
```diff
@@ -26,6 +26,119 @@ Within each group, extensions are listed in alphabetical order by their
...
+
+Function pointer typedefs should be declared for all extensions that add API
+entrypoints.
+These typedefs are a required part of the extension interface, and should be
+provided in an appropriate header, such as `cl_ext.h`.
+
+The following convention should be followed for all extensions affecting the
+host API:
+
+[source,opencl]
```
oddhack, line 87: Are you not generating vendor extension interfaces for the headers? I thought all that stuff was in the XML already but haven't looked in quite a while.

man/static/clGetExtensionFunctionAddressForPlatform.txt
```diff
@@ -11,7 +11,7 @@ include::{config}/attribs.txt[]
 
 clGetExtensionFunctionAddressForPlatform - Returns the address of the extension function named by _funcname_ for a given _platform_.
 
-== C Specification
+== Specification
```
oddhack: Could this be integrated in a refpage block in the API spec, along with the extensions?

api/appendix_extensions.asciidoc
bashbaug: Good idea, I added a note to the section about vendor extension conventions:

> Vendor extensions are not currently included in the OpenCL specifications, but vendor extension specifications are frequently included in the online Registry of extensions.

api/appendix_extensions.asciidoc
bashbaug: We are, but it doesn't look like we include any information about the XML file in the spec, currently.

I'm open to adding this, but I would prefer to do so in a separate PR.

man/static/deadLinks.txt
bashbaug: I've gone ahead and removed the deadLinks page for now because I don't think it is needed anymore.

man/static/clGetExtensionFunctionAddressForPlatform.txt
bashbaug: Good idea, done!  Since the page is auto-generated now, I've removed the static page.

api/appendix_extensions.asciidoc
oddhack: For reference, for Vulkan we consider the XML to be a normative part of the spec (hard not to, since a substantial amount of the spec is generated from the XML), but the exact XML schema / toolchain are documented separately in the RNC schema and in a separate https://registry.khronos.org/vulkan/specs/latest/registry.html human-readable description of the schema. So in the spec we just point to the registry.html and mention the XML occasionally as a source of truth, but little more.

PR #1522, opened 2026-01-20
Spec source and XML changes for the cl_intel_kernel_allocations_info extension

ivvenevt:
Signed-off-by: Igor Venevtsev <igor.venevtsev@intel.com>

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -8041,7 +8041,7 @@ server's OpenCL/api-docs repository.
                 <type name="CL/cl.h"/>
             </require>
             <require>
-                <type name="cl_unified_shared_memory_intel"/>
+                <type name="cl_unified_shared_memory_type_intel"/>
             </require>
             <require>
                 <type name="cl_kernel_allocation_info_intel"/>


xml/cl.xml
```diff
@@ -335,6 +334,12 @@ server's OpenCL/api-docs repository.
             <member><type>cl_uint</type>                                <name>count</name></member>
             <member><type>char</type> <name>name</name>[<enum>CL_QUEUE_FAMILY_MAX_NAME_SIZE_INTEL</enum>]</member>
         </type>
+        <type category="struct" name="cl_kernel_allocation_info_intel">
+            <member><type>void*</type>                                  <name>base</name></member>
```
bashbaug: Minor XML issue: `void` is the type, and the pointer part `*` should be outside of the type tag:

```suggestion
            <member><type>void</type>*                                  <name>base</name></member>
```

xml/cl.xml
```diff
@@ -8029,5 +8036,19 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_COMMAND_QUEUE_SCHEDULING_WORK_GROUP_EXECUTE_COUNT_IMG"/>
             </require>
         </extension>
+        <extension name="cl_intel_kernel_allocations_info" supported="opencl">
+            <require>
+                <type name="CL/cl.h"/>
+            </require>
+            <require>
+                <type name="cl_intel_unified_shared_memory"/>
```
bashbaug: I think this should be:

```suggestion
                <type name="cl_unified_shared_memory_type_intel"/>
```

xml/cl.xml
```diff
@@ -8029,5 +8036,19 @@ server's OpenCL/api-docs repository.
                 <enum name="CL_COMMAND_QUEUE_SCHEDULING_WORK_GROUP_EXECUTE_COUNT_IMG"/>
             </require>
         </extension>
+        <extension name="cl_intel_kernel_allocations_info" supported="opencl">
```
bashbaug: We need to include a revision here.  Based on the doc, I think this should be 1.0.0?

```suggestion
        <extension name="cl_intel_kernel_allocations_info" revision="1.0.0" supported="opencl">
```

xml/cl.xml
ivvenevt: Fixed

xml/cl.xml
ivvenevt: Fixed

xml/cl.xml
ivvenevt: Fixed

xml/cl.xml
bashbaug: Can you check that this change has been applied?  I still see the previous text in this PR.

ivvenevt: Yes, I made the change but I've missed suggested type is cl_unified_shared_memory_type_**intel**, not cl_unified_shared_memory_type
Fixed.
Thanks Ben!

PR #1529, opened 2026-02-22
break extension conventions into two reference pages

bashbaug:
Apparently the reference page markup cannot span multiple sections without generating asciidoctor rendering issues.  So, break the "extension conventions" reference page into two pages - one for "naming conventions", and another for "header conventions", where each page consists of a single section.


kpet (approved):
As discussed, not ideal, but still better than the broken document we currently have.

bashbaug:
Merging as discussed in the March 3rd teleconference.

PR #1531, opened 2026-02-26
fix the SPIR-V Image Data Types table header row

bashbaug:
Fixes an error in the table header row for the "Image Data Types" table in the OpenCL SPIR-V environment spec.

I suspect this was a copy-paste error from a previous table.  Regardless, this table column describes the image channel data type, not the image channel order.


kpet (approved):

bashbaug:
Merging as discussed in the March 3rd teleconference.

PR #1535, opened 2026-03-05
move cluster size restrictions for cl_khr_subgroup_clustered_reduce

### LINKED ISSUE #1534: clustersize is clarified subgroup_rotate, but not for subgroup_reduce operations
I wonder if this is intentional.

For `sub_group_clustered_rotate` the CLC spec states:

> clustersize must be an integer constant expression and a power of two, smaller than or equal to the maximum sub-group size, otherwise the behavior is undefined.

but not for any other subgroup operation with a clustersize parameter.

bashbaug:
fixes #1534

Instead of documenting cluster size restrictions in the section header, describe the restrictions in the description for each of the functions. This aligns with the way cluster size restrictions are documented for cl_khr_subgroup_rotate and makes the restriction easier to spot.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -14959,9 +14959,9 @@ gentype sub_group_clustered_rotate(
     The value of _delta_ is required to be dynamically-uniform for all work
     items in the sub-group, otherwise the behavior is undefined.
 
-    _clustersize_ must be an integer constant expression and a power of two,
-    smaller than or equal to the maximum sub-group size, otherwise the
-    behavior is undefined.
+    Behavior is undefined if _clustersize_ is not an integer constant
+    expression, is not a power-of-two, or is greater than the maximum size of a
+    sub-group within the dispatch.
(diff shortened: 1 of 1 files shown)


karolherbst:
Mhh looks good. I just wonder if we want to converge the wording, because for rotate it says "clustersize must be an integer constant expression and a power of two, smaller than or equal to the maximum sub-group size, otherwise the behavior is undefined."

Picking either wording, or leaving it as is is fine with me.

bashbaug:
> I just wonder if we want to converge the wording

Sure, that's easy enough to do.

I've gone with the clustered reduce wording for now instead of the rotate wording because we usually use "must" to describe conditions that are statically checked (or at least, checkable), and that isn't the case for any of these conditions.

CC @kpet for awareness.

kpet (approved):

bashbaug:
Merging as discussed in the March 31st teleconference.

PR #1536, opened 2026-03-07
update and pin github actions

bashbaug:
Pins GitHub actions to a specific commit SHA, as per best practices.  Note that our dependabot job will still automatically keep these up-to-date, including the version comment.


Kerilk (approved):
Looks good to me, thanks.

bashbaug:
Merging as discussed in the March 10th teleconference.

PR #1540, opened 2026-03-27
clean up a few more descriptions of custom devices

bashbaug:
This is my attempt at a follow-up to PR #1388.  It removes even more specific text related to "custom devices" from the spec.  After this change, custom devices are basically only described in their glossary entry and in the description of the CL_DEVICE_TYPE_CUSTOM device type.

### DIFF
--- a/api/glossary.asciidoc
+++ b/api/glossary.asciidoc
@@ -171,8 +171,9 @@ Converged control flow ::
 Custom Device ::
     A _custom device_ is a specialized _device_ that supports a subset of the
     OpenCL runtime APIs for directed tasks but is not OpenCL conformant.
-    A _custom device_ must implement all of the OpenCL runtime APIs, but may
-    return implementation-defined error codes for unsupported functionality.
+    A _custom device_ must implement all of the OpenCL runtime APIs, but may not
+    support all of the required minimum device capabilities and may return
+    implementation-defined error codes for unsupported functionality.
     See also _Device_.
 
 Data Parallel Programming Model ::
--- a/api/glossary.asciidoc
+++ b/api/glossary.asciidoc
@@ -83,13 +83,11 @@ Buffer Object ::
 --
 
 Built-in Kernel ::
-    A _built-in kernel_ is a _kernel_ that is executed on an OpenCL _device_
-    or _custom device_ by fixed-function hardware or in firmware.
-    _Applications_ can query the _built-in kernels_ supported by a _device_
-    or _custom device_.
-    A _program object_ can only contain _kernels_ written in OpenCL C or
-    _built-in kernels_ but not both.
-    See also _Kernel_ and _Program_.
+    A _built-in kernel_ is a _kernel_ that is provided by an OpenCL
+    implementation.
+    A _built-in kernel_ is enqueued for execution like other _kernels_, but may
+    execute on specialized hardware that is unavailable to non-built-in kernels.
+    _Applications_ can query the _built-in kernels_ supported by a _device_.
 
 Child kernel ::
     See _Device-side enqueue_.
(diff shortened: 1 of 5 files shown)


api/glossary.asciidoc
```diff
@@ -171,17 +169,13 @@ Converged control flow ::
...
-    that allow OpenCL programs to be created from source (if an online
-    compiler is supported) and/or binary, or from _built-in kernels_
-    supported by the _device_.
+    A _custom device_ is a specialized _device_ that supports a subset of the
+    OpenCL runtime APIs for directed tasks but is not OpenCL conformant.
+    A _custom device_ must implement all of the OpenCL runtime APIs, but may
+    return implementation-defined error codes for unsupported functionality.
+    _Custom devices_ may support an online compiler.
+    When an online compiler is not available, OpenCL _programs_ may be created
+    from binaries or for _built-in kernels_ supported by the device.
```
karolherbst: This reads a bit weird. Maybe:

> When an online compiler is not available, OpenCL _programs_ may still be created from binaries or _built-in kernels_ supported by the device.

karolherbst (approved):
Looks good overall

api/glossary.asciidoc
bashbaug: Honestly, I almost removed this sentence, since it's true for any device that does not support an online compiler, not just a custom device.  Would it be better to remove it completely?

Or, maybe we should move it to the description of [CL_DEVICE_COMPILER_AVAILABLE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_COMPILER_AVAILABLE) instead?

karolherbst: Yeah, I think moving it to [CL_DEVICE_COMPILER_AVAILABLE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_COMPILER_AVAILABLE) makes more sense.

bashbaug: Cool, I've moved this to the CL_DEVICE_COMPILER_AVAILABLE description instead, and switched the wording closer to your suggestion.  Lemme know what you think.

api/glossary.asciidoc
```diff
@@ -171,17 +169,10 @@ Converged control flow ::
...
-    very power efficient and performant for directed tasks or hardware with
-    limited programmable capabilities such as specialized DSPs.
-    Custom devices are not OpenCL conformant.
-    Custom devices may support an online compiler.
-    Programs for custom devices can be created using the OpenCL runtime APIs
-    that allow OpenCL programs to be created from source (if an online
-    compiler is supported) and/or binary, or from _built-in kernels_
-    supported by the _device_.
+    A _custom device_ is a specialized _device_ that supports a subset of the
+    OpenCL runtime APIs for directed tasks but is not OpenCL conformant.
```
karolherbst, line 173: Do we want to clarify that this also means that queries might return non conformant values? E.g. return 2 for `CL_DEVICE_MAX_SAMPLERS`?

karolherbst (approved):
Looks good overall

api/glossary.asciidoc
bashbaug: Do you think we should say this explicitly?  IMHO this falls naturally out of "is not OpenCL conformant".

I guess we could add it easily enough to the following sentence, though, something like:

> A _custom device_ must implement all of the OpenCL runtime APIs, but may return device capabilities that are less than required minimum values or implementation-defined error codes for unsupported functionality.

bashbaug: I went with something like this, see below.

karolherbst: yeah looks good. I don't think we have to make this explicit either, but we also don't have to do it for error returned error codes. And I think it's better to be explicit about those two aspects at least.

lakshmih (approved):

bashbaug:
Merging as discussed in the April 7th teleconference.

Issue #1541, opened 2026-04-01
missing clEnqueueNDRangeKernel error condition


bashbaug:
I think when we did https://github.com/KhronosGroup/OpenCL-Docs/pull/1501 we missed an error condition, or perhaps the error condition was missing all along and it's just easier to see after the error code consistency changes.

The specific error conditions I'm looking at are:

> CL_INVALID_WORK_GROUP_SIZE
> * if local_work_size is not NULL, if the work-group size must be uniform, and if the global_work_size is not evenly divisible by the local_work_size
> * if local_work_size is not NULL and if the total number of work-items in the work-group is greater than the maximum work-group size supported for kernel on the device associated with command_queue (is greater than the value returned for CL_KERNEL_WORK_GROUP_SIZE
> * if local_work_size is not NULL and if the local_work_size does not match the required work-group size for kernel
> * if local_work_size is not NULL and if the local_work_size is not consistent with the required number of sub-groups for kernel

The specific error condition that I believe is missing is the case where the local_work_size is not NULL, the work-group size is zero, and the work-group size may be non-uniform.  I think this case should be an error, but it is not covered by any of these CL_INVALID_WORK_GROUP_SIZE error conditions, or by any other clEnqueueNDRangeKernel error conditions.

Note also, it's kind of only disallowed for the "work-group size must be uniform" case indirectly, since it's impossible for the global work size to be evenly evenly divisible by zero (the local work size).

One way to fix this is to add an explicit error condition, something like:

> * if local_work_size is not NULL and the total number of work-items in the work-group is zero

Do we agree that this should be an error?  If so, is this the best way to document this error condition?

bashbaug:
Closing, fixed by https://github.com/KhronosGroup/OpenCL-Docs/pull/1542.

PR #1542, opened 2026-04-01
document error for kernel enqueue with a work-group size of zero

bashbaug:
possible fix for #1541

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -12160,6 +12160,7 @@ Otherwise, it returns one of the following errors:
     {size_t_TYPE} on the device associated with _command_queue_
   * {CL_INVALID_WORK_GROUP_SIZE}
     ** if _local_work_size_ is not `NULL`, if the work-group size must be uniform, and if the _global_work_size_ is not evenly divisible by the _local_work_size_
+    ** if _local_work_size_ is not `NULL` and if the total number of work-items in the work-group is zero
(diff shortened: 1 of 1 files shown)


lpavank (approved):
Looks good to me.

joselopeqti (approved):

karolherbst:
I wonder if requiring this could break applications that simply specify a `[0, 0, 0]` local size.

But the spec also states `If local_work_size is specified, the values specified in global_work_size[0], …​, global_work_size[work_dim - 1] must be evenly divisible by the corresponding values specified in local_work_size[0], …​, local_work_size[work_dim - 1]` and that would be a division by 0. So if we have some major vendor that don't handle explicit local sizes of 0 I think it's safe to merge.

However if those are handled it's difficult to rule out that it wouldn't break things. But I think this is good to clarify, I just know that in rusticl enqueue with local_work_size == [0, 0, 0] is the same as specifying `NULL`. Not because the intention was to support 0 local sizes, but because it made sense to implement it that way logistically.

bashbaug:
> So if we have some major vendor that don't handle explicit local sizes of 0 I think it's safe to merge.

Mentioning for completeness: we currently return the error `CL_INVALID_WORK_GROUP_SIZE` when the local work-group size is explicitly zero for both CPU and GPU.

```
>>>> clEnqueueNDRangeKernel( Test ): queue = 0x5aefdf259830, kernel = 0x5aefe19b8810, global_work_size = < 512 x 1 >, local_work_size = < 0 x 0 >
<<<< clEnqueueNDRangeKernel -> CL_INVALID_WORK_GROUP_SIZE
```

PR #1557, opened 2026-05-11
clarify version requirements in the OpenCL C spec are for OpenCL C

### LINKED ISSUE #1556: Misleading references to OpenCL versions in the OpenCL C spec
In the OpenCL C spec, we usually use text similar to "[Requires](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#unified-spec) support for OpenCL C 1.2 or newer" when a feature requires a specific OpenCL C version, but in a few places we refer to the OpenCL version instead.  See, for example the description of get_enqueued_local_size, which uses the text "[Requires](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#unified-spec) support for OpenCL 2.0 or newer".  We should update these places to correctly use OpenCL **C** instead.

Specific places to fix:

* get_enqueued_local_size
* get_global_linear_id
* get_local_linear_id
* ctz

bashbaug:
fixes #1556

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -5075,7 +5075,7 @@ identifier of each work-item when this kernel is being executed on a device.
       Valid values of _dimindx_ are 0 to *get_work_dim*() - 1.
       For other values of _dimindx_, *get_enqueued_local_size*() returns 1.
 
-      <<unified-spec, Requires>> support for OpenCL 2.0 or newer.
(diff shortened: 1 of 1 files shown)


Kerilk (approved):
Thanks for this, the wording threw us off wrt SPIR-V extended instruction support.

PR #1558, opened 2026-05-12
revert clGetEventInfo returning CL_COMPLETE is a sync point

bashbaug:
This PR changes the behavior of `clGetEventInfo(CL_EVENT_COMMAND_EXECUTION_STATUS)` returning `CL_COMPLETE` back to the behavior in OpenCL 3.0.  This avoids a potential performance regression when the stronger host synchronization point is not needed, for example to determine if the event is `CL_COMPLETE` to query event profiling data.


lakshmih (approved):

bashbaug:
Merging as discussed in the May 19th teleconference.

PR #1563, opened 2026-05-18
fix erroneous leading underscore from halfn type names

bashbaug:
A few places in the OpenCL C spec erroneously included a leading underscore in several places referring to the "halfn" vector type.


bashbaug:
Editorial change, merging immediately.

PR #1572, opened 2026-06-17
Document Docker build

EwanC:
The repository README doesn't currently mention Docker, but this is by far the easiest way to build the OpenCL-Docs specs, as opposed to manually configuring some quite dated ruby dependencies. It is also how CI is testing the project, so the most faithful to reproduce any CI errors.

This PR updates the README to alert to developers that a Docker image is available help them build the project.

It is worth noting that the build instructions in both [Vulkan-Docs BUILD.adoc](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/BUILD.adoc#khronos-provided-docker-image) and [SYCL-Docs README](https://github.com/KhronosGroup/SYCL-Docs#building-using-the-khronos-docker-image) only document building the project using a docker image. We should consider whether we want a larger rewrite of the OpenCL-Docs REAME to only describe building the repo in terms of Docker, but for now the PR makes an incremental improvement.


bashbaug (approved):
LGTM. No reason to wait further to merge this change IMHO, since it's only affecting the README. We can always make additional changes later.

Issue #1552, opened 2026-05-04
memory_scope_work_item clarification


VerenaBeckham:
Quoting from https://github.com/KhronosGroup/OpenCL-Docs/blob/154cb2b0e9ac3e94e411dcee9f0a739af93d1a58/OpenCL_C.txt#L8115

> memory_scope_work_item can only be used with atomic_work_item_fence with flags set to CLK_IMAGE_MEM_FENCE.

Does that mean:

1. `memory_scope_work_item` can only be used with `atomic_work_item_fence` AND when it is used in this way the flags need to be set to `CLK_IMAGE_MEM_FENCE` or
2. If you use `atomic_work_item_fence` with `memory_scope_work_item` you must use the CLK_IMAGE_MEM_FENCE flag (no comment on any other APIs' usage of `memory_scope_work_item`)?

VerenaBeckham:
Some more background for this question.
I am trying to work out whether `atomic_load_explicit` is allowed to use `memory_scope_work_item`.
That's because SYCL allows the equivalent: using `sycl::atomic_ref::load()` with `sycl::memory_scope::work_item`. With an OpenCL backend this translates into `atomic_load_explicit` with `memory_scope_work_item`.
If OpenCL does not support this, then the SYCL behaviour needs to be clarified as undefined behaviour. (See the SYCL issue https://github.com/KhronosGroup/SYCL-Docs/issues/1004)
Either way, this sentence could be clarified in the OpenCL spec, since it is not clear to me. I'm happy to propose a clarified sentence, if that's the outcome.

VerenaBeckham:
I just noticed the OpenCL spec also mentions the supported scopes. In section 3.3.7 Memory Consistency Model for OpenCL 2.x there is a list of scopes and memory_scope_work_item has a footnote, and the footnote actually says "This value for memory_scope can only be used with atomic_work_item_fence with flags set to CLK_IMAGE_MEM_FENCE.", which is pretty clear.
So, I am quite confident that the answer is Option 1.
However, in my pdf version of the spec (downloaded from the Khronos website), the list of scopes is on page 38, whereas the footnote is on page 56! That's the end of the chapter. Might want to reconfigure that to be end of the page instead?
Anyway, I do feel the OpenCL C spec could improve its wording to make this clearer.

VerenaBeckham:
I have put up https://github.com/KhronosGroup/OpenCL-Docs/pull/1574/, which would address this ticket.

PR #1574, opened 2026-06-22
Clarify usage of memory_scope_work_item

VerenaBeckham:
Addresses https://github.com/KhronosGroup/OpenCL-Docs/issues/1552.

### DIFF
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -8107,8 +8107,8 @@ The following table lists the enumeration constants:
 |====
 | Memory Scope | Additional Notes
 | `memory_scope_work_item`
-    | `memory_scope_work_item` can only be used with both `atomic_work_item_fence`
-      and flags set to `CLK_IMAGE_MEM_FENCE`.
+    | `memory_scope_work_item` can only be used with `atomic_work_item_fence`
+      and in that case the flags must be set to `CLK_IMAGE_MEM_FENCE`.
       <<unified-spec, Requires>> support for OpenCL C 2.0 or newer.
 | `memory_scope_sub_group`
     | <<unified-spec, Requires>> support for
--- a/OpenCL_C.txt
+++ b/OpenCL_C.txt
@@ -8108,7 +8108,7 @@ The following table lists the enumeration constants:
 | Memory Scope | Additional Notes
 | `memory_scope_work_item`
     | `memory_scope_work_item` can only be used with `atomic_work_item_fence`
-      with flags set to `CLK_IMAGE_MEM_FENCE`.
+      and with flags set to `CLK_IMAGE_MEM_FENCE`.
       <<unified-spec, Requires>> support for OpenCL C 2.0 or newer.
 | `memory_scope_sub_group`
     | <<unified-spec, Requires>> support for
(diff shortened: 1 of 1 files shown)


VerenaBeckham:
@kpet

kpet:
This reads as if it is `atomic_work_item_fence` that is set to `CLK_IMAGE_MEM_FENCE`. I still think the original was clearer :).

VerenaBeckham:
Last option @kpet

OpenCL_C.txt
```diff
@@ -8108,7 +8108,7 @@ The following table lists the enumeration constants:
 | Memory Scope | Additional Notes
 | `memory_scope_work_item`
     | `memory_scope_work_item` can only be used with `atomic_work_item_fence`
-      with flags set to `CLK_IMAGE_MEM_FENCE`.
+      and in that case the flags must be set to `CLK_IMAGE_MEM_FENCE`.
       <<unified-spec, Requires>> support for OpenCL C 2.0 or newer.
```
bashbaug, line 8112: We usually use "must" to refer to conditions that are statically checkable and where some sort of error diagnostic is possible / desirable if the condition is violated. This condition isn't one that we can statically check, though, since the memory scope does not need to be a compile-time constant (even though typically it is).

As expected, violating this condition is currently allowed by Clang (no diagnostic), see: https://godbolt.org/z/cP3Yd368b

How about instead (can't leave this as a suggestion since it modifies the prior line, hopefully this is clear enough):

> Requires support for OpenCL C 2.0 or newer.  Behavior is undefined if `memory_scope_work_item` is used with an atomic function other than `atomic_work_item_fence` or with memory flags other than `CLK_IMAGE_MEM_FENCE`.

We don't have a similar restriction documented in the OpenCL SPIR-V environment spec. Should we?

VerenaBeckham: Thanks for following up!
I like the change you suggested and applied it.

The env spec (https://github.com/KhronosGroup/OpenCL-Docs/blob/1f2a44a4ecc2a76696410b3a110511ff2dc0ae49/env/validation_rules.asciidoc?plain=1#L126) does not describe the scopes individually, only their usage. `Invocation` scope, which is the equivalent of `work_item_scope` as I understand it, is explicitly excluded from the list of valid scopes for atomic instructions. So, I think that is all clear enough.

However, I also found a footnote in the OpenCL spec, which I changed to match.

api/footnotes.asciidoc
```diff
@@ -67,7 +67,7 @@ To create an image object from another image object that share the data store be
 ]
 
 :fn-image-mem-fence: pass:n[ \
-This value for *memory_scope* can only be used with *atomic_work_item_fence* with flags set to `CLK_IMAGE_MEM_FENCE`. \
+Behavior is undefined if this value for *memory_scope* is used with an atomic function other than *atomic_work_item_fence* or with memory flags other than `CLK_IMAGE_MEM_FENCE`. \
 ]
```
bashbaug, line 71: Not going to block these changes, but I would prefer to just remove this footnote, for several reasons:

1. Mentioning OpenCL C functions (`atomic_work_item_fence`) in the OpenCL API specification is usually a sign that something is documented in the wrong place.
2. This same restriction (probably?) applies to SPIR-V programs in addition to OpenCL C programs, in some form.
3. This seems like normative behavior that belongs somewhere in the main spec vs. in a footnote.

The updated text is an improvement over the prior text, which has been around since the OpenCL 2.x specifications, so I'm OK merging it in its current form if we decide not to just remove it outright, but if we do this let's file an issue to make further improvements in a separate PR.

PR #1579, opened 2026-06-24
Fix trailing spaces in enum names

Exanite:
Hello, I'm working on OpenCL bindings for C# as part of the Silk.NET project.

While working on the bindings, I found some trailing spaces in the names of 2 enums.


kpet (approved):
Trivial fix. Merging.

Issue #1145, opened 2024-04-04
[XML] Consistently group `<enum>`-s under their `<type>`


SunSerega:
### The problem

High-level bindings benefit from grouping enum values under a common type name.
This allows verbose IntelliSense suggestions and avoids oversaturating the namespace with all the enum values in one place.

The `cl.xml` file in this repo already defined its own enum type for every enum group:
```
        <type category="define">typedef <type>cl_uint</type>          <name>cl_device_local_mem_type</name>;</type>
...
    <enums name="cl_device_local_mem_type" vendor="Khronos">
        <enum value="0x1"           name="CL_LOCAL"/>
        <enum value="0x2"           name="CL_GLOBAL"/>
            <unused start="0x3" end="9999"/>
    </enums>
...
        <require comment="cl_device_local_mem_type">
            <enum name="CL_LOCAL"/>
            <enum name="CL_GLOBAL"/>
        </require>
```
However, there is currently no consistent way in which XML links the enum type (`cl_device_local_mem_type` above) and its valid values.

---

### Old solutions

The `<enums>` blocks seem to have been initially intended for this purpose.

However, this linkage was not maintained. At some point people just started adding all their new enums to:
```
    <enums start="0x1000" end="0x107F" name="cl_device_info" vendor="Khronos">
```
#779 has tried to rectify this, but was abandoned.

Initially, I also tried to continue the work in #779, because I was already using (and un-messing in other ways) the `<enums>` tags.

But then realized that the moderate abuse of `comment` attributes on `<require>` tags, of all things, was the most maintained form of such linkage between enums and their types.
I guess it's just a more natural form for it (in this repo at least).

---

### My solution

With that, I made #926, which currently replaces all the `comment` attributes referring to an enum type with a `group` attribute.

A few odd-balls of the `comment` attributes currently follow one of these 3 patterns:
```
        # The "- bitfield" and the end (missing on some bitfields, I think I caught all of those)
        <require comment="cl_device_type - bitfield">
...
        # 2 enum types sharing <require> tag
        <require comment="cl_mem_flags and cl_svm_mem_flags - bitfield">
...
        # Human-only text describing things like deprecation, also containing the actual enum type name
        <require comment="OpenCL 1.2 cl_image_info enums that were deprecated in OpenCL 2.0">
```
I've converted them into:
```
        # Add etype (enum type/kind/usage), attribute
        <require group="cl_device_type" etype="bitfield">
...
[...]

Issue #1585, opened 2026-07-14
Inconsistent enum labeling for `cl_khr_dx9_media_sharing` enums


NogginBops:
In [cl_khr_dx9_media_sharing](https://github.com/KhronosGroup/OpenCL-Docs/blob/main/api/cl_khr_dx9_media_sharing.asciidoc) the following enum constants are defined as being part of the `cl_context_info` enumeration. 

But looking at https://github.com/KhronosGroup/OpenCL-Docs/blob/main/man/static/enums.txt which documents all valid enum values for each enumeration these values seem to be listed under `cl_context_properties`.

However, in `cl.xml` these enums are included under a `cl_context_info` comment.
https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml#L5914-L5917

It does seem likely from [other specification wording](https://registry.khronos.org/OpenCL/specs/unified/refpages/man/html/clCreateContextFromType.html#:~:text=CL_INVALID_DX9_MEDIA_,created%20may%20interoperate) that the correct enum is indeed `cl_context_properties` but it's a little hard to know.

If the `cl_khr_dx9_media_sharing` file is wrong it and `cl.xml` should be updated to reflect the correct enumeration for these values.

PR #1587, opened 2026-07-14
Mark all enum constants with their enumeration

NogginBops:
After many attempts at making enum group metadata available in `cl.xml` (#779, #926, #1145), I've decided to give it a try too.

In this PR I went with the original plan of marking the `<enums>`-tags themselves with the enum groups the containing enums have, but also allow individual `<enum>`-tags be marked with additional enum groups that they are part of. 
This is similar to how this is handled in `gl.xml` where the group association is done directly on the enum entries themselves.

Before marking this PR as not a draft I'll probably undo my excessive splitting of `<enums>`-tags. 
My original goal was to only have `group=""` attributes on the `<enums>`-tags and not on `<enum>`-tags but in hindsight this was not necessarily a good idea due to most entries not being grouped by enumeration.

Personally I think I'm leaning towards #926 being the better approach as it is features and extensions themselves that define the typedefs for enum groups so the association between enum entry and type name seems more appropriate there.

The argument for doing the approach this PR does is that it matches how this data is described in `gl.xml` and there this enum grouping has been very successful. Important to note is that the enum group situation in OpenGL seems to much more complicated than it is in OpenCL and I feel it should be possible to have very accurate enum groups in `cl.xml`, with enough information the https://github.com/KhronosGroup/OpenCL-Docs/blob/main/man/static/enums.txt file could even be machine generated to always be kept up to date.

### DIFF
--- a/xml/cl.xml
+++ b/xml/cl.xml
@@ -1438,16 +1438,28 @@ server's OpenCL/api-docs repository.
         <enum value="0x1074"        name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_8BIT_KHR"/>
         <enum value="0x1075"        name="CL_DEVICE_INTEGER_DOT_PRODUCT_ACCELERATION_PROPERTIES_4x8BIT_PACKED_KHR"/>
             <unused start="0x1076" end="0x107F" comment="Reserved for cl_device_info"/>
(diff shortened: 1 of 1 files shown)


NogginBops:
Also, during this work I've found the following issue (#1585) the made it hard to know where to group some enums.

PR #1588, opened 2026-07-17
Add missing error check in cl_khr_command_buffer

kpet:
When a command buffer is enqueued with queues set to NULL, the queues provided at command buffer creation time are used but these were never checked against the device's advertised queue properties support and requirements.

Change-Id: I5803d11c1e077e0ed525472dedb5265e4f4a8685

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -15010,6 +15010,12 @@ returned in _errcode_ret_:
   * {CL_INVALID_PROPERTY} if values specified in _properties_ are valid but
     are not supported by all the devices associated with command-queues in
     _queues_.
+  * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any command-queue
+    in _queues_ contains a property not specified by
+    {CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR}.
+  * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any
+    command-queue in _queues_ does not contain the minimum properties
+    specified by {CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR}.
   * {CL_OUT_OF_RESOURCES} if there is a failure to allocate resources
     required by the OpenCL implementation on the device.
   * {CL_OUT_OF_HOST_MEMORY} if there is a failure to allocate resources
@@ -15172,11 +15178,12 @@ execution was successfully queued, or one of the errors below:
   * {CL_INVALID_COMMAND_QUEUE} if any element of _queues_ is not a valid
     command-queue.
   * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any command-queue
-    in _queues_ contains a property not specified by
+    in _queues_, when present, contains a property not specified by
     {CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR}.
   * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any
-    command-queue in _queues_ does not contain the minimum properties
-    specified by {CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR}.
+    command-queue in _queues_, when present, does not contain the minimum
+    properties specified by
+    {CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR}.
   * {CL_INVALID_DEVICE} if any element of _queues_ does not have the same
     device as the command-queue set on _command_buffer_ creation at the
     same list index.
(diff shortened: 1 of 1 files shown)


api/opencl_runtime_layer.asciidoc
```diff
@@ -15010,6 +15010,12 @@ returned in _errcode_ret_:
   * {CL_INVALID_PROPERTY} if values specified in _properties_ are valid but
     are not supported by all the devices associated with command-queues in
     _queues_.
+  * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any command-queue
+    in _queues_ contains a property not specified by
+    {CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR}.
+  * {CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if the properties of any
+    command-queue in _queues_ does not contain the minimum properties
+    specified by {CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR}.
```
EwanC: I'm not sure about this part of the change. There is currently the spec wording (emphasis my own)

> When constructing a command-buffer by appending commands, the queue parameter passed for the command being created is used to set the device with which the command will be associated with, and also inform the scheduling of the command. If the queue is an in-order queue, then an additional dependency is created on the last command appended to the command-buffer using the same queue parameter. If the queue is an out-of-order queue, then no extra dependencies on previous commands using the same queue are created. **All queue properties other than [CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) are ignored for the purposes of command creation**, with the exception of any vendor extension defined queue properties that explicitly define semantics for this purpose.

So these new error conditions place extra restrictions on the properties of command-buffers used for command recording, even though the spec say we're ignoring most of the queue properties anyway.

We might be better off instead adding extra error conditions to clEnqueueCommandBufferKHR, e.g
`{CL_INCOMPATIBLE_COMMAND_QUEUE_KHR} if _queues_ is NULL and the properties of any command-queue used creation of _command_buffer_ contains a property not specified by {CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR}.`

EwanC (approved):
LGTM

bashbaug:
Confirming: This is the expected behavior?

1. Recording into a command-buffer created from a command-queue with missing [CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR) is valid, as long as a different command-queue with the required queue properties is explicitly specified when the command-buffer is executed.
2. Recording into a command-buffer created from a command-queue with unsupported properties not in [CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR) is valid, again as long as a command-queue without the unsupported properties is explicitly specified when the command-buffer is executed.

Do we have tests for these two scenarios?  Note, these are positive tests, not negative tests, though scenario (1) requires non-zero required queue properties, and scenario (2) requires extra un-supported queue properties, either for this device or for some other device in the context.

EwanC:
> Confirming: This is the expected behavior?
> 
>     1. Recording into a command-buffer created from a command-queue with missing [CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR) is valid, as long as a different command-queue with the required queue properties is explicitly specified when the command-buffer is executed.
> 
>     2. Recording into a command-buffer created from a command-queue with unsupported properties not in [CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR) is valid, again as long as a command-queue without the unsupported properties is explicitly specified when the command-buffer is executed.
> 
> 
> Do we have tests for these two scenarios? Note, these are positive tests, not negative tests, though scenario (1) requires non-zero required queue properties, and scenario (2) requires extra un-supported queue properties, either for this device or for some other device in the context.

I think the closest test we have to this is one in [command_buffer_queue_substitution.cpp](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp) where we create a command-buffer with an out-of-order queue and execute it on an in-order queue. We don't verify whether or not `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` is in command buffer supported queue properties in that tests, so I think this could stress your second testing scenario by chance if a device happens to not support it.

We should have dedicated positive tests for both these situations if we go ahead with this change though. So I've extended the CTS PR https://github.com/KhronosGroup/OpenCL-CTS/pull/2757 I have open to cover this spec change with these two scenarios.

bashbaug:
Merging as discussed in the August 11th teleconference.

PR #1593, opened 2026-07-29
Add dependency on PYAPIMAP to API includes, so the 'Provided by' comments

### LINKED ISSUE #1591: Missing Dependency in the OpenCL-Docs Makefile
There's a missing dependency somewhere in the OpenCL-Docs Makefile that I haven't been able to completely untangle.

This happens to work, and the dependency is somehow being satisfied:

```sh
python3 makeSpec -clean -spec khr+ext OUTDIR=out.test -v manhtmlpages apihtml
```

This doesn't work, though, and the generated specs are missing API dependency information (e.g. `// Provided by CL_VERSION_1_0`):

```sh
python3 makeSpec -clean -spec khr+ext OUTDIR=out.test -v apihtml
```

I think the issue has to do with the `apimap.py` file, which is being generated in the "good" case, but not in the "bad" case?  I'm not 100% sure, though, and I'm not sure which dependency to add and where, assuming that this is the problem.

oddhack:
are generated when just the API spec (and not refpages) is being built.

@bashbaug give this a try, it was a simple oversight.

Fixes #1591


makeSpec
```diff
@@ -11,6 +11,7 @@
 #   -genpath path - directory for generated files and outputs
 #   -spec core - make a spec with no extensions (default)
 #   -spec khr - make a spec with all KHR extensions
+#   -spec khr - make a spec with all KHR + all EXT extensions
```
bashbaug: This seems to contradict the previous line - maybe this was intended to be an edit rather than an addition?

bashbaug (approved):
Thanks, this worked for me, appreciate the fast fix!

makeSpec
oddhack:
```suggestion
#   -spec khr+ext - make a spec with all KHR + all EXT extensions
```

Oops.

PR #1594, opened 2026-07-29
SPIR-V integer image sampling restriction

bashbaug:
The main purpose of this change is to add a missing restriction to the SPIR-V environment spec when reading from an image with a sampler: If the image is an integer image, then the sampler must be using nearest filtering or behavior is undefined.  This restriction is present in OpenCL C, e.g.:

> The read_image{i|ui} calls support a nearest filter only. The filter_mode specified in sampler must be set to CLK_FILTER_NEAREST; otherwise the values returned are undefined.

It is also consistent with restrictions in the graphics APIs, so I do not believe this was ever intended to work via SPIR-V.

Additionally:

* Switch to describing the restrictions in terms of defined and undefined behavior rather than valid and invalid behavior.  Defined and undefined behavior is more accurate, since these conditions cannot be statically verified.
* Editorial updates to improve table appearance.

I'd like to make similar changes to the OpenCL C spec eventually.  Right now, we're describing these rules for each of the image read and write overloads, which is redundant and error-prone.  We should be able to describe these restrictions in one place instead, which would shorten the spec and improve understanding.

### DIFF
--- a/env/image_addressing_and_filtering.asciidoc
+++ b/env/image_addressing_and_filtering.asciidoc
@@ -1278,18 +1278,14 @@ layer = v
 
 === Coordinate Format for Reading and Writing Images
 
-This section describes valid types for the _Coordinate_ operand used by image
+This section describes supported types for the _Coordinate_ operand used by image
 read instructions (such as *OpImageRead*) or image write instructions (such as
 *OpImageWrite*).
-The valid operand types are determined by the *OpImageType* for the image and
+The supported operand types are determined by the *OpImageType* for the image and
 whether the image is being read from or written to.
 
-The following table describes the valid types for the _Coordinate_ operand when
-reading from the specified image type.
-The integer types for the _Coordinate_ operand are only valid when the image is
-read without a sampler (such as *OpImageRead*), or with a sampler using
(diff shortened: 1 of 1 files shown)


env/image_addressing_and_filtering.asciidoc
```diff
@@ -1505,6 +1516,8 @@ endif::cl_ext_image_unsigned_10x6_12x4_14x2[]
 
 |====
 
+For the integer image data types, when reading from an image with a sampler,
```
bashbaug, line 1519: Discussed in the August 4th teleconference.  Need to use a more precise term here than "integer image data types".

bashbaug: I've updated the wording to:

> For the image channel data types that return integer data, when reading from an image with a sampler, behavior is undefined unless the sampler is using Nearest filtering.

The set of image formats that return integer data is described in the table immediately above this sentence.

@Kerilk does this work for you?

Kerilk: Yes, I think so.

Kerilk (approved):
LGTM

bashbaug:
Merging as discussed in the August 11th teleconference.

PR #1602, opened 2026-08-20
clarify clSetProgramSpecializationConstant behavior

### LINKED ISSUE #553: clSetProgramSpecializationConstant description needs clarification
At the moment, [`clSetProgramSpecializationConstant`](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/api/opencl_runtime_layer.asciidoc#setting-spir-v-specialization-constants) only documents what happens if user uses `clCreateProgramWithIL` -> `clBuildProgram` sequence of APIs, however, it is possible to do separate compilation and linking even for SPIR-V inputs via `clCreateProgramWithIL` -> `clCompileProgram` -> `clBuildProgram`.

For the second usage model, the spec doesn't document which exact API call (`clCompileProgram` or `clBuildProgram`) uses values set by `clSetProgramSpecializationConstant`:

> The data pointed to by spec_value are copied and can be safely reused by the application after {clSetProgramSpecializationConstant} returns. This specialization value will be used by subsequent calls to {clBuildProgram} until another call to {clSetProgramSpecializationConstant} changes it. 

> Calling this function multiple times for the same specialization constant shall cause the last provided value to override any previously specified value. The values are used by a subsequent {clBuildProgram} call for the program.

bashbaug:
fixes #553

The values set for specialization constants affect the next time the program is compiled using clCompileProgram or built using clBuildProgram.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -9480,29 +9480,18 @@ include::{generated}/api/version-notes/clSetProgramSpecializationConstant.asciid
     set.
   * _spec_size_ specifies the size in bytes of the data pointed to by
     _spec_value_.
-    This should be 1 for boolean constants.
-    For all other constant types this should match the size of the
-    specialization constant in the module.
-  * _spec_value_ is a pointer to the memory location that contains the value of
-    the specialization constant.
-    The data pointed to by _spec_value_ are copied and can be safely reused by
-    the application after {clSetProgramSpecializationConstant} returns.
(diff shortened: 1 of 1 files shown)


bashbaug:
Note that the test additions were merged, so we should merge the spec changes also to stay in sync.  Still need a review approval before merging, though.

CC @kpet @lakshmih @gowtham-sarc

kpet (approved):

bashbaug:
Thanks! I'm going to go ahead and merge this. We can always make additional fixes later.

PR #1605, opened 2026-08-28
clarify usage of the CL_MEM_KERNEL_READ_AND_WRITE memory flags

### LINKED ISSUE #1599: Clarification on intended behavior when CL_MEM_KERNEL_READ_AND_WRITE is passed into clCreateBuffer
For reference, Section 5.2.1 Table 19. List of supported memory flag values states this about CL_MEM_KERNEL_READ_AND_WRITE:
"This flag is only used by [clGetSupportedImageFormats](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clGetSupportedImageFormats) to query image formats that may be both read from and written to by the same kernel instance. To create a memory object that may be read from and written to use [CL_MEM_READ_WRITE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_READ_WRITE)."

This table of values is referenced earlier in the description for clCreateBuffer:
"flags is a bit-field that is used to specify allocation and usage information about the buffer memory object being created and is described in the [supported memory flag values](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#memory-flags-table) table."

While it is clear that CL_MEM_KERNEL_READ_AND_WRITE is not intended to be passed into clCreateBuffer, it seems implied that passing it would be supported since it is in the table of supported flags. This behavior is not tested by the official CTS, but it is tested by piglit, which expects the clCreateBuffer calls to succeed. This causes piglit tests to fail on Mali DDK. This piglit merge request would have added mutual exclusion checks between CL_MEM_KERNEL_READ_AND_WRITE and CL_MEM_READ_ONLY/CL_MEM_WRITE_ONLY/ CL_MEM_READ_WRITE to pass with Mali: https://gitlab.freedesktop.org/mesa/piglit/-/merge_requests/1004. This change was rejected, as this behavior is inconsistent with the rusticl implementation which simply ignores CL_MEM_KERNEL_READ_AND_WRITE.

Admittedly, this behavior probably shouldn't be an issue as long as flags are set directly in the clCreateBuffer call. However, one could fathom a developer being lazy, assigning flags to a variable (or getting lazy with copy+paste), and then using that variable to call both clGetSupportedImageFormats and clCreateBuffer. In this case, clCreateBuffer should still have consistent behavior across implementations. I think the current specification implies that these calls should succeed and ignore the flag. I read "flag is only used by clGetSupportedImageFormats" as "an OpenCL implementation should not use the flag to affect the execution logic (including validation) of any function except clGetSupportedImageFormats." However, "flag is only used" may be interpreted to mean "flag is valid only when used by" like with the Mali DDK. 

This difference in interpretation probably justifies an extra sentence or two in the spec explicitly stating what should happen in order to help implementers align the behavior. It probably makes the most sense to enforce the more permissive interpretation to ensure that existing applications don't break. Speaking of conformance, the official CTS could also use a test to validate this behavior if the spec is made more explicit.

bashbaug:
fixes #1599

Clarifies that the `CL_MEM_KERNEL_READ_AND_WRITE` flag is only used to query supported image formats and is not valid for use when creating memory objects.

### DIFF
--- a/api/opencl_runtime_layer.asciidoc
+++ b/api/opencl_runtime_layer.asciidoc
@@ -870,10 +870,11 @@ include::{generated}/api/version-notes/CL_MEM_HOST_NO_ACCESS.asciidoc[]
 | {CL_MEM_KERNEL_READ_AND_WRITE_anchor}
 
 include::{generated}/api/version-notes/CL_MEM_KERNEL_READ_AND_WRITE.asciidoc[]
-  | This flag is only used by {clGetSupportedImageFormats} to query image
+  | This flag is only valid for {clGetSupportedImageFormats} to query image
     formats that may be both read from and written to by the same kernel
     instance.
-    To create a memory object that may be read from and written to use
+    It must not be used to create a memory object.
+    To create a memory object that may be read from and written to, use
     {CL_MEM_READ_WRITE}.
 
 ifdef::cl_ext_immutable_memory_objects[]


bashbaug:
If we merge this change then we should add a negative CTS test to ensure an error (`CL_INVALID_VALUE`?) is returned if this flag is used to create a buffer or image memory object.

NosajHS:
Looks good to me

lakshmih (approved):

lakshmih:
https://github.com/KhronosGroup/OpenCL-CTS/issues/2791

PR #1611, opened 2026-09-03
clarify what is meant by a null string for CL_PROGRAM_SOURCE

bashbaug:
See discussion: https://github.com/KhronosGroup/OpenCL-CTS/pull/2784#discussion_r3905619503

This PR clarifies that CL_PROGRAM_SOURCE can return a null string (no characters) or an empty string (just the null terminator) when the program object being queried is created from a binary and no program source code is stored in the binary.

Note that this is the one place in the spec where we use the term "null string", so this is the only place that I believe needs a clarification.

Alternatively, I suppose we could decide that "null string" is the wrong term, and this should be "empty string" like other queries, but the "null string" term has been used since OpenCL 1.1 and there are implementations in the wild that return zero characters for the query so IMHO it's too late to change it.

For Khronos folks, some original internal discussion about this scenario can be found in Bugzilla 6719.


rjodinchr (approved):

joselopeqti (approved):

bashbaug:
Merging as discussed in the September 15th teleconference.

